PR #23, opened 2022-03-09
use decorations for ordered lists

YousefED:
This is an alternative version that uses decorations. I pushed some other changes to :main on your repo, my bad (but those might also be interesting to review)


YousefED:
I don't think this is an issue any more in the latest version of BlockNote

PR #32, opened 2022-03-22
Position resize fix

### LINKED ISSUE #15: DragHandle positioning
We need to clean up the position of the draghandle:

- [ ] Fix vertical position: the draghandle is now too far to the top. Check for both headings and regular elements
- [ ] Fix horizontal position (see below)

_Horizontal position_
The draghandle now displays right before elements, even when they are indented. This is ugly as it overlaps with the vertical line ("the indentation guide"). 

Design-wise; I don't think there's an easy option where we can have both indentation guides and a draghandle right next to the block.

So there are two options here:

a) remove indentation guides (notion also doesn't have this, it's more inspired by Roam) and display draghandles right next to the element similar to Notion
b) always show the draghandles on the left side of the document (so at the same horizontal position as root-level elements)

image

Preferably, indentation lines can be configured by the consumer library, and if enabled  we show Draghandles at position B, otherwise at position A

17Amir17:
#15 
The issue was that the horizontal anchor was set only when the page opened. I added an event handler on the 'resize' event to update the horizontal anchor. This brought a new issue that the anchor is updated but there is no transaction and as a result the drag handle on re-renders after the mouse is moved. I fixed this by moving the render functionality into a function and created a transaction each time the page resize, and calling the render function.
thanksnoam

### DIFF
--- a/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.tsx
+++ b/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.tsx
@@ -101,7 +101,6 @@ function dragStart(e: DragEvent, view: EditorView) {
 export const createDraggableBlocksPlugin = () => {
   let dropElement: HTMLElement | undefined;
 
-  let prevCoords: Coordinates; // Prev coords draghandle was at
   let horizontalAnchor: number; // Where to anchor the draghandle on the x axis
 
   const WIDTH = 48; // Width of drag handle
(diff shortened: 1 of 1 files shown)


horacioh:
the solution I came up with on my app was to "disable" the handle for events outside the editor (window resize). I only show it when the user is moving the mouse over the editor. I control everything with a state machine, if you like I can share it with y'all!

YousefED:
> the solution I came up with on my app was to "disable" the handle for events outside the editor (window resize). I only show it when the user is moving the mouse over the editor. I control everything with a state machine, if you like I can share it with y'all!

Sounds interesting! I think more examples are always welcome / useful!

YousefED:
fixed by #92

PR #41, opened 2022-09-28
fix copy-paste

richmengsix:
The copy-paste functionality within the app was broken because there was no way for the parsing rule to distinguish between `block`/`blockgroup`/`content`.

In this PR, added extra rule conditions for `block` and `content` to disambiguate.

_NOTE: Copy-pasting external document "into" the editor is still problematic after this fix._

### DIFF
--- a/packages/core/src/extensions/Blocks/nodes/Block.ts
+++ b/packages/core/src/extensions/Blocks/nodes/Block.ts
@@ -103,6 +103,7 @@ export const Block = Node.create<IBlock>({
     return [
       {
         tag: "div",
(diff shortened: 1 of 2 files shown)


richmengsix:
Well, I found this fix is still buggy for list recursions. Looking to see if there are ways around it :/

YousefED:
Thanks for this @richmengsix ! Appreciate the PR. @matthewlipski took this as a starting point and made a more extensive fix here: https://github.com/YousefED/BlockNote/pull/46. Hope this works for you as well!

PR #42, opened 2022-09-28
fix order list counter

richmengsix:
Now order list counter correctly calculates per `blockgroup` layer.


YousefED:
Thanks @richmengsix! Closes https://github.com/YousefED/BlockNote/issues/62

PR #45, opened 2022-10-05
poc: Blocknote lexical

YousefED:
This PR is a PoC of an implementation of BlockNote on top of Lexical: https://blocknote-48r0ou3yn-yousefed.vercel.app/

https://user-images.githubusercontent.com/368857/193570335-4402de8b-9707-40f6-94d3-21f981da019f.mov

We're currently investigating the structure of "nexted" blocks. In editors like Notion and also in BlockNote ProseMirror, indented blocks are considered children of their parent blocks. This has some advantages:

- makes it possible to create container elements such as "toggles" or "callouts", while allowing rich children
- easier to work with drag and drop (dragging a parent node also drags the contents)
- in general; it makes tabs / indents part of the structure of the document, instead of purely a UI aspect. With this, editing a block-based document feels more like editing a coherent structure instead of a freeform document (like word / google docs)
- more ideas [here](https://www.notion.so/blog/data-model-behind-notion)

I think there are two ways to go about this:

### Extended schema
Make all core nodes (paragraph, quote, heading, etc. `block` types, which contain `content` and an optional `childgroup`, that represent nested items. This is how BlockNote currently works ([read more](https://github.com/YousefED/BlockNote/blob/main/packages/core/ARCHITECTURE.md)).

```xml
  <block type="heading">
      <content><h1><text>Parent element 1</text></h1></content>
      <childgroup>
          <block type="paragraph">
              <content><p><text>Nested / child / indented item</text></p></content>
          </block>
      </childgroup>
  </block>
```

Pros:
- this worked ok for prosemirror
- the "schema" is close to the DOM structure that we'd like

Cons:
- It doesn't seem clean. We introduce a lot of extra structure in the schema (editor state), only to achieve the concept of children. Instead of designing it so that "a paragraph can have children", we are now saying "a paragraph is represented as a block, with a paragraph as a child"; i.e.: extra level of complexity
- All core nodes are now represented as a `block`. This will probably break compatibility with other Lexical plugins in the future
- It's going to be confusing for developers, as it differs significantly with how Lexical normally works (this is feedback we got on our PM implementation as well)

### Clean schema
Alternatively, (which is what we do in this PR) we can try to keep the schema / structure cleaner:

```xml
<h1>
      <text>Parent element 1</text>
      <childgroup>
          <p><text>Nested / child / indented item</text></p>
      </childgroup>
  </h1>
```

Pros:
- cleaner schema
- We keep more semantic node types, hopefully easier to work with other plugins in the future

Cons:
- We have to modify / extend nodes to add a "childgroup"
- We need to modify how the elements are rendered in the DOM. The DOM schema that we like is closer to "Extended schema" mentioned above (i.e.: we want a root div with the h1 and childgroup as children, instead of having the childgroup inside the DOM `<h1>`).

**Current implementation**

This PR tries to implement the "clean schema":
- We patch the main nodes to output a different DOM structure. The patching is a bit hacky, but we can change this to extending the original nodes (and perhaps also use the Proxy concept). The DOM structure is achieved with https://github.com/facebook/lexical/pull/3062 
- A number of extra fixes were required for handling selection and linebreaks: https://github.com/YousefED/lexical/pull/2/files. This is currently quite hacky (Lexical should not depend on "childgroup") so we need to see if we can achieve the same in a cleaner way and / or improve Lexical so it makes less "implicit assumptions" of the document structure

We should decide whether we think this approach makes sense, and if so, how we can make this as clean as possible.


shinzui:
Is this an experiment, or are you planning to switch to Lexical? I ask since I'm evaluating Lexical and tiptap and discovered your project on the Lexical discord.

YousefED:
> Is this an experiment, or are you planning to switch to Lexical? I ask since I'm evaluating Lexical and tiptap and discovered your project on the Lexical discord.

Hi @shinzui! So far, this is mostly an experiment. If we can cleanly resolve the issues mentioned in this PR we'll consider migrating to Lexical, it seems quite promising

PR #46, opened 2022-10-18
Fix Copy/Paste behaviour

### LINKED ISSUE #40: Bug: keep styles on copy / paste
See https://twitter.com/steventey/status/1567956854801833984

matthewlipski:
Currently, pasting content in the editor results in improper formatting in most cases. This PR fixes these bugs, so that the formatting of clipboard content is preserved when pasting it into the editor.

Closes #40


packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -60,6 +60,7 @@ export const Block = Node.create<IBlock>({
     return {
       listType: {
         default: undefined,
+        // Causes nesting issues pasting outside editor
```
YousefED, line 63: can you explain this / is this still accurate?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -98,25 +99,83 @@ export const Block = Node.create<IBlock>({
...
     return [
+      // For all content copied from within the editor.
       {
         tag: "div",
+        getAttrs: (element) => {
+          if (typeof element === "string") {
+            return false;
+          }
+
+          if (element.getAttribute("node") === "block") {
```
YousefED, line 112: My proposal would be to use data- attributes for this, I think it's a best practice; https://www.w3schools.com/tags/att_data-.asp

i.e.: data-bn-type (`bn-type` indicates it's the blocknote type)

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -98,25 +99,83 @@ export const Block = Node.create<IBlock>({
...
+            return false;
+          }
+
+          let ancestor = element.parentElement;
+
+          if (ancestor === null) {
+            return false;
+          }
+
+          while (ancestor.tagName !== "B") {
```
YousefED, line 151: this while doesn't make sense? ancestor is never changed?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -98,25 +99,83 @@ export const Block = Node.create<IBlock>({
...
     ];
   },
 
-  renderHTML({ HTMLAttributes }) {
+  renderHTML({HTMLAttributes}) {
     return [
       "div",
       mergeAttributes(this.options.HTMLAttributes, HTMLAttributes, {
         class: styles.blockOuter,
+        node: "block-outer"
```
YousefED, line 172: (same about the data- attribute)

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -149,89 +208,89 @@ export const Block = Node.create<IBlock>({
...
   },
 
   addCommands() {
     return {
       setBlockHeading:
         (attributes) =>
-        ({ tr, dispatch }) => {
-          return setBlockHeading(tr, dispatch, attributes.level);
-        },
+          ({tr, dispatch}) => {
```
YousefED, line 219: I think there are still some formatting-only changes here

packages/core/src/extensions/Blocks/nodes/BlockGroup.ts
```diff
@@ -13,14 +13,30 @@ export const BlockGroup = Node.create({
...
+    return [
+      {
+        tag: "div",
+        getAttrs: (element) => {
+          if(typeof element === "string") {
+            return false;
+          }
+
+          if(element.getAttribute("node") === "block-group") {
+            return null;
```
YousefED, line 25: also comment why `return null` here (like you've done for the other nodes)

YousefED (changes requested):
Nice, this seems to work well :)

Would be great if we can add playwright tests for this as well. Have you been able to get those working locally?

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Yeah, lists copied from BlockNote don't get formatted properly in other apps - I'm pretty sure it's down to the fact that we don't use _ul_ or _ol_ tags to mark the start and end of a list. Copying a list from BlockNote and pasting it in e.g. Google Docs therefore makes list  items get reformatted as paragraphs, so nesting also doesn't work. Not sure if this is currently relevant right now though.

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Whoops, it was supposed to traverse up the DOM tree until it finds an ancestor that's a list. I guess it's not needed either way since list items always have a list as a parent.

YousefED (approved):
Great stuff!

sunillarsson:
when using copy/paste functions from different websites the results seem to be a bit so so. Using chatgpts copy function strips the line breaks etc. in blocknote.

PR #52, opened 2022-10-24
Mantine migration

### LINKED ISSUE #50: UI migration to Mantine
Currently, BlockNote uses [AtlasKit](https://atlaskit.atlassian.com/) to style most of its UI components. However, AtlasKit is meant more for internal use by Atlassian, and it's less suited for use in BlockNote. Therefore, the UI should be migrated to [Mantine](https://ui.mantine.dev/).

matthewlipski:
This PR converts all AtlasKit UI elements with Mantine elements, currently including the bubble and slash menus.

Issues:

- [x] Clicking bubble menu buttons causes the menu to disappear for 1-2 frames which looks unpolished.
- [x] Bubble menu dropdown selector length does not adapt to the currently selected item's text length.
- [x] Text gets temporarily deselected when Bubble menu dropdown is open.
- [x] Opening the slash menu causes editor to lose focus, meaning results can't be filtered.

Closes #50


packages/core/src/extensions/BubbleMenu/component/BubbleMenu.tsx
```diff
@@ -59,123 +59,157 @@ export const BubbleMenu = (props: { editor: Editor }) => {
...
-            icon={RiH1}
-            isSelected={currentBlockName === "Heading 1"}
-            onClick={() =>
+      <SimpleToolbarDropdown
+        text={currentBlockName}
+        icon={blockIconMap[currentBlockName]}
+        items={[
+          {
+            onClick: () => {
+              props.editor.view.focus();
```
YousefED, line 79: is this necessary? as there is already a focus in the chain below?

packages/core/src/extensions/Hyperlinks/menus/helpers/PanelTextInput.tsx
```diff
@@ -30,6 +29,7 @@ export interface State {
   value?: string;
 }
 
+const isMacOS = () => /Mac/.test(navigator.platform);
```
YousefED, line 32: Let's move this to a "util" file. We don't want this logic spread around the codebase

packages/core/src/extensions/Hyperlinks/menus/helpers/PanelTextInputStyles.ts
```diff
@@ -1,8 +1,10 @@
-import { N400, N800 } from "@atlaskit/theme/colors";
 import styled from "styled-components";
 
 // code taken from https://bitbucket.org/atlassian/design-system-mirror/src/master/editor/editor-core/src/ui/PanelTextInput/styles.ts
 
+export const N400 = "#505F79";
```
YousefED, line 5: (about line 1)

Let's remove the dependency on styled-components, as mantine uses emotion, and there's no reason to use both afaik

packages/core/src/extensions/helpers/formatKeyboardShortcut.ts
```diff
@@ -1,7 +1,10 @@
-import browser from "../../lib/atlaskit/browser";
+const isAppleOS = () =>
```
YousefED, line 1: Move to util file (same comment as above)

Or, alternatively, keep the old "browser" file from atlaskit (just move it somewhere else)

packages/core/src/shared/components/toolbar/SimpleToolbarButton.tsx
```diff
@@ -29,27 +29,27 @@ export const SimpleToolbarButton = forwardRef(
...
-                  (props.isSelected ? styles.isSelected : "") +
-                  " " +
-                  (props.isDisabled ? styles.isDisabled : "")
-                }
-              />
-            )
-          }>
-          {props.children}
-        </Button>
+        {props.children ? (
```
YousefED, line 32: when do we use the variant with children?

packages/core/src/shared/plugins/suggestion/components/SuggestionGroup.tsx
```diff
@@ -70,23 +52,47 @@ function SuggestionComponent<T extends SuggestionItem>(
...
 
+  const Icon = props.item.icon;
+
   return (
-    <div className={styles.buttonItem}>
-      <ButtonItem
-        isSelected={isButtonSelected} // This is needed to navigate with the keyboard
-        iconBefore={getIcon(props.item, isButtonSelected)}
-        onClick={(_e) => {
+    <MantineProvider
```
YousefED, line 58: As we use this in two places, should we maybe refactor to something like:

`<BlockNoteMantineProvider>...</BlockNoteMantineProvider>`?

packages/core/src/shared/plugins/suggestion/components/SuggestionList.tsx
```diff
@@ -59,13 +59,15 @@ export function SuggestionList<T extends SuggestionItem>(
...
   return (
     <div className={styles.menuList + " " + rootStyles.bnRoot}>
-      <MenuGroup>
-        {renderedGroups.length > 0 ? (
-          renderedGroups
-        ) : (
-          <Section title={"No match found"}> </Section>
-        )}
-      </MenuGroup>
+      <Menu defaultOpened={true} trigger={"hover"} closeDelay={10000000}>
```
YousefED, line 62: let's always make sure to comment workarounds like these (why is this necessary?), so we can easily understand it later :)

YousefED (commented):
Nice! This seems to map quite well :)

I think the animations when changing the block type (for example, from paragraph to heading) are broken now though

packages/core/src/extensions/BubbleMenu/component/BubbleMenu.tsx
matthewlipski: For some reason chaining commands still causes that flicker in the bubble menu, even though the editor should be focused before calling the command. It's still definitely a focus related issue, since I had a different fix before using `event.preventDefault()` on `MouseDown` events, but this was a cleaner solution since I also had to deal with some of Mantine Menu's custom behaviour. Either way, I'll remove the `focus()` calls from the chain.

packages/core/src/extensions/Hyperlinks/menus/helpers/PanelTextInputStyles.ts
matthewlipski: Agreed, I'll wrap the `PanelTextInput` component in a `MantineProvider`.

packages/core/src/extensions/helpers/formatKeyboardShortcut.ts
matthewlipski: I'd rather not keep the whole browser file since we were only using it to check if `cmd` or `ctrl` should be used for keyboard shortcuts, so I'll probably just move that function to a utils file.

packages/core/src/shared/components/toolbar/SimpleToolbarButton.tsx
matthewlipski: In `HyperlinkBasicMenu.tsx`, it's used for an edit button for existing hyperlinks.

matthewlipski: Could be changed but this is how it was in the AtlasKit UI.

packages/core/src/shared/plugins/suggestion/components/SuggestionGroup.tsx
matthewlipski: I think we still want to use a separate `MantineProvider` component for the Suggestions plugin since the Menu component used in the Suggestions has different styling to e.g. the Menu component used in the Toolbar. That being said, the `MantineProvider` component should probably wrap the whole Suggestions dropdown, not just `SuggestionGroup`.

packages/core/src/BlockNoteTheme.ts
```diff
@@ -0,0 +1,131 @@
...
+          },
+        },
+      }),
+    },
+  },
+  fontFamily: "Inter",
+  primaryColor: "brandFinal",
+  primaryShade: 9,
+  other: {
+    defaultMenuStyles: (theme: MantineTheme) => ({
```
YousefED, line 123: Can't we just set this in `{ Component: Menu {...} }`?

packages/core/src/shared/components/toolbar/ToolbarDropdownItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Menu.Item
+      key={props.text}
+      onClick={props.onClick}
+      icon={ItemIcon && <ItemIcon size={16} />}
+      rightSection={
+        props.isSelected ? (
+          <TiTick size={16} />
+        ) : (
+          // Ensures space for tick even if item isn't currently selected.
```
YousefED, line 27: nicely commented :)

packages/core/src/shared/plugins/suggestion/components/SuggestionGroupItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const isKeyboardSelected =
+      props.selectedIndex !== undefined && props.selectedIndex === props.index;
+    const isMouseSelected = itemRef.current?.matches(":hover");
+
+    return isKeyboardSelected || isMouseSelected;
+  }
+
+  function updateSelection() {
+    isSelected()
+      ? itemRef.current?.setAttribute("data-hovered", "true")
```
YousefED, line 32: explain that data-hovered is mantine related

packages/core/src/shared/plugins/suggestion/components/SuggestionGroupItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  const Icon = props.item.icon;
+
+  return (
+    <Menu.Item
+      className={classes.root}
+      icon={Icon && <Icon size={18} />}
+      onClick={() => {
+        setTimeout(() => {
```
YousefED, line 61: why do all these need a setTimeout?

YousefED (commented):
Great. Some minor comments.

+ I found one small bug I think regarding the tooltips which look a bit weird

https://user-images.githubusercontent.com/368857/199564703-0954dc6d-944f-43e9-9536-19aeee9de165.mov

packages/core/src/BlockNoteTheme.ts
matthewlipski: For the primaryColor & primaryShade, they're supposed to get applied by default to stuff like buttons but it's not really working anyway so tbh I'll probably remove them.

The styles created by the defaultMenuStyles function are used in a bunch of different menus, so it's pretty much just there to reduce code duplication.

YousefED: But can't we target those menus in the normal Mantine way? `{ components: Menu {...} }` would target all Menus right? Why not add the styles to that?

matthewlipski: Yep, you're right, I thought I might be using different menu styles somewhere else but it seems like all menus use the same styles.

packages/core/src/shared/plugins/suggestion/components/SuggestionGroupItem.tsx
matthewlipski: I honestly have no idea, but for whatever reason it doesn't work properly without, even though the timeout is set to 0.

packages/core/src/BlockNoteTheme.ts
matthewlipski: Ah, I forgot that the Toolbar component a Group element instead of a Menu.dropdown like the suggestions, headings, and drag handle menus. So I can put the styles in Menu.dropdown for those but I would have to duplicate them in Toolbar.root. Should I do this or keep it as is?

YousefED: yes let's do it I think

packages/core/src/shared/plugins/suggestion/components/SuggestionGroupItem.tsx
YousefED: it works here without it, it seems? (chrome on mac). Are you sure?

matthewlipski: Oh you're right, I was looking at the onMouseLeave callback, which for sure does need it (otherwise if you hover the mouse over the item selected using the keyboard and then move the mouse elsewhere, the keyboard selection disappears). I'll remove the setTimeout from the onClick callback though👍.

YousefED:
Great stuff!

PR #56, opened 2022-10-28
Block animation fix

matthewlipski:
BlockNote keeps track of what node type a block currently is (paragraph, heading, list, etc), as well as what type it was previously. This is done using custom HTML attributes using kebab-case (e.g. `data-heading-type` and `prev-data-heading-type`) to animate when the block type changes. Previously, these attributes used improper naming conventions (actually `data-headingType` and `prev-data-headingType`). PR #46 improved the naming scheme for the current block type attribute but not the previous type, which broke animations. This PR changes the previous block type attribute to also comply with kebab-case naming conventions, and subsequently fixes the animations.

### DIFF
--- a/packages/core/src/extensions/Blocks/PreviousBlockTypePlugin.ts
+++ b/packages/core/src/extensions/Blocks/PreviousBlockTypePlugin.ts
@@ -5,7 +5,7 @@ import {
 } from "@tiptap/core";
 import { Plugin, PluginKey } from "prosemirror-state";
 import { Decoration, DecorationSet } from "prosemirror-view";
-import { BlockAttributes } from "./BlockAttributes";
+import BlockAttributes from "./BlockAttributes";
 
 const PLUGIN_KEY = new PluginKey(`previous-blocks`);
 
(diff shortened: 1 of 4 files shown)


packages/core/src/extensions/Blocks/PreviousBlockTypePlugin.ts
```diff
@@ -8,6 +8,13 @@ import { Decoration, DecorationSet } from "prosemirror-view";
 
 const PLUGIN_KEY = new PluginKey(`previous-blocks`);
 
+// Used to change HTML attribute string from camelCase to kebab-case to better follow naming conventions.
+const changeAttributeCase = (str: string) =>
```
YousefED, line 12: why do we need to changeAttributeCase? attributes will never be camelcase anymore right?

matthewlipski: TipTap node attributes still use camelCase as a convention (makes sense since hyphens are reserved symbols in JS), so they only get changed to kebab-case when rendering them to HTML (also because we prepend custom HTML attributes with `data-*`). E.g. `headingType` gets rendered out to `data-heading-type`.

TipTap internal representation:
```
{
  "type": "block",
  "attrs": {
    **"headingType": "1"**
  },
  "content": [
    {
      "type": "content",
      "attrs": {},
      "content": [
        {
          "type": "text",
          "text": "Heading"
        }
      ]
    }
  ]
}
```

Rendered HTML:
```
<div data-heading-type="1" class="..." data-node-type="block-outer">
  <div data-heading-type="1" class="..." data-node-type="block">
    <div class="..." data-node-type="block-content">
      <div>
        This is a heading
      </div>
    </div>
  </div>
</div>
```

In hindsight I should have used a regex check for all HTML attribute rendering & parsing in `Block.tsx` since it's cleaner than mapping TipTap node attributes to HTML custom attributes, but either approach works.

YousefED (approved):
Great, this is a lot clearer than earlier!!

PR #57, opened 2022-11-07
Multiple block dragging fix

### LINKED ISSUE #51: Drag multiple elements at once
It would be great to drag more than 1 block at a time. If you multi-select right now and drag, it only drags the top element.

matthewlipski:
This PR allows users to drag all blocks spanned by a selection, if the drag handle of any of those blocks is used.

Closes #51


packages/core/src/extensions/Blocks/MultipleNodeSelection.ts
```diff
@@ -0,0 +1,100 @@
+import { Selection } from "prosemirror-state";
+import { Fragment, Node, Slice } from "prosemirror-model";
+import { Mappable } from "prosemirror-transform";
+
+/**
+ * This class represents an editor selection which spans multiple nodes/blocks. It's currently only used to allow users
+ * to drag multiple blocks at the same time.
```
YousefED, line 7: explain that it's based on the original selection + url.

(i.e.: what was your reasoning / design considerations?)

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.tsx
```diff
@@ -98,8 +100,24 @@ function dragStart(e: DragEvent, view: EditorView) {
   };
   let pos = blockPosAtCoords(coords, view);
   if (pos != null) {
+    let selection = view.state.selection;
+
+    const multipleSelection = MultipleNodeSelection.create(
```
YousefED, line 105: shouldn't we only create the applicatble Selection?

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.tsx
```diff
@@ -110,6 +128,7 @@ function dragStart(e: DragEvent, view: EditorView) {
     e.dataTransfer.setData("text/plain", text);
     e.dataTransfer.effectAllowed = "move";
     const block = getDraggableBlockFromCoords(coords, view);
+    // TODO: Fix drag image
```
YousefED, line 131: is this easy to fix?

matthewlipski: As it turns out, it's quite an annoying one. The `setDragImage` function takes a single HTML element as an argument, and that element must be visible in the browser when the drag starts. With only 1 selected node that's no problem, but with multiple not so much since they're no longer a single element, and you can't use the parent element since not all children might be selected. So the only real solution is to create a div somewhere the user can't see with all the selected nodes in it, and use that for the image. It seems quite hacky to me but AFAIK there really isn't really another option.

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.tsx
```diff
@@ -96,11 +183,30 @@ function dragStart(e: DragEvent, view: EditorView) {
...
+    const doc = view.state.doc;
+
+    const { from, to } = blockPositionsFromSelection(selection, doc);
+
+    const draggedBlockInSelection = from <= pos && pos < to;
+    const multipleBlocksSelected = !selection.$anchor
+      .node()
+      .eq(selection.$head.node());
+
+    if (draggedBlockInSelection && multipleBlocksSelected) {
```
YousefED, line 199: so much clearer!

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.tsx
```diff
@@ -87,6 +73,107 @@ function getDraggableBlockFromCoords(
...
+    }
+    return desc.posBefore;
+  }
+  return null;
+}
+
+function blockPositionsFromSelection(
+  selection: Selection,
+  doc: Node
+): Record<string, number> {
```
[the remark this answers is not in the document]
YousefED: fixed

YousefED (approved):

PR #59, opened 2022-11-09
Move block `id` attribute rendering code

matthewlipski:
This PR moves code for rendering the `id` attribute of blocks to the same place as all other block-specific attributes, while it was previously done separately as part of the UniqueId extension. This should also be a more robust fix for ensuring the `id` node attribute/`data-id` HTML attribute does not get accidentally removed from blocks in the future.


YousefED:
The previous decoupling (all "id" logic in UniqueID) also had it's pros, but I think it's ok like this as well.

PR #64, opened 2022-11-16
Block Restructuring

matthewlipski:
Currently, the way block types are handled by BlockNote is through setting custom attributes on blocks. This heavily restricts how extensible blocks can be as the custom attributes only affect styling and block contents are always expected to be made up of `div`s. This PR restructures blocks so that it's possible to easily extend what kind of content they can contain.


packages/core/src/extensions/BubbleMenu/component/BubbleMenu.tsx
```diff
@@ -22,42 +22,32 @@ import { findBlock } from "../../Blocks/helpers/findBlock";
...
-    li: "Bullet List",
-    oli: "Numbered List",
-  };
-  // A heading that's also a list, should show as Heading
-  if (currentBlockHeading) {
-    return headings[currentBlockHeading - 1];
-  } else if (currentBlockListType) {
-    return lists[currentBlockListType];
-  } else {
+function getBlockName(blockContentNode: Node) {
```
YousefED, line 27: we should find a way to define this as part of the block / node. But can also do that when we work on the Block API

packages/core/src/extensions/DraggableBlocks/components/DragHandle.tsx
```diff
@@ -20,59 +16,28 @@ export const DragHandle = (props: {
...
-        props.view.state.tr.setSelection(
-          new TextSelection(props.view.state.tr.doc.resolve(endOfBlock + 1))
-        )
-      );
-    }
-    // Focus and activate slash menu
-    props.view.focus();
-    props.view.dispatch(
-      props.view.state.tr.scrollIntoView().setMeta(SlashMenuPluginKey, {
+    if (props.onAddClicked) props.onAddClicked();
```
YousefED, line 29: nice cleanup!!

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -173,264 +123,388 @@ export const Block = Node.create<IBlock>({
...
-            // BlockA
-            //     BlockB
-            //
-            // joinBackward has been patched with our custom version to fix this (see commands/joinBackward)
-            .joinBackward()
-            .run(),
-
-        () => commands.selectNodeBackward(), // (source: tiptap)
+        // Regular backspace behaviour.
+        () => commands.deleteSelection(),
```
YousefED, line 410: shouldn't deleteSelection still come first?

YousefED (commented):
Looks like a great improvement, but still some quirks to iron out. I'm running into some issues in the vercel build:

- list items are quite funky on chrome I think 
- I also got into a state where nothing happens when clicking +. Not sure how I can reproduce 
- pressing Tab focusses the URL bar of the browser

I think many of these are related to the commands in AddKeyboardShortcuts. Maybe we should go over these changes together?

I've also added some other thoughts / suggestions in this review.

packages/core/src/extensions/BubbleMenu/component/BubbleMenu.tsx
matthewlipski: This function can be replaced by `getBlockInfoFromPos`, but I think we also want to find a better way of integrating that function into the block code.

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Actually seems like this is redundant anyway, since I guess ProseMirror just uses the default backspace handler if none of your custom handlers return true.

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -173,264 +123,388 @@ export const Block = Node.create<IBlock>({
   addInputRules() {
     return [
       ...["1", "2", "3"].map((level) => {
-        // Create a heading when starting with "#", "##", or "###""
-        return textblockTypeInputRuleSameNodeType({
-          find: new RegExp(`^(#{1,${level}})\\s$`),
-          type: this.type,
-          getAttributes: {
-            headingType: level,
+        // Creates a heading of appropriate level when starting with "#", "##", or "###".
```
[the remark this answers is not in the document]
matthewlipski: Oh, yep you're right. For some reason I thought input rules declared in a node type only trigger if the selection is in that same node type, but actually they trigger regardless of what node you're in.

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -173,264 +123,388 @@ export const Block = Node.create<IBlock>({
...
+        return new InputRule({
+          find: new RegExp(`^(#{${parseInt(level)}})\\s$`),
+          handler: ({ state, commands, range }) => {
+            commands.BNSetContentType(state.selection.from, "headingBlock", {
```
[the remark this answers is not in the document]
matthewlipski: Yeah true, mixing commands and transactions is pretty bad practice so will fix. IMO using `BNSetContentType` is a bit cleaner and saves you from having to go to a different file to see what it does, and seems to work the same.

packages/core/src/extensions/Blocks/nodes/ListItemBlock.ts
```diff
@@ -0,0 +1,246 @@
...
+        "data-node-type": "block-content",
+        "data-content-type": this.name,
+        class: styles.blockContent,
+      },
+      ["li", 0],
+    ];
+  },
+});
+
+// ProseMirror Plugin which automatically assigns
```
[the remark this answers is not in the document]
matthewlipski: Yeah probably a good idea. I'm still not sure about the file structure since text and headings only need 1 file while lists also need a plugin which should probably go in a separate file. So it seems a bit clunky to have a separate directory for each block type since some will only have 1 file, but I'm also not a fan of having both nodes and plugins defined in the same directory.

packages/core/src/extensions/Blocks/nodes/HeadingBlock.ts
```diff
@@ -0,0 +1,65 @@
...
+  addAttributes() {
+    return {
+      level: { default: "1" },
+    };
+  },
+
+  addNodeView() {
+    return (props: NodeViewRendererProps) => {
+      const element = document.createElement("div");
+      element.setAttribute("data-node-type", "block-content");
```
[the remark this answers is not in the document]
matthewlipski: Honestly I need to see if we even check for `block-content` anywhere. This and the `content-type` attributes probably need a bit of cleaning up, since it's likely we don't need both of them.

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -46,38 +43,34 @@ declare module "@tiptap/core" {
...
+
   addOptions() {
     return {
       HTMLAttributes: {},
     };
   },
 
   // A block always contains content, and optionally a blockGroup which contains nested blocks
-  content: "content blockgroup?",
+  content: "(textBlock | headingBlock | listItemBlock) blockgroup?",
```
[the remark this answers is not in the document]
matthewlipski: I think it's good to keep referring to blocks the same way we do now (i.e. the nodes which have drag handles, ids, etc) and change the naming of the inner blocks (previously block content) to something else since now they're also called xyzBlock which is confusing.

packages/core/src/extensions/Blocks/nodes/HeadingBlock.ts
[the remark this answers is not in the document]
matthewlipski: So it seems like both attributes are only really uesful for styling, might be good to discuss later what the plan is for that.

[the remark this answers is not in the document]
matthewlipski: Another thing to add is that `block-content` isn't that useful even for styling since `.block > div:first-child` also works, though at the expense of readability.

packages/core/src/extensions/BubbleMenu/component/BubbleMenu.tsx
matthewlipski: I think we should wait until we start on the API, since `getBlockInfoFromPos` is another useful function defined outside `Block.tsx` that we want users to have access to, and I'm not sure yet where to put it either.

packages/core/src/extensions/Blocks/nodes/ListItemBlock.ts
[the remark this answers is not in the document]
YousefED: Let's create a separate directory per block type for now

packages/core/src/extensions/Blocks/nodes/HeadingBlock.ts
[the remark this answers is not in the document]
YousefED: Styling uses classnames, I think `data-node-type` is only used for testing right? **if** that's the case I'd rather have "data-test" as prefix

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -46,38 +43,31 @@ declare module "@tiptap/core" {
...
         default: undefined,
       },
-      headingType: {
-        default: undefined,
-        keepOnSplit: false,
-      },
     };
   },
 
   parseHTML() {
```
YousefED, line 69: maybe remove this? Let's do it after tests work, and also create a test for pasting a <span> or other unsupported tag

packages/core/src/extensions/Blocks/PreviousBlockTypePlugin.ts
```diff
@@ -43,6 +43,10 @@ export const PreviousBlockTypePlugin = () => {
       },
 
       apply(transaction, prev, oldState, newState) {
+        if (transaction.getMeta("orderedListIndexing") !== undefined) {
```
YousefED, line 53: this check might not be necessary anymore I think

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -130,10 +130,10 @@ export const Block = Node.create<IBlock>({
 
           return true;
         },
-      // Deletes a block at a given position.
+      // Deletes a block at a given position and sets the selection to where the block was.
       BNDeleteBlock:
         (posInBlock) =>
-        ({ state }) => {
+        ({ state, view }) => {
```
YousefED, line 136: I think if we want to do things the tiptap / PM way, we should check for "dispatch" and don't issue the transactions if false?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
+            new TextSelection(state.doc.resolve(startPos + 1))
+          );
+          view.focus();
+
+          return true;
         },
+      // Appends the text contents of a block to the nearest previous block, given a position between them. Children of
+      // the merged block are moved out of it first, rather than also being merged.
+      //
+      // In the example below, the position passed into the function is between Block1 and Block2.
```
YousefED, line 155: don't you mean betwee Block2 and Block3?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
   addKeyboardShortcuts() {
     // handleBackspace is partially adapted from https://github.com/ueberdosis/tiptap/blob/ed56337470efb4fd277128ab7ef792b37cfae992/packages/core/src/extensions/keymap.ts
     const handleBackspace = () =>
       this.editor.commands.first(({ commands }) => [
-        // Maybe the user wants to undo an auto formatting input rule (e.g.: - or #, and then hit backspace) (source: tiptap)
+        // Deletes the selection if it's not empty.
+        () => commands.deleteSelection(),
+        // Undoes an input rule if one was triggered in the last editor state change.
         () => commands.undoInputRule(),
-        // maybe convert first text block node to default node (source: tiptap)
```
YousefED, line 293: this functionality was removed, it seems?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
-
-            if (
-              !empty ||
-              !isAtStart ||
-              !parent.type.isTextblock ||
-              parent.textContent.length
-            ) {
-              return false;
+          commands.command(({ state }) => {
+            const { depth } = getBlockInfoFromPos(
```
YousefED, line 335: should we get doc and selection from `state` or from `tr`? My guess is the latter, just how it was before

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
+              state.doc,
+              state.selection.from
+            )!;
+
+            const selectionAtBlockStart =
+              state.selection.$anchor.parentOffset === 0;
+            const selectionEmpty =
```
YousefED, line 342: this can be removed, because we would have executed `deleteSelection` if not empty

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
+              state.selection.anchor === state.selection.head;
+            const blockIndented = depth > 2;
```
YousefED, line 344: Do we need this check? My guess is `liftListItem` would return false if it can't lift the current block

packages/core/src/extensions/Blocks/nodes/Block.ts
YousefED: It also doesn't work, compare with main branch when you backspace at start of  oldparent

image

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
             return false;
           }
 
-          // If current blocks content is empty dont create a new block
-          if (currentBlock.node.firstChild?.textContent.length === 0) {
-            if (dispatch) {
-              tr.setNodeMarkup(currentBlock.pos, undefined, attributes);
+          // Finds the nearest previous block, prioritizing higher nesting levels.
+          while (prevBlockInfo.numChildBlocks > 0) {
+            prevBlockEndPos--;
```
YousefED, line 204: this seems inefficient?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
-        () => commands.splitListItem("block"),
-        // Otherwise, maybe we are in an empty list item. "Enter" should remove the list bullet
-        ({ tr, dispatch }) => {
-          const $from = tr.selection.$from;
-          if ($from.depth !== 3) {
-            // only needed at root level, at deeper levels it should be handled already by splitListItem
+        // Removes a level of nesting if the block is empty & indented, while the selection is also empty & at the start
+        // of the block.
+        () =>
+          commands.command(({ state }) => {
```
YousefED, line 387: can't we remove this (387-409), and use the code from before the change? `() => commands.splitListItem("block")`

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -162,275 +109,408 @@ export const Block = Node.create<IBlock>({
...
+                .run();
+
+              return true;
             }
-            return true;
-          }
-          return false;
-        },
-        // Otherwise, we might be on an empty line and hit "Enter" to create a new line:
-        ({ tr, dispatch }) => {
```
YousefED, line 409: Can't we use this code instead of line 414-461?

packages/core/src/extensions/Blocks/helpers/getBlockInfoFromPos.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @param doc The ProseMirror doc.
+ * @param posInBlock A position somewhere within a block node.
+ * @returns A BlockInfo object for the block the given position is in, or undefined if the position is not in a block
+ * for the given doc.
+ */
+export function getBlockInfoFromPos(
+  doc: Node,
+  posInBlock: number
+): BlockInfo | undefined {
+  const maxDepth = doc.resolve(posInBlock).depth;
```
YousefED, line 25: Let's do `doc.resolve` only once?

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Yep, you're right, position should be between Block2 and Block3. Not sure why it's not working for you though, I tried it just now and the behavior for me is the same as on main. What does it do for you?

YousefED: hmm weird, works now. I'll let you know if I come across it again

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: I don't really understand what this does. As far as I understand, calls `clearNodes` if the selection is empty and at the start of the doc, and if the block it's in is empty. But TipTap docs say that `clearNodes` changes all node types to "what’s in the content attribute of your Document", which in our case is a blockGroup, so I'm not sure what effect it's supposed to have on the editor.

YousefED: I think this:
firstbackspace

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Yep, you're absolutely right, not sure why I overcomplicated this so much.

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: To some extent, but we still need to make sure the selection is empty and at the start of an empty block. The enter behavior is a little different to before since before it didn't consider whether or not the selection is empty. Now, if the selection isn't empty, the content spanned by it is deleted first and then the Enter keypress is handled normally, which is what happens in most other editors. 

So for example ([] denotes selection):

12[3
qw]e

Old:
12
3
qwe

New:
12
e

This new behavior is what lines 442-460 are for.

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: That's a fair point, I added quite a few of these checks/conditions to the keyboard commands since it was easier for me to follow along with what's happening and catch bugs, though it means some of them are redundant. I do still think it helps with readability and the performance hit is negligible, but I'll remove them if you'd rather have less code.

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Basically I can use the transaction instead of the chained commands but the checks are still necessary, unless I move lines 442-460 to the top and add a selection not empty check.

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Yep, agreed, looking deeper into the PM docs it does seem to add functionality that we want.

packages/core/src/extensions/Blocks/nodes/Block.ts
YousefED: I think it comes from `state` so probably it doesn't really matter: https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/CommandManager.ts

packages/core/src/extensions/Blocks/nodes/Block.ts
YousefED: Let's comment them out for now, to make sure it doesn't conflict with anything. If it's ok we can remove it later.

packages/core/src/extensions/Blocks/nodes/Block.ts
YousefED: Maybe we should discuss this together

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Aha, the old implementation isn't compatible with the animations as it doesn't preserve block ID. Is that worth the extra complexity?

matthewlipski:
Summary of `joinBackward` functionality vs `BNMergeBlocks`:
- `joinBackward` actually does more than just joining 2 nodes together, some of which is already handled by other functions and beyond the scope of what we use `BNMergeBlocks` for (just backspace handling at the start of a block).
- `joinBackward` will lift a node if there are no nodes before it (not needed).
- `joinBackward` will also lift a node if it cannot be joined in order to move it closer to the one before. This was previously useful to unindent empty blocks but does not preserve IDs, and is now handled by another function (not needed).
- `joinBackward` has edge cases for isolating and atomic nodes. Atomic nodes might come up when dealing with custom blocks, but the implementation might not be what we want. If the previous block contains an image and the current block contains text, for example, `joinBackward` takes the GDocs approach and removes the image. Notion, however, joins the current block with the first previous text block it finds, so it effectively ignores images. Not sure which approach we want, but it's not relevant for this PR since image/custom/non-text blocks aren't implemented yet (not needed right now).
- `joinBackward` checks if the nodes being joined have compatible content types. Again, we only have text blocks right now so this doesn't do anything, but will be useful when we implement image/custom/non-text blocks (not needed right now).

I think that for this PR, `BNMergeBlocks` does everything that we need from `joinBackward`, but we'll still have to change it once we start implementing other block types and a custom block API. `joinBackward` would maybe still be fine to use with non-text blocks, but it's hard to say at the moment ,and I think theres more value in sticking with `BNMergeBlocks` since it'll make working with the block code easier long term.

packages/core/src/extensions/Blocks/PreviousBlockTypePlugin.ts
```diff
@@ -93,21 +91,40 @@ export const PreviousBlockTypePlugin = () => {
...
               const oldAttrs = {
-                // listType: oldNode.node.attrs.listType,
+                listItemType: oldContentNode.attrs.listItemType,
+                listItemIndex: oldContentNode.attrs.listItemIndex,
                 headingLevel: oldContentNode.attrs.headingLevel,
                 type: oldContentNode.type.name,
                 depth: oldState.doc.resolve(oldNode.pos).depth,
               };
 
+              // Hacky fix to avoid processing certain transactions created by ordered list indexing plugin.
```
YousefED, line 109: Can't we re include the above then?

```
if (transaction.getMeta("orderedListIndexing") !== undefined) {
          return prev;
        }
```

YousefED: If we need to keep it, can you comment (explain what they do) indexInitialized / depthChanged / update?

packages/core/src/extensions/Blocks/nodes/Block.ts
YousefED: We were not able to find a built-in Tiptap / PM function for this?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -241,46 +254,54 @@ export const Block = Node.create<IBlock>({
...
-              newBlockContentPos,
-              state.schema.node(contentType).type,
-              contentNode.attrs
-            );
+          if (dispatch) {
+            state.tr.insert(newBlockInsertionPos, newBlock);
+            state.tr.insertText(secondBlockContent, newBlockContentPos);
+
+            if (keepType) {
+              dispatch(
```
YousefED, line 262: wrong use of dispaych

packages/core/src/extensions/Blocks/PreviousBlockTypePlugin.ts
matthewlipski:
```
if (transaction.getMeta("orderedListIndexing") !== undefined) {
    return prev;
}
```

This prevents all transactions from the `OrderedListItemIndexPlugin`, but we actually want to process some of them for the animations to work, which are let through by `indexInitialized` and `depthChanged`.

matthewlipski:
[Referring to this](https://github.com/YousefED/BlockNote/pull/64#discussion_r1040848540)
Yes, we looked at using `findParentNodeClosestToPos` but found out that it has a similar implementation to `getBlockFromPos` (i.e. it loops and gets a block at a higher depth each iteration), so IIRC we decided not to change it since it didn't really make it more efficient.

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Since copy/pasting from other sites doesn't work very well after the restructuring, I think we should just leave it as is for now and try to fix the issue later, since it's quite a bit more complex than originally expected.

PR #68, opened 2022-11-22
upgrade to react 18

### LINKED ISSUE #66: TailwindCSS and React18 Support
Hey folks awesome project. 

Really, liked the notion style UI and the way it's all coming together.

Is there a way to make this work on React 18? Right now it's erroring a bit since it depends on ReactDom.render();

Also TailwindCSS support would be very much appreciated. Though, do love mantine.

YousefED:
upgrade to react 18, closes #66

We could choose to keep react 17 compatibility, but as this library is in it's infancy, let's upgrade to v18 until we're sure non-17 compatibility is a breaking issue for users


YousefED:
@matthewlipski are you sure the move of deps fixes the build? I think the correct way would still be that React should be peerDeps instead of deps, and it seems like the failure of the test was due to a playwright error, not a build error, right?

matthewlipski:
> @matthewlipski are you sure the move of deps fixes the build? I think the correct way would still be that React should be peerDeps instead of deps, and it seems like the failure of the test was due to a playwright error, not a build error, right?

Turns out I was wrong about playwright being the issue. The actual issue is a react version mismatch in the built environment, for which adding react dependencies in blocknote/core seems to fix. But you're right, it should read the peerDeps and figure out that react 18 is a requirement, yet after building, package-lock.json still includes a dependency on react 17 for blocknote/core. I don't know much about dependency management, so I'm not sure why this is happening, and this fix doesnt make much sense to me (since react isnt needed in blocknote/core), but it does indeed cause react 17 to stop showing up in package-lock.json.

YousefED:
closed by https://github.com/YousefED/BlockNote/pull/74

PR #72, opened 2022-12-09
`BNSplitBlock` fix

matthewlipski:
Since the `BNSplitBlock` function only copies text content to the new block, if the block being split has nested blocks, they aren't preserved and their text is just concatenated together. This PR fixes the issue, and now uses `Split` objects to replace content in both the original and new block, whereas before they were only used for the original block.


YousefED:
Nice, is there a test for this as well?

PR #74, opened 2022-12-13
Separate React package

matthewlipski:
BlockNote is currently written for React, though the `core` package should be decoupled from our implementation of the UI so it can be more easily integrated into non-React projects. This PR moves all React-based functionality to a new `react` package, namely the custom `useEditor` hook and all menu implementations (bubble menu, suggestions menu, drag handle menu, hyperlink menus). This will also allow users to add their own custom UI to BlockNote.


examples/editor/src/main.tsx
```diff
@@ -1,14 +1,24 @@
-import React from "react";
-import ReactDOM from "react-dom";
-import App from "./App";
 import "./index.css";
+import styles from "./App.module.css";
+import { mountBlockNoteEditor } from "../../../packages/core/src/BlockNoteEditor";
```
YousefED, line 3: imports should reference @blocknote/core or @blocknote/react.

examples/editor/src/main.tsx
```diff
@@ -1,14 +1,24 @@
...
+// type WindowWithProseMirror = Window &
+//   typeof globalThis & { ProseMirror: Editor };
 
-ReactDOM.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-  document.getElementById("root")
-);
+mountBlockNoteEditor({
```
YousefED, line 9: If we want this example, to be a "react" example, then I think the previous code, with `<App />` and `useEditor` and `<EditorContent>` is better

packages/react/src/BubbleMenu/BubbleMenuFactory.tsx
```diff
@@ -0,0 +1,52 @@
+import { Editor, isNodeSelection, posToDOMRect } from "@tiptap/core";
+import { createRoot } from "react-dom/client";
+import { MantineProvider } from "@mantine/core";
+import { BlockNoteTheme } from "../../../core/src/BlockNoteTheme";
+import { BubbleMenu } from "./components/BubbleMenu";
+import tippy from "tippy.js";
+import rootStyles from "../../../core/src/root.module.css";
+
+export const BubbleMenuFactory = (editor: Editor) => {
```
YousefED, line 9: should this just return an element? Or an object with methods like: updatePosition, show, close, etc? Or is this still WIP / we'll figure this out later?

examples/editor/src/main.tsx
matthewlipski: Would it would make sense to have 2 examples, one for vanilla JS and one for React?

examples/editor/src/main.tsx
matthewlipski: I might need your help with @blocknote/react since AFAIK it has to be published to npm, but I can change the local imports from @blocknote/core to npm imports now.

packages/react/src/BubbleMenu/BubbleMenuFactory.tsx
matthewlipski: I was also thinking about that, but I'm still not sure what the best approach regarding menu factory parameters and return types is.

That's something I experimented with in `MenuFunctions`, which would get passed to a factory and provide all the functions necessary for someone's custom UI to figure out on its own what & when to update. This approach IMO gives users much more flexibility with the UI.

The main concern I have is figuring out all the functions users might want when making their UI. For example, if someone wanted to display the names of editor's extensions for some reason, it doesn't really make sense to add a function which gets the names of the editor's extensions to `MenuFunctions` just for this niche use case. If we want to allow that level of customization, we'd be better off passing the whole editor instance to the custom UI like we do now.

It's a similar situation with that the factory should return. If the user wants to make a static menu, for example, it doesn't make sense for BlockNote to decide when the menu should show/hide.

So really, both the factory parameters and return types depend on how much flexibility we want to give the user. On one hand we can pass the editor instance as an argument and expect to receive a HTML element, letting (but also forcing) the user to decide how the menu should behave. On the other, we pass nothing and expect to receive not only a HTML element, but also show/hide/update functions for it, leaving it entirely up to the `BubbleMenuPlugin` to decide how the menu should behave.

matthewlipski: I think this is something we should discuss once the PR is closer to being finished, but it's also possibly something we might want to discuss with the Fermat guys.

YousefED: Let's now focus on the scenario where someone wants to build the existing functionality, but without React. That way we have a "locked" set of things that should be necessary. If then later we notice a lot more different customizations are needed, we can implement a more versatile solution like "editor" integration etc.

imo the show / hide / update functions are necessary in this case for clients to show the menu when text is highlighted, and removed when the focus changes. You're right that this might be different if creating a menu that is always visible, I'm wondering if we can think of an API that works for both (maybe "onSelectionChanged" or onBlur or sth, not sure what makes sense exactly

matthewlipski: In that case, can we just use a regular class to make the object:
```
{
    element: HTMLElement,
    show: () => void,
    hide: () => void,
    update: (params: BubbleMenuUpdateParams) => void
}
```
instead of a factory? Since we don't need to redefine `show`, `hide`, and `update`, we just potentially need to create, destroy, or update `element`.

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
```diff
@@ -217,52 +222,131 @@ function dragStart(e: DragEvent, view: EditorView) {
...
+  // Declares additional listeners to attach to the drag handle element for drag/drop & menu opening.
+  const dragStartCallback = (event: DragEvent) =>
+    dragStart(event, options.editor.view);
+  const dragEndCallback = (_event: DragEvent) => unsetDragImage();
+  const clickCallback = (_event: MouseEvent) => {
+    dragHandleMenu.show(dragHandleMenuParams);
+    blockButtonsFrozen = true;
+  };
+
+  function addDragHandleListeners() {
```
YousefED, line 318: I would remove the dependency on `element` here, but pass `dragStartCallback` and similar handlers to the factory. It's then the responsibility of the implementation (e.g.: blocknote/react) to attach the event listeners and call the callbacks

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
```diff
@@ -341,33 +421,43 @@ export const createDraggableBlocksPlugin = (editor: Editor) => {
...
-          );
+          // Gets bounding box of relevant block.
+          const blockBoundingBox = blockContent.getBoundingClientRect();
+          blockBoundingBox.x = horizontalPosAnchoredAtRoot
+            ? getHorizontalAnchor()
+            : blockBoundingBox.left;
+
+          // Updates element params.
+          addBlockButtonParams.addBlock = () =>
+            addBlock({
```
YousefED, line 432: I don't think the functions addblock / deleteblock should depend on coords. We already know which block it is right? Can't we pass the block or blockid?

YousefED (commented):
My main feedback on the draghandles is that I think it should be 1 Factory (I'd suggest the name `BlockHoverSideMenuFactory` and rename things from `dragHandleMenu` to `hoverSideMenu`.

I think it's the responsibility of the UI layer (blocknote/react) to determine how to render the buttons, and how / if a menu should expand. No need to have this logic in core

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
matthewlipski: Having the block's underlying PM node itself doesn't really help here since the actual editor functions to add and delete blocks (`BNCreateBlock()` and `BNDeleteBlock()`) take in a position inside a block as an argument. Since PM doesn't treat blocks as unique objects, even if we had the PM node, we would have to find it in the document, which wouldn't be any more efficient. So `view.posAtCoords()` turns out to be the easiest way of getting the correct position.

packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
```diff
@@ -85,6 +95,111 @@ export function findCommandBeforeCursor(
...
+          range: { from: 0, to: 0 },
+        });
+      },
+      queryStartBoundingBox: new DOMRect(),
+      editorElement: this.editor.options.element,
+    };
+  }
+
+  updateSuggestionsMenuParams(pluginState: any) {
+    this.suggestionsMenuParams.items = pluginState.items;
```
YousefED, line 182: @matthewlipski I don't think there's use to store this as `suggestionsMenuParams`. You can just return it and use that. The local state (private field) is not used and can only cause confusion

packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
```diff
@@ -85,6 +95,111 @@ export function findCommandBeforeCursor(
...
+      this.suggestionsMenu.show(this.suggestionsMenuParams);
+
+      // Listener stops focus moving to the menu on click.
+      this.suggestionsMenu.element!.addEventListener("mousedown", (event) =>
+        event.preventDefault()
+      );
+    }
+  }
+
+  initSuggestionsMenuParams() {
```
YousefED: Ideally, I think we should also get rid of this. This relates to a bigger issue with the Factories; we now need code to initialize "fake data", which is not used anyway. We discussed this before, but while review this (plugin -> class conversion) it now becomes clear it's annoying. More a question of _when_ than _if_ we want to fix it imo

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
```diff
@@ -237,14 +238,13 @@ export class BlockMenuView {
   // When false, the drag handle with be just to the left of the element
   horizontalPosAnchoredAtRoot: boolean;
 
-  blockMenuParams: BlockSideMenuParams;
   blockMenu: BlockSideMenu;
 
+  hoveredBlock: HTMLElement | undefined;
```
YousefED, line 243: I'm not sure how safe this is. We call `getBoundingClientRect` at a moment that we're not sure thism element still exists. Consider the following scenario:

- hover element
- click draghandle
- a different plugin, let's say a spellchecker or whatsoever, changes an attribute of the block
-  now we click "add"

I think this won't work, as the htmlelement of the block can have changed

YousefED (approved):
Nice. 1 comment, which I don't think is high prio. I also think we need to put some extra thought on API design (setBlockType). Will discuss on discord

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
matthewlipski: I think the safest way to do it would be to save a hoveredBlockId instead and each time we need to get the block HTML element & bounding box for a function we could call:
```
document.querySelectorAll('[data-id=this.hoveredBlockId]')
```
This would imo be relatively clean but pretty inefficient.

We could also save the hovered block's bounding box instead which would be nicer and more efficient, but could still cause problems if another extension changes e.g. the block type in the meantime.

Any thoughts on which solution is better?

YousefED: Maybe the easiest is to stick to the current solution, but use your `document.querySelectorAll('[data-id=this.hoveredBlockId]')` idea if `hoveredBlockId.isConnected` is false?

PR #81, opened 2023-01-17
Suggestion plugin refactor

matthewlipski:
This PR refactors the suggestions plugin to hopefully make it more readable and provide better documentation.

### DIFF
--- a/packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
+++ b/packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
@@ -325,8 +325,10 @@ export function createSuggestionPlugin<T extends SuggestionItem>({
 
     props: {
       handleKeyDown(view, event) {
-        // Shows the menu if the default trigger character was pressed.
-        if (event.key === defaultTriggerCharacter) {
+        const menuIsActive = (this as Plugin).getState(view.state).active;
+
+        // Shows the menu if the default trigger character was pressed and the menu isn't active.
+        if (event.key === defaultTriggerCharacter && !menuIsActive) {
           view.dispatch(
             view.state.tr
               .insertText(defaultTriggerCharacter)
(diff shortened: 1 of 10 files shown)


packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
```diff
@@ -382,57 +325,72 @@ export function createSuggestionPlugin<T extends SuggestionItem>({
...
-                .setMeta(pluginKey, { activate: true, type: "slash" })
-            );
-            // return true to cancel the original event, as we insert / ourselves
-            return true;
-          }
-        } else {
-          const { items, range, selectedItemIndex } = pluginKey.getState(
-            view.state
+        // Shows the menu if the default trigger character was pressed.
+        if (event.key === defaultTriggerCharacter) {
```
YousefED, line 329: minor detail, I think we should only do this when not active (now, the menu moves to the right when typing multiple slashes)

YousefED (changes requested):
Nice. 1 comment and 1 bug I found:

- The menu doesn't fade out anymore (it does on main)

packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
matthewlipski: Yeah I think you're right. I originally preferred this as with the current implementation, writing more "/" characters doesn't filter the items and you can do it endlessly. I'll change it so it's more in line with what Notion does, i.e. treat "/" characters as regular characters when the menu is open, so typing many in a row will return no items and eventually close the menu.

matthewlipski: Also the fade animation is a weird one, but I think I figured it out. The problem is two-fold:

1. We show and hide the React UI but mounting/unmounting the menu components and adding/removing the root element from the DOM, so the component doesn't animate when hiding because it just gets removed from the DOM straight away.
2. The animation is set by Tippy. While we used Tippy before and still use it now, the way we create a Tippy tooltip for the menus to go in is different. Before, we used the `tippy()` function to setup the tooltip for each menu to go in, whereas now we use the `<Tippy />` React component.

matthewlipski:
> Nice. 1 comment and 1 bug I found:
> 
> * The menu doesn't fade out anymore (it does on main)

Fixed in PR #87

PR #86, opened 2023-01-20
Block API Improvements

matthewlipski:
This PR improves on BlockNote editor functions that get exposed to APIs such as the editor element factory API. It mainly simplifies the naming conventions for block content types, adds a shorthand way of setting a block's content type, and enforces stricter typing when setting a block's content type. This should lead to a more intuitive developer experience.


packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -292,13 +311,16 @@ export const Block = Node.create<IBlock>({
           }
 
           const { startPos, contentNode } = blockInfo;
+          const contentTypeName = typeof type !== "string" ? type.name : type;
+          const contentTypeAttrs =
```
YousefED, line 315: This means type attrs will always be optional right? So for example, just passing "heading" will be valid, and we'll probably set it to heading level 1?

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -496,8 +518,16 @@ export const Block = Node.create<IBlock>({
     return {
       Backspace: handleBackspace,
       Enter: handleEnter,
-      Tab: () => this.editor.commands.sinkListItem("block"),
-      "Shift-Tab": () => this.editor.commands.liftListItem("block"),
+      // Always returning true for tab key presses endures they're not captured by the browser. Otherwise, they blur the
```
YousefED, line 521: typo, "ensures"

packages/core/src/extensions/Blocks/nodes/Block.ts
```diff
@@ -506,49 +536,49 @@ export const Block = Node.create<IBlock>({
...
         ),
       "Mod-Shift-7": () =>
         this.editor.commands.BNSetContentType(
           this.editor.state.selection.anchor,
           {
-            name: "listItemContent",
+            name: "listItem",
             attrs: {
-              listItemType: "unordered",
+              ordered: "false",
```
YousefED, line 571: Is there a specific reason you went from "type" to "ordered"?

I think "type" makes more sense, as we might also have a "check" type for a check (todo) list, etc.

packages/core/src/extensions/Blocks/nodes/BlockTypes/BlockContentType.ts
```diff
@@ -0,0 +1,7 @@
+export type BlockContent<
```
YousefED, line 1: I think the following names make more sense:

`BlockType`

and then `HeadingType`, `HeadingBlockType`. Or maybe `HeadingBlock`. this one is more difficult

packages/core/src/extensions/FormattingToolbar/FormattingToolbarFactoryTypes.ts
```diff
@@ -23,7 +23,7 @@ export type FormattingToolbarDynamicParams = {
   // BlockContentType is mostly used to set a block's type, so the attr field is optional as block content types have
   // default values for attributes. However, it means that a block type's attributes field will never be undefined due to
   // these default values, which the Required type enforces.
-  activeBlockType: Required<BlockContentType>;
+  activeBlockType: Required<BlockContent>;
```
YousefED, line 26: Hmm, I'm wondering about the comment above. If I understand it correctly, then I think we're using types a bit wrong here.

I understand:
- `BlockContent` marks `attrs` as optional, because it's mostly used for setting
- `Required<BlockContent>` is the actual type, because the attrs should always be set.

Based on this, I think we should revise this, and make sure the type `BlockContent` is always "complete". e.g.: it sounds like we both think a `heading` without a `level` is not a valid type. Then this should be reflected in the HeadingBlockType. 

If you want to make it possible to create a `HeadingBlockType` without specifying a `level`, then that concern should be handled somewhere that has that specific responsibility (to create a `HeadingBlockType` with a sensible default level. We should not be passing semi-complete types around in the rest of the code base (and then at some point mark them `Required` when needed.

YousefED (changes requested):

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: Yeah the headings and list items have default attributes, they're defined in `Heading.ts` and `ListItem.ts` (level: 1 for headings and ordered: false for list items).

packages/core/src/extensions/Blocks/nodes/Block.ts
matthewlipski: The reason is because the naming clashes with the type parameter, and also because I think there could be a lot of attributes for other block types in the future which would also want to use a type attribute, so imo it's good to have it be called something unique to list items. Though the checked lists are a good point. Maybe `listType` wouldn't be bad?

packages/core/src/extensions/Blocks/nodes/BlockTypes/BlockContentType.ts
matthewlipski: I thought about this for a bit and I think what I like best is having a `BlockContainer` (currently `Block`) and `BlockContent` (currently also `BlockContent`, and `Heading`/`ListType`/`Paragraph` are types of `BlockContent`). I think they best describe what the nodes actually do and also best describe what we refer to as a block, as imo they're all parts of a block, not a full block themselves. Thoughts on this?

packages/core/src/extensions/FormattingToolbar/FormattingToolbarFactoryTypes.ts
matthewlipski: I just realized that since we can pass in either `BlockContent` or `BlockContent["name"]`, having the attributes field be optional is totally redundant. E.g. if we want to make a heading with default attributes we can just call `BNSetContentType(pos, "heading")` so giving users the option to do the same thing with `BNSetContentType(pos, { name: "heading" })` is pretty pointless.

matthewlipski: Another update, the original idea I had was flawed anyway since the `attrs` field actually needs to be `Partial`, not optional, when setting a block type. Getting a block type should still always return the full `attrs`. The only way I've found of doing this is making a second, settable type with `Partial<Attrs>` for `Paragraph`s, `Heading`s, and `ListItem`s, then declaring a new type like `type SettableBlockContent = SettableParagraph | SettableHeading | SettableListItem`, but this results in a huge amount of duplicated code. I can't do something similar after already declaring `type BlockContent = Paragraph | Heading | ListItem` as it will allow type names and attributes to be mixed and matched (e.g. it won't warn you if you try pass `{ name: "paragraph", attrs: { level: "1" }}`).

packages/core/src/extensions/Blocks/nodes/BlockContent/ListItemBlockContent/NumberedListItemBlockContent/NumberedListItemBlockContent.ts
```diff
@@ -0,0 +1,95 @@
...
+          }
+
+          // Case for BlockNote list structure.
+          if (parent.getAttribute("data-content-type") === "numberedListItem") {
+            return {};
+          }
+
+          // Case for regular HTML list structure.
+          if (parent.tagName === "OL") {
+            return { ordered: "true" };
```
YousefED, line 75: this needs to be changed?

packages/core/src/extensions/UniqueID/UniqueID.ts
```diff
@@ -75,6 +75,12 @@ const UniqueID = Extension.create({
         attributes: {
           [this.options.attributeName]: {
             default: null,
+            parseHTML: (element) =>
```
YousefED, line 78: why is this added here?

matthewlipski: Without it, we have to use `BlockAttributes` to translate node attributes to HTML attributes, which can mess with `blockContainer` attributes and should really just be done using the `parseHTML` & `renderHTML` functions. Adding those to `UniqueID` greatly simplifies the `renderHTML` function of `blockContainer` and makes it less prone to causing bugs if new attributes are added in the future.

YousefED (approved):

PR #87, opened 2023-01-20
React editor element factory overhaul

matthewlipski:
This PR rewrites the editor element factories in the `react` package. The actual elements (menus & toolbars) remain the same, only the way they are created and rendered/hidden is changed. These changes come with a few notable improvements:

1. When an element is shown or hidden, the React component responsible for it is no longer unmounted. Instead, the `Tippy` component that wraps the element, which was already being used to handle where in the viewport the menu should be shown, is now also used to actually show/hide the menu. This results in:
a. No longer having to create a new React root, `MantineProvider` component, and `Tippy` component each time an element is shown, which increases performance.
b. The fade out animation working again, since that's a feature of Tippy, which now handles hiding elements.

2. Creation of the `ElementFactory` function in the `react` package, which allows the same boilerplate code to be used for each React editor element factory, drastically reducing code duplication.


packages/react/src/ElementFactory.tsx
```diff
@@ -0,0 +1,77 @@
...
+      </MantineProvider>
+    );
+  };
+
+  root.render(<EditorElementComponentWrapper />);
+
+  return {
+    element: rootElement,
+    render: (dynamicParams: ElementDynamicParams, isHidden: boolean) => {
+      if (isHidden) {
```
YousefED, line 68: these hacks are not valid react, I think it's pretty dangerous (I'm surprised it even works). State is something that should be set from inside the component, props can be passed from the outside.

So the pattern you can use, is something like:

`root.render(<EditorElementComponentWrapper isOpen={false} {...dynamicParams} />);`

packages/react/src/ElementFactory.tsx
```diff
@@ -0,0 +1,77 @@
...
+  ) => JSX.Element,
+  tippyProps?: TippyProps
+): EditorElement<ElementDynamicParams> => {
+  const rootElement = document.createElement("div");
+  const root = createRoot(rootElement);
+
+  let setElementDynamicParams: (dynamicParams: ElementDynamicParams) => void;
+  let setElementIsOpen: (isOpen: boolean) => void;
+
+  const EditorElementComponentWrapper = () => {
```
YousefED, line 24: can we extract this to a separate file?

packages/react/src/ElementFactory.tsx
```diff
@@ -0,0 +1,77 @@
+import { useState } from "react";
+import { MantineProvider } from "@mantine/core";
+import { BlockNoteTheme } from "./BlockNoteTheme";
+import Tippy, { TippyProps } from "@tippyjs/react";
+import { createRoot } from "react-dom/client";
+import { EditorElement, RequiredDynamicParams } from "@blocknote/core";
+
+export const ElementFactory = <
```
YousefED, line 8: Pretty nice that you've extracted this to a single element for all the different factories, that cleans up code quite a bit. I don't see any per-factory exceptions (you've been able to keep it generic), so that's really nice.

Please do at some comments to what ElementFactory is, because it's now a pretty important centerpiece of the library

packages/react/src/HyperlinkToolbar/HyperlinkToolbarFactory.tsx
```diff
@@ -1,61 +1,19 @@
...
   HyperlinkToolbarDynamicParams,
   HyperlinkToolbarFactory,
   HyperlinkToolbarStaticParams,
 } from "@blocknote/core";
-import { MantineProvider } from "@mantine/core";
-import Tippy from "@tippyjs/react";
 import { HyperlinkToolbar as ReactHyperlinkToolbar } from "./components/HyperlinkToolbar";
-import { BlockNoteTheme } from "../BlockNoteTheme";
-// import rootStyles from "../../../core/src/root.module.css";
+import { ElementFactory as ReactElementFactory } from "../ElementFactory";
```
YousefED, line 8: Why not just rename it ReactElementFactory? Then you don't need `ElementFactory as ReactElementFactory` all the time anymore

YousefED (changes requested):
nice

packages/react/src/ElementFactory.tsx
matthewlipski: In that case I think I'll have to use the non-React version of Tippy, since re-rendering the whole component means a new Tippy instance has to be created each time, and I have no other way of externally setting the React Tippy's content of referenceRect.

YousefED (approved):
nice!

PR #88, opened 2023-01-24
Lock block editing

matthewlipski:
This PR exposes `blockEditable: boolean` and `setBlockEditable: (blockEditable: boolean) => void` fields to the block side menu factory API, which allows any implementation of the block side menu to lock and unlock editing for a given block. It also adds a button to the drag handle menu in the React block side menu implementation to do exactly this.


rdewolff:
Nice!

YousefED:
Let's close this in favor of #135

PR #89, opened 2023-01-24
WIP: Text alignment

matthewlipski:
The PR exposes `textAlignment: string` and `setTextAlignment: (textAlignment: string) => void' fields to the formatting toolbar factory API, which allows any implementation of the formatting toolbar to set text alignment for any given paragraph, heading, or list item block content. It also adds buttons to the React formatting toolbar implementation to do exactly this.

This PR is currently WIP as it fiddles with block content node (paragraph, heading, & list item) specs, making it likely to change based on feedback from PR #86. It would also be good to finalize PR #88 before this one, as it also includes some changes that would be beneficial here. Overall, the functionality introduced by this PR is working, it would just make more sense to merge #86 and #88 before finishing this one.

### DIFF
--- a/packages/core/src/extensions/Blocks/nodes/Block.ts
+++ b/packages/core/src/extensions/Blocks/nodes/Block.ts
@@ -508,6 +508,7 @@ export const Block = Node.create<IBlock>({
           {
             name: "headingContent",
             attrs: {
+              textAlignment: "left",
               headingLevel: "1",
             },
           }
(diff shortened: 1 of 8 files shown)


YousefED:
Testing feedback: I think one of the primary use-cases is that the user selects multiple blocks and changes the alignment.

Code feedback (can discuss tonight): I don't think this should be part of the "type" attrs. I.e. a paragraph that's left aligned vs a paragraph that's right aligned, are both the same block type ("paragraph"). The way it's architected now, it seems like they'd be different block types. I think we should step back a bit and think about how we should model this correctly

PR #91, opened 2023-02-03
Text colors, background colors, and text alignment

matthewlipski:
This PR exposes functions to element factories for setting inline/block text and background colors, as well as setting text alignment. It also adds corresponding buttons & menus in the React element factories.

- [X] Block background color extension
- [X] Block text color extension
- [X] Background color mark
- [X] Text color mark
- [X] Text alignment extension
- [x] Text alignment for all selected block content


packages/core/src/extensions/Blocks/nodes/BlockContainer.ts
```diff
@@ -87,6 +76,7 @@ export const BlockContainer = Node.create<IBlock>({
   },
 
   renderHTML({ HTMLAttributes }) {
+    console.log(HTMLAttributes);
```
YousefED, line 79: Todo: remove

packages/core/src/extensions/TextAlignment/TextAlignmentExtension.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+
+export const TextAlignmentExtension = Extension.create({
+  name: "textAlignment",
+
+  addGlobalAttributes() {
+    return [
+      {
+        types: ["paragraph", "heading", "bulletListItem", "numberedListItem"],
```
YousefED, line 20: Is there a reason we set this on specific block types, instead of blockcontainer? If so, maybe good to comment this

packages/react/src/BlockNoteTheme.ts
```diff
@@ -161,6 +185,48 @@ export const BlockNoteTheme: MantineThemeOverride = {
...
+    }
+
+    return {
+      ...textColors,
+      ...backgroundColors,
+      ...textAlignments,
+    };
+  },
+  other: {
+    colorIndex: {
```
YousefED, line 217: Why "colorIndex"? The numbers aren't used right?

packages/react/src/SharedComponents/ColorPicker/components/ColorPicker.tsx
```diff
@@ -0,0 +1,284 @@
...
+  setBackgroundColor: (color: string) => void;
+}) => {
+  const Target = props.targetElement;
+
+  return (
+    <Menu trigger={"hover"} position={"right"}>
+      <Menu.Target>{Target}</Menu.Target>
+      <Menu.Dropdown>
+        <Menu.Label>Text</Menu.Label>
+        <Menu.Item
```
YousefED, line 20: can we do this in a loop?

YousefED (commented):
most looks great, some small comments. + the questions on discord ofc

packages/core/src/extensions/TextAlignment/TextAlignmentExtension.ts
matthewlipski: Yeah it's because applying it to `blockContainer` will apply it to all child blocks, which I think isn't great from a UX perspective. Will add comment

packages/react/src/BlockNoteTheme.ts
matthewlipski: Oh I was implementing the colors in `BlockNoteTheme` to see how it would work but I think we agreed to do the text/background color styling in the CSS file anyway.

YousefED: but still, why do we need "colorIndex"? can't it be an array for example?

PR #92, opened 2023-02-06
Drag handle position fix

matthewlipski:
This PR fixes the drag handle not updating when the editor changes position or dimensions. It also fixes hovered block updating when the editor changes position.

### DIFF
--- a/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
+++ b/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
@@ -18,18 +18,7 @@ import { MultipleNodeSelection } from "./MultipleNodeSelection";
 const serializeForClipboard = (pv as any).__serializeForClipboard;
 // code based on https://github.com/ueberdosis/tiptap/issues/323#issuecomment-506637799
 
-let horizontalAnchor: number;
 let dragImageElement: Element | undefined;
-
-function getHorizontalAnchor() {
-  const firstBlockGroup = document.querySelector(
-    ".ProseMirror > [class*='blockGroup']"
-  ) as HTMLElement | undefined; // first block group node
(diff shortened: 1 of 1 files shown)


packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
```diff
@@ -22,14 +22,12 @@ let horizontalAnchor: number;
...
 function getHorizontalAnchor() {
-  if (!horizontalAnchor) {
-    const firstBlockGroup = document.querySelector(
-      ".ProseMirror > [class*='blockGroup']"
-    ) as HTMLElement | undefined; // first block group node
-    if (firstBlockGroup) {
-      horizontalAnchor = absoluteRect(firstBlockGroup).left;
-    } // Anchor to the left of the first block group
-  }
+  const firstBlockGroup = document.querySelector(
```
YousefED, line 25: I don't think we need this function `getHorizontalAnchor` at all. in the mousemove listener, we already have the left position of the editor, and we can just use that I think?

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
```diff
@@ -184,8 +182,10 @@ function dragStart(e: DragEvent, view: EditorView) {
     return;
   }
 
+  const editorBoundingBox = view.dom.getBoundingClientRect();
```
YousefED, line 185: fyi, this call to getBoundingClientRect _might_ be expensive (because it forces a reflow), but I couldn't see any performance issues using chrome inspector. I was thinking we might need some smart caching, but let's keep it as-is until we really see there's a bottleneck

YousefED:
closes https://github.com/YousefED/BlockNote/issues/15

PR #97, opened 2023-02-12
BlockNote Api 

YousefED:
In this PR we add an API to access and manipulate blocks 🥳


packages/core/src/extensions/Blocks/api/Block.ts.bak
```diff
@@ -0,0 +1,55 @@
+// alternative interface
```
YousefED, line 1: let's remove this one

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,416 @@
...
+
+  constructor(private tiptapEditor: TiptapEditor) {}
+
+  public createBlock(blockSpec: BlockSpec): Block {
+    const props = this.tiptapEditor.schema.node(
+      blockSpec.type,
+      blockSpec.props
+    ).attrs;
+
+    let textContent: string;
```
YousefED, line 28: let's extract this text handling to a separate (unit tested) function

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,416 @@
...
+      textContent: textContent,
+      styledTextContent: styledTextContent,
+      children: blockSpec.children || [],
+    } as Block;
+  }
+
+  private blockToNode(block: Block) {
+    const content: Node[] = [];
+
+    for (const styledText of block.styledTextContent) {
```
YousefED, line 60: let's also extract this to a separate (unit tested) function

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,416 @@
...
+
+    return { block: this.nodeToBlock(node) };
+  }
+
+  public insertBlocks(
+    blocksToInsert: Block[],
+    referenceBlockId: string,
+    placement?: "before" | "after" | "nested"
+  ): void {
+    if (!placement) {
```
YousefED, line 166: you can use `placement: "before" | "after" | "nested" = "before"`

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,416 @@
...
+      placement = "before";
+    }
+
+    let blockFound = false;
+
+    this.tiptapEditor.state.doc.descendants((node, pos) => {
```
YousefED, line 172: don't we have (or can we make?) a separate function that just get's the tiptap node by ID?

packages/core/src/extensions/Blocks/api/styleTypes.ts
```diff
@@ -0,0 +1,42 @@
+export type StyleTemplate<
```
YousefED, line 1: nice!

YousefED (commented):
Overall looks great, let's see if we can:

- use smaller functions in Editor.ts
- unit test the API using #101 
- add JSDoc /*** for the API functions

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,416 @@
...
+    /**
+     * Rehype plugin which converts the HTML output rendered by BlockNote into a simplified structure which better
+     * follows HTML standards. It does several things:
+     * - Removes all block related div elements, leaving only the actual content inside the block.
+     * - Lifts nested blocks to a higher level for all block types that don't represent list items.
+     * - Wraps blocks which represent list items in corresponding ul/ol HTML elements and restructures them to comply
+     * with HTML list structure.
+     * @param options Options for specifying which block types represent ordered and unordered list items.
+     */
+    function simplifyBlocks(options: SimplifyBlocksOptions) {
```
YousefED, line 247: Let's extract this plugin to a separate file so we can easily unittest it

examples/editor/src/App.tsx
```diff
@@ -1,14 +1,498 @@
 // import logo from './logo.svg'
+import { Block, Editor } from "@blocknote/core";
```
YousefED, line 2: this file needs to be cleaned up / reverted, or if it's useful, add a different example or test

packages/core/package.json
```diff
@@ -66,7 +68,14 @@
     "@tiptap/extension-text": "2.0.0-beta.217",
     "@tiptap/extension-underline": "2.0.0-beta.217",
     "@tiptap/pm": "2.0.0-beta.217",
+    "@types/hast": "^2.3.4",
```
YousefED, line 71: should be devdep

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+export function blockToNode(block: PartialBlock, schema: Schema) {
+  let content: PMNode[] = [];
+
+  if (typeof block.content === "string") {
+    content.push(schema.text(block.content));
+  } else if (typeof block.content === "object") {
+    for (const styledText of block.content) {
+      const marks = [];
+
+      for (const style of styledText.styles) {
```
YousefED, line 38: did we decide on richtext vs styledtext?

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+  Block,
+  BlockProps,
+  blockProps,
+  PartialBlock,
+} from "../extensions/Blocks/api/blockTypes";
+import { TextCursorPosition } from "../extensions/Blocks/api/cursorPositionTypes";
+import { simplifyBlocks } from "./simplifyBlocksRehypePlugin";
+import { removeUnderlines } from "./removeUnderlinesRehypePlugin";
+
+export function blockToNode(block: PartialBlock, schema: Schema) {
```
YousefED, line 29: let's move the helper functions somewhere else

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+    children.push(nodeToBlock(blockInfo.node.lastChild!.child(i)));
+  }
+
+  return {
+    id: blockInfo.id,
+    type: blockInfo.contentType.name,
+    props: props,
+    content: content,
+    children: children,
+  } as Block;
```
YousefED, line 147: this `as` is unsafe. I would use

`const ret: Block = {....}; return ret;`

This way we get help from the TS compiler in case anything changes

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+
+  const content: StyledText[] = [];
+  blockInfo.contentNode.content.forEach((node) => {
+    const styles: Style[] = [];
+
+    for (const mark of node.marks) {
+      styles.push({
+        type: mark.type.name,
+        props: mark.attrs,
+      } as Style);
```
YousefED, line 127: is `as Style` needed here?

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+      const newDocSize = this.tiptapEditor.state.doc.nodeSize;
+      removedSize += oldDocSize - newDocSize;
+
+      return false;
+    });
+
+    if (idsOfBlocksToRemove.size > 0) {
+      let notFoundIds = "";
+
+      for (const id of Array.from(idsOfBlocksToRemove)) {
```
YousefED, line 319: idsOfBlocksToRemove.join(",") could work?

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+   * to each other, the new blocks are inserted at the position of the first block in the array. Throws an error if any
+   * of the blocks could not be found.
+   * @param blocksToRemove An array of blocks that should be replaced.
+   * @param blocksToInsert An array of blocks to replace the old ones with.
+   */
+  public replaceBlocks(
+    blocksToRemove: Block[],
+    blocksToInsert: PartialBlock[]
+  ) {
+    this.insertBlocks(blocksToInsert, blocksToRemove[0]);
```
YousefED, line 341: Let's say we call this API. What happens if we hit Cmd+Z for undo after this?

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+    this.tiptapEditor.on("update", callback);
+  }
+
+  /**
+   * Serializes a list of blocks into an HTML string. The output is not the same as what's rendered by the editor, and
+   * is simplified in order to better conform to HTML standards. Block structuring elements are removed, children of
+   * blocks which aren't list items are lifted out of them, and list items blocks are wrapped in `ul`/`ol` tags.
+   * @param blocks The list of blocks to serialize into HTML.
+   */
+  public async blocksToHTML(blocks: Block[]): Promise<string> {
```
YousefED, line 359: I think it's worth considering splitting this file further apart, and for example extracting these functions to separate files (one for html conversions, one for markdown conversions, etc)

packages/core/src/api/Editor.ts
```diff
@@ -0,0 +1,433 @@
...
+  })) {
+    if (!(blockInfo.contentType.name in blockProps)) {
+      throw Error(
+        "Block is of an unrecognized type: " + blockInfo.contentType.name
+      );
+    }
+
+    const validAttrs = blockProps[blockInfo.contentType.name as Block["type"]];
+
+    if (validAttrs.has(attr as BlockProps)) {
```
YousefED, line 114: log a warning if not valid?

packages/core/src/extensions/Blocks/api/blockTypes.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export type ParagraphBlock = BlockTemplate<"paragraph", GlobalProps>;
+
+export type Block =
+  | ParagraphBlock
+  | HeadingBlock
+  | BulletListItemBlock
+  | NumberedListItemBlock;
+
+// @ts-ignore
```
YousefED, line 63: why ts-ignore?

packages/core/src/extensions/Blocks/nodes/BlockContent/ListItemBlockContent/NumberedListItemBlockContent/NumberedListItemBlockContent.ts
```diff
@@ -65,18 +66,35 @@ export const NumberedListItemBlockContent = Node.create({
...
             return {};
           }
 
-          // Case for regular HTML list structure.
-          if (parent.tagName === "OL") {
+          return false;
+        },
+        node: "blockContainer",
+      },
+      // Case for BlockNote list structure.
```
YousefED, line 77: How does this work again / why is this used?

YousefED (commented):
This is coming together nicely :) I've left some comments, I think besides that the main todo is how to expose the API right? Maybe that's best to look into together

packages/core/src/api/Editor.ts
matthewlipski: I think `validAttrs` is just a bad name - all node attributes are technically valid, but when converting a block to a node, we don't necessarily want all the attributes to be added to the block props (e.g. ordered list item index). So this checks for that.

packages/core/src/api/Editor.ts
matthewlipski: Seems to work fine, reverts the both inserting the new blocks and removing the old ones.

packages/core/src/api/Editor.ts
matthewlipski: Don't think we made the final call on that yet, should probably jump on a quick call for that.

packages/core/src/extensions/Blocks/api/blockTypes.ts
matthewlipski: It's because otherwise TypeScript doesn't know for sure that `PartialBlock` doesn't recursively define itself. AFAIK it's ok when defining an object type to have fields in it of the same type, which is why it's fine that `Block`s contain more `Block`s inside them. For some reason though, `PartialBlock` being conditional breaks this.

The reason this didn't come up before is because `PartialBlock["children"]` was of type `Block[]`, which was incorrect.

packages/core/src/extensions/Blocks/nodes/BlockContent/ListItemBlockContent/NumberedListItemBlockContent/NumberedListItemBlockContent.ts
matthewlipski: It's just so that ordered/unordered list items are parsed correctly from HTML. Since BlockNote doesn't use `ol`/`ul` tags to represent lists, we need 2 separate HTML parse rules - one for list items coming from BlockNote, and the other for list items coming from everywhere else. This was changed a bit since I changed BlockNote list items to have `p` tags instead of `li` (having `li` elements without a parent `ol`/`ul` element isn't valid HTML).

packages/core/src/api/Editor.ts
YousefED: Interesting, that surprises me. How does PM know it's "1 transaction"?

YousefED: notes after call: for now it works (PM-history automatically groups transactions), but might not work with yjs plugin or in other specific scenarios

PR #98, opened 2023-02-15
fix draghandle bug

YousefED:
Fixes issue reported on Discord:

image


matthewlipski (approved):
Looks good! While it feels a bit hacky, IIRC other people also seemed to be running into this issue and suggested similar fixes when I was originally fixing the drag preview to work with multiple block dragging.

PR #103, opened 2023-02-20
initial website

YousefED:
Adds `packages/website` 🥳

still some wip, but merging so we can deploy to the domain


YousefED:
@matthewlipski nice to see the style change, next time let's make this a separate PR :)

Only downside I see now is that the size of the document (JSON structure) is significantly increased, with not very useful data (just specifying everywhere that text has no styles). Do you see good ways to prevent this?

matthewlipski:
> @matthewlipski nice to see the style change, next time let's make this a separate PR :)
> 
> Only downside I see now is that the size of the document (JSON structure) is significantly increased, with not very useful data (just specifying everywhere that text has no styles). Do you see good ways to prevent this?

We could make each style optional, that should work. So a style is considered active if it's present in the `Styles` object. The only weird thing about this is that for togglable styles (bold, italic, etc - not colors), we only care if it's present as they don't really have a value like color styles do, so we would probably assign them a type of `true` instead of the `boolean`. Seems ok to me but I'm wondering if that's not more confusing to the average dev, since the value is basically `true | undefined` instead of `true | false`.

YousefED:
> We could make each style optional, that should work. So a style is considered active if it's present in the `Styles` object. The only weird thing about this is that for togglable styles (bold, italic, etc - not colors), we only care if it's present as they don't really have a value like color styles do, so we would probably assign them a type of `true` instead of the `boolean`. Seems ok to me but I'm wondering if that's not more confusing to the average dev, since the value is basically `true | undefined` instead of `true | false`.

I suppose devs will mostly do `if(bold)`, so in that case it wouldn't matter. I think they should be optional indeed, let's do it!

PR #105, opened 2023-02-27
fix linktoolbar doesn't disappear when click outer;ssr problem fixed

exdj158:
this pr fix problem that linktoolbar doesn't disappear when we click outer of the editor.
fix problem that navigator is undefined when using on ssr.


exdj158:
I need to close this pr because it will bring new bugs

PR #106, opened 2023-03-02
make api accessible via editor, hide tiptap initialization options

YousefED:
This PR:

- removes the difference between BlockNoteEditor and EditorAPI. API methods are directly accessible on blocknoteeditor.
- renames tiptapEditor to _tiptapEditor to clarify it's only for internal use
- We only allow our own `BlockNoteEditorOptions` instead of having it extend from Tiptap's options. This way we can design + document our own API. (and if we want to at some point, we can remove TipTap later)

### DIFF
--- a/packages/core/src/api/nodeConversions/nodeConversions.ts
+++ b/packages/core/src/api/nodeConversions/nodeConversions.ts
@@ -7,8 +7,15 @@ import {
   PartialBlock,
 } from "../../extensions/Blocks/api/blockTypes";
 import { Style, StyledText } from "../../extensions/Blocks/api/styleTypes";
+import UniqueID from "../../extensions/UniqueID/UniqueID";
 
 export function blockToNode(block: PartialBlock, schema: Schema) {
+  let id = block.id;
+
+  if (id === undefined) {
+    id = UniqueID.options.generateID();
+  }
+
   let content: Node[] = [];
 
   if (typeof block.content === "string") {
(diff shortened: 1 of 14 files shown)


matthewlipski (approved):
Tests are fixed, ready to merge:)

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -60,14 +70,19 @@ export function getNodeById(
     if (node.type.name !== "blockContainer" || node.attrs.id !== id) {
       return true;
     }
+    console.log("grdgrdgrdgdr")
```
YousefED, line 73: always double check the PR ;)

packages/core/test/blockManipulation.test.ts
```diff
@@ -1,81 +1,111 @@
...
-    const existingBlock = editorAPI.allBlocks[0];
-
-    editorAPI.insertBlocks(multipleBlocks, existingBlock, placement);
-
-    return editorAPI.allBlocks;
-  }
-
   it("Insert before existing block", async () => {
-    const output = insert("before");
+    await waitForEditor()
```
YousefED, line 91: why is this needed?

matthewlipski: When you create a new `BlockNoteEditor()`, that only starts the creation of the editor. It's not ready to be used straight away. For example, if you pass an `onCreate` function where you log smth then immediately after calling `BlockNoteEditor()`, you terminate the app, nothing will get logged.

This is basically what's happening in the tests as well - the editor isn't 100% ready when we start accessing and changing blocks. So this causes, e.g. the UniqueID plugin to not be initialized yet and we see blocks with null IDs.

`waitForEditor()` waits until the editor is actually ready, then the tests continue as normal.

PR #109, opened 2023-03-06
refactor links a bit

YousefED:
Addresses this feedback: https://github.com/YousefED/BlockNote/pull/108#discussion_r1126348393

It adds one extra loop + duplicate node instantiation in:

```
const nodes = styledTextArrayToNodes(link.content, schema).map((node) => {
    return node.mark([...node.marks, linkMark]);
  });
```

However, it's probably a bit cleaner. wdyt @matthewlipski ?

### DIFF
--- a/packages/core/src/api/nodeConversions/nodeConversions.ts
+++ b/packages/core/src/api/nodeConversions/nodeConversions.ts
@@ -9,6 +9,7 @@ import {
   InlineContent,
   Link,
   PartialInlineContent,
+  PartialLink,
   Style,
   StyledText,
 } from "../../extensions/Blocks/api/inlineContentTypes";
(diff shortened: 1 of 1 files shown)


packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -55,14 +79,7 @@ export function inlineContentToNodes(
...
     if (content.type === "link") {
-      // special case for links, because in ProseMirror, links are marks
-      // in BlockNote, links are a type of inline content
-      const linkMark = schema.marks.link.create({
-        href: content.href,
-      });
-      nodes.push(
-        ...styledTextArrayToNodes(content.content, schema, [linkMark])
-      );
+      nodes.push(...linkToNodes(content, schema));
```
matthewlipski, line 82: I think it might still be a good idea to change the if/else if chain to a lookup map:
```
export function inlineContentToNodes(
  blockContent: PartialInlineContent[],
  schema: Schema
): Node[] {
  let nodes: Node[] = [];
  
  const lookup = {
    "text": (content) => nodes.push(...styledTextArrayToNodes([content], schema)),
    "link": (content) => nodes.push(...linkToNodes(content, schema))
  }

  for (const content of blockContent) {
    if (content.type in lookup) {
      lookup[content.type](content)
    } else {
      throw new UnreachableCaseError(content);
    }
  }
  return nodes;
}

```

Would make it easier to add support for more inline content types in the future.

matthewlipski (commented):
Looks good! Much cleaner than before. Left one comment with a proposed change, but other than that seems good to go.

packages/core/src/api/nodeConversions/nodeConversions.ts
YousefED: Let's do this only when it becomes relevant. At this point the downside is it would break the typesafe union-extensiveness check (`UnreachableCaseError`)

PR #110, opened 2023-03-07
fix dragimage + allow dragging next to editor

YousefED:
- add support for dragging / dropping next to the editor
- fixes drag previews for nested items

drag


matthewlipski (approved):
Very nice, great that you cleaned up the event handlers too. Since we don't call preventDefault I guess that we get 2 drop and drag over events being triggered? Don't think it's an issue either way but might be something to remember for the future.

YousefED:
I think the prosemirror handlers will call preventDefault. Not 100% sure, but good to keep in mind indeed

PR #112, opened 2023-03-07
Added BlockIdentifier & small changes

YousefED:


packages/core/src/api/Editor.ts
```diff
@@ -38,26 +43,77 @@ export class Editor {
...
+      if (node.type.name !== "blockContainer" || node.attrs.id !== id) {
+        return true;
+      }
+
+      newBlock = nodeToBlock(node, this.blockCache);
+
+      return false;
+    });
+
+    if (typeof newBlock === "undefined") {
```
YousefED, line 69: I would return `undefined` here, I don't think it's intuitive to have a "get by id" function that throws an error if it's not found.

This means we do need to change the other functions (`updateBlock`, `replaceBlock`), because I do think those should throw an error if the block isn't found

PR #117, opened 2023-03-11
fix umd build for sandpack

YousefED:
The sandpack demos fail when Tippy components appear. The issue is that it doesn't correctly resolve the Tippy default import from `@tippyjs/react`

This is a regression from v0.3.0, which was built with an older version of Rollup. 

Old code shows `.default`, where new builds don't

image


YousefED:
works! (0.4.3-alpha.0)

PR #128, opened 2023-03-14
fix: externalize unified plugins by adding them to package.json

YousefED:
I think this might have been breaking the sandpack examples. in `vite.config.ts`, we have the following setting:

```
// make sure to externalize deps that shouldn't be bundled
      // into your library
      external: (source: string) => {
        if (deps.includes(source)) {
          return true;
        }
        return source.startsWith("prosemirror-");
      },

```

We imported several packages that were not explicitly in "dependencies". this then caused the remark plugins to end up in our `blocknote.umd.js` built file. My guess is this broke the examples


YousefED:
confirmed to fix the dependency errors

PR #129, opened 2023-03-14
Fix node html parse

YousefED:
This seems to fix the HTML -> Node conversions. @matthewlipski Any reason why `node` was always set to `blockContainer`?

I'm not sure why this worked locally btw in the unit tests. But the published version now at least also works in codesandbox

### DIFF
--- a/examples/editor/package.json
+++ b/examples/editor/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@blocknote/example-editor",
   "private": true,
-  "version": "0.4.6-alpha.1",
+  "version": "0.4.6-alpha.2",
   "scripts": {
     "dev": "vite",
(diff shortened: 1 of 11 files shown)


matthewlipski:
> This seems to fix the HTML -> Node conversions. @matthewlipski Any reason why `node` was always set to `blockContainer`?
> 
> I'm not sure why this worked locally btw in the unit tests. But the published version now at least also works in codesandbox

This was written a while ago but I think it was to force PM to parse each element as a separate block. I think I was having some issues at that time with blocks being nested when they shouldn't be when pasting, but maybe something in the codebase changed since then? Either way glad that it works now.

matthewlipski (approved):

Issue #104, opened 2023-02-24
Support Vue3?


HalleyGao:
Is so great. May you support vue3?

howtopick:
same ask

Achrou:
+1

YousefED:
Hi!

Thanks for sharing your interest. As the UI components are written in React, there's two ways we could work on this:

1) recreate UI components completely in Vue. This would have the disadvantage of needing to maintain 2 UI interfaces
2) have the Vue implementation for now bridge to the React components. This mean there will be a dependency on React as well.

I think option 2 is feasible at the moment. Would this be a ok for you?

howtopick:
> 1. recreate UI components completely in Vue. 
Hi!
I'm trying the first one, but still encountered some difficulties, eg: I'm trying to create my slashMenu extension by SlashMenuExtension, but don't know how to create a SuggestionsMenuFactory use vue, so ask a simple example for this.
Many thanks！

YousefED:
We're working on a website with documentation. This page should help to explain how to integrate without React: https://www.blocknotejs.org/docs/vanilla-js

ManUtopiK:
Hi! I would like to use it in Vue.
I'm trying vanilla-js but nothing appear on the screen :/

In the console :
![image](https://user-images.githubusercontent.com/188172/224762005-c8e09951-b8ae-469a-9200-4fdcde45fb7a.png)

But the dom is not filled. There is still the root div.

YousefED:
@ManUtopiK did you pass `parentElement` when creating the editor?

ManUtopiK:
I just run the vanilla-js example.

ManUtopiK:
As I said, the console output informs that the editor is well created.

I tried to add options without success :
```
_tiptapOptions: {
    content: 'Some content',
    autofocus: true,
  }
```

Fun fact, in the devtools console, i can store the editor in the temp1 variable and get the content :
![image](https://user-images.githubusercontent.com/188172/225090986-4f8b56ac-de55-49ac-9c6b-8ba5e864308f.png)

ManUtopiK:
Awesome. It works !
Have you started a branch with Vue.js ?

ManUtopiK:
Ok, I just created a working branch with vue3 #130 !

YousefED:
Actually, I just realized we have "option 2" (bridge to React) already working for the website, which is based on Vitepress / Vue.

[...]

PR #130, opened 2023-03-14
Basic mount in vue3

ManUtopiK:
`npm init vue@latest` and basic implementation


examples/vue/src/BlockNoteView.vue
```diff
@@ -0,0 +1,51 @@
...
+}>()
+
+const emit = defineEmits(['update:modelValue'])
+
+const editor = ref()
+onMounted(async () => {
+  // Convert md to html
+  const Editor = new BlockNoteEditor({})
+  const blocks = await Editor.markdownToBlocks(props.modelValue)
+  const content = await Editor.blocksToHTML(blocks)
```
ManUtopiK, line 20: @YousefED This is ugly. How to pass initial content ?

YousefED: We'll add a new method + demo for this next week! Best to always work with `JSON.stringify(blocks)` btw, instead of storing markdown

See https://github.com/TypeCellOS/BlockNote/issues/139

YousefED:
Hi! 

You mentioned here #104 you had some trouble getting a separate package to work. I had a quick look, and my recommendations first would be:

- Don't use `pnpm`: this repository currently works with `npm`, so I'm not sure what the effect is of adding a different package manager
- make sure `npm run build` in the root directory works

After resolving those errors, it will be easier to diagnose!

PR #131, opened 2023-03-14
fix github link in nav

cuire:
Sincerely, Link Keeper.


YousefED:
Thanks! I'll probably want to move the repository, but for now I'll merge this :)

PR #135, opened 2023-03-15
Editor Locking

matthewlipski:
This PR introduces editor functions for getting & setting whether the editor is editable/locked, as well as documentation for that.


packages/core/src/BlockNoteEditor.ts
```diff
@@ -250,6 +252,22 @@ export class BlockNoteEditor {
     }
   }
 
+  /**
```
YousefED, line 255: If we use a getter for this, shouldn't we use a setter as well? wdyt?

YousefED (approved):

PR #137, opened 2023-03-15
`BNMergeBlocks` fix

matthewlipski:
Previously, `BNMergeBlocks`, which is used to combine 2 blocks when pressing backspace at the start of one, only copied text from the latter block. This PR changes it to copy all content, meaning all `InlineContent` is preserved.


YousefED (approved):
@matthewlipski nice. This seems like important functionality for the core though, should we have a unit / e2e test for this?

PR #142, opened 2023-03-17
feat: customize formatting toolbar

YousefED:
Work in progress: I think FormattingToolbar can be really simple. If so, we could then expose ToolbarButton to consumers, and they can just use JSX to create a custom formattingtoolbar. I think this is a more extensible approach than for example asking for a custom array defining which buttons to show in what order

- [x] simplify logic of FormattingToolbar
- [x] Expose ToolbarButton / LinkToolbarButton
- [x] Make it possible to pass a custom FormattingToolbar
- [x] Docs & demo
- [ ] Does the API `FormattingToolbarProps` still make sense? or should we (partially) move this to `BlockNoteEditor`?


matthewlipski:
> Work in progress: I think FormattingToolbar can be really simple. If so, we could then expose ToolbarButton to consumers, and they can just use JSX to create a custom formattingtoolbar. I think this is a more extensible approach than for example asking for a custom array defining which buttons to show in what order
> 
> * [x]  simplify logic of FormattingToolbar
> * [ ]  Expose ToolbarButton / LinkToolbarButton
> * [ ]  Make it possible to pass a custom FormattingToolbar
> * [ ]  Docs & demo
> * [ ]  Does the API `FormattingToolbarProps` still make sense? or should we (partially) move this to `BlockNoteEditor`?

I'm not sure about letting consumers just pass JSX directly into the formatting toolbar, mainly because it seems to be the opposite approach to the slash menu. I agree it makes it more extensible, but idk if extensibility is something we want to focus on for the premade menu, moreso consistency. Might need to have a longer discussion about this since it's relevant for all menus.

Alternative:

```
type FormattingToolbarButton = {
  name: string;
  execute: (editor: BlockNoteEditor => void);
  icon: JSX.Element;
  shortcut?: string;
}

type FormattingToolbarDropdownItem = {
  name: string;
  execute: (editor: BlockNoteEditor => void);
  active: (editor: BlockNoteEditor => boolean);
  icon: JSX.Element;
}

type FormattingToolbarDropdown = {
  items: FormattingToolbarDropdownItem[];
}

type FormattingToolbarItem = (FormattingToolbarButton | FormattingToolbarDropdown)[];
```

packages/core/src/BlockNoteEditor.ts
```diff
@@ -299,6 +308,135 @@ export class BlockNoteEditor {
     replaceBlocks(blocksToRemove, blocksToInsert, this._tiptapEditor);
   }
 
+  public getActiveStyles() {
+    const styles: Styles = {};
+    const marks = this._tiptapEditor.state.selection.$to.marks();
+    console.log(marks);
```
YousefED: todo: remove

packages/core/src/BlockNoteEditor.ts
```diff
@@ -299,6 +308,135 @@ export class BlockNoteEditor {
...
+      "bold",
+      "italic",
+      "underline",
+      "strike",
+    ]);
+    const colorStyles = new Set<ColorStyle>(["textColor", "backgroundColor"]);
+
+    for (const mark of marks) {
+      if (toggleStyles.has(mark.type.name as ToggledStyle)) {
+        styles[mark.type.name as ToggledStyle] = true;
```
YousefED, line 419: if the mark exists, it's always actuve ("true") right?

packages/core/src/BlockNoteEditor.ts
```diff
@@ -299,6 +308,135 @@ export class BlockNoteEditor {
     replaceBlocks(blocksToRemove, blocksToInsert, this._tiptapEditor);
   }
 
+  public getActiveStyles() {
```
YousefED, line 405: lets add docs for all public funcs

packages/react/src/FormattingToolbar/components/DefaultButtons/CreateLinkButton.tsx
```diff
@@ -0,0 +1,43 @@
+import { useCallback } from "react";
+import { BlockNoteEditor } from "@blocknote/core";
+import { RiLink } from "react-icons/ri";
+import LinkToolbarButton from "../LinkToolbarButton";
+import { formatKeyboardShortcut } from "../../../utils";
+
+export const CreateLinkButton = (props: { editor: BlockNoteEditor }) => {
+  const setLink = useCallback(
```
YousefED, line 8: I don't think useCallback is useful in this function, especially with [props] as parameter.

The reason is that most functions are invoked immediately anyway (e.g. `linkIsActive()`)

packages/core/src/BlockNoteEditor.ts
matthewlipski: Yep, if a toggled style is active, in PM its represented as a mark with no attrs, in the API it's always true in the styles object.

packages/core/src/BlockNoteEditor.ts
```diff
@@ -361,6 +402,170 @@ export class BlockNoteEditor {
...
+        this._tiptapEditor.commands.toggleMark(style, { color: value });
+      }
+    }
+  }
+
+  /**
+   * Gets the URL of the link at the current selection, and the currently selected text. If no link is active, the URL
+   * is an empty string.
+   */
+  public getActiveLink() {
```
YousefED, line 495: split

YousefED (commented):
see inline + double check useCallbacks

packages/react/src/FormattingToolbar/FormattingToolbarFactory.tsx
```diff
@@ -3,9 +3,24 @@ import {
   FormattingToolbarFactory,
   FormattingToolbarStaticParams,
   FormattingToolbarDynamicParams,
+  BlockNoteEditor,
 } from "@blocknote/core";
 import { FormattingToolbar as ReactFormattingToolbar } from "./components/FormattingToolbar";
 import { ReactElementFactory } from "../ElementFactory/components/ReactElementFactory";
+import { FC } from "react";
+
+export const createReactFormattingToolbarFactory = (
```
YousefED, line 12: same question as in sidemenu PR: the function call below seems very similar to what happens here

PR #143, opened 2023-03-17
feat: customize sidemenu

YousefED:
For customizing the BlockSideMenu, I think we can follow a similar approach as described in https://github.com/TypeCellOS/BlockNote/pull/142. We can make the default menu contents customizable allowing users to pass custom content for `<Menu.Dropdown className={classes.root}>`. This would make the entire dropdown menu customizable, while keeping it very simple. 

Note that in this design, if users would want to customize more (i.e.: add a button besides the draghandle / plus button), they'd still need to provide a custom factory.

As a first start I simplified the Color Menu. @matthewlipski I know you were having some issues with this earlier, could you check if this approach doesn't have any of the bugs you encountered earlier?

- [x] check color menu refactor
- [x] Provide a custom `Menu.Item` wrapper (e.g.: `BlockNoteMenuItem`) that automatically calls`setDragHandleMenuOpened(false); props.unfreezeMenu();` in `onClick`, so users don't need to do this manually
- [x] extract dropdown contents and allow users to easily customize this
- [x] docs + demo


packages/react/src/BlockSideMenu/components/DefaultButtons/BlockColorsButton.tsx
```diff
@@ -0,0 +1,80 @@
+import { ReactNode, useCallback, useRef, useState } from "react";
+import { Box, Menu } from "@mantine/core";
+import { BlockNoteEditor } from "@blocknote/core";
+import { HiChevronRight } from "react-icons/hi";
+import { ColorPicker } from "../../../SharedComponents/ColorPicker/components/ColorPicker";
+
+export const BlockColorsButton = (props: {
```
YousefED: I think this is the old code. Have you reviewed my changes / fixes to the color menu? (see PR)

packages/react/src/BlockSideMenu/components/BlockSideMenu.tsx
```diff
@@ -69,24 +71,24 @@ export const BlockSideMenu = (props: BlockSideMenuProps) => {
...
+            editor={props.editor}
+            closeMenu={() => {
               setDragHandleMenuOpened(false);
               props.unfreezeMenu();
             }}
-            {...props}
           />
-        </Menu.Dropdown>
+        ) : (
+          <DragHandleMenu>
```
YousefED, line 83: Let's make this a little cleaner by separating this to a separate file, and then use
`DragHandleMenu = props.dragHandleMenu || DefaultDragHandleMenu`.

packages/core/src/BlockNoteEditor.ts
```diff
@@ -137,6 +141,11 @@ export class BlockNoteEditor {
         options.onEditorReady?.(this);
         options.initialContent &&
           this.replaceBlocks(this.topLevelBlocks, options.initialContent);
+        document.addEventListener(
```
YousefED, line 144: I'm not entirely sure we want this here. Can't we keep this in the DraggableBlocksPlugin?

packages/react/src/BlockSideMenu/components/DefaultButtons/BlockColorsButton.tsx
matthewlipski: Are you sure? I looked at it again and it seems the same as in commit [30018f5](https://github.com/TypeCellOS/BlockNote/pull/143/commits/30018f50ad930adb5c518308a654b7f1b9af90dd), besides the functions for getting/setting colors, which changed as the props are different.

packages/core/src/BlockNoteEditor.ts
matthewlipski: We could, and then ig we would pass the cursor pos as a prop to the menu. I think that if we want to allow consumers to check the currently mouse-hovered block outside the DraggableBlocksPlugin though, we would still need smth like this. Imo it's better to init the listener in BlockNoteEditor and pass the mouse coords down to DraggableBlocksPlugin, rather than relying on DraggableBlocksPlugin to update mouse coords. Wdyt?

matthewlipski: I think there is a use case for getting the block hovered by the mouse outside the DraggableBlocksPlugin btw, e.g. a slash menu command that involves 2 blocks, where you can select the 2nd block with the mouse.

packages/react/src/BlockSideMenu/components/DefaultButtons/BlockColorsButton.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Box, Menu } from "@mantine/core";
+import { HiChevronRight } from "react-icons/hi";
+import { ColorPicker } from "../../../SharedComponents/ColorPicker/components/ColorPicker";
+import {
+  DragHandleMenuItem,
+  DragHandleMenuItemProps,
+} from "../DragHandleMenuItem";
+
+export const BlockColorsButton = (props: DragHandleMenuItemProps) => {
+  const [block] = useState(props.editor.getMouseCursorPosition()?.block);
```
YousefED, line 11: why useState?

packages/react/src/BlockSideMenu/components/DefaultButtons/BlockColorsButton.tsx
YousefED: You're right!

packages/react/src/BlockSideMenu/components/DefaultButtons/BlockColorsButton.tsx
matthewlipski: It's because the menu has to update whenever we hover on/off the colors button, so without `useState`, `block` would get updated to whichever block the cursor is next to and the color might get applied to the wrong one.

packages/core/src/extensions/DraggableBlocks/BlockSideMenuFactoryTypes.ts
```diff
@@ -1,22 +1,22 @@
...
   unfreezeMenu: () => void;
-
-  setBlockTextColor: (color: string) => void;
-  setBlockBackgroundColor: (color: string) => void;
 };
 
 export type BlockSideMenuDynamicParams = {
-  blockTextColor: string;
-  blockBackgroundColor: string;
+  editor: BlockNoteEditor;
```
YousefED, line 18: should editor be part of both static and dynamic params?

packages/website/docs/docs/side-menu.md
```diff
@@ -1,8 +1,142 @@
-# Side Menu
+# Block Side Menu & Drag Handle Menu
```
YousefED, line 1: remove the coming soon and comments

YousefED, line 1: for title, I think Block Side Menu is fine. (drag handle is part of side menu)

packages/website/docs/docs/side-menu.md
```diff
@@ -1,8 +1,142 @@
...
+const editor = useBlockNote({
+  uiFactories: {
+    blockSideMenuFactory: createReactBlockSideMenuFactory(
+      CustomBlockSideMenu
+    ),
+  },
+});
+```
+
+If all this sounds confusing, you can find out more about the `uiFactories` option in [Creating Your Own UI Elements](/docs/vanilla-js#creating-your-own-ui-elements), but it's only useful if you want to use BlockNote in a non-React app.
```
YousefED, line 99: `but it's only useful if you want to use BlockNote in a non-React app.`, then let's remove this paragraph

packages/website/docs/docs/side-menu.md
```diff
@@ -1,8 +1,142 @@
...
+
+## Default Items
+
+It might seem daunting to create your own Drag Handle Menu from scratch, which is why BlockNote provides React components for everything you see in the default layout - both the menu itself and the items in it. Below are all the default components you can use to build your custom menu:
```
YousefED, line 103: nice!

YousefED (commented):
some small comments

packages/react/src/BlockSideMenu/BlockSideMenuFactory.tsx
```diff
@@ -6,6 +7,27 @@ import {
 } from "@blocknote/core";
 import { BlockSideMenu as ReactBlockSideMenu } from "./components/BlockSideMenu";
 import { ReactElementFactory } from "../ElementFactory/components/ReactElementFactory";
+import { DragHandleMenuProps } from "./components/DragHandleMenu";
+
+export const createReactBlockSideMenuFactory = (
```
YousefED, line 12: this is pretty much a duplicate as the functon below? Can we simplify this?

Issue #144, opened 2023-03-17
Customizing BlockNote Editor Styles


steven-tey:
I've been trying to customize the editor's fonts using `_tiptapOptions` but it doesn't seem to work:

```
const editor: BlockNoteEditor | null = useBlockNote({
    onEditorContentChange: debouncedChanges,
    _tiptapOptions: {
      editorProps: {
        attributes: {
          class: "prose font-vercel text-red-600",
        },
      },
    },
  });
```

The goal is to handle all the styling via `@tailwind/typography`.

mkarbo:
+1 for tailwind support

YousefED:
You should now be able to pass `class: "prose font-vercel text-red-600",` to `editorDOMAttributes`, and set `defaultStyles` to false!

steven-tey:
Wonderful! Thanks a lot @YousefED!

steven-tey:
Works flawlessly <picture data-single-emoji=":chefs-kiss:" title=":chefs-kiss:"><img class="emoji" width="20" height="auto" src="https://single-emoji.vercel.app/api/emoji/eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIn0..9Y0_eaxm29yIh0kM.59-5S4laxYjTmyuHLRl8RfNi7TE6-q2U6yLn4zUdZRcwTIqMBx1OPkQVeWgf9eaR3SZHVurT5W-PDADW6flnVhBDzUdDLtthn09FUWcfN16Mz1Ygqgo5dog.Xs_lXM29S5em0jAKVrcVvA" alt=":chefs-kiss:" align="absmiddle"></picture> thank you!

PR #148, opened 2023-03-21
add defaultStyles option, reorganize styles

YousefED:
Fixes https://github.com/TypeCellOS/BlockNote/issues/144. With this change, users can prevent the "css reset" of block styles

Todo:

- [x] validate there are no regressions on Playwright
- [ ] I noticed a bug in the formattingtoolbar, double check if that's resolved
- [x] make it work for dragPreview as well

### DIFF
--- a/packages/core/src/BlockNoteEditor.ts
+++ b/packages/core/src/BlockNoteEditor.ts
@@ -34,13 +34,48 @@ export type BlockNoteEditorOptions = {
   // TODO: Figure out if enableBlockNoteExtensions/disableHistoryExtension are needed and document them.
   enableBlockNoteExtensions: boolean;
   disableHistoryExtension: boolean;
+  /**
+   * Factories used to create a custom UI for BlockNote
+   */
   uiFactories: UiFactories;
+  /**
+   * TODO: why is this called slashCommands and not slashMenuItems?
+   *
+   * @default defaultSlashMenuItems from `./extensions/SlashMenu`
+   */
   slashCommands: BaseSlashMenuItem[];
+
+  /**
+   * The HTML element that should be used as the parent element for the editor.
+   *
+   * @default: undefined, the editor is not attached to the DOM
+   */
   parentElement: HTMLElement;
+  /**
+   * An object containing attributes that should be added to the editor's HTML element.
+   *
+   * @example { class: "my-editor-class" }
+   */
   editorDOMAttributes: Record<string, string>;
+  /**
+   *  A callback function that runs when the editor is ready to be used.
+   */
   onEditorReady: (editor: BlockNoteEditor) => void;
+  /**
+   * A callback function that runs whenever the editor's contents change.
+   */
   onEditorContentChange: (editor: BlockNoteEditor) => void;
+  /**
+   * A callback function that runs whenever the text cursor position changes.
+   */
   onTextCursorPositionChange: (editor: BlockNoteEditor) => void;
+
+  /**
+   * Use default BlockNote font and reset the styles of <p> <li> <h1> elements etc., that are used in BlockNote.
+   *
+   * @default true
+   */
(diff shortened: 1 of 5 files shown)


packages/website/docs/docs/editor.md
```diff
@@ -1,6 +1,6 @@
 # Customizing the Editor
 
-There are a number of options that you can pass to `useBlockNote()`, which you can use to customize the editor. You can find the full list of these below:
+There are a number of options (all optional) that you can pass to `useBlockNote()`, which you can use to customize the editor. You can find the full list of these below:
```
matthewlipski, line 3: I think it makes more sense to just do `BlocknoteEditorOptions = Partial<{...}>`

packages/core/src/extensions/Blocks/nodes/Block.module.css
```diff
@@ -3,21 +3,11 @@ BASIC STYLES
...
 
 .blockOuter {
-  /* padding-bottom: 15px; */
-  font-size: 16px;
   line-height: 1.5;
-  transition: all 0.2s;
-  font-weight: normal;
-}
-
-.block {
```
matthewlipski, line 13: I didn't see this moved to a different file, was the transition just redundant?

packages/core/src/BlockNoteEditor.ts
```diff
@@ -34,14 +34,49 @@ export type BlockNoteEditorOptions = {
...
+   *
+   * @default: undefined, the editor is not attached to the DOM
+   */
   parentElement: HTMLElement;
+  /**
+   * An object containing attributes that should be added to the editor's HTML element.
+   *
+   * @example { class: "my-editor-class" }
+   */
   editorDOMAttributes: Record<string, string>;
```
matthewlipski, line 59: I'm thinking we might want to separate this into `editorClass` and `editorDOMAttributes` instead of adding the `defaultStyles`, the reason being we want to apply the same styling to the drag preview but probably don't want to set the same attributes since they might break it. That way, we can just have `editorClass: string` and set it to `defaultStyles` if it's undefined. Then the drag preview can also just inherit `editorClass`.

matthewlipski: So to just remove the default styles, you would do `editorClass: ""` instead of `defaultStyles: false`. Though in reality idk if someone would remove the default styles without adding their own using a custom class anyway.

matthewlipski (commented):
Looks like a good start! Left a few comments, mainly ideas to fix the drag preview styling.

packages/core/src/extensions/Blocks/nodes/Block.module.css
YousefED: yes I think so, but maybe you can double check if all animations still work

packages/core/src/BlockNoteEditor.ts
YousefED: Hmm, that will have the disadvantage that it's impossible to keep the default styles but add a different class (I think that's what you mean in your last comment).

We can still infer the classnames needed for the dragpreview from `editorDOMAttributes` right?

Or maybe the best way to go about this is directly grabbing the class from the editor element in the DraggableBlocksPlugin. That way, we also don't need to pass the class attributes all the way down

YousefED:
> What was the formatting toolbar bug you found?

When toggling bold / indent, I think the icon would occasionally dissappear

packages/core/src/BlockNoteEditor.ts
matthewlipski: Yeah that's a good point, I was thinking about what if someone wanted to add a `styles` attribute to the editor (as it would need to be applied to the drag preview too), but my idea doesn't fix that issue either. Getting the editor class from the DOM element should be ok.

PR #164, opened 2023-04-12
wip: custom blocks typing and API

YousefED:
I'm curious what you think if this approach. I think this direction will make it possible to register custom blocks while keeping a strongly typed API.

This PR does:

- refactor of registration of built-in block types. 
- makes it possible to have a strongly typed `BlockNoteEditor` which API matches the Blocks that have been defined. This will make it possible to add support for custom blocks, while keeping a strongly typed API
- As part of this, `globalProps` and `blockProps` have been removed, because they should be dynamically configurable as well (based on the "registered blocks" in the editor).

TODO:
- make it compile (nodeToBlock, etc)
- use the new BlockSpec types instead of `globalProps` / `blockProps`
- how to handle markdown / html
- design + implement custom block interface
- test / add tests
- ...

### DIFF
--- a/packages/core/src/BlockNoteEditor.ts
+++ b/packages/core/src/BlockNoteEditor.ts
@@ -19,19 +19,24 @@ import { getNodeById } from "./api/util/nodeUtil";
 import { getBlockNoteExtensions, UiFactories } from "./BlockNoteExtensions";
 import styles from "./editor.module.css";
 import {
-  Block,
   BlockIdentifier,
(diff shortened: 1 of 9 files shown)


dukesx:
Looks promising but don't forget to add support for "react render", a function that tiptap exposes so that we can render react components directly inside Tiptap.

KashifAhmed:
Great feature, well what the estimate time to implement this feature

PR #166, opened 2023-04-14
fix: slashmenu dropdown positioning on edge of page

tensor-tian:
martine menu dropdown position style is set to "absolute". 
The tippy container's ( the parent of dropdown menu ) computed width and height are 0.
This makes popper overflow not work when change the type of last several blocks.

before fix, popper's overflow detection not work:
Screenshot 2023-04-14 at 18 17 10

after fix, popper's default flip behavior work:
Screenshot 2023-04-14 at 18 22 11


matthewlipski (approved):
Looks good!

YousefED:
Thanks @tensor-tian! We'll make this part of the next release!

Issue #165, opened 2023-04-13
When scrolling, the slash menu and plus sign doesn't scroll


the-watchmaker:
**Issue** 
![Screen Recording 2023-04-12 at 5 05 56 PM](https://user-images.githubusercontent.com/4682613/231612593-98933c72-f163-460a-8f8b-adf18c21a5c9.gif)

When you scroll down the menu is stuck on the screen rather than following you (positioned by position: absolute; and transform)

I think this is ok and can be intended design if we have
- onSlashMenuOpen() to freeze the scroll altogether and only let the user scroll within the slash menu, just like Notion
- and make it pop up on top of the editing area when there is no more space underneath

I've been playing around with all the positioning tricks to fix it but no luck. Let me know if any solutions.

PR #167, opened 2023-04-14
fix: menu positioning when scrolling

matthewlipski:
The menus and toolbars currently stay in place when the page is scrolled. This PR makes makes their positions update on scroll.

closes https://github.com/TypeCellOS/BlockNote/issues/165


YousefED (approved):
Best we can do for now.

A more ideal solution would be if we the menus would be locate relatively in the editor. In that case CSS should be able to take care of the positioning, and we don't have to manually update it.

However, this would be a bit cumbersome to fix now I think as menus are located at a different position in the HTML tree. We should create a separate issue for this

PR #183, opened 2023-05-01
Custom blocks api changes

matthewlipski:


packages/react/src/FormattingToolbar/components/DefaultDropdowns/BlockTypeDropdown.tsx
```diff
@@ -10,7 +15,38 @@ import {
...
+  for (let i = 0; i < a.length; i++) {
+    if (a[i] !== b[i]) {
+      return false;
+    }
+  }
+
+  return true;
+};
+
+const shouldShow = (schema: BlockSchema) => {
```
YousefED, line 32: It's not clear to me why we use the `arrayEquals` on schema.

Wouldn't it make more sense to just show a dropdown for all the levels in the propschema?

YousefED (commented):
Great stuff!

- I think the way schemas (props / blocks) now work is a good improvement
- I'm not 100% convinced about the `NumberedListItemBlockContent`, `TipTapNodeConfig`, `TypesMatch`. It feels like this is quite some code (complexity) for a small (maybe I'm wrong here) gain in type-correctness. I think for these scenarios it's find to do a quick runtime check

Overall, great work. I've left some minor comments.  I'll review the parse / renderhtml etc.  when the first custom block tests have been added

packages/core/src/BlockNoteEditor.ts
```diff
@@ -133,36 +121,36 @@ export class BlockNoteEditor<
     this._tiptapEditor.view.focus();
   }
 
-  constructor(options: Partial<BlockNoteEditorOptions<Block>> = {}) {
+  constructor(options: Partial<BlockNoteEditorOptions<BSchema>> = {}) {
     console.log("test");
     // apply defaults
-    options = {
+    const newOptions: Omit<typeof options, "defaultStyles" | "blockSchema"> & {
```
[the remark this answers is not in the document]
matthewlipski: It's because `options` is `Partial`, so `defaultStyles` and `blockSchema` might be undefined. Since we're applying defaults to them this changes those 2 fields to always be defined.

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -132,6 +153,47 @@ export function createBlockSpec<
...
+        blockContent.setAttribute("data-content-type", blockConfig.type);
+        // Add props as HTML attributes in kebab-case with "data-" prefix
+        for (const [attribute, value] of Object.entries(HTMLAttributes)) {
+          blockContent.setAttribute(attribute, value);
+        }
+
+        // Gets BlockNote editor instance
+        const editor = this.options.editor!;
+        // Gets position of the node
+        const pos = typeof getPos === "function" ? getPos() : undefined;
```
YousefED, line 171: throw error if pos is undefined, so you don't need `pos!` below

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -110,16 +123,24 @@ export function createBlockSpec<
...
+          id: "",
+          type: "",
+          props: {},
+          content: [],
+          children: [],
+        } as Block<BlockSchema>);
 
       // Render elements
-      const rendered = blockConfig.render(props);
+      const rendered = blockConfig.render(getDummyBlock, editor);
```
YousefED, line 143: let's discuss this

YousefED:
superseded by https://github.com/TypeCellOS/BlockNote/pull/191

Issue #156, opened 2023-03-27
Website: Fix titles and image previews


YousefED:
currently, pasted links just show up like this:

<img width="668" alt="image" src="https://user-images.githubusercontent.com/368857/227912371-7dc3a01f-8de6-44b7-a78d-07d27bbdd13c.png">

We should at least be able to add a title / metadata and a standard og:image. 

Or, if we want to go more advanced a prerendered image:

https://vercel.com/docs/concepts/functions/edge-functions/og-image-generation

PR #187, opened 2023-05-05
fix: title and description

PhilipWillms:
Closes https://github.com/TypeCellOS/BlockNote/issues/156


PhilipWillms:
Discord example of how it looks now.
Screenshot 2023-05-05 at 19 55 54

PhilipWillms:
> This is great @PhilipWillms! Thanks ❤️
> 
> Two questions I hope you can help with now you've already figured this out:
> 
> * We can still override the title / description per page right? Because while this is a great improvement, all the pages will still have the same metadata
> * Same question for the og image. It's a bit overkill now to have it rendered like this, because there's no "dynamic" element. We might as well link to a static asset? Or, what would be even cooler is to add a dynamic title as well (like in [this example](https://vercel.com/docs/concepts/functions/edge-functions/og-image-generation/og-image-examples#dynamic-text-generated-as-image), but that might be quite a bit more work

I liked the idea of having images with dynamic titles so I implemented something like that. Now each page can have custom meta tags. I abstracted that in some way so that you do not need to set all meta tags on each page but rather just have to add some values in the frontmatter and the meta tags get created based on that. Not sure if we want it this way or rather be able to set all the meta tags for each site individually like it kinda got started in [this commented code](https://github.com/TypeCellOS/BlockNote/blob/main/packages/website/docs/index.md?plain=1#L4). 


I was not sure about how the preview image should look like. I did not like how it was looking when using the whole banner with text below so I chose to just use the Icon like this.
Screenshot 2023-05-07 at 18 34 15

 If no `imageTitle` got set the whole banner gets used as the `og:image` [like this](https://github.com/TypeCellOS/BlockNote/pull/187#issuecomment-1536589734)

packages/website/docs/.vitepress/config.ts
```diff
@@ -207,6 +232,75 @@ function getHeadTags(env: NodeJS.ProcessEnv): HeadConfig[] {
...
+        : METADATA_DEFAULT.image,
+    },
+  ]);
+
+  head.push([
+    "meta",
+    {
+      property: "twitter:title",
+      content:
+        `${METADATA_DEFAULT.title} - ${pageData.frontmatter.title}` ||
```
YousefED, line 280: I don't think this syntax will work? if `frontmatter.title` is undefined, I think it will just render `BlockNote - undefined` right?

(there's a couple more occurences in this file)

packages/website/api/og.tsx
```diff
@@ -0,0 +1,111 @@
...
+              gradientUnits="userSpaceOnUse">
+              <stop stop-color="#00EBE7" />
+              <stop offset="1" stop-color="#6923BA" />
+            </linearGradient>
+          </defs>
+        </svg>
+      </div>
+    );
+
+    return new ImageResponse(title ? iconWithTitle(title) : banner, {
```
YousefED, line 102: I would probably prefer banner + title. Or alternatively, a quick fix could be to prefix the title with `BlockNote - `. Currently, there's no way from the image to understand to which site / project the image refers to which I think is not ideal

packages/website/api/og.tsx
```diff
@@ -0,0 +1,111 @@
...
+              y2="93.632"
+              gradientUnits="userSpaceOnUse">
+              <stop stop-color="#00EBE7" />
+              <stop offset="1" stop-color="#6923BA" />
+            </linearGradient>
+          </defs>
+        </svg>
+        <div
+          style={{
+            fontSize: 60,
```
YousefED, line 49: Can we use the "Gilroy" font that's also used on the website? https://vercel.com/docs/concepts/functions/edge-functions/og-image-generation/og-image-examples#using-a-custom-font

packages/website/api/og.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { ImageResponse } from "@vercel/og";
+
+export const config = {
+  runtime: "edge",
+};
+
+export default async function handler(request) {
+  try {
+    const { searchParams } = new URL(request.url);
+    const title = searchParams.get("title")?.slice(0, 100);
```
YousefED, line 11: Add ellipsis? … (single utf character) if it's too long?

YousefED (commented):
This is great @PhilipWillms, thanks!! I've checked the code and it looks great. Some final suggestions in the review 💪

PR #188, opened 2023-05-06
fix: placeholder defined in code instead of css

achrefBoukhili:


packages/core/src/extensions/Blocks/nodes/Block.module.css
```diff
@@ -27,11 +27,11 @@ NESTED BLOCKS
   margin-left: 1.5em;
 }
 
-.blockGroup .blockGroup > .blockOuter {
```
matthewlipski, line 30: I think it's better to keep the formatting as it was. Admittedly the CSS isn't very readable at the moment, but it would make it much easier to keep track of the more important changes in the PR.

packages/core/src/extensions/Placeholder/localisation/translation.ts
```diff
@@ -0,0 +1,52 @@
...
+    translation: {
+      placeholder: "Geben Sie Text ein oder geben Sie '/' für Befehle ein",
+      heading: "Überschrift",
+      list: "Aufführen",
+    },
+  },
+  it: {
+    translation: {
+      placeholder: "Inserisci il testo o digita '/' per i comandi",
+      heading: "intestazione",
```
matthewlipski, line 20: Should be capitalized

matthewlipski (commented):
Thanks for your contribution! I'm not sure it makes sense to merge these changes at the moment though, since we're still not set on how we want to implement default styling for BlockNote. Right now we're ok with using CSS modules, though I think this might change. We'll probably revisit this when we're sorting out localization for the whole editor, including menus. @YousefED Thoughts?

YousefED:
This is great @achrefBoukhili! My thoughts:
- I think we should do i18n in a different PR. It's definitely a nice addition, but then I feel we should also translate other parts of the codebase (not just the placeholder).
- Why don't we pass the block type to `this.options.placeholder` and allow this placeholder function to be passed in from the BlockNoteEditor options? That way, everything is user customizable 🤙 (including different languages)

Let me know what you think!

packages/core/src/extensions/Placeholder/localisation/translation.ts
```diff
@@ -0,0 +1,52 @@
...
+      placeholder: "Enter text or type '/' for commands",
+      heading: "heading",
+      list: "List",
+    },
+  },
+  de: {
+    translation: {
+      placeholder: "Geben Sie Text ein oder geben Sie '/' für Befehle ein",
+      heading: "Überschrift",
+      list: "Aufführen",
```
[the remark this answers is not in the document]
schulzf: noticed that localization is misspelled (see path name)

achrefBoukhili:
@YousefED thanks a lot for the feedback and all what you said it makes a sense I think.
So far, let's keep this PR open until we define the final localization mechanism 🙏

YousefED:
Closing this. It's now possible to change the placeholders, although this still needs to be documented. We'll also add support for i18n soon

PR #191, opened 2023-05-10
feat: Custom Blocks

matthewlipski:
This PR enables users to add Custom Blocks! 🥳 

See https://www.blocknotejs.org/docs/block-types#custom-block-types

Besides the API surface, the biggest change was rearchitecting BlockNote for custom schemas without compromising on Type-safety. A BlockNote editor will now have a specific type (using generics) based on the Schema (the blocks registered to it). 

---
Two remaining issues will be picked up separately:

- [ ] work out API surface for cleaning up vanilla custom blocks #220 
- [ ] work on clipboardserializer #221 

(see also https://github.com/TypeCellOS/BlockNote/pull/164)


examples/custom/src/App.tsx
```diff
@@ -0,0 +1,96 @@
...
+    editorDOMAttributes: {
+      class: styles.editor,
+      "data-test": "editor",
+    },
+    blockSchema: {
+      ...defaultBlockSchema,
+      image: Image,
+    },
+    slashCommands: [
+      ...defaultReactSlashMenuItems<typeof customSchema>(),
```
YousefED, line 85: I'm not sure if passing the `schema` to `defaultReactSlashMenuItems` makes any sense

playwright-ct.config.ts
```diff
@@ -0,0 +1,46 @@
+import { defineConfig, devices } from "@playwright/experimental-ct-react";
```
YousefED, line 1: do we need a separate config, or can we just use the same config? (are there reasons to separate this?)

tests/component/example.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+//   });
+//
+//   // Give tests a way to get prosemirror instance
+//   // (window as WindowWithProseMirror).ProseMirror = editor?._tiptapEditor;
+//
+//   return <BlockNoteView editor={editor} />;
+// }
+
+test("basic editor", async ({ mount }) => {
+  const component = await mount(<App />);
```
YousefED, line 44: Nice, I think the way to go now is that in this directory, we can create simple example test components, for example EditorWithImage.tsx, or EditorWithoutDefaultBlocks.tsx, etc. i.e. we don't necessarily need to use `import App from "@blocknote/example-editor/src/App";`

examples/custom/src/App.tsx
matthewlipski: I feel like there must be a better way of doing it but I couldn't figure it out. Since the slash commands call editor functions, we need some way of passing the block schema to them.

playwright-ct.config.ts
matthewlipski: I would have thought so, but the Playwright docs say you should have separate configs for e2e and component tests: https://playwright.dev/docs/test-components#q-can-i-use-playwrighttest-and-playwrightexperimental-ct-reactsveltevuesolid

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -0,0 +1,259 @@
...
+        },
+      ]
+    : [
+        {
+          tag: "div[data-content-type=" + blockConfig.type + "]",
+        },
+      ];
+}
+
+export function render<
```
YousefED, line 85: let's add a comment on what this function does and for which scenarios it does / doesn't work

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -0,0 +1,259 @@
...
+              [camelToDataKebab(name)]: attributes[name],
+            }
+          : {},
+    };
+  });
+
+  return tiptapAttributes;
+}
+
+export function parse<
```
YousefED, line 52: similar as with `render`; let's comment

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -246,29 +251,34 @@ export function nodeToBlock(
...
     }
 
-    const validAttrs = blockProps[blockInfo.contentType.name as Block["type"]];
+    const propSchema = blockSpec.propSchema;
 
-    if (validAttrs.has(attr)) {
+    if (attr in propSchema) {
       props[attr] = value;
+    } else {
+      console.warn("Block has an unrecognized attribute: " + attr);
```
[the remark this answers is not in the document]
matthewlipski: Yep good point, since we treat ID as a separate thing to props in BN, I forgot that in TipTap it's still just an attribute. Will fix

packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
```diff
@@ -53,27 +54,29 @@ export class FormattingToolbarView {
 
   public prevWasEditable: boolean | null = null;
 
-  public shouldShow: Exclude<FormattingToolbarPluginProps["shouldShow"], null> =
-    ({ view, state, from, to }) => {
-      const { doc, selection } = state;
-      const { empty } = selection;
+  public shouldShow: Exclude<
+    FormattingToolbarPluginProps<BSchema>["shouldShow"],
```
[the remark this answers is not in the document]
matthewlipski: Yeah it's very odd, surprised we didn't see this earlier. `shouldShow` can either be a function or `null`, so it just makes sure its a function. But since, `shouldShow` is also optional, I think it's unnecessary

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -0,0 +1,259 @@
...
+  //  probably won't work and we'll need an optional parseHTML option.
+  return blockConfig.parse
+    ? [
+        {
+          getAttrs: (node: HTMLElement | string) => {
+            if (typeof node === "string") {
+              return false;
+            }
+
+            return blockConfig.parse!(node);
```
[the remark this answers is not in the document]
matthewlipski: Good point we don't, I added it a long time ago before we went down the whole serialization rabbit hole. None of the example use cases we came up with need a custom parser so I think it's safe to remove it until we deal with clipboard parsing, though I'd like to hear what you think about that also.

packages/react/src/hooks/useBlockNote.ts
```diff
@@ -41,7 +46,7 @@ export const useBlockNote = (
     //  but it would have to be on several different classes (BlockNoteEditor, BlockNoteEditorOptions, UiFactories) and
     //  gets messy quick.
     let newOptions: Record<any, any> = {
-      slashCommands: defaultReactSlashMenuItems,
+      slashCommands: defaultReactSlashMenuItems(),
```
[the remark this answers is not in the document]
matthewlipski: No point keeping it a function since it always returns the same thing. There are typing issues that I tried to solve by making it a function with a block schema type argument, but it was still bad. The issue I'm talking is this:

Example using custom alert block:
```
const editor = useBlockNote({
    blockSchema: {
      ...defaultBlockSchema,
      alert: Alert
    },
    slashCommands: [
      // Expects editor schema to be defaultBlockSchema. Structurally, it is, but also includes the 
      // alert block. TypeScript for some reason thinks that the schema is therefore incompatible
      // with the default slash commands.
      ...defaultReactSlashMenuItems,
      insertAlert
    ]
});
```

This will cause a type error in your IDE. Remove the alert block and its slash command, the error goes away. Not sure why this is happening since the custom schema extends the default schema.

tests/component/copypaste-internal.spec.tsx
```diff
@@ -0,0 +1,159 @@
...
+  await page.waitForTimeout(50);
+
+  const button = await page.locator("button");
+  await button.first().click();
+  await button.last().click();
+
+  await compareDocToSnapshot(page, "button-internal");
+});
+
+test("Embed Copy/Paste Internal", async ({ browserName, mount, page }) => {
```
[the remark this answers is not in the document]
matthewlipski: Nope, we've only really used screenshot tests before to make sure styling props and marks worked properly since that's the main thing they're for. Here imo we mostly care about the underlying TipTap/PM doc structure.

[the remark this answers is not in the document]
matthewlipski: I suppose to check rendering we could add screenshot tests but again imo it's not necessary

packages/react/src/ReactBlockSpec.tsx
```diff
@@ -0,0 +1,126 @@
...
+        const blockContainer = tipTapEditor.state.doc.resolve(pos!).node();
+        // Gets block identifier
+        const blockIdentifier = blockContainer.attrs.id;
+        // Function to get the block
+        const getBlock: () => Block<BSchema> = () =>
+          editor.getBlock(blockIdentifier)!;
+
+        return (
+          <NodeViewWrapper>
+            <div
```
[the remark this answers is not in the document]
YousefED: If I'm not mistaken, the `NodeViewWrapper` already adds an extra div. Can you validate that the DOM structure is now actually the same?

packages/core/src/extensions/Blocks/api/block.ts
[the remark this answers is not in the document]
YousefED: If the code is never run at this moment, let's remove it -> we can readd it if / when we need it when fixing the clipboard

tests/component/copypaste-external.spec.tsx
```diff
@@ -14,13 +14,16 @@ test.beforeEach(async ({}, testInfo) => {
   testInfo.snapshotSuffix = "";
 });
 
+// TODO: These tests should not be used for now. While they do pass, the
```
YousefED, line 17: wdym with "should not be used"?

tests/utils/customblocks/Alert.tsx
```diff
@@ -142,112 +137,3 @@ export const insertAlert = new ReactSlashMenuItem<{
   <RiAlertFill />,
   "Insert an alert block to emphasize text"
 );
-
-export const ReactAlert = createReactBlockSpec({
```
YousefED: where did this go?

tests/component/copypaste-external.spec.tsx
matthewlipski: The functionality they test doesn't work yet, so the tests and snapshots aren't useful atm

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -262,7 +263,12 @@ export function nodeToBlock<BSchema extends BlockSchema>(
 
     if (attr in propSchema) {
       props[attr] = value;
-    } else {
+    }
+    // Block ids are stored as node attributes the same way props are, so we
+    // need to ensure we don't attempt to read block ids as props. Also, props
+    // which apply to child blocks are set on the block container, but not all
+    // block types use them.
+    else if (attr !== "id" && !(attr in defaultProps)) {
```
YousefED, line 271: Why do we have `!(attr in defaultProps)`?

packages/react/src/hooks/useBlockNote.ts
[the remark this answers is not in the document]
YousefED: Tried to reproduce this in `custom-blocks-example-try-reproduce-type-issue` with a simpler example, but without luck.

Issue #182, opened 2023-04-29
Uncaught TypeError: Cannot read properties of undefined (reading 'size') - Prosemirror


chrisvasey:
Hello!

I just upgraded to v0.6.2 of the package and have had some issues using the package.
The editor seems to initialise but then throws an error in the prosemirror dep.

![Screenshot 2023-04-29 at 16 33 09](https://user-images.githubusercontent.com/7483460/235310819-9bfc8599-7d07-48fa-867e-d0dd2f023b46.png)

```
Uncaught TypeError: Cannot read properties of undefined (reading 'size')
    at NodeViewDesc.posBeforeChild (webpack-internal:///(:3000/app-client)/./node_modules/prosemirror-view/dist/index.js:726:24)
 ```
NPM Version: 8.11.0
Node Version: v16.16.0
Chrome: Version 112.0.5615.137 (arm64)/Mac M1, Ventura 13.3

Initially, seeing this in a React/Inertia.js project with SSR, I tried with SSR off and hit the same issues.
To test, I have spun up a new Nextjs (v13.3.1) project with the example from [this doc](https://www.blocknotejs.org/docs/editor#demo-saving-restoring-editor-contents) I am hitting the same issue.
I am rendering the Editor on the client side.

You can replicate the issue on [repo](https://github.com/chrisvasey/nextjs-blocknote-issue-182-example).
The editor component is here`/src/components/Editor.tsx` and is loaded on the root page of the project with this page component: `src/app/page.tsx`.

Is anyone else experiencing this issue on the most recent release?

I noticed a similar error in issue #133. Could this be related?

chrisvasey:
Just tested with v8.11.0 and Yarn v1.22.19 after seeing some discussion on Discord.
Same across both.

chrisvasey:
![Screenshot 2023-04-29 at 19 10 03](https://user-images.githubusercontent.com/7483460/235317784-669115e7-240a-43f1-9830-a7b2d5894350.png)

Yousef replied to me in Discord and recommended downgrading `prosemirror-view`.
I have tested this, and it works, I added `"prosemirror-view": "1.30.2"` to my projects package.json dependencies and the error is gone:
![Screenshot 2023-04-29 at 19 11 07](https://user-images.githubusercontent.com/7483460/235317821-fdef18ef-b74b-4c1d-8c19-2c94acf0d22f.png)

Great to have a workaround while we wait for a fix on Prosemirror

YousefED:
[...]

PR #193, opened 2023-05-15
chore: update dependencies

maxswjeon:
This updates tiptap dependencies to 2.0.3 in @blocknote/core and @blocknote/react, resolving #182.


maxswjeon:
Found out removing `node_modules` and reinstalling it solves the problem more easily.

PR #195, opened 2023-05-18
poc: custom blocks react

YousefED:
PoC of how react blocks could work. 

Maybe it's smart to re-use another React Nodeview bridge like:
- https://github.com/nytimes/react-prosemirror/blob/main/src/nodeViews/createReactNodeViewConstructor.tsx 
- https://github.com/Saul-Mirone/prosemirror-adapter 
- or https://remirror.io/blog/create-node-view-with-content/
 
to be more fool-proof and support more advanced scenarios


matthewlipski:
Looks good! Though I have one question, since TipTap already provides a React renderer for node views (https://tiptap.dev/guide/node-views/react), so why not just use that?

YousefED:
> Looks good! Though I have one question, since TipTap already provides a React renderer for node views (https://tiptap.dev/guide/node-views/react), so why not just use that?

whoops 😬 just me being stupid

packages/react/src/ReactBlockSpec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export function createReactBlockSpec<
+  BType extends string,
+  PSchema extends PropSchema,
+  ContainsInlineContent extends boolean
+>(
+  blockConfig: ReactBlockConfig<BType, PSchema, ContainsInlineContent>
+): BlockSpec<BType, PSchema> {
+  const rootEl = document.createElement("div");
+  const root = createRoot(rootEl);
```
timsu, line 30: From testing this code, it seems like it'll only support one instance of this react component at a time? I'm curious why create this element here instead of in the render block.

PR #200, opened 2023-05-22
feat: Multiplayer API and homepage collaboration

YousefED:
- clean package json (needed for proper yjs installation)
- added multiplayer / collaborative API
- added demo to homepage that runs on partykit!


packages/core/src/editor.module.css
```diff
@@ -37,7 +37,6 @@ Tippy popups that are appended to document.body directly
 .defaultStyles h2,
 .defaultStyles h3,
 .defaultStyles li {
-  all: unset !important;
```
YousefED, line 40: we can't remove this one, as it's used to make sure styles from the host page are not carried over into blocknote

PR #202, opened 2023-05-23
Custom blocks example typing improvements

YousefED:
I managed to at least get the correct type of `block` in the render functions. Unfortunately, getting the correct type of "editor" is impossible afaik, unless we decide to separate the "render" functions from the other spec definitions. Theoretically, it's not necessary to define them at the same time as they serve different purposes

Upgraded TS as part of this process to make sure I wasn't running into issues that have meanwhile been fixed


packages/core/src/extensions/Blocks/api/blockTypes.ts
```diff
@@ -73,8 +73,21 @@ export type BlockConfig<
   containsInlineContent: ContainsInlineContent;
   parse?: (element: HTMLElement) => Props<PSchema>;
   render: (
-    block: Block<BSchema>,
-    editor: BlockNoteEditor<BSchema>
+    /**
+     * The custom block to render
+     */
+    block: SpecificBlock<
```
matthewlipski, line 79: Could this be simplified to
```
block: Block<{ [k in Type]: BlockSpec<Type, PSchema> }>
```
?

matthewlipski: Also as a side note - right now you could, if you wanted to, update `block` to a different type. Not relevant for now, but will be when we need to sort out cleaning up listeners, since I think that a listener for a custom block should be removed if the block changes type.

packages/core/src/extensions/Blocks/api/blockTypes.ts
```diff
@@ -141,6 +149,13 @@ export type Block<BSchema extends BlockSchema> =
     children: Block<BSchema>[];
   };
 
+export type SpecificBlock<
```
matthewlipski, line 152: Might be worth adding a comment for a bit of clarification (the other types are now also commented on the `custom-blocks-example` branch)

packages/core/src/extensions/Blocks/api/blockTypes.ts
```diff
@@ -73,8 +73,21 @@ export type BlockConfig<
...
+    block: SpecificBlock<
+      BSchema & { [k in Type]: BlockSpec<Type, PSchema> },
+      Type
+    >,
+    /**
+     * The BlockNote editor instance
+     * This is typed generically. If you want an editor with your custom schema, you need to
+     * cast it manually, e.g.: `const e = editor as BlockNoteEditor<typeof mySchema>;`
+     */
+    editor: BlockNoteEditor<BSchema & { [k in Type]: BlockSpec<Type, PSchema> }>
```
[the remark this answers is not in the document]
YousefED: Yes, and BlockSchema was super generic. Now, at least we know it countains the type of the block being added

packages/core/src/extensions/Blocks/api/blockTypes.ts
YousefED: I don't think it can, because the `children` of the block should be able to be of other types

matthewlipski: Ah, good point, missed that

PR #210, opened 2023-05-25
fix: Address React Warning when clicking DragHandleMenuItem

fogle:
Filter out closeMenu from the props so React stops warning about it.

To generate the warning without this change, simply click on the drag handle menu with your dev console open.

Screenshot 2023-05-24 at 8 26 20 PM


YousefED (approved):
Great, thanks!

PR #217, opened 2023-05-30
Custom blocks docs

matthewlipski:
This PR adds documentation for custom blocks and fixes a few smaller issues with them. There are still a few things that need changing:

- [ ] Automatically setting `contentEditable="false"` on all elements except the `contentDOM`/`InlineContent` and its ancestors (https://discuss.prosemirror.net/t/contenteditable-islands-prevents-selection/3938).
- [ ] Finalize naming for types & functions.
- [ ] Make it so you don't have to call `createBlockSpec` manually? Not sure if this is something we want since we probably want to keep the ability to create BlockSpecs from scratch and add them to the schema.

### DIFF
--- a/packages/react/src/FormattingToolbar/components/DefaultButtons/TextAlignButton.tsx
+++ b/packages/react/src/FormattingToolbar/components/DefaultButtons/TextAlignButton.tsx
@@ -1,9 +1,10 @@
 import {
   BlockNoteEditor,
-  DefaultBlockSchema,
+  BlockSchema,
   DefaultProps,
+  PartialBlock,
 } from "@blocknote/core";
-import { useCallback } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { IconType } from "react-icons";
 import {
   RiAlignCenter,
@@ -22,10 +23,31 @@ const icons: Record<TextAlignment, IconType> = {
   justify: RiAlignJustify,
 };
 
-export const TextAlignButton = (props: {
-  editor: BlockNoteEditor<DefaultBlockSchema>;
+export const TextAlignButton = <BSchema extends BlockSchema>(props: {
+  editor: BlockNoteEditor<BSchema>;
   textAlignment: TextAlignment;
 }) => {
+  const [show, setShow] = useState(true);
+
+  useEffect(() => {
+    const selection = props.editor.getSelection();
+
+    if (selection) {
+      for (const block of selection.blocks) {
+        if (!("textAlignment" in block.props)) {
+          setShow(false);
+        }
+      }
+    } else {
+      const block = props.editor.getTextCursorPosition().block;
+      console.log(block);
+
+      if (!("textAlignment" in block.props)) {
+        setShow(false);
+      }
+    }
+  }, [props.editor]);
+
   const setTextAlignment = useCallback(
     (textAlignment: TextAlignment) => {
       props.editor.focus();
@@ -36,16 +58,22 @@ export const TextAlignButton = (props: {
         for (const block of selection.blocks) {
           props.editor.updateBlock(block, {
             props: { textAlignment: textAlignment },
-          });
+          } as PartialBlock<BSchema>);
         }
       } else {
-        props.editor.updateBlock(props.editor.getTextCursorPosition().block, {
+        const block = props.editor.getTextCursorPosition().block;
+
+        props.editor.updateBlock(block, {
           props: { textAlignment: textAlignment },
-        });
+        } as PartialBlock<BSchema>);
       }
     },
     [props.editor]
   );
+
+  if (!show) {
+    return null;
+  }
 
   return (
     <ToolbarButton
--- a/packages/website/docs/docs/block-types.md
+++ b/packages/website/docs/docs/block-types.md
@@ -110,19 +110,25 @@ In addition to the default block types that BlockNote offers, you can also make 
 import {
   BlockNoteEditor,
   defaultBlockSchema,
+  defaultProps,
 } from "@blocknote/core";
 import {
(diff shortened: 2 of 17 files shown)


packages/react/src/FormattingToolbar/components/DefaultButtons/TextAlignButton.tsx
```diff
@@ -22,10 +23,30 @@ const icons: Record<TextAlignment, IconType> = {
...
 
-export const TextAlignButton = (props: {
-  editor: BlockNoteEditor<DefaultBlockSchema>;
+export const TextAlignButton = <BSchema extends BlockSchema>(props: {
+  editor: BlockNoteEditor<BSchema>;
   textAlignment: TextAlignment;
 }) => {
+  const [show, setShow] = useState(true);
+
+  useEffect(() => {
```
YousefED, line 32: I think this is cleaner as `useMemo`, and get rid of `useState`. Because you never call `setShow` outside of the effect anyway

packages/website/docs/docs/block-types.md
```diff
@@ -102,4 +102,244 @@ type DefaultBlockProps = {
...
+  InlineContent,
+  ReactSlashMenuItem,
+  defaultReactSlashMenuItems,
+} from "@blocknote/react";
+import "@blocknote/core/style.css";
+import { RiImage2Fill } from "react-icons/ri";
+
+export default function App() {
+  // Configuration for a custom image block.
+  const ImageBlockConfig = {
```
YousefED, line 128: add `createReactBlockSpec` here

packages/website/docs/docs/block-types.md
```diff
@@ -102,4 +102,244 @@ type DefaultBlockProps = {
...
+```typescript
+type PropSchema = Record<
+  string,
+  {
+    default: string;
+    values?: string[];
+  };
+>
+
+type ReactBlockConfig<
```
YousefED, line 219: I think it will be a lot easier if we keep the term `BlockConfig` internally, or only in the context of `createReactBlockSpec`. I think we should focus first on saying you need to create a new block spec using `createReactBlockSpec`, and that that function takes an argument of:

```typescript
type: Type;
  propSchema: PropSchema;
  containsInlineContent: boolean;
  render: (props: {
    block: Block,
    editor: BlockNoteEditor
  }) => JSX.Element;
```

packages/website/docs/docs/block-types.md
```diff
@@ -102,4 +102,244 @@ type DefaultBlockProps = {
...
+
+For our custom image block, we use a parent `div` which contains the image and caption. Since `block` will always be an `image` block, we also know it contains a `src` prop, and can pass it to the child `img` element.
+
+But what's this `InlineContent` component? Since we set `containsInlineContent` to `true`, it means we want to include an editable rich-text field somewhere in the image block. You should use the `InlineContent` component to represent this field in your `render` component. Since we're using it to create our caption, we add it below the `img` element.
+
+While the `InlineContent` component can be put anywhere inside the component you pass to `render`, you should make sure to only have one. If `containsInlineContent` is set to false, `render` shouldn't contain any.
+
+### Adding Custom Blocks to the Editor
+
+After creating `BlockConfig`s for all of our custom blocks, we need to convert them into `BlockSpec`s. A `BlockSpec` just tells BlockNote how your custom block should be defined internally, so we don't need to know much about it. We just need to convert our `BlockConfig` to a `BlockSpec` using the `createReactBlockSpec` function:
```
YousefED, line 299: remove this (related to above)

packages/website/docs/docs/block-types.md
```diff
@@ -102,4 +102,244 @@ type DefaultBlockProps = {
...
+  blockSchema: {
+    // Adds all default blocks.
+    ...defaultBlockSchema,
+    // Adds the custom image block.
+    image: ImageBlock,
+  }
+})
+```
+
+Notice three details about this code snippet. First, since we still want the editor to use the [Built-In Block Types](/docs/block-types#built-in-block-types), we add `defaultBlockSchema` to our custom block schema. Second, the key which we use for the custom image block is the same string we use for its type. Make sure that this is always the case for your own custom blocks. Finally, we provide the type of our schema as a type argument to `BlockNoteEditor`.
```
YousefED, line 326: is the last part necessary? I thought it would infer the type automatically

packages/website/docs/docs/block-types.md
```diff
@@ -102,4 +102,244 @@ type DefaultBlockProps = {
...
+
+And we're done! You now know how to create custom blocks and add them to the editor. Head to [Manipulating Blocks](/docs/manipulating-blocks) to see what you can do with them in the editor.
+
+### Limitations
```
YousefED, line 330: Let's replace the entire Limitations (including the notes on multiple rich text fields and contenteditable) to an "alert"

> While custom blocks open a lot of doors for what you can do with BlockNote, we're still working on improving the API and there are a few limitations for the kinds of blocks you can create.
> We'd love to hear your feedback on Github or in our Discord community!

Reasoning: we don't know yet what people will run into / which use-cases they envision. The below examples might only scare them away

packages/website/docs/docs/block-types.md
matthewlipski: For pretty much everything else the type is inferred automatically, but the type argument for `BlockNoteEditor` defaults to `DefaultBlockSchema`, so if no argument is provided, it uses that instead of inferring the type. Maybe that's something we should change?

YousefED: apparently typescript is smart enough for this. I'll remove the type annotation from the other branch

PR #229, opened 2023-06-01
fix: Error caused by converting `hardBreak` nodes

### LINKED ISSUE #184: RangeError: Empty text nodes are not allowed (on editor.blocksToMarkdown)
Hi there!

Great project! I really like the UI and the direction you're going.

I played a bit around in a NextJS project and found a couple of quirky issues that might have gone unnoticed. I'll report them in separate issues but here is the first one.

```
Unhandled Runtime Error
RangeError: Empty text nodes are not allowed

Source
new TextNode
node_modules/prosemirror-model/dist/index.js (1533:0)
Schema.text
node_modules/prosemirror-model/dist/index.js (2415:0)
Qt
node_modules/@blocknote/core/dist/blocknote.js (245:0)
ut
node_modules/@blocknote/core/dist/blocknote.js (258:0)
ht
node_modules/@blocknote/core/dist/blocknote.js (267:0)
R
node_modules/@blocknote/core/dist/blocknote.js (286:0)
mt
node_modules/@blocknote/core/dist/blocknote.js (498:0)
ie
node_modules/@blocknote/core/dist/blocknote.js (515:76)
No.blocksToMarkdown
node_modules/@blocknote/core/dist/blocknote.js (2801:0)
components/md-editor.tsx (68:46) @ blocksToMarkdown

  66 | // saves them.
  67 | const saveBlocksAsMarkdown = async () => {
> 68 |   const markdown: string = await editor.blocksToMarkdown(
     |                                        ^
  69 |     editor.topLevelBlocks
  70 |   );
  71 |   setValue(markdown);
```
  
  I followed the documentation on  converting blocks to Markdown with this code snippet:

```
// Stores the editor's contents as Markdown.
  const [markdown, setMarkdown] = useState<string>("");

  // Creates a new editor instance.
  const editor: BlockNoteEditor | null = useBlockNote({
    // Listens for when the editor's contents change.
    onEditorContentChange: (editor: BlockNoteEditor) => {
      // Converts the editor's contents from Block objects to Markdown and 
      // saves them.
      const saveBlocksAsMarkdown = async () => {
        const markdown: string = 
          await editor.blocksToMarkdown(editor.topLevelBlocks);
        setMarkdown(markdown);
      };
      saveBlocksAsMarkdown();
    }
  });
```
  
At first everything seemed to work as intended, but when copying a couple of websites to the editor (which converted the HTML to blocks in the background (excellent!)) in some cases I got the above error message.

Indeed, when investigating the block structure after pasting, I found this content block:
```
  {
    "id": "70b1274a-c071-4311-b4f2-9a854190406b",
    "type": "paragraph",
    "props": {
      "textColor": "default",
      "backgroundColor": "default",
      "textAlignment": "left"
    },
    "content": [
      { "type": "text", "text": "sometext", "styles": { "bold": true } },
      { "type": "text", "text": "", "styles": { "bold": true } }
    ],
    "children": []
  },
```
I am not sure what the HTML structure looked like that caused this. However, this doesn't seem to be intended. 

I could solve the issue with manually checking for empty content nodes like this:  
```
const removeEmptyTextBlocks = (blocks: Block[]) => {
  // Visit all nodes
  for (let i = 0; i < blocks.length; i++) {
    const block = blocks[i];

    // Recursively remove empty text nodes from children
    if (block.children) {
      removeEmptyTextBlocks(block.children);
    }

    // Remove empty text nodes from content
    if (block.content) {
      const filteredContent = block.content.filter(
        (contentNode) =>
          !(contentNode.type === 'text' && contentNode.text === '')
      );
      block.content = filteredContent.length ? filteredContent : [];
    }
  }
};

    // Listens for when the editor's contents change.
    onEditorContentChange: (editor: BlockNoteEditor) => {
      // Converts the editor's contents from Block objects to Markdown and
      // saves them.
      const saveBlocksAsMarkdown = async () => {
        const blocks = editor.topLevelBlocks;
        removeEmptyTextBlocks(blocks);

        const markdown: string = await editor.blocksToMarkdown(blocks);
        setValue(markdown);
      };
      saveBlocksAsMarkdown();
    }

```

Probably though this problem should be fixed either in the `editor.blocksToMarkdown` function or whatever function is responsible for pasting HTML. Or even a check when creating new content blocks. I am not too familiar with your code base, so I am not creating a pull request myself.

Keep up the great work!

### LINKED ISSUE #216: [bug]Empty text nodes are not allowed
I encountered an error when reloading saved JSON data after using shift+enter to insert a line break in the text.

Error prompt: Empty text nodes are not allowed.

The input I entered was: Test Report 2023-05-30 (use shift+enter to line break after the date)
An additional {"type":"text","text":"","styles":{}} was included, which caused the JSON to fail to load.

Here is the JSON data I received:

[{"id":"b1250752-20c3-487b-b6ce-70fbbda51673","type":"heading","props":{"textColor":"default","backgroundColor":"blue","textAlignment":"center","level":"2"},"content":[{"type":"text","text":"Test Report","styles":{}},`{"type":"text","text":"","styles":{}}`,{"type":"text","text":"2023-05-30","styles":{}}],"children":[]},{"id":"9d682c8e-d27f-4e26-ac8a-021d852b9969","type":"paragraph","props":{"textColor":"default","backgroundColor":"default","textAlignment":"left"},"content":[],"children":[]}]

// Add a '\n' in the text   {"type":"text","text":"\n","styles":{}} , it works fine .

matthewlipski:
When you press Shift+Tab in a block, it creates a new line in the same block which TIpTap/ProseMirror store as a `hardBreak` node. It turns out we completely forgot about these when doing inline content conversions, so this PR remedies that.

This also fixes the `Empty text nodes are not allowed` error that some people have been getting, mentioned in issues #184 and #216.

Closes #184
Closes #216

### DIFF
--- a/packages/core/src/api/nodeConversions/nodeConversions.ts
+++ b/packages/core/src/api/nodeConversions/nodeConversions.ts
@@ -76,7 +76,10 @@ function linkToNodes(link: PartialLink, schema: Schema): Node[] {
       return node.mark([...node.marks, linkMark]);
     }
 
-    return node;
+    if (node.type.name === "hardBreak") {
+      return node;
+    }
+    throw new Error("unexpected node type");
   });
 }
 
@@ -189,20 +192,18 @@ function contentNodeToInlineContent(contentNode: Node) {
     // hardBreak nodes do not have an InlineContent equivalent, instead we
     // add a newline to the previous node.
     if (node.type.name === "hardBreak") {
-      // Current content exists.
       if (currentContent) {
-        // Current content is text.
+        // Current content exists.
         if (currentContent.type === "text") {
+          // Current content is text.
           currentContent.text += "\n";
-        }
-        // Current content is a link.
-        else if (currentContent.type === "link") {
+        } else if (currentContent.type === "link") {
+          // Current content is a link.
           currentContent.content[currentContent.content.length - 1].text +=
             "\n";
         }
-      }
-      // Current content does not exist.
-      else {
+      } else {
+        // Current content does not exist.
         currentContent = {
           type: "text",
           text: "\n",
@@ -233,16 +234,15 @@ function contentNodeToInlineContent(contentNode: Node) {
     if (currentContent) {
       // Current content is text.
       if (currentContent.type === "text") {
-        // Node is text (same type as current content).
         if (!linkMark) {
-          // Styles are the same.
+          // Node is text (same type as current content).
           if (
             JSON.stringify(currentContent.styles) === JSON.stringify(styles)
           ) {
+            // Styles are the same.
             currentContent.text += node.textContent;
-          }
-          // Styles are different.
-          else {
+          } else {
+            // Styles are different.
             content.push(currentContent);
             currentContent = {
               type: "text",
@@ -250,9 +250,8 @@ function contentNodeToInlineContent(contentNode: Node) {
               styles,
             };
           }
-        }
-        // Node is a link (different type to current content).
-        else {
+        } else {
+          // Node is a link (different type to current content).
           content.push(currentContent);
           currentContent = {
             type: "link",
@@ -266,11 +265,10 @@ function contentNodeToInlineContent(contentNode: Node) {
             ],
           };
         }
-      }
-      // Current content is a link.
-      else if (currentContent.type === "link") {
-        // Node is a link (same type as current content).
+      } else if (currentContent.type === "link") {
+        // Current content is a link.
         if (linkMark) {
+          // Node is a link (same type as current content).
           // Link URLs are the same.
           if (currentContent.href === linkMark.attrs.href) {
             // Styles are the same.
@@ -281,18 +279,16 @@ function contentNodeToInlineContent(contentNode: Node) {
             ) {
               currentContent.content[currentContent.content.length - 1].text +=
                 node.textContent;
-            }
-            // Styles are different.
-            else {
+            } else {
+              // Styles are different.
               currentContent.content.push({
                 type: "text",
                 text: node.textContent,
                 styles,
               });
             }
-          }
-          // Link URLs are different.
-          else {
+          } else {
+            // Link URLs are different.
             content.push(currentContent);
             currentContent = {
               type: "link",
@@ -306,9 +302,8 @@ function contentNodeToInlineContent(contentNode: Node) {
               ],
             };
           }
-        }
-        // Node is text (different type to current content).
-        else {
+        } else {
+          // Node is text (different type to current content).
           content.push(currentContent);
           currentContent = {
             type: "text",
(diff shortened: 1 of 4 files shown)


YousefED:
Nice find!!

I think programmability wise, it would definitely be easier if we could just represent it with `\n` as you suggested. Do you think this will be doable? Maybe we should first double check what prosemirror does when you pass \n in text of a node - just to make sure we won't break any other already existing behavior around \n

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -30,11 +29,32 @@ const toggleStyles = new Set<ToggledStyle>([
 ]);
 const colorStyles = new Set<ColorStyle>(["textColor", "backgroundColor"]);
 
+function splitTextByNewline(text: string): string[] {
```
YousefED, line 32: why not use text.split("\n")?

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -163,28 +187,42 @@ export function blockToNode<BSchema extends BlockSchema>(
...
-        currentLink = undefined;
+      // Current content exists.
+      if (currentContent) {
+        // Current content is text.
+        if (currentContent.type === "text") {
+          currentContent.text += "\n";
+        }
+        // Current content is a link.
+        else if (currentContent.type === "link") {
+          const endText = currentContent.content.pop();
```
YousefED, line 206: why pop and push? can just do `currentContent.content.text += "\n"`?

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -182,37 +235,128 @@ function contentNodeToInlineContent(contentNode: Node) {
...
-        styles,
-      });
-    } else if (linkMark) {
-      // if the node is a link that doesn't match the current link, create a new link
-      currentLink = {
-        type: "link",
-        href: linkMark.attrs.href,
-        content: [
-          {
+    // Parsing links and text.
```
YousefED, line 238: can you explain these changes? e.g.: why are so many changes necessary?

YousefED (changes requested):
added some questions. Also, can you add unit tests similar like we have for the links?

packages/core/src/api/nodeConversions/nodeConversions.ts
matthewlipski: Since I had to change `currentLink` to `currentContent`, which can be either a link or styled text, I had to add more conditions to the existing code, which ended up being really difficult to read and understand.

YousefED (approved):
I moved the comments to inside the branches, imo more readable.

Besides that I think it's good to go

Issue #185, opened 2023-05-01
Block Side Menu still visible on hover when editor parent has visibility:hidden


empiriker:
I used the BlockNote editor in my NextJS project where I placed the editor inside the content of an MUI Accordion. 

When the Accordion is closed, I still get the BlockSideMenu when hovering over the (invisible) blocks in the (invisible) editor.

Here is a screenshot of when my mouse hover over where the block element would be:

![image](https://user-images.githubusercontent.com/46843049/235469917-0f737aaa-3e64-44f9-ac8b-551481d959f5.png)

The fact that it's an MUI Accordion doesn't matter. It's still the same behaviour whenever the editor is a descendent of an element with visibility:hidden.

Here is a minimal example:
```
import { BlockNoteEditor } from '@blocknote/core';
import { BlockNoteView, useBlockNote } from '@blocknote/react';
import '@blocknote/core/style.css';

const App = () => {
  // Creates a new editor instance.
  const editor: BlockNoteEditor | null = useBlockNote({});

  // Renders the editor instance using a React component.
  return (
    <div style={{ visibility: 'hidden' }}>
      <BlockNoteView editor={editor} />
    </div>
  );
};

export default App;
```

My hacky solution is to use a useEffect hook based on the property that controls the accordion state (and thereby the css property visibility:hidden of the editors ancestor). 

  ```
useEffect(() => {
    const tippyElement: HTMLElement | null = document.querySelector('#tippy-1');
    if (tippyElement && !expanded) {
      tippyElement.style.visibility = 'hidden';
    }
  }, [expanded]);

```
But ideally this issue should be dealt with by BlockNote itself.

Thanks!

Edit: Issue #172 should probably fix what I describe here as well.

YousefED:
Thanks, we'll look into this!

PR #230, opened 2023-06-02
fix: Side menu showing when editor isn't focused.

### LINKED ISSUE #201: mousemove event kept executed, even when not needed.
ezgif com-video-to-gif

As you can see in above gif, mousemove event is kept called even when cursor is on other component.
I wonder if this is intended. If it's not, how can I stop mousemove event to be called when the cursor is on other components?

matthewlipski:
It probably doesn't make sense to show the block side menu if the editor isn't focused, since the user could be interacting with elements somewhere completely different in the page and it's just distracting to have the side menu moving around. See issue #201 for an example of this. This PR makes it so that the side menu is hidden when the editor or its menus aren't focused.

There is another approach to this, i.e. exposing a `forceSideMenuHide` function in the API, though imo the former solution is more intuitive and requires less from consumers.

Closes #201
Also helps with #185, but doesn't solve the issue completely

### DIFF
--- a/packages/core/src/BlockNoteEditor.ts
+++ b/packages/core/src/BlockNoteEditor.ts
@@ -154,6 +154,10 @@ export class BlockNoteEditor<BSchema extends BlockSchema = DefaultBlockSchema> {
     return this._tiptapEditor.view.dom as HTMLDivElement;
   }
 
+  public isFocused() {
+    return this._tiptapEditor.view.hasFocus();
+  }
+
   public focus() {
     this._tiptapEditor.view.focus();
   }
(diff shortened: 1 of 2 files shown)


YousefED:
I'm not sure if I agree with this in terms of UX. For example, if I check the preview in https://blocknote-website-git-side-menu-behaviour-change-typecell.vercel.app, when I'm browsing the website, and hovering over the editor, the side menu would normally give me an indication that I can edit the document and / or take certain actions

matthewlipski:
Do you think adding functions to e.g. force hide the menus would make more sense? I agree that it's nice to have the side menu appear even when the editor isn't focused, but I don't think there's a different condition that it makes sense to use if we automatically want to hide it in cases like #201.

matthewlipski:
Ok so this currently breaks because the side menu itself is not a descendant of the editor. To get the current behaviour, AFAIK the only way to fix this is to make the side menu a descendant of the editor. So fixing #233 will also fix this PR and vice-versa

PR #235, opened 2023-06-06
Fixed dependency using `require`

matthewlipski:
The `rehype-parse` dependency hasn't been updated in a while and a sub-dependency, `parse5`, has moved from `require` to `import` statements in the meantime. This PR overrides `parse5` to use a newer version so that Node JS is not required.


YousefED:
Closing issue until we can reproduce it

YousefED:
Addressed in https://github.com/TypeCellOS/BlockNote/pull/274

PR #250, opened 2023-06-19
fix: Scrolling responsiveness & UI element overflow

matthewlipski:
Right now, every time we scroll, each `uiElement` is re-rendered. In general, I do think we want to leave the responsibility of figuring out when to render/hide `uiElement`s completely up to BlockNote, Tippy has it's own listeners to figure out when it should update its position. Trying to re-render the whole component on scroll instead of letting Tippy sort it out itself leads to some issues, namely overflow handling and bad responsiveness.

https://discord.com/channels/928190961455087667/1015169282444894219/1120228729600348181
https://discord.com/channels/928190961455087667/1015169282444894219/1120355999828680795

### DIFF
--- a/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
+++ b/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
@@ -250,6 +250,8 @@ export class BlockMenuView<BSchema extends BlockSchema> {
   menuOpen = false;
   menuFrozen = false;
 
+  private lastPosition: DOMRect | undefined;
+
   constructor({
     tiptapEditor,
     editor,
@@ -271,9 +273,6 @@ export class BlockMenuView<BSchema extends BlockSchema> {
 
     // Shows or updates menu position whenever the cursor moves, if the menu isn't frozen.
     document.body.addEventListener("mousemove", this.onMouseMove, true);
-
-    // Makes menu scroll with the page.
-    document.addEventListener("scroll", this.onScroll);
 
     // Hides and unfreezes the menu whenever the user selects the editor with the mouse or presses a key.
     // TODO: Better integration with suggestions menu and only editor scope?
@@ -463,20 +462,6 @@ export class BlockMenuView<BSchema extends BlockSchema> {
     }
   };
 
-  onScroll = () => {
-    // Editor itself may have padding or other styling which affects size/position, so we get the boundingRect of
-    // the first child (i.e. the blockGroup that wraps all blocks in the editor) for a more accurate bounding box.
-    const editorBoundingBox = (
-      this.ttEditor.view.dom.firstChild! as HTMLElement
-    ).getBoundingClientRect();
-
-    this.horizontalPosAnchor = editorBoundingBox.x;
-
-    if (this.menuOpen) {
-      this.blockMenu.render(this.getDynamicParams(), false);
-    }
-  };
-
   destroy() {
     if (this.menuOpen) {
       this.menuOpen = false;
@@ -487,7 +472,6 @@ export class BlockMenuView<BSchema extends BlockSchema> {
     this.ttEditor.view.dom.removeEventListener("dragstart", this.onDragStart);
     document.body.removeEventListener("drop", this.onDrop);
     document.body.removeEventListener("mousedown", this.onMouseDown);
-    document.removeEventListener("scroll", this.onScroll);
     document.body.removeEventListener("keydown", this.onKeyDown);
   }
 
@@ -558,11 +542,12 @@ export class BlockMenuView<BSchema extends BlockSchema> {
       },
       getReferenceRect: () => {
         if (!this.menuOpen) {
-          return;
+          return this.lastPosition;
         }
 
         const blockContent = this.hoveredBlock!.firstChild! as HTMLElement;
         const blockContentBoundingBox = blockContent.getBoundingClientRect();
+        this.lastPosition = blockContentBoundingBox;
 
         return new DOMRect(
           this.horizontalPosAnchoredAtRoot
--- a/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
+++ b/packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
@@ -556,6 +556,23 @@ export class BlockMenuView<BSchema extends BlockSchema> {
       unfreezeMenu: () => {
         this.menuFrozen = false;
       },
+      getReferenceRect: () => {
+        if (!this.menuOpen) {
+          return;
+        }
+
+        const blockContent = this.hoveredBlock!.firstChild! as HTMLElement;
+        const blockContentBoundingBox = blockContent.getBoundingClientRect();
+
+        return new DOMRect(
+          this.horizontalPosAnchoredAtRoot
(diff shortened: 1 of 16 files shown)


packages/core/src/extensions/HyperlinkToolbar/HyperlinkToolbarPlugin.ts
```diff
@@ -255,6 +255,14 @@ class HyperlinkToolbarView {
...
         this.hyperlinkToolbar.hide();
       },
+      getReferenceRect: () =>
+        this.hyperlinkMark
+          ? posToDOMRect(
+              this.editor.view,
+              this.hyperlinkMarkRange!.from,
+              this.hyperlinkMarkRange!.to
+            )
+          : undefined,
```
YousefED, line 265: when would this be undefined?

is this expected, or should it throw an error?

packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
```diff
@@ -195,6 +195,17 @@ class SuggestionPluginView<
   getStaticParams(): SuggestionsMenuStaticParams<T> {
     return {
       itemCallback: (item: T) => this.itemCallback(item),
+      getReferenceRect: () => {
+        const decorationNode = document.querySelector(
+          `[data-decoration-id="${this.pluginState.decorationId}"]`
+        );
+
+        if (!decorationNode) {
```
YousefED, line 203: is this expected, or should it throw an error?

packages/react/src/ElementFactory/components/ReactElementFactory.tsx
```diff
@@ -30,15 +34,43 @@ export const ReactElementFactory = <
...
+  let isOpen = false;
 
   return {
     element: rootElement,
     render: (dynamicParams: ElementDynamicParams, _isHidden: boolean) => {
+      // Tippy should handle the reference bounding box changing positions
+      // itself, otherwise it's less responsive and isn't able to handle
+      // page overflows. Therefore, we want to avoid re-rendering the element
+      // when only its position changes.
+      const prevDynamicParamsCopy: Record<string, any> = {};
```
YousefED, line 50: this code is not very readable and using JSON.stringify for this is hacky.

Can't we just loop through all keys and compare them for equality (except `referenceRect`)?

packages/core/src/extensions/HyperlinkToolbar/HyperlinkToolbarPlugin.ts
YousefED: do we still need `referenceRect` in dynamicParams?

YousefED (commented):
Trying to understand why this works;

The theory is that Tippy calls getReferenceRect() at the "right" moment, and that previously the data that we return to getReferenceRect can be outdated?

To understand it:
- what are the scenarios where the data passed was outdated?
- at which moments does tippy call `getReferenceRect` that fixes the issues?

Are we only talking about `onScroll`? and does Tippy attach it's own listeners to window?

packages/core/src/extensions/HyperlinkToolbar/HyperlinkToolbarPlugin.ts
matthewlipski: It's expected, returns `undefined` only when the element is hidden, so the bounding box for it doesn't actually exist and it doesn't make sense to try to access it.

packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
matthewlipski: This one is also ok, again if `decorationNode` is `undefined`, it means the element is hidden so the bounding box for it doesn't exist. That means the Tippy is also hidden so it doesn't try to listen for position changes.

packages/react/src/ElementFactory/components/ReactElementFactory.tsx
matthewlipski: True, that probably makes more sense

matthewlipski:
> Trying to understand why this works;
> 
> The theory is that Tippy calls getReferenceRect() at the "right" moment, and that previously the data that we return to getReferenceRect can be outdated?
> 
> To understand it:
> 
> * what are the scenarios where the data passed was outdated?
> * at which moments does tippy call `getReferenceRect` that fixes the issues?
> 
> Are we only talking about `onScroll`? and does Tippy attach it's own listeners to window?

I think the issue wasn't necessarily due to outdated data, moreso that the entire component was being re-rendered on scroll.  This would explain why you couldn't scroll down to see it if it was overflowing (since each scroll event would render a new component) and ofc why it wasn't as responsive.

I'm not 100% sure how Tippy's `getReferenceClientRect` works, but it must have its own listeners as it can figure out when to update the popup's position on its own, you just have to provide the function which gets the current reference element position. According to Tippy docs, this should happen whenever the reference element moves, including on scroll.

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
```diff
@@ -556,23 +540,30 @@ export class BlockMenuView<BSchema extends BlockSchema> {
       unfreezeMenu: () => {
         this.menuFrozen = false;
       },
+      getReferenceRect: () => {
+        if (!this.menuOpen) {
+          return this.lastPosition;
```
YousefED, line 545: lastPosition is the `blockContentBoundingBox`, but if we don't use lastPosition we create a new DOMRect. This looks inconsistent

packages/react/src/ElementFactory/components/EditorElementComponentWrapper.tsx
```diff
@@ -55,8 +50,11 @@ export function EditorElementComponentWrapper<
             />
           ) : undefined
         }
+        // Type cast is needed as getReferenceRect will return `undefined` when
```
YousefED, line 53: in that case, we never expect it to be called right? i.e.: getReferenceRect() should only be called when it can return a value?

If that's the case, it's better to throw an Error (as it's an unexpected situation), this will also allow you to change the return type to remove `undefined`

packages/react/src/ElementFactory/components/ReactElementFactory.tsx
```diff
@@ -30,8 +34,9 @@ export const ReactElementFactory = <
   const rootElement = document.createElement("div");
   const root = createRoot(rootElement);
 
-  // Used when hiding the element. If we were to pass in undefined instead, the element would be immediately cleared, not
-  // leaving time for the fade out animation to complete.
+  // Used when hiding the element. If we were to pass in undefined instead, the
+  // element would be immediately cleared, not leaving time for the fade out
+  // animation to complete.
```
YousefED, line 39: do we still need `prevDynamicParams` even if the `getBoundingclientRect` is now always available?

matthewlipski: Yeah we do, at least for the side menu and slash menus, since they need dynamic params to render as they fade out. Otherwise the side menu throws an error while the drag handle menu is open, and the slash menu renders with no items. Maybe these could be cached like the menu position is, though I feel like that's a bit beyond the scope of this PR.

YousefED (approved):

PR #257, opened 2023-06-22
feat: Custom block serialization

matthewlipski:
This PR fixes how custom blocks are converted into HTML by implementing a custom `DOMSerializer`. The custom `DOMSerializer` allows us to use the HTML rendered by a custom block to the editor using its `render` function also when copying or exporting HTML. Alternatively, consumers can implement a `serialize` function for their custom blocks, which lets them override how the HTML conversion is done when exporting and copying to clipboard.

The PR also adds a new method for converting the HTML rendered by BlockNote, which includes many `div` elements for structuring the document, to HTML that more closely complies with the spec (removes structuring `div`s, wraps list elements in `ul`/`ol` elements, and removes non-list-item nesting). This method is called `cleanHTML`, but its implementation is basically the same as the old `blocksToHTML`. There is also a new `markdown` method which converts "clean" HTML to Markdown, similarly to the old `blocksToMarkdown`.

In addition, copying and pasting content within BlockNote is improved. Now, BlockNote will attempt to read `blocknote/html` from the clipboard when pasting content, otherwise it will use `text/html` or `text/plain`. Copying from BlockNote will write the selected content to the clipboard in the following formats:

- `text/plain`: Markdown
- `text/html`: "clean" HTML
- `blocknote/html`: "structured" HTML

Since other editors & paste areas will generally read from `text/plain` or `text/html`, this ensures interoperability when copying content from BlockNote.

TODOs:

- [x] Implement custom `DOMSerializer` which uses a block's `render` function to serialize it, if the block has one (i.e. if it's a custom block).
- [x] Allow consumers to override the `render` function when serializing to export or clipboard.
- [x] Implement functions to convert a "structured" HTML string to "clean" HTML, and "clean" HTML to Markdown.
- [x] Write unit tests to ensure the custom `DOMSerializer` works with both default and custom blocks (just image for now).
- [x] Write unit tests to ensure "structured" HTML is converted to "clean" HTML correctly.
- [x] Write unit tests to ensure that overriding a block's `render` function when exporting and copying works.
- [x] Handle copy/paste in the editor to read content from & write content to the correct formats on the clipboard.
- [x] Ensure copy/paste handling works in Firefox and Safari.
- [ ] Ensure the "clean" HTML is sanitized from unnecessary classes & attributes.


tests/utils/customblocks/ReactImage.tsx
```diff
@@ -30,15 +30,17 @@ export const ReactImage = createReactBlockSpec({
           alt={"Image"}
           contentEditable={false}
         />
-        <InlineContent />
+        <InlineContent style={{ flexGrow: 1 }} />
```
YousefED, line 33: why was this added?

packages/react/src/ReactBlockSpec.tsx
```diff
@@ -76,57 +120,57 @@ export function createReactBlockSpec<
...
+          }
+        )(props);
     },
   });
 
   return {
     node: node,
     propSchema: blockConfig.propSchema,
+    serialize: (block, editor) => {
+      // TODO: This doesn't seem like a great way of doing things - any better
```
YousefED, line 157: why do you think this? isn't this exactly the purpose of `renderToString`?

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -70,57 +70,94 @@ export function parse<
...
         dom: blockContent,
-        contentDOM: contentDOM,
+        contentDOM: rendered.contentDOM,
       }
     : {
         dom: blockContent,
       };
 }
 
+export function getBlockFromPos<
```
YousefED, line 121: don't we have this helper function somewhere else?

packages/core/src/extensions/Blocks/api/serialization.ts
```diff
@@ -1,29 +1,114 @@
...
+  editor: BlockNoteEditor<BSchema>
+) => {
+  const customSerializer = DOMSerializer.fromSchema(schema) as DOMSerializer & {
+    serializeNodeInner: (
+      node: Node,
+      options: { document?: Document }
+    ) => HTMLElement;
+  };
+
+  customSerializer.serializeNodeInner = (
```
YousefED, line 23: can we extend DomSerializer (create a new class), instead of overwriting a method like this? Not sure if it's possible in this case

YousefED (commented):
Quite nice!!

- Can you look into unit tests (not e2e) for the different scenarios (e.g.: vanilla / react, with / without inlinecontent, with / without children). We worked on these earlier, right?
- Can we use this for markdown and html serialization as well (besides just copy to clipboard)?
- What about parsing?

tests/utils/customblocks/ReactImage.tsx
matthewlipski: It's just to make sure that the regular `Image` and `ReactImage` have identical styling, makes it easier to compare the serialized HTML to check that it's the same for vanilla and React.

packages/react/src/ReactBlockSpec.tsx
matthewlipski: Haven't really used `renderToString` before tbh, in general converting React to HTML feels a bit like something you shouldn't do. But if this is actually what it's supposed to be for then ig it's not a problem here.

packages/core/src/extensions/Blocks/api/block.ts
matthewlipski: We have `getBlockInfoFromPos` but it's not that useful here since it:
- Deals more with the PM API, so it gives you nodes and positions which you'd still have to convert to a block.
- Is mostly useful since it will find the block that the position is in regardless of the positions depth and exact location inside the block. We don't really need that here since `getPos` always returns the position just before the `blockContent` node in the block.

packages/core/src/extensions/Blocks/api/serialization.ts
matthewlipski: Don't really even need to extend it, the constructor for `DOMSerializer` allows you to specify serialization rules for different node types, so I can just set one for `blockContainer`

matthewlipski: I tried 2 different methods of refactoring the `DOMSerializer`. First was using the existing `DOMSerializer` constructor and adding a custom serialization rule for `BlockContainer` (see [commit b4f3fb6](https://github.com/TypeCellOS/BlockNote/pull/257/commits/b4f3fb6433c6a9e00b90a4f3a6f7d1ff87bc1758)), and the second by extending the DOMSerializer class (see the [custom-block-serialization-class-extension](https://github.com/TypeCellOS/BlockNote/tree/custom-block-serialization-class-extension) branch).

While both of these work for copy/paste, for some reason the serializer still tries to read `toDOM` for custom blocks in other cases like converting to Markdown & HTML. I don't understand why this is the case, especially since AFAIK the second method should literally do the same thing as overwriting `serializeNodeInner` like how it's done on this branch. Maybe you could have a look and see if there's anything obvious that's wrong?

packages/core/src/api/formatConversions/formatConversions.ts
```diff
@@ -12,16 +12,19 @@ import { Block, BlockSchema } from "../../extensions/Blocks/api/blockTypes";
...
+  schema: Schema,
+  editor: BlockNoteEditor<BSchema>
 ): Promise<string> {
   const htmlParentElement = document.createElement("div");
-  const serializer = DOMSerializer.fromSchema(schema);
+  const serializer = customBlockSerializer(schema, editor);
 
   for (const block of blocks) {
-    const node = blockToNode(block, schema);
+    const node = blockToNode(block, editor._tiptapEditor.schema);
```
YousefED, line 27: this can just be `schema` right?

packages/core/src/api/formatConversions/formatConversions.ts
```diff
@@ -131,3 +135,31 @@ export async function markdownToBlocks<BSchema extends BlockSchema>(
 
   return HTMLToBlocks(htmlString.value as string, blockSchema, schema);
 }
+
+// Takes structured HTML and makes it comply with the HTML spec using the
+// `simplifyBlocks` plugin.
+// TODO: Remove classes?
+export async function cleanHTML(structuredHTMLString: string) {
```
YousefED, line 142: blocks should be able to specify their own "clean html", right? So this can never be a `string -> string` method?

packages/core/src/api/serialization/__snapshots__/serialization.test.ts.snap
```diff
@@ -0,0 +1,73 @@
...
...s=\\"_resizeHandle_1c48ad\\" style=\\"right: 4px;\\"></div></div><p class=\\"_caption_1c48ad\\" style=\\"padding: 4px;\\"></p></div></div></div><div class=\\"_blockGroup_1c48ad\\" data-node-type=\\"blockGroup\\"><div class=\\"_blockOuter_1c48ad\\" data-node-type=\\"block-outer\\"><div class=\\"_block_1c48ad\\" data-node-type=\\"blockContainer\\"><div class=\\"_blockContent_1c48ad\\" data-content-type=\\"simpleImage\\" data-url=\\"exampleURL\\" data-caption=\\"Caption\\" data-width=\\"256\\"><div class=\\"_wrapper_1c48ad\\" style=\\"align-items: flex-start;\\"><div class=\\"_addImageButton_1c48ad\\" style=\\"display: none;\\"><div class=\\"_addImageButtonIcon_1c48ad\\"></div><p class=\\"_addImageButtonText_1c48ad\\"></p></div><div class=\\"_imageAndCaptionWrapper_1c48ad\\"><div class=\\"_imageWrapper_1c48ad\\"><img class=\\"_image_1c48ad\\" src=\\"exampleURL\\" alt=\\"placeholder\\" draggable=\\"false\\" style=\\"width: 0px;\\"><div class=\\"_resizeHandle_1c48ad\\" style=\\"left: 4px;\\"></div><div class=\\"_resizeHandle_1c48ad\\" style=\\"right: 4px;\\"></div></div><p class=\\"_caption_1c48ad\\" style=\\"padding: 4px;\\"></p></div></div></div></div></div></div></div></div></div>"`;
```
YousefED, line 53: can we prevent things like the resizehandle from being exported? How would we go about this?

packages/core/src/api/serialization/serialization.test.ts
```diff
@@ -0,0 +1,296 @@
...
+  blockGroup: Node
+) {
+  const serializeNodeHTML = (
+    serializer.serializeNode(blockGroup) as HTMLElement
+  ).outerHTML;
+  const serializeFragmentHTML = (
+    serializer.serializeFragment(Fragment.from(blockGroup)) as HTMLElement
+  ).firstElementChild!.outerHTML;
+
+  expect(serializeNodeHTML).toMatchSnapshot();
```
YousefED, line 63: should `serializeNodeHTML` and `serializeFragmentHTML` be the same? In that case, let's only store 1 snapshot, and just compare them to eachother

packages/core/src/api/serialization/serialization.test.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  expect(serializeNodeCleanHTML).toMatchSnapshot();
+  expect(serializeFragmentCleanHTML).toMatchSnapshot();
+}
+
+describe("Convert paragraphs to structured HTML", () => {
+  it("Convert paragraph to structured HTML", async () => {
+    const serializer = customBlockSerializer(tt.schema, editor);
+
+    const text = tt.schema.text("Paragraph");
```
YousefED, line 77: Is there a reason we use the pm syntax to instantiate nodes / a document, instead of the blocknote API? Better to "eat your own dogfood"?

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -148,106 +246,61 @@ export function createBlockSpec<
...
-          );
-        }
-        // Add elements to blockContent
-        blockContent.appendChild(rendered.dom);
-
-        return "contentDOM" in rendered
-          ? {
-              dom: blockContent,
-              contentDOM: rendered.contentDOM,
-              destroy: rendered.destroy,
```
YousefED, line 238: `destroy` functionality seems removed?

packages/react/src/ReactBlockSpec.tsx
```diff
@@ -7,36 +7,42 @@ import {
...
 import { createContext, ElementType, FC, HTMLProps, useContext } from "react";
 
 // extend BlockConfig but use a React render function
 export type ReactBlockConfig<
-  Type extends string,
+  BType extends string,
   PSchema extends PropSchema,
   ContainsInlineContent extends boolean,
-  BSchema extends BlockSchema
+  BSchema extends BlockSchema & {
```
YousefED, line 33: Do we need a shorthand for this? I think the meaning of this change is `BlockSchemaWithBlock<...>`?

packages/react/src/ReactBlockSpec.tsx
```diff
@@ -68,13 +74,55 @@ export const InlineContent = <Tag extends ElementType>(
   );
 };
 
+// Function that wraps the React component returned from 'blockConfig.render' in
+// a `NodeViewWrapper` which also acts as a `blockContent` div. It contains the
```
YousefED, line 78: nice comments!

YousefED (commented):
reviewed the rest as well. Looks very nice already!

packages/core/src/extensions/Blocks/api/serialization.ts
```diff
@@ -1,94 +1,79 @@
...
-              options
+        // Checks if the `blockContent is custom and the node has to be
+        // serialized manually, or it can be serialized normally using `toDOM`.
+        if (node.firstChild!.type.name in customNodes) {
+          // Serializes the `blockContent` node using the custom `blockSpec`'s
+          // `serialize` function.
+          const blockContentElement = DOMSerializer.renderSpec(
+            document,
+            editor.schema[node.firstChild!.type.name as keyof BSchema]
+              .serialize!(
```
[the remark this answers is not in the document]
matthewlipski: It's because default blocks don't implement it (they use `renderHTML`/`toDOM` instead)

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -58,71 +63,164 @@ export function parse<
...
+  }
+
+  return block;
+}
+
+// Function that wraps the `dom` element returned from 'blockConfig.render' in a
+// `blockContent` div, which contains the block type and props as HTML
+// attributes. If `blockConfig.render` also returns a `contentDOM`, it also adds
+// an `inlineContent` class to it.
+export function wrapInBlockStructure<
```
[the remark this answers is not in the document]
matthewlipski: Yeah the code was kinda bad before and there was a lot of duplication in `ReactBlockSpec`, so I cleaned it up with `wrapInBlockStructure`

packages/core/src/api/formatConversions/formatConversions.ts
matthewlipski: I think we need some more tests now that you mention it. Since right now we're only testing image custom blocks, everything works. This is because for structured HTML, we only need the `blockContent` div to recreate the block, anything inside is not relevant.

Therefore, for copy/paste, we can just put the HTML from the image block's `serialize` function inside the `blockContent` div instead of the output from `render`. But image blocks don't have inline content, which should be found somewhere inside the `blockContent` div.

So you're right, this might have to be changed. I think right now we should add unit tests for custom blocks with inline content (e.g. an alert block).

packages/core/src/api/serialization/__snapshots__/serialization.test.ts.snap
```diff
@@ -0,0 +1,73 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Convert images to structured HTML > Convert add image button to structured HTML 1`] = `"<div class=\\"_blockGroup_1c48ad\\" data-node-type=\\"blockGroup\\"><div class=\\"_blockOuter_1c48ad\\" data-node-type=\\"block-outer\\"><div class=\\"_block_1c48ad\\" data-node-type=\\"blockContainer\\"><div class=\\"_blockContent_1c48ad\\" data-content-type=\\"image\\"><p>Add Image</p></div></div></div></div>"`;
```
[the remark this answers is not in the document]
matthewlipski: Agreed, think CSS modules overcomplicate editor styling/theming for consumers in general.

packages/core/src/api/serialization/__snapshots__/serialization.test.ts.snap
matthewlipski: Right now we use `display: none` to hide the resize handles, but we could instead unmount them from the DOM. Then they also wouldn't show up in the tests. I think I did it this way originally bc it was simpler and more performant, but adds more clutter to the DOM.

packages/core/src/api/serialization/serialization.test.ts
matthewlipski: 2 reasons:
1. The BlockNote API doesn't let you create blocks outside the editor, i.e. when creating a block, you have to insert it somewhere into the editor. We already have unit tests for block insertion, and I think since these ones are meant to test serialization specifically, it's better that they don't fail if something non-serialization related goes wrong.

packages/core/src/api/serialization/serialization.ts
```diff
@@ -0,0 +1,182 @@
...
+                  const parent = document.createElement("div");
+                  parent.appendChild(selectedHTML);
+
+                  const structured = parent.innerHTML;
+                  const clean = await cleanHTML(structured);
+                  const plain = await markdown(clean);
+
+                  event.clipboardData!.setData("text/plain", plain);
+                  event.clipboardData!.setData("text/html", clean);
+                  // TODO: Writing to other MIME types not working in Safari for
```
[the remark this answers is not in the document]
matthewlipski: Ah good point, in any case that's still a problem since the HTML cleaning itself is async

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -58,71 +63,164 @@ export function parse<
...
-): ParseRule[] {
-  return [
+) {
+  const rules: ParseRule[] = [
     {
       tag: "div[data-content-type=" + blockConfig.type + "]",
     },
   ];
+
+  if (blockConfig.parse) {
```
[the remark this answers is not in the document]
matthewlipski: Yes, I'll comment it out for now since it's for S3 (rn just doing S1 & 2)

packages/core/src/api/serialization/serialization.ts
```diff
@@ -0,0 +1,182 @@
...
+        // Renders block content using the custom `blockSpec`'s `serialize`
+        // function.
+        const blockContent = DOMSerializer.renderSpec(
+          doc(options),
+          editor.schema[node.firstChild!.type.name as keyof BSchema].serialize!(
+            nodeToBlock(
+              node,
+              editor.schema,
+              editor.blockCache
+            ) as SpecificBlock<BlockSchema, string>,
```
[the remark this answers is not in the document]
matthewlipski: I think the typing for `nodeToBlock` can just be changed to return a `SpecificBlock`. We'll also need to cast stuff to do this but it's kinda inevitable since PM nodes aren't strongly typed.

packages/core/src/api/serialization/__snapshots__/serialization.test.ts.snap
matthewlipski: Ah also keep in mind that the default image block implements `serialize` and outputs a `figure`. This one is modified to not implement `serialize`, therefore uses the same HTML for the clipboard as for rendering.

packages/core/src/api/formatConversions/formatConversions.ts
YousefED: I think this could also be the case for blocks without inline content, let's discuss

packages/core/src/api/serialization/serialization.ts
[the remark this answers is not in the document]
YousefED: true, but first let's see if this was the issue or not

packages/core/src/extensions/Blocks/api/serialization.ts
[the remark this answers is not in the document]
YousefED: Let's see if we can / want to streamline this

packages/core/src/api/serialization/serialization.ts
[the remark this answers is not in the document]
matthewlipski: I made an effort to improve this but the casting is still pretty nasty. The underlying issue is that when we convert a ProseMirror `Node` to a `Block`, we have basically 0 type information. I managed to work around this issue to some extent for the custom blocks API with the `TipTapNode` type and `createTipTapBlock` function, but this only works because we're creating the node in the first place.

packages/core/src/api/serialization/serialization.ts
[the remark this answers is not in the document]
matthewlipski: This was indeed the issue for Firefox. Safari, like Chrome, doesn't seem to care if the clipboard contents are set asynchronously, regardless it won't write to `blocknote/html`.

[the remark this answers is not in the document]
YousefED: Judging from notion web, when using on safari it also doesn't add custom data to the clipboard. (only sanitized html).

I'm not sure yet how they do it, because when I copy and paste within notion on safari (even across tabs or between safari private mode and regular), colored text works (even though that info is not on the clipboard)

examples/editor/src/App.tsx
```diff
@@ -13,7 +13,7 @@ function App() {
     },
     domAttributes: {
       editor: {
-        class: styles.editor,
+        class: "editor",
```
YousefED, line 16: if we're not using CSS modules, we need to have a good strategy to prevent collisions, for example, prefix everything with `bn-`

packages/core/src/BlockNoteEditor.ts
```diff
@@ -777,47 +773,47 @@ export class BlockNoteEditor<BSchema extends BlockSchema = DefaultBlockSchema> {
...
-   * Creates a list of blocks from a Markdown string. Tries to create `Block` and `InlineNode` objects based on
-   * Markdown syntax, though not all symbols are recognized. If BlockNote doesn't recognize a symbol, it will parse it
-   * as text.
-   * @param markdown The Markdown string to parse blocks from.
-   * @returns The blocks parsed from the Markdown string.
-   */
-  public async markdownToBlocks(markdown: string): Promise<Block<BSchema>[]> {
-    return markdownToBlocks(markdown, this.schema, this._tiptapEditor.schema);
-  }
+  // /**
```
YousefED, line 776: TODO, right?

packages/core/src/api/formatConversions/formatConversions.ts
```diff
@@ -1,133 +1,140 @@
...
-    type: "element",
-    tagName: "pre",
-    properties: {},
-    children: [result],
-  };
-  state.patch(node, result);
-  return result;
-}
-
-export async function markdownToBlocks<BSchema extends BlockSchema>(
```
YousefED, line 115: I understand it's temporarily commented out (similar to some stuff above), but we need to make a rollout plan:
- do an intermediate release with the new code for use-case 1+2, and the old code for markdown / html / etc
- don't do an intermediate release, release only until we have all use-cases solved. maybe still make a separate branch where we can squash use-case 1+2 into

packages/core/src/api/formatConversions/simplifyBlocksRehypePlugin.ts
```diff
@@ -22,6 +22,19 @@ export function simplifyBlocks(options: SimplifyBlocksOptions) {
   ]);
 
   const simplifyBlocksHelper = (tree: HASTParent) => {
+    // Checks whether blocks in the tree are wrapped by a parent `blockGroup`
```
YousefED, line 25: why is this new / we didn't have this earlier?

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -349,11 +350,14 @@ function contentNodeToInlineContent(contentNode: Node) {
 /**
  * Convert a TipTap node to a BlockNote block.
  */
-export function nodeToBlock<BSchema extends BlockSchema>(
+export function nodeToBlock<
+  BSchema extends BlockSchema,
+  BType extends keyof BSchema
```
YousefED, line 355: BType now functions as an "unsafe" cast. I'm not sure that's desirable, because we don't know for sure whether the node is of this type, right?

Should we pass a parameter with `blockType: BType`, so we can do a run-time check on whether it's a matching node?

packages/core/src/api/serialization/html/internalHTMLSerializer.ts
```diff
@@ -0,0 +1,148 @@
+// TODO: IMO this should be part of the formatConversions file since the custom
```
YousefED, line 1: not sure about this? I think separating is nice?

packages/core/src/api/serialization/html/externalHTMLExporter.ts
```diff
@@ -0,0 +1,166 @@
...
+// `exportBlocks`: Exports an array of blocks to HTML.
+// `exportFragment`: Exports a ProseMirror fragment to HTML. This is mostly
+// useful if you want to export a selection which may not start/end at the
+// start/end of a block.
+export interface ExternalHTMLExporter<BSchema extends BlockSchema> {
+  exportBlocks: (blocks: PartialBlock<BSchema>[]) => Promise<string>;
+  exportProseMirrorFragment: (fragment: Fragment) => Promise<string>;
+}
+
+// TODO: This implementation obviously sucks. For now we're just doing the
```
YousefED, line 45: `exact same thing` not true, you're also calling `toExternalHTML` instead of internal

Can you also comment what a solution that "does not suck" (i.e.: a better solution) would look like?

Does it really "suck" or is just suboptimal?

packages/core/src/api/serialization/html/externalHTMLExporter.ts
```diff
@@ -0,0 +1,166 @@
...
+      Fragment.from(blockGroup)
+    );
+  };
+
+  customSerializer.exportProseMirrorFragment = async (fragment) => {
+    const internalHTML = customSerializer.serializeFragment(fragment);
+    const parent = document.createElement("div");
+    parent.appendChild(internalHTML);
+
+    const externalHTML = await unified()
```
YousefED, line 153: if this is important logic / different than internal, comment it to explain

packages/core/src/extensions/Blocks/nodes/BlockContent/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
```diff
@@ -133,4 +124,6 @@ const BulletListItemBlockContent = createTipTapBlock<"bulletListItem", true>({
 export const BulletListItem = {
   node: BulletListItemBlockContent,
   propSchema: bulletListItemPropSchema,
+  toInternalHTML: serializeBlockToHTMLDefault,
```
YousefED, line 127: I don't think we want to specify this on the blocks themselves right? `toInternalHTML` should just be optional to implement I think, and if it's not implemented, use the default impl

packages/core/src/api/formatConversions/formatConversions.ts
matthewlipski: I think we should just roll everything out at once tbh. For one, we need to do S3 (paste from external) to finish fixing copy/paste. Once we do that I think it's largely a matter of extracting existing code into functions for the remaining scenarios, which shouldn't take long.

packages/core/src/api/nodeConversions/nodeConversions.ts
matthewlipski: I think that shouldn't be necessary. The idea here is that you can get `BType` from `node.type.name`, and provided that the node came from an editor with the same `blockSchema` as you passed to `nodeToBlock`, you should end up with a `Block` of type `BType`. Unless there's a scenario where the node comes from a different editor to the schema, I think it should be ok.

There is one caveat, being that TypeScript has no way to infer `BType` from `node.type.name` since `Node` is not that strictly typed. So you have to provide the type arguments yourself if you need them.

packages/core/src/api/serialization/html/internalHTMLSerializer.ts
matthewlipski: Ah yeah this is an old comment, will remove.

packages/core/src/extensions/Blocks/nodes/BlockContent/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
matthewlipski: Imo it's actually cleaner this way since then there's no distinction between default and custom blocks in the `BlockSchema`. Because keep in mind that the default implementation between the default and custom blocks is also different (one should use `node.config.renderHTML` and the other `blockConfig.render`). I think it causes more confusion (and more logic in `createBlockSpec`) if we don't explicitly add `toInternalHTML` to the default blocks, so I think it's worth it.

packages/core/src/api/serialization/html/externalHTMLExporter.ts
matthewlipski: I sorted out the code duplication issue that it was causing between `toInternalHTML` and `toExternalHTML`. So now the only issue is the fact that it uses the rehype plugin in `serializeProseMirrorFragment` to convert internal HTML to external, causing the function to be async. After looking at alternative solutions I realized that doing the "cleaning" (un-nesting blocks, wrapping list items in list elements, etc), is only really possible after converting to HTML anyway, else it won't comply with the PM schema. So I think now the main issue is just the async - I'll see if I can get around the issue in Firefox, since that's where it's causing issues.

packages/core/src/api/formatConversions/simplifyBlocksRehypePlugin.ts
matthewlipski: We were using it a bit differently before. Before, we were calling `unified.process` with this HTML:

```ts
<div data-node-type="blockGroup">
  <div data-node-type="blockContainer">
    ...
  </div>
  <div data-node-type="blockContainer">
    ...
  </div>
  ...
</div>
```

Notice how the `blockContainer` elements aren't wrapped in a `blockGroup` element. But we ideally want to just use `serializeProseMirrorFragment(...)` to get the internal HTML that we pass to `unified.process`, which looks like this:

packages/core/src/extensions/Blocks/nodes/BlockContent/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
```diff
@@ -94,33 +94,18 @@ const BulletListItemBlockContent = createTipTapBlock<"bulletListItem", true>({
...
-            "bn-inline-content",
-            inlineContentDOMAttributes.class
-          ),
-        },
-        0,
-      ],
-    ];
+    return defaultRenderHTML(
+      this.name,
+      // We use a <p> tag, because for <li> tags we'd need an <il> element to
```
YousefED, line 99: `<ul>`  instead of `<il>`

packages/core/src/extensions/Blocks/nodes/BlockContent/defaultRenderHTML.ts
```diff
@@ -0,0 +1,40 @@
+import { mergeCSSClasses } from "../../../../shared/utils";
+
+export function defaultRenderHTML(
```
YousefED, line 3: let's stay focused on naming. "default" doesn't say a lot, and "renderHTML" is very much focused on where it's used (i.e., on the "outside"), instead of explaining what the function itself is. Can you come up with a better name?

YousefED (commented):
great! change definitely helps with a more consistent codebase. fyi, validated with marijn here: https://discuss.prosemirror.net/t/motivation-for-domoutputspec/5997/3

packages/core/src/api/serialization/html/externalHTMLExporter.ts
matthewlipski: Nvm, there's literally a `unified.processSync` method we can use to run the rehype plugin instead of `unified.process`, which fixes the async issue.

packages/react/src/ReactBlockSpec.tsx
```diff
@@ -97,12 +104,18 @@ export function reactWrapInBlockStructure<
...
       )}
       // Sets content type attribute
-      data-content-type={block.type}
+      data-content-type={blockType}
       // Add props as HTML attributes in kebab-case with "data-" prefix
       {...Object.fromEntries(
-        Object.entries(block.props).map(([prop, value]) => {
-          return [camelToDataKebab(prop), value];
-        })
+        Object.entries(blockProps)
```
YousefED, line 110: can you add a comment why this was changed?

YousefED (approved):

PR #270, opened 2023-07-10
feat: React re-architecture

matthewlipski:
There are a number of issues with the current rendering implementation in React:

- [X] Each menu/toolbar is rendered under a separate root.
- [x] The editor is not immediately initialized upon calling `useBlocknote`, causing issues.
- [X] Each menu is attached to `document.body` instead of being a sibling to the editor.
- [X] The way in which factories are designed makes sense for a vanilla implementation, but using them for React is unintuitive.
- [X] Plugins have too much responsibility - they should only provide information for UI elements to render, but instead they also render the UI elements themselves.
- [X] Supplying custom menu/formatting components is done using the `uiFactories` option, which is kind of clunky.

This PR is a major rethink of how we create, render, and pass elements to the editor in React (also changes the vanilla implementation a bit).


packages/react/src/SlashMenu/components/SlashMenu.tsx
```diff
@@ -63,3 +67,51 @@ export function SlashMenu<BSchema extends BlockSchema>(
...
+
+  useEffect(() => {
+    createSlashMenu<ReactSlashMenuItem<DefaultBlockSchema>>(
+      props.editor as any,
+      ({ referencePos, ...state }) => {
+        setState(state);
+        referenceClientRect.current = referencePos;
+      },
+      (query) =>
+        defaultReactSlashMenuItems.filter(
```
YousefED, line 93: should this be defaultReactSlashMenuItems?

YousefED (commented):
added number of comments. Will also create a separate PR with some fixes / changes. 

About the overall architecture, I have following questions:
- does it make sense that "static" things like "freezeMenu" are part of "state"? (sorry for opening up this discussion again, seems we're going a bit back and forth)
- Plugin views are recreated indeed when other plugins are changed / added. This is by design in prosemirror, probably so that plugins can react to changes to the view that other plugins make. Although it works (see my separate PR that fixes the issues you ran into with this), I think it will be nicer (and faster) to register all plugins on editor instantation

packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
```diff
@@ -110,28 +100,31 @@ export class FormattingToolbarView<BSchema extends BlockSchema> {
...
+    if (this.formattingToolbarState?.show) {
+      this.formattingToolbarState.show = false;
+      this.updateFormattingToolbar();
     }
   };
 
   scrollHandler = () => {
-    if (this.toolbarIsOpen) {
-      this.formattingToolbar.render(this.getDynamicParams(), false);
+    if (this.formattingToolbarState?.show) {
```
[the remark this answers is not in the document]
matthewlipski: I think it's better for the plugin to update the position on scroll than leave it to the components. While Tippy needs a function which returns a position, I don't think this approach is super common and imo, in the majority of implementations, the components would just end up having to init the scroll listeners themselves.

packages/core/src/extensions/DraggableBlocks/DraggableBlocksPlugin.ts
```diff
@@ -328,49 +332,52 @@ export class BlockMenuView<BSchema extends BlockSchema> {
...
-    }
-
-    if (this.menuOpen) {
-      this.menuOpen = false;
-      this.blockMenu.hide();
-    }
-
-    this.menuFrozen = false;
+  onMouseDown = (_event: MouseEvent) => {
+    // TODO: Fix
```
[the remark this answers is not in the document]
matthewlipski: Don't think so, we still want it to hide the side menu on mouse down

packages/react/src/FormattingToolbar/components/DefaultButtons/ColorStyleButton.tsx
```diff
@@ -1,13 +1,36 @@
...
   editor: BlockNoteEditor<BSchema>;
 }) => {
+  const [currentTextColor, setCurrentTextColor] = useState<string>(
+    props.editor.getActiveStyles().textColor || "default"
+  );
+  const [currentBackgroundColor, setCurrentBackgroundColor] = useState<string>(
+    props.editor.getActiveStyles().backgroundColor || "default"
+  );
+
+  useEditorContentChange(props.editor, () => {
```
YousefED, line 20: @matthewlipski I would use a `forceUpdate` mechanism in `useEditorContentChange`. And maybe rename to `useUpdateOnContentChange` or sth. That way we can keep the old code and get rid of the separate state fields, which will be a lot cleaner I think

examples/editor/src/App.tsx
```diff
@@ -20,7 +31,39 @@ function App() {
   // Give tests a way to get prosemirror instance
   (window as WindowWithProseMirror).ProseMirror = editor?._tiptapEditor;
 
-  return <BlockNoteView editor={editor} />;
+  return (
+    <BlockNoteView editor={editor}>
```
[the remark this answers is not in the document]
matthewlipski: - I'm not usually a fan of contexts since i feel like they're not as explicit as props, though here I think it's worth it since it would give a cleaner experience for users.
- I would rather have them as children but I don't know if it's possible since you can have multiple children, whereas we should only allow a single child per wrapper. I'll look into it though since I agree it's nicer.
- Also agree with the naming not being that great. I'm not sure what else to call it though since the menu or toolbar element goes inside it. So just calling it e.g. Formatting toolbar would be weird if you also pass a formatting toolbar to it.

matthewlipski:
Getting a linting issue with dependencies (see most recent commits). The linter claims they're useless but removing them breaks the menu and toolbar positions. Any ideas how to fix this?

packages/core/src/extensions/SlashMenu/defaultSlashMenuItems.ts
```diff
@@ -0,0 +1,122 @@
...
+export const getDefaultSlashMenuItems = <BSchema extends BlockSchema>(
+  schema: BSchema
+) => {
+  const slashMenuItems: BaseSlashMenuItem<BSchema>[] = [];
+
+  if ("heading" in schema && "level" in schema.heading.propSchema) {
+    if (schema.heading.propSchema.level.values?.includes("1")) {
+      slashMenuItems.push(
+        // Command for creating a level 1 heading
+        createBaseSlashMenuItem(
```
YousefED, line 36: I don't think this needs to be a function, we can get rid of `createBaseSlashMenuItem` all together I think?

packages/react/src/FormattingToolbar/components/FormattingToolbarPositioner.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <Tippy
+      appendTo={props.editor.domElement.parentElement!}
+      content={formattingToolbarElement}
+      getReferenceClientRect={getReferenceClientRect}
+      interactive={true}
+      visible={show}
+      animation={"fade"}
+      placement={"top-start"}
+      sticky={true}
+      plugins={[sticky]}
```
YousefED, line 60: what is the sticky plugin (couldn't find a clear description on the tippy website, do you have a link / can you explain)?

YousefED:
> Getting a linting issue with dependencies (see most recent commits). The linter claims they're useless but removing them breaks the menu and toolbar positions. Any ideas how to fix this?

The warning given is `Mutable values like 'referencePos.current' aren't valid dependencies because mutating them doesn't re-render the component`

This makes sense, I think we need to make referencePos a state on the component (and make sure performance is still ok / there are no unnecessary rerenders)

matthewlipski:
> > Getting a linting issue with dependencies (see most recent commits). The linter claims they're useless but removing them breaks the menu and toolbar positions. Any ideas how to fix this?
> 
> The warning given is `Mutable values like 'referencePos.current' aren't valid dependencies because mutating them doesn't re-render the component`
> 
> This makes sense, I think we need to make referencePos a state on the component (and make sure performance is still ok / there are no unnecessary rerenders)

Hmm yeah that's kinda what I wanted to avoid, since we don't wanna re-render on scroll. I think the reason I made it a `ref` in the first place is because scrolling was kinda laggy when it was a `state`.

packages/react/src/FormattingToolbar/components/FormattingToolbarPositioner.tsx
matthewlipski: It's needed as Tippy only repositions itself if the reference element/DOMRect moves on scroll, while the sticky plugin also allows it to move, for example, when the text alignment changes. However, this isn't performant, so I added an alternative approach on the `tippy-sticky-alt` branch.

packages/website/docs/docs/formatting-toolbar.md
```diff
@@ -18,9 +18,180 @@ The Formatting Toolbar appears whenever you highlight text in the editor, and is
 
 <img style="max-width:600px" :src="isDark ? '/img/screenshots/formatting_toolbar_dark.png' : '/img/screenshots/formatting_toolbar.png'" alt="image">
 
-## Custom Formatting Toolbar
+## Custom Formatting Toolbar Element
 
-You can create a custom Formatting Toolbar using a React component. This component should take the following props:
+You can create a custom Formatting Toolbar using a React component. The example below shows a basic custom formatting toolbar with four items. The first three are default items to toggle bold, italic, and underline, while the last one is a custom button which toggles the text and background colors to be blue.
```
YousefED, line 23: We should make it clear what "custom" means.

proposal:


> By default, the formatting toolbar looks like the screenshot above. However, if you like, you can fully customize the appearance and functionality by creating a custom formatting toolbar. The example below...

packages/website/docs/docs/formatting-toolbar.md
```diff
@@ -18,9 +18,180 @@ The Formatting Toolbar appears whenever you highlight text in the editor, and is
...
+      <ToolbarButton
+      mainTooltip={"Blue Text & Background"}
+      onClick={() => {
+        props.editor.toggleStyles({
+          textColor: "blue",
+          backgroundColor: "blue",
+        });
+      }}
+      isSelected={
+        props.editor.getActiveStyles().textColor === "blue" &&
```
YousefED, line 58: does this even work? I don't think isSelected will properly update

packages/website/docs/docs/formatting-toolbar.md
```diff
@@ -18,9 +18,180 @@ The Formatting Toolbar appears whenever you highlight text in the editor, and is
...
+
+If you're looking to customize this to create, for example, a static Formatting Toolbar, head to [Custom Formatting Toolbar Positioner](/docs/formatting-toolbar#custom-formatting-toolbar-positioner). If you only want to change the toolbar element itself, you can use the `FormattingToolbarPositioner` to apply the default behaviour.
+
+TODO: Text below is not final. We should change how we remove & customize the menus & toolbars. Take the scenarios:
+- Customizing 1 UI element
+- Customizing all UI elements
+- Removing 1 UI element
+- Removing all UI elements
+
+By default, if you don't provide any children the `BlockNoteView`, BlockNote uses the default UI. However, if you want to provide a custom Formatting Toolbar, you should also add the default positioners for them (default components are used if custom ones aren't provided), as BlockNote won't render them otherwise.
```
YousefED, line 182: I think we should create a separate page about BlockNoteView, that explains the defaults

packages/website/docs/docs/formatting-toolbar.md
```diff
@@ -18,9 +18,180 @@ The Formatting Toolbar appears whenever you highlight text in the editor, and is
...
+}
+```
+
+```css-vue /styles.css [hidden]
+{{ getStyles(isDark) }}
+```
+
+:::
+
+Let's go over this code, starting with the `CustomFormattingToolbar` component, which is used to render the toolbar and takes the following props:
```
YousefED, line 94: I'm in doubt whether any of the docs below this are necessary. I'm starting to think that we benefit more from keeping the docs simple and focus on the core functionality, combined with a clear code base and some examples that we can point people to.

packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
```diff
@@ -1,146 +1,101 @@
...
         item: item,
         editor: editor,
       });
     };
 
-    this.suggestionsMenu = suggestionsMenuFactory(this.getStaticParams());
+    document.addEventListener("scroll", this.handleScroll);
   }
 
+  handleScroll = () => {
```
YousefED, line 88: why is this necessary in the new version, but not in the old one?

packages/react/src/FormattingToolbar/components/FormattingToolbarPositioner.tsx
YousefED: let's add this info as comment in the code. Comments in PRs will get lost eventually after merging

packages/react/src/HyperlinkToolbar/components/DefaultHyperlinkToolbar.tsx
```diff
@@ -1,36 +1,42 @@
...
 
   if (isEditing) {
     return (
       <EditHyperlinkMenu
         url={props.url}
         text={props.text}
-        update={props.editHyperlink}
+        update={(url, text) => props.editHyperlink(url, text)}
+        // TODO: Better way of waiting for fade out
+        onBlur={(event) =>
```
YousefED, line 21: why wasn't this necessary before?

packages/react/src/HyperlinkToolbar/components/DefaultHyperlinkToolbar.tsx
```diff
@@ -1,36 +1,42 @@
...
+        }
+        ref={editMenuRef}
       />
     );
   }
 
   return (
-    <Toolbar>
+    <Toolbar
+      // TODO: This should go back in the plugin.
```
YousefED, line 36: this still relevant?

packages/react/src/SideMenu/components/DefaultSideMenu.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { DefaultDragHandleMenu } from "./DragHandleMenu/DefaultDragHandleMenu";
+
+export const DefaultSideMenu = <BSchema extends BlockSchema>(
+  props: SideMenuProps<BSchema> & {
+    dragHandleMenu?: FC<DragHandleMenuProps<BSchema>>;
+  }
+) => {
+  const dragHandleRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
```
YousefED, line 18: I think this should be a callback instead of useEffect

packages/react/src/SideMenu/components/SideMenuPositioner.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+import { DefaultSideMenu } from "./DefaultSideMenu";
+
+export type SideMenuProps<BSchema extends BlockSchema = DefaultBlockSchema> =
+  Omit<SideMenuCallbacks, keyof BaseUiElementCallbacks> &
+    Omit<SideMenuState<BSchema>, keyof BaseUiElementState> & {
+      editor: BlockNoteEditor<BSchema>;
+    };
+
+// export type a<BSchema extends BlockSchema> = {
```
YousefED, line 22: remove?

packages/react/src/hooks/useBlockNote.ts
```diff
@@ -2,106 +2,25 @@ import {
...
-      instance._tiptapEditor.destroy();
-      isMounted = false;
-    };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, deps);
-
-  return editor;
+  console.log("create new blocknote instance");
+
+  return new BlockNoteEditor<BSchema>({
```
YousefED, line 20: let's remove the log above (see #268 )

packages/react/src/hooks/useBlockNote.ts
```diff
@@ -2,106 +2,25 @@ import {
...
  */
 export const useBlockNote = <BSchema extends BlockSchema = DefaultBlockSchema>(
-  options: Partial<
-    BlockNoteEditorOptions<BSchema> & {
-      customElements: CustomElements<BSchema>;
-    }
-  > = {},
-  deps: DependencyList = []
+  options: Partial<BlockNoteEditorOptions<BSchema>> = {},
+  _deps: DependencyList = []
```
YousefED, line 16: I think we still need to check this? Now we'll create a new instance whenever the hook is called

YousefED: I think a usememo should be enough

examples/editor/src/App.tsx
```diff
@@ -20,7 +27,35 @@ function App() {
   // Give tests a way to get prosemirror instance
   (window as WindowWithProseMirror).ProseMirror = editor?._tiptapEditor;
 
-  return <BlockNoteView editor={editor} />;
+  // Use mock ID if tests are running.
+  if ((window as any).__TEST_OPTIONS === undefined) {
+    // (window as any).__TEST_OPTIONS = {};
```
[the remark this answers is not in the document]
matthewlipski: I was testing am ID bug which has since been fixed, so will remove

packages/react/src/BlockNoteView.tsx
```diff
@@ -1,15 +1,52 @@
...
   return (
-    // TODO: Should we wrap editor in MantineProvider? Otherwise we have to duplicate color hex values.
-    // <MantineProvider theme={BlockNoteTheme}>
-    <EditorContent editor={props.editor?._tiptapEditor || null} />
-    // </MantineProvider>
+    <MantineProvider theme={getBlockNoteTheme()}>
+      <EditorContent
+        editor={props.editor?._tiptapEditor}
+        // TODO: Better class name
+        className={"BlockNote"}>
```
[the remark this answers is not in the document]
matthewlipski: Maybe we don't need it since you can add it as a prop to `BlockNoteView`. I figured it might make sense to add some kind of default class otherwise it's not immediately obvious that it's a wrapper for BlockNote.

packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
matthewlipski: We used to provide a function that returns the `referenceRect` in `staticParams`, but this was pretty much just because of quirks with Tippy that we didn't fully understand at the time (basically so it wouldn't trigger an update). However, it makes more sense from an API design perspective to just provide the `referenceRect` in the `dynamicParams`/`state`, especially since now the components aren't forced to update.

packages/core/src/BlockNoteEditor.ts
```diff
@@ -70,7 +55,7 @@ export type BlockNoteEditorOptions<BSchema extends BlockSchema> = {
    *
    * @default defaultSlashMenuItems from `./extensions/SlashMenu`
    */
-  slashCommands: BaseSlashMenuItem<any>[];
+  slashMenuItems: BaseSlashMenuItem<any>[];
```
YousefED, line 58: For this breaking change, we should:

a) check in the constructor whether `(options as any).slashCommands` has been passed, and log an error to explain this has been renamed
b) search the project for `slashcommands`, I think there might be some unrenamed values
c) make sure to mention it in changelog

packages/website/docs/docs/side-menu.md
```diff
@@ -25,40 +25,38 @@ You can also click the drag handle in the Block Side Menu (`⠿`) to open the Dr
 
 ## Custom Drag Handle Menu
 
-BlockNote lets you customize which items appear in the Drag Handle Menu. Have a look at the example below, in which the color picker item is replaced with a custom item that opens an alert.
+If you want to change the items in the Drag Handle Menu, or replace it altogether, you can do that using a React component.
```
YousefED, line 28: on line 22, change

You can also click the drag handle in the Block Side Menu (`⠿`)

to

You can also click the drag handle (`⠿`) in the Block Side Menu

YousefED, line 28: this document is confusing. We don't explain how to customize the sidemenu, but the draghandle menu instead. agree?

I think maybe we should first explain how to customize the side menu, and only then how to change the drag menu

packages/website/docs/docs/side-menu.md
```diff
@@ -85,67 +92,23 @@ export default function App() {
...
+## Components
 
-```typescript
-const editor = useBlockNote({
-  customElements: {
-    blockSideMenuFactory: CustomBlockSideMenu
-  },
-});
-```
+It might seem daunting to create your own Formatting Toolbar from scratch, which is why BlockNote provides React components that you can use which match the default styling.
```
YousefED, line 101: copy / paste from ft

packages/website/docs/docs/ui-elements.md
```diff
@@ -0,0 +1,134 @@
...
+export default function App() {
+  // Creates a new editor instance.
+  const editor: BlockNoteEditor = useBlockNote({
+    theme: "{{ getTheme(isDark) }}",
+  });
+
+  // Renders the editor instance.
+  return (
+    <BlockNoteView editor={editor}>
+      <FormattingToolbarPositioner editor={editor} />
```
YousefED, line 66: At this point, the user hasn't seen the code for the sidemenupositioner. So it's completely unclear that something has been removed.

We should first explain the default children, and then explain that it's possible to customize them

packages/website/docs/docs/ui-elements.md
```diff
@@ -0,0 +1,134 @@
...
+Along with the editor itself, BlockNote includes a few additional UI elements in the forms of menus and toolbars:
+
+- [Formatting Toolbar](/docs/formatting-toolbar)
+- Hyperlink Toolbar
+- [Slash Menu](/docs/slash-menu)
+- [Side Menu](/docs/side-menu)
+
+By default, these are all included in the editor, but you can remove or replace each of them with your own React components.
+
+## Positioner Components
```
YousefED, line 27: I would remove this part in favor of first showing the default contents of the <BlockNoteView>. Then you can use 1 sentence to explain what the positioners do, but it doesn't have to be a section

packages/react/src/BlockNoteView.tsx
```diff
@@ -1,15 +1,52 @@
...
+import { SideMenuPositioner } from "./SideMenu/components/SideMenuPositioner";
 
 export function BlockNoteView<BSchema extends BlockSchema>(props: {
-  editor: BlockNoteEditor<BSchema> | null;
+  editor: BlockNoteEditor<BSchema>;
+  children?: ReactNode;
 }) {
+  const [ready, setReady] = useState(false);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
matthewlipski: Actually it's necessary due to the bug in TipTap which fires event listeners before the editor is ready. But the ready check should be moved to the listener hooks instead of the `BlockNoteView` either way.

packages/react/src/hooks/useBlockNote.ts
```diff
@@ -25,15 +25,16 @@ export const useBlockNote = <BSchema extends BlockSchema = DefaultBlockSchema>(
...
+  const ret = useMemo(() => {
     if (editorRef.current) {
-      editorRef.current && editorRef.current._tiptapEditor.destroy();
+      editorRef.current._tiptapEditor.destroy();
     }
 
     editorRef.current = initEditor(options);
-  }, [options, ...deps]);
+    return editorRef.current;
+  }, [deps, options]); //eslint-disable-line react-hooks/exhaustive-deps
```
YousefED, line 37: this options dependency is still wrong

packages/website/docs/docs/side-menu.md
```diff
@@ -94,21 +159,28 @@ export default function App() {
...
 `CustomDragHandleMenu` is the component we use to replace the default Drag Handle Menu. You can see it's made up of a bunch of other components that are exported by BlockNote. Read on to [Components](/docs/side-menu#components) to find out more about these.
 
+Here, we're passing it as a prop to the `DefaultSideMenu` component, but you can also pass it to the `DragHandle` component we saw in the [previous demo](/docs/side-menu#custom-side-menu), if you don't want to use the default Side Menu layout.
+
 After creating `CustomDragHandleMenu`, we tell BlockNote to use it inside `BlockNoteView`. [Changing UI Elements](/docs/ui-elements) has more information about how this is done.
 
 ## Components
 
-It might seem daunting to create your own Formatting Toolbar from scratch, which is why BlockNote provides React components that you can use which match the default styling.
+It might seem daunting to create your own Side Menu or Drag Handle Menu from scratch, which is why BlockNote provides React components that you can use which match the default styling.
```
YousefED, line 168: personal take, think this paragraph is a bit too much. I would focus on the core and just make it

> BlockNote provides React components that you can reuse to easily create your menus.

packages/react/src/hooks/useBlockNote.ts
matthewlipski: I'm confused on how to do this - I tried another approach:
```ts
export const useBlockNote = <BSchema extends BlockSchema = DefaultBlockSchema>(
  options: Partial<BlockNoteEditorOptions<BSchema>> = {},
  deps: DependencyList = []
): BlockNoteEditor<BSchema> => {
  const editorRef = useRef<BlockNoteEditor<BSchema>>();
  const newDeps = [...deps, options];

  const ret = useMemo(() => {
    ...
  }, newDeps);

  return ret;
};
```
But this also gives me a linting error that options is not in deps.

YousefED: there's //eslint-disable-line react-hooks/exhaustive-deps right?

matthewlipski: Yeah, but I'm still confused. I get that now we just pass an already created array to deps so React doesn't have to keep recreating it, but how come it's not an issue anywhere else we use hooks? Since pretty much everywhere else, we do declare the array in deps.

Issue #241, opened 2023-06-14
 Error [ERR_REQUIRE_ESM]: require() of ES Module while using NX monorepo


hitwill:
**Describe the bug**
Error on compile while using NX mono-repo
![image](https://github.com/TypeCellOS/BlockNote/assets/2501079/22f0540d-b86e-42d2-bf9d-2249b7274c20)

This issue does not happen with a vanilla NextJS app. However, it happens on NX  Monorepo

**To Reproduce**
Install [NX](https://nx.dev/getting-started/intro)
Create a new app: `nx g @nx/next:app my-app`
Install BLocknote: `npm i @blocknote/react`

Add the code to index.tsx of your app
```javascript
import styles from './index.module.css';
import { Inter } from 'next/font/google';
import { BlockNoteView, useBlockNote } from '@blocknote/react';
import '@blocknote/core/style.css';

const inter = Inter({ subsets: ['latin'] });

export function Index() {
  const editor = useBlockNote({
    onEditorContentChange: (editor) => {
      // Log the document to console on every update
    },
  });
  return (
    <div className={styles.page}>
      <div className="wrapper">
        <div className="container">
          <div id="welcome">
            <h1>
              <span> Hello there, </span>
            </h1>
          </div>
          <div>
            {' '}
            <BlockNoteView editor={editor} />
          </div>
        </div>
      </div>
    </div>
  );
}

export default Index;
````

**Misc**
- Node version:18.11.0 and 20.3.0
- Package manager: tried with NPM and Yarn

hitwill:
I am not sure whether to list this as a bug since it happens specifically with NX.

guptabhaskar:
I'm also facing the same issue, is there a good way to fix/avoid this?

I added this to my next.config.js which fixed the issue but I don't think this is a good idea for the long term:
`experimental: { esmExternals: true }`

YousefED:
This should be addressed here! https://github.com/TypeCellOS/BlockNote/pull/274

Let me know if you're still running into this issue

YousefED:
Are you using the latest version of blocknote @Ianduha13 ?

hitwill:
@Ianduha13 If you are on NextJS, please see this: https://www.blocknotejs.org/docs/nextjs

PR #274, opened 2023-07-12
nextjs docs

YousefED:


moaazsidat (commented):
Looks good to me, thanks for digging in and a quick resolution!

packages/website/docs/docs/nextjs.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Next.js and BlockNote
+description: Details on integrating BlockNote with Next.js
+imageTitle: Next.js and BlockNote
+path: /docs/nextjs
+---
+
+# Next.js and BlockNote
+
+(only for next.js users)
```
matthewlipski, line 10: Don't think this line is necessary, the page is pretty clearly already targeted at NextJS users

packages/website/docs/docs/quickstart.md
```diff
@@ -46,6 +46,10 @@ function App() {
 
 As well as `BlockNoteView` and `useBlockNote`, we import `@blocknote/core/style.css` to provide default styling for the editor.
 
+::: warning Next.js usage (or other server-side React frameworks)
```
matthewlipski, line 49: I think it might make sense to broaden this to all serverside React frameworks since it's also applicable to those (we've had a couple of people having issues with NX too). We can have the last line be smth like "If you're using NextJS, read our guide on setting it up with BlockNote".

matthewlipski (changes requested):
Looks good! Just some minor things that I'd recommend changing

packages/website/docs/docs/quickstart.md
YousefED: NX is a framework for monorepos. My guess is those users were also using nextjs.

Agree that users of other server-side frameworks (e.g.: remix) might run into this as well, but let's wait until we actually get those reports :)

packages/website/docs/docs/nextjs.md
YousefED: agree, rephrased

matthewlipski (approved):

hitwill:
Confirmed it solves the original problem: https://github.com/TypeCellOS/BlockNote/issues/241

PR #288, opened 2023-07-31
simplify ui elements

YousefED:
Proposal for a simplified setup of the ui elements


packages/core/src/shared/plugins/suggestion/SuggestionPlugin.ts
```diff
@@ -64,24 +40,6 @@ class SuggestionsMenuView<
       updateSuggestionsMenu(this.suggestionsMenuState);
     };
 
-    this.itemCallback = (item: T) => {
```
YousefED, line 67: I don't think this was used

packages/react/src/SlashMenu/components/SlashMenuPositioner.tsx
```diff
@@ -36,21 +31,15 @@ export const SlashMenuPositioner = <
     useState<number>();
 
   const referencePos = useRef<DOMRect>();
-  const callbacks =
-    useRef<SuggestionsMenuCallbacks<ReactSlashMenuItem<BSchema>>>();
 
   useEffect(() => {
-    callbacks.current = props.editor.createSlashMenu((slashMenuState) => {
+    return props.editor.slashMenu.on("update", (slashMenuState) => {
```
matthewlipski, line 36: Is there any point having the first param? Isn't it always just "update"? Same for the other UI elements

packages/core/src/shared/EventEmitter.ts
```diff
@@ -0,0 +1,58 @@
...
+type CallbackType<
+  T extends Record<string, any>,
+  EventName extends StringKeyOf<T>
+> = T[EventName] extends any[] ? T[EventName] : [T[EventName]];
+type CallbackFunction<
+  T extends Record<string, any>,
+  EventName extends StringKeyOf<T>
+> = (...props: CallbackType<T, EventName>) => any;
+
+export class EventEmitter<T extends Record<string, any>> {
```
matthewlipski, line 13: Is it necessary to have this file? What's the benefit vs the previous approach?

matthewlipski (commented):
Nice! Much less code than before. As far as I can see it's mostly due to grouping the UI plugins in a single extension and and moving logic from `BlockNoteEditor` to `EventEmitter`. I am wondering though about using `EventEmitter` vs just adding similar logic to `BlockNoteEditor`, and I have a few questions about that.

PR #305, opened 2023-08-04
feat: Color scheme customization

matthewlipski:
This PR overhauls how users can style the BlockNote editor. This is done in 3 ways:

- Allowing users to change the colors, fonts, and border radii of the default UI (includes editor itself, and all menus/toolbars)
- Allowing users to add HTML class names to elements within the editor.
- Overriding the CSS for UI components.

There are still a few points of contention:

One of the main limitations is the extended theming customizability. I think most people are interested in just changing colors and fonts, for which we have a good solution, but overriding CSS could probably use some improvements for the future.

Another limitation is that the HTML classes set on elements cannot be dynamically changed atm. I can see this being a blocker for some as setting `hidden`, `locked`, or `selected` classes is a pretty common use case. However, the implementation is actually pretty complex so not sure it's worth our time atm.


examples/editor/src/App.tsx
```diff
@@ -1,26 +1,150 @@
...
-    editorDOMAttributes: {
-      class: styles.editor,
-      "data-test": "editor",
+    blockSchema: schema,
+    slashMenuItems: [
+      ...getDefaultReactSlashMenuItems(schema),
+      insertImage,
+      insertReactImage,
+    ],
+    domAttributes: {
```
YousefED, line 118: should we be able to pass custom domattributes for blocks here?

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -151,8 +151,20 @@ export function createBlockSpec<
       return ({ HTMLAttributes, getPos }) => {
         // Create blockContent element
         const blockContent = document.createElement("div");
-        // Sets blockContent class
-        blockContent.className = styles.blockContent;
+        // Add custom HTML attributes
+        const blockContentDOMAttributes =
```
YousefED, line 155: why not blockContentDOMAttributes.blockContent.class?

packages/core/src/extensions/Blocks/nodes/BlockContainer.ts
```diff
@@ -89,11 +86,16 @@ export const BlockContainer = Node.create<IBlock>({
...
         "div",
-        mergeAttributes(HTMLAttributes, {
-          // TODO: maybe remove html attributes from inner block
-          class: styles.block,
-          "data-node-type": this.name,
-        }),
+        mergeAttributes(
+          {
+            ...domAttributes,
+            class: domAttributes.class
```
YousefED, line 92: let's start using either https://www.npmjs.com/package/classnames for concatenation, or create a simple helper function ourselves

examples/editor/src/App.tsx
```diff
@@ -1,26 +1,150 @@
...
+import { BlockSchema, defaultBlockSchema } from "@blocknote/core";
+import { Image, insertImage } from "../../../tests/utils/customblocks/Image";
+import {
+  insertReactImage,
+  ReactImage,
+} from "../../../tests/utils/customblocks/ReactImage";
 
 type WindowWithProseMirror = Window & typeof globalThis & { ProseMirror: any };
 
+const redTheme: Partial<Theme> = {
```
YousefED, line 20: I think adding dark and light modes as part of a theme creates a more complex API, and also more work (if you're creating a "red" theme, you'll need to implement both dark and light mode for it. In practice, what I think will happen, is that people will just insert duplicates or dummy modes for one).

I think what we should do instead is making BlockNoteLightTheme and BlockNoteDarkTheme default themes that we expose.

The `theme` prop on the editor could then take: `Theme | { dark: Theme, light: Theme}` and defaults to `{dark: BlockNoteDarkTheme, light: BlockNoteLightTheme}`. 

If a dark/light config is passed, we should auto-switch to the right theme based on browser settings

packages/react/src/BlockNoteTheme.ts
```diff
@@ -1,82 +1,110 @@
...
+export type Theme = {
+  colors: ColorScheme;
+  borderRadius: number;
+  fontFamily: string;
+};
 
-  const sideMenu = useDarkTheme
-    ? blockNoteColorScheme[4]
-    : blockNoteColorScheme[2];
+export const blockNoteToMantineTheme = (
```
YousefED, line 43: My suggestion for dark / light would also make the code in this method a lot easier

packages/react/src/BlockNoteTheme.ts
```diff
@@ -85,33 +113,69 @@ export const getBlockNoteTheme = (
...
+          : fullTheme.colors.highlightColors.blue.background.light,
+        useDarkTheme
+          ? fullTheme.colors.highlightColors.purple.background.dark
+          : fullTheme.colors.highlightColors.purple.background.light,
+        useDarkTheme
+          ? fullTheme.colors.highlightColors.pink.background.dark
+          : fullTheme.colors.highlightColors.pink.background.light,
       ],
     },
     components: {
```
YousefED, line 181: I'm wondering if we should make it easy to customize component styles as well for consumers.

I realize this will be mantine specific, so it should be an advanced option for sure. I also think we should only do it if it's easy to implement. The benefit I see is that it will basically allow almost full styleability of UI components (without creating your own) - even though users would probably need to dive into the source / mantine to see what exactly is stylable.

I think the main thing we need to know for this is whether we can extend from a mantine theme and keep both our `components` settings, and add new ones? i.e.: is the `components` setting extendable?

packages/react/src/BlockNoteView.tsx
```diff
@@ -1,33 +1,66 @@
...
 export function BlockNoteView<BSchema extends BlockSchema>(
   props: {
     editor: BlockNoteEditor<BSchema>;
+    useDarkTheme?: boolean;
+    customTheme?: Partial<Theme>;
     children?: ReactNode;
   } & HTMLAttributes<HTMLDivElement>
 ) {
-  const { editor, children, ...rest } = props;
+  const { classes } = createStyles({ root: {} })(undefined, {
```
YousefED, line 50: I think you can also pass the theme to `createStyles`? In that case the extra `UnThemedBlockNoteView` is not necessary

packages/react/src/SharedComponents/ColorPicker/components/ColorIcon.tsx
```diff
@@ -7,21 +7,27 @@ export const ColorIcon = (
...
 
   return (
     <Box
+      className={classes.root}
       sx={(theme) => {
         return {
           backgroundColor:
             theme.colors.backgroundColors[
-              theme.other.colors.indexOf(backgroundColor)
+              theme.other.highlightColorNames.indexOf(backgroundColor)
```
YousefED, line 25: should `highlightColorNames` be part of the mantine theme? it seems unrelated tbh

YousefED (commented):
Looks neat! Added some points to discuss, but I think we can quickly move to docs (keep it simple) & tests

packages/react/src/SharedComponents/ColorPicker/components/ColorIcon.tsx
matthewlipski: I think it's important since they should really be different for the light and dark scheme, otherwise you get contrast issues and it generally looks bad. Maybe values for light & dark should just be hardcoded though?

examples/editor/src/App.tsx
matthewlipski: Don't see why not, I think most people will still just add classes but imo it adds very little complexity so I think we may as well give people the option.

examples/editor/src/App.tsx
matthewlipski: Yep I agree with this. How about the `useDarkMode` prop, do you think we should still keep it and have it default to the user's browser/OS setting? Or remove it entirely and just decide which to use based on browser/OS setting?

packages/react/src/BlockNoteTheme.ts
matthewlipski: I think it should be possible, though I don't think we should spend much time on it. You can technically already do this with just CSS, e.g.:
```css
.mantine-Menu-item {
  background-color: blue !important;
}
```
So it's only really a semantic difference for a use case that we don't expect to be too common.

examples/editor/src/App.tsx
YousefED: I think we can remove it. If someone wants to manually control it, they can just pass the light or darktheme directly as “theme”

packages/core/src/extensions/Blocks/api/block.ts
matthewlipski: For each of the DOM element types, you can add custom attributes, not just a custom class

packages/react/src/BlockNoteView.tsx
matthewlipski: Hmm don't think so - instead of an object containing styles (in this case `{ root: {} }`), you can provide a function which takes a theme and returns an object containing styles. But that theme param is still taken from the Context, I don't think there's a way to actually pass a theme in yourself:

```ts
export declare function createStyles<
  Key extends string = string, 
  Params = void, 
  Input extends Record<Key, CSSObject> = Record<Key, CSSObject>
>(
  input: 
    | (theme: MantineTheme, params: Params, createRef: (refName: string) => string) => Input
    | Input
): (
  params: Params, 
  options?: UseStylesOptions<Key>
) => {
  classes: { [key in keyof Input]: string; };
  cx: (...args: any) => string;
  theme: MantineTheme;
};
```

examples/editor/src/App.tsx
```diff
@@ -1,26 +1,162 @@
...
+} satisfies Partial<Theme>;
+
+const schema = {
+  ...defaultBlockSchema,
+  image: Image,
+  reactImage: ReactImage,
+} satisfies BlockSchema;
+
+function App() {
+  const [theme, setTheme] = useState<
```
YousefED, line 79: This should note be here right? if the consumer passes dark + light to the theme prop, blocknote should do the selection?

packages/core/src/extensions/Blocks/api/block.ts
YousefED: Ok, you could do this:

constr { class, ...otherAttributes} = this.options.domAttributes?.blockContent || {};

packages/react/src/BlockNoteTheme.ts
```diff
@@ -1,117 +1,160 @@
...
+    dark: darkDefaultTheme,
+  },
+  componentStyles?: BlockNoteComponentStyles
 ): MantineThemeOverride => {
-  const boxShadow = `0px 4px 8px ${
-    useDarkTheme ? blockNoteColorScheme[8] : blockNoteColorScheme[2]
-  }, 0px 0px 1px ${
-    useDarkTheme ? blockNoteColorScheme[6] : blockNoteColorScheme[1]
-  }`;
+  const fullTheme: Theme = shouldUseDarkTheme(theme)
```
YousefED, line 87: This code and `shouldUseDarkTheme` is difficult to read. To fix: first check if { light:..., dark: ... } was passed, or whether a single theme was passed.

i.e.:

```
if ('light' in theme && 'dark' in theme) {
    theme = browserWantsDarkTheme() ? theme.dark : theme.light;
}
```

packages/react/src/BlockNoteTheme.ts
```diff
@@ -1,117 +1,160 @@
...
+        ...darkDefaultTheme,
+        ...("dark" in theme ? theme.dark : theme),
+      }
+    : {
+        ...lightDefaultTheme,
+        ...("light" in theme ? theme.light : theme),
+      };
+  console.log(`.${blockStyles.isEmpty} .${blockStyles.inlineContent}`);
+
+  const editorText = fullTheme.colors.editor.text;
```
YousefED, line 98: do we need line 98 - 114?

packages/react/src/BlockNoteTheme.ts
```diff
@@ -246,22 +325,24 @@ export const getBlockNoteTheme = (
                 backgroundColor: hoveredBackground,
               },
             },
+            ...componentStyles?.Toolbar,
```
YousefED, line 328: The difficulty now of course is that if I want to change sth to `.mantine-UnstyledButton-root`, it will override the stuff from line 290. I suppose there's no easy way to prevent this? (again, let's not spend too much time on this)

packages/react/src/ReactBlockSpec.tsx
```diff
@@ -38,16 +48,36 @@ export type ReactBlockConfig<
...
-    className={`${props.className ? props.className + " " : ""}${
-      blockStyles.inlineContent
-    }`}
-  />
-);
+) => {
+  const inlineContentDOMAttributes =
+    useContext(BlockNoteDOMAttributesContext).inlineContent || {};
+
+  const classNames = useMemo(
```
YousefED, line 59: I think a usememo is overkill here as it's just a simple string concatenation

packages/react/src/defaultTheme.ts
```diff
@@ -0,0 +1,124 @@
+import { Theme } from "./index";
```
YousefED, line 1: nice that this file is extracted! super clear. Maybe rename to defaultThemes or create separate files for light / dark

packages/react/src/SharedComponents/ColorPicker/components/ColorIcon.tsx
YousefED: I understand they should be themeable, I just don't understand the entire `indexOf` construction. Aren't we overcomplicating things?

examples/editor/src/App.tsx
matthewlipski: It does, but it needs to also trigger a re-render if the browser theme changes, though this should probably be in `BlockNoteView` either way.

packages/react/src/BlockNoteTheme.ts
matthewlipski: We would have to deep merge the existing styles with the new styles, I think theres a function from lodash actually which might do that.

packages/core/src/extensions/Blocks/api/block.ts
matthewlipski: Actually this also doesn't really work since `class` is a reserved word for class definitions

packages/react/src/BlockNoteTheme.ts
```diff
@@ -1,316 +1,385 @@
...
+    purple: CombinedColor;
+    pink: CombinedColor;
+  };
+};
 
-  const border = `1px solid ${
-    useDarkTheme ? blockNoteColorScheme[7] : blockNoteColorScheme[1]
-  }`;
+export type Theme = {
+  type: "light" | "dark";
```
YousefED, line 34: why should a theme have a type?

packages/website/docs/docs/theming.md
```diff
@@ -0,0 +1,261 @@
...
+
+`blockContainer:` The main container element for blocks. Contains both the block's content and its nested blocks.
+
+`blockGroup:` The wrapper element for all top-level blocks in the editor and nested blocks.
+
+`blockContent:` The wrapper element for a block's content.
+
+`inlineContent:` The wrapper element for a block's rich-text content.
+
+## Overriding CSS
```
YousefED, line 176: maybe add "(advanced)" to this

packages/react/src/BlockNoteView.tsx
```diff
@@ -1,33 +1,136 @@
...
 import { FormattingToolbarPositioner } from "./FormattingToolbar/components/FormattingToolbarPositioner";
 import { HyperlinkToolbarPositioner } from "./HyperlinkToolbar/components/HyperlinkToolbarPositioner";
 import { SideMenuPositioner } from "./SideMenu/components/SideMenuPositioner";
 import { SlashMenuPositioner } from "./SlashMenu/components/SlashMenuPositioner";
+import { darkDefaultTheme, lightDefaultTheme } from "./defaultThemes";
+
+// Helper function which takes a partial theme provided by the user, and turns
+// it into a full theme. Either the light or dark theme is used, depending on
+// which ones the user provided and the browser settings.
+const getFullTheme = (baseTheme: "light" | "dark", theme?: PartialTheme) => {
```
YousefED, line 20: I still don't really get this, let's discuss on discord

YousefED (commented):
some more comments

packages/website/docs/docs/theming.md
```diff
@@ -0,0 +1,261 @@
...
+export default function App() {
+  // Creates a new editor instance.
+  const editor: BlockNoteEditor = useBlockNote();
+
+  // Renders the editor instance using a React component.
+  return (
+    <BlockNoteView
+      editor={editor}
+      theme={"{{ getTheme(isDark) }}"}
+      componentStyles={(theme) => ({
```
YousefED, line 198: Should we make this part of the theme instead of a separate prop?

packages/website/docs/docs/theming.md
```diff
@@ -0,0 +1,261 @@
...
+
+BlockNote allows you to change how the editor UI looks. You can change the theme of the default UI, or override its CSS styles.
+
+## Theming
+
+If you want to make small adjustments to BlockNote's UI, you can create a custom theme, which will let you change colors, fonts, and border radii. You can do this by passing partial `Theme` objects to `BlockNoteView`.
+
+Take a look at how this is done in the demo below:
+
+::: sandbox {template=react-ts}
```
YousefED, line 25: For sake of simplicity, I would just make one "red theme" and not a red light / dark theme

packages/core/src/extensions/Blocks/api/block.ts
YousefED: Check. We could decide to rename `class` to `className`, that's also inline with how React does this

packages/react/src/BlockNoteView.tsx
YousefED: I think you're right. In that case I have a small preference to rename `UnThemedBlockNoteView` to something else (`InnerBlockNoteView`)? While I get what you mean, "UnThemed" sounds to me like it's not themed, which is not really the case

packages/core/src/extensions/Blocks/api/block.ts
matthewlipski: Not sure if it's a good idea tbh, since we're not only applying the class to React components. Basic nodes like `blockContainer`, `blockGroup`, etc, are defined outside of React, so I think it's better to keep HTML naming conventions for this rather than React ones

packages/website/docs/docs/theming.md
matthewlipski: I think it's nice to have both since it both shows off the auto switching from browser settings and saves us having to explain how to pass in a light & dark theme instead of just a single theme. Theming is also not really complicated compared to a lot of the other stuff, since 90% of it is defining colors, so imo we shouldn't run into the issue of people getting overloaded with information.

packages/react/src/BlockNoteView.tsx
```diff
@@ -1,33 +1,97 @@
...
     children?: ReactNode;
   } & HTMLAttributes<HTMLDivElement>
 ) {
-  const { editor, children, ...rest } = props;
+  const {
+    theme = { light: lightDefaultTheme, dark: darkDefaultTheme },
+    ...rest
+  } = props;
+
+  const [browserTheme, setBrowserTheme] = useState<"light" | "dark">(
```
YousefED, line 61: maybe use a separate hook like https://github.com/rfoel/use-prefers-color-scheme/blob/main/src/index.ts to isolate this logic?

packages/website/docs/docs/theming.md
```diff
@@ -0,0 +1,259 @@
...
+  Theme,
+  useBlockNote,
+} from "@blocknote/react";
+import "@blocknote/core/style.css";
+
+// Default dark theme with additional component styles.
+const theme = {
+  ...darkDefaultTheme,
+  componentStyles: (theme) => ({
+    // Adds basic styling to the editor.
```
YousefED, line 198: You're also setting the font to comic sans, but don't explain that anywhere

YousefED (approved):
Looks great!

PR #307, opened 2023-08-11
Fix problem about custom block with `containsInlineContent: false`

dlwocks31:
This pull request attempts to fix the problem regarding custom block with `containsInlineContent: false`.

image

Problem: For a custom block without inline content, its corresponding nodeSize is 3. The original code queried `doc.resolve(pos)` starting from pos = 2, which is larger than outerBlockGroupEndPos, which caused the error "Position 2 out of range". Therefore, I think querying `doc.resolve(pos)` from pos = 1 will fix the problem. I verified this works in example project with custom block added.

I am not familiar with ProseMirror and BlockNote source code, so I am not sure whether this is a good way to fix this problem, so I need some advice from maintainers.

### DIFF
--- a/examples/editor/package.json
+++ b/examples/editor/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@blocknote/example-editor",
   "private": true,
-  "version": "0.8.5",
+  "version": "0.9.0",
   "scripts": {
     "dev": "vite",
     "build": "tsc && vite build",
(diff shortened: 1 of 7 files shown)


dlwocks31:
@matthewlipski : I found out that this code is recently modified by you, so could you please look into this and give me your opinion?

PR #312, opened 2023-08-14
fix: Custom blocks without inline content

### LINKED ISSUE #300: getBlockInfoFromPos is still buggy in 0.8.5
**Describe the bug**
I was incredibly excited about the menus refactor and I decided to play around with it today. 

Would like to apologize for not being able to provide a proper sandbox, but I think the bug is easy to show.

getBlockInfoFromPos() is called by nodeToBlock() which is called with arbitrary nodes.

```
//console.log() messages:

outerBlockGroupEndPos 1
in pos <= outerBlockGroupStartPos
pos is 2
doc nodeSize 3
Uncaught RangeError: Position 2 out of range
    at ResolvedPos.resolve
    at ResolvedPos.resolveCached
```

0.8.5 resolved collaboration (was struggling with that too), but it broke some of this other functionality. The confusion was likely caused by the fact that getBlockInfoFromPos()'s param is called doc and not node.

matthewlipski:
A recent change has broken custom blocks which don't contain inline content. This is because we added edge case handling in `getBlockInfoFromPos`, for situations where the selection is outside of any blocks. However, `nodeToBlock` was unnecessarily and incorrectly calling `getBlockInfoFromPos` with the `blockContainer` node instead of the `doc`, and was causing errors when inserting custom blocks without inline content.

Closes #300

### DIFF
--- a/packages/core/src/api/nodeConversions/nodeConversions.ts
+++ b/packages/core/src/api/nodeConversions/nodeConversions.ts
@@ -16,7 +16,6 @@ import {
   Styles,
   ToggledStyle,
 } from "../../extensions/Blocks/api/inlineContentTypes";
-import { getBlockInfoFromPos } from "../../extensions/Blocks/helpers/getBlockInfoFromPos";
 import UniqueID from "../../extensions/UniqueID/UniqueID";
 import { UnreachableCaseError } from "../../shared/utils";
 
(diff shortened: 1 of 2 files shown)


YousefED (approved):
Why was the call incorrect?

Nice that it's fixed. I do think we can improve this by making `getBlockInfoFromPos` and `nodeToBlock` call a method `blockContainerNodeToBlockInfo` or something? That would remove some duplicate code right?

matthewlipski:
> Why was the call incorrect?
> 
> Nice that it's fixed. I do think we can improve this by making `getBlockInfoFromPos` and `nodeToBlock` call a method `blockContainerNodeToBlockInfo` or something? That would remove some duplicate code right?

Well thing is that the main point of `getBlockInfoFromPos` is finding the nearest `blockContainer` node. Getting stuff like child count, depth, etc is pretty trivial after that. So it doesn't really make sense to call it if you already have the `blockContainer` node imo. Your suggestion makes sense though, just having a `getBlockInfo` function which takes a `blockContainer` node instead of a position.

PR #324, opened 2023-08-18
feat: Improved block type dropdown customization

matthewlipski:
This PR makes the block type dropdown more customizable. Both it, and the default formatting toolbar, take an `items` prop which you can replace the items with. The object type has also been much simplified, and a default list of items is provided.

### DIFF
--- a/packages/react/src/FormattingToolbar/components/DefaultDropdowns/BlockTypeDropdown.tsx
+++ b/packages/react/src/FormattingToolbar/components/DefaultDropdowns/BlockTypeDropdown.tsx
@@ -1,10 +1,5 @@
 import { useMemo, useState } from "react";
-import {
-  Block,
-  BlockNoteEditor,
-  BlockSchema,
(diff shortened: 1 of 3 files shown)


packages/react/src/FormattingToolbar/components/DefaultDropdowns/BlockTypeDropdown.tsx
```diff
@@ -15,114 +10,112 @@ import {
...
+          return false;
+        }
+      }
+
+      return true;
+    });
+  }, [props.editor, props.items]);
+
+  const shouldShow: boolean = useMemo(
+    () => filteredItems.map((item) => item.type).includes(block.type as string),
```
YousefED, line 99: minor detail, you can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find for this

YousefED (approved):

Issue #328, opened 2023-08-23
BlockTypeDropdown can not select Heading


jerrycompal:
BlockTypeDropdown cannot select Heading

![1692799174614](https://github.com/TypeCellOS/BlockNote/assets/39256241/be1cff2c-4f60-42c1-9a18-2394db39cf21)

matthewlipski:
Fixed in 0.9.3!

PR #332, opened 2023-08-27
fix: BlockTypeDropdown can not select Heading

prathameshjchavan:
BlockTypeDropdown can now able to select the appropriate heading as per the level of the heading.

fixes: #328

Changes:
- Changed the logic for isSelected prop
- If the block.type is 'heading', it will also check for the level of heading in block.props.level

https://github.com/TypeCellOS/BlockNote/assets/74180320/8eccfe54-8502-4219-955c-46d303f7bf95


matthewlipski:
Fixed in #322

Issue #333, opened 2023-08-28
BlockTypeDropdown cannot update to different levels of heading


prathameshjchavan:
**Describe the bug**
Once we defined a block for a particular level of heading, let's say we have added a block of type heading of level 1. 
Then, if we try to change the block to the heading of level 2, it doesn't update.

**To Reproduce**
1. Open Editor
2. Select the block type as Heading (H1) by typing  '/heading' and clicking enter
3. Enter some text into it
4. Select the whole text and the formatting toolbar appears
5. Select Heading 2 (H2) from the BlockType Dropdown

- Node version: v18.17.1
- Package manager: npm
- Browser: Chrome Version 116.0.5845.111 (Official Build) (64-bit)

matthewlipski:
Fixed in 0.9.3!

PR #334, opened 2023-08-28
fix: block type dropdown cannot update to different levels of heading

prathameshjchavan:
BlockTypeDropdown can now able to update different levels of heading.

fixes: #333

Changes:

- Changed the updateBlock function to check the block type as 'heading'
- If it is heading then, it will pass the level of heading as props instead of empty object {}.

https://github.com/TypeCellOS/BlockNote/assets/74180320/54303bc9-4e7d-4036-bc20-647bbb189187


kaianuar:
Would be good to let this go through as it's a bug that's affecting my implementation of BlockNote as well.

matthewlipski:
Fixed in #322

PR #336, opened 2023-09-04
fix: custom blocks may error when it is re-created

teyc:
The custom block wrapper attempts to read the block id of the inner block using getPos(), but the inner block has not been initialized by tiptap yet. Do not attempt to render this block until it is ready

### DIFF
--- a/packages/react/src/ReactBlockSpec.tsx
+++ b/packages/react/src/ReactBlockSpec.tsx
@@ -127,6 +127,14 @@ export function createReactBlockSpec<
         // Gets position of the node
         const pos =
           typeof props.getPos === "function" ? props.getPos() : undefined;
+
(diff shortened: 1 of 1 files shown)


teyc:
Thanks for having a look: sample here: https://github.com/teyc/custom-blocknote-bug

matthewlipski:
Alright I've found the root cause of the issue, so should be able to fix it without needing a workaround. Basically, the editor's content is set before the underlying TipTap editor is done initializing, so any custom blocks which use the BlockNote editor's functions will fail to render. Should be able to fix it by just waiting for the TipTap editor before setting the initial content, just varifying with @YousefED👍

matthewlipski:
Gonna close this PR and merge #348 since it fixes the root issue and works for both vanilla and React custom blocks, but thanks for bringing this bug to light and helping with this!

PR #341, opened 2023-09-20
Bugfix/code block in parentheses

### LINKED ISSUE #338: Code Block is not generated inside parentheses
**Describe the bug**
when I make Code Block using backtick in parentheses, then Code Block not generated
 
code block not created

code block not working cases

I don't know why it behaves like this because I haven't figured out the behavior of the code yet. However, if I have enough time, I will try to figure out the problem. First of all, I think it's important to note that there is a bug, so I'm writing this down.

**To Reproduce**
This bug can demonstrate on BlockNote's demo site. (The gif posted earlier was also created on the demo site)

1. Access https://www.blocknotejs.org/ in Chrome (Chrome 116.0.5845.180)
2. Write the string between parentheses with backticks.

marwan37:
### Description
In the current implementation of the [@tiptap/extension-code](https://github.com/ueberdosis/tiptap/blob/39cf6979c49e953118bbbe4b894a1dc296128932/packages/extension-code/src/code.ts#L4), the regex pattern does not correctly format text enclosed in backticks when they are immediately preceded or followed by any character other than whitespace.

fixes: [#338](https://github.com/TypeCellOS/BlockNote/issues/338)

**Changes**
- Created a new folder and file at: `packages/code/extensions/CustomCode/CustomCodeExtension.ts`.
- Copied the `code.ts` file from TipTap's extension and updated the `inputRegex` and `pasteRegex` patterns (see below)
- The regex uses negative lookbehind and lookahead assertions to ensure that only single backticks are considered for formatting to avoid conflicts with potential code block syntax or other uses of multiple backticks.

```ts
export const inputRegex = /(?<!`)`([^`]+)`(?!`)/;
export const pasteRegex = /(?<!`)`([^`]+)`(?!`)/g;
```

**Testing**
- The updated regex pattern has been tested in various scenarios to ensure that it correctly formats text as inline code blocks. 

BlockNoteCustomCode


matthewlipski:
Nice find! Though if this is an issue in TipTap's code base, wouldn't it be better to make a PR in their repo? It seems like a pretty major workaround to pretty much copy a file from TipTap to BlockNote, when the change could be made directly at the source of the issue

marwan37:
@matthewlipski You're absolutely right. My initial approach was to provide a quick fix for BlockNote users, but I understand the concerns about maintaining a separate copy of the file. I'll look into submitting a PR to TipTap directly.

matthewlipski:
Awesome, let me know how it goes - if for some reason the TipTap guys don't approve the changes, we can sort out a fix in BlockNote

marwan37:
@matthewlipski Thanks for the guidance! I've submitted a PR to the TipTap repo addressing the issue [here](https://github.com/ueberdosis/tiptap/pull/4468#issuecomment-1730087569). I'll keep you posted on its progress.

PR #343, opened 2023-09-21
Fix: CreateLinkButton does not work when click event propagation prevented

YoshiyukiKato:
Hi, thank you for providing great library!

I found a case that `CreateLinkButton` does not work when `BlockNoteView` wrapped by a component with stoping click event propagation, such as modal component, like following:

```tsx
...
<div onClick={(e) => { e.stopPropagation();}}>
  <BlockNoteView editor={editor} />
</div>
...
```

This is because, currently `LinkToolbarButton` captures click event via event listener of `document.body`.
So, this PR fix the problem by changing the way to capture click event of the button as setting `onClick` handler to the button directly.

### DIFF
--- a/packages/react/src/FormattingToolbar/components/LinkToolbarButton.tsx
+++ b/packages/react/src/FormattingToolbar/components/LinkToolbarButton.tsx
@@ -1,5 +1,5 @@
 import Tippy from "@tippyjs/react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useRef, useState } from "react";
 import {
   ToolbarButton,
   ToolbarButtonProps,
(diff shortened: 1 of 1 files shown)


matthewlipski:
Nice fix! No idea what I was thinking when I originally wrote that, `menuRef` isn't even attached to anything😅

YoshiyukiKato:
@matthewlipski Thank you for comment!
As you said, `menuRef` and `buttonRef` is not used anymore. I removed it😌

And I add `onHide` handler to fix behavior when a toolbar component hidden by change of focus.

YousefED:
@matthewlipski I think this has been fixed in the meantime, but please check

PR #346, opened 2023-09-22
small fixes

YousefED:


YousefED (commented):
Some small suggestions.

Main change is to change default uploader to tmpfiles.org, as it doesn't require an API key, and also works for non-images that we might want to support in the future.

packages/core/src/BlockNoteEditor.ts
```diff
@@ -217,25 +215,24 @@ export class BlockNoteEditor<BSchema extends BlockSchema = DefaultBlockSchema> {
...
-      return fetch("https://api.imgbb.com/1/upload?expiration=600", {
-        method: "POST",
-        body: body,
-      })
-        .then((response) => response.json())
-        .then((data) => data.data.url);
-    };
+    this.uploadFile =
+      newOptions.uploadFile ||
+      (async (file) => {
```
YousefED, line 220: I think we should move this example upload function to a separate file, and name it sth like `uploadToTmpFilesOrg_DEV_ONLY`. Then, it becomes clear that this should be replaced by the user. I also don't think we should set it as a default, but that the user should always pass `uploadFile` as part of the options.

matthewlipski: Instead of a separate file, I think it's better to just add it as an example to the docs

YousefED: Hmm, then the code to get started (both for dev and in docs) will be quite cumbersome, because they’ll always need to cooy and paste that code

matthewlipski (approved):

PR #348, opened 2023-09-26
fix: Error when custom blocks are in `initialContent`

matthewlipski:
This PR fixes an error that's thrown when custom blocks are present in teh `initialContent` editor option, as they are rendered before the TipTap editor is initialized.


packages/core/src/BlockNoteEditor.ts
```diff
@@ -216,28 +216,36 @@ export class BlockNoteEditor<BSchema extends BlockSchema = DefaultBlockSchema> {
...
-          schema.node("blockGroup", undefined, ic)
+          schema.node("blockGroup", undefined, [
+            blockToNode({ id: "initialBlock", type: "paragraph" }, schema),
+          ])
         );
-        // override the initialcontent
         editor.editor.options.content = root.toJSON();
       },
+      onCreate: () => {
+        newOptions.onEditorReady?.(this);
```
YousefED, line 239: I think we should move this to after replaceBlocks. Because the user will expect the blocks to have the content of `initialContent` when `onEditorReady` is called

Issue #351, opened 2023-09-27
Option to hide side menu when cursor is outside of the editor


danew:
Currently the side menu will appear next to the block where the cursors mouse Y position overlaps. It would be great if we could configure this behaviour to only appear when the mouse cursors position is inside of the editor.

This behaviour is currently controlled within the `SideMenuPlugin` under the `onMouseMove` method.

PR #352, opened 2023-09-27
Option to hide side menu when cursor is outside of the editor

danew:
This is in response to #351. By no means is this a final solution, need help on location of the configuration as well as the naming. Putting this up as an example of how we can achieve the desired functionality


YousefED:
Let me

PR #362, opened 2023-10-04
fix: Boolean/number props not getting parsed correctly

matthewlipski:
The last release made block props able to be booleans or numbers, not just strings. While the type definitions for this are all correct, there is an issue when parsing block props. All props are rendered to HTML attributes, so they must be converted to strings. This means when parsing them, we must also figure out whether to parse they should actually be booleans or numbers, which we weren't doing before. E.g. copy/pasting an image would cause the second image's `width` prop to be a string instead of a number.


packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -38,7 +38,35 @@ export function propsToAttributes<
...
+          }
+
+          if (value === "false") {
+            return false;
+          }
+
+          return null;
+        }
+
+        if (typeof spec.default === "number") {
```
YousefED, line 60: a number can also be a float.

Safest way:

use `parseFloat`

Check if result is numeric: `!isNaN(parseFloat(n)) && isFinite(n)` (https://stackoverflow.com/questions/9716468/pure-javascript-a-function-like-jquerys-isnumeric)

YousefED (commented):
nice. I think it's smart to add tests for this as this is core functionality

packages/core/src/extensions/Blocks/api/block.ts
```diff
@@ -38,7 +38,38 @@ export function propsToAttributes<
...
+          const isNumeric =
```
YousefED, line 61: th is parses the number 3 times. I think you can just do

```const asNum = parseFloat(value);

if(!Number.isNaN(asNum) && Number.isFinite(asNum)) {
return asNum;
}
return null;
```

YousefED (approved):
nice that you added the test!

PR #363, opened 2023-10-04
feat: Add Drag Handle Menu item `closeMenuOnClick` prop

### LINKED ISSUE #359: Adding the closeMenuOnClick prop for DragHandleMenuItem
If i put an input in DragHandleMenuItem, then typing every symbol in it causes the menu to close.
But in Mantine Menu (on the basis of which DragHandleMenu is made) it's enough to set closeMenuOnClick={false} for Menu.Item and everything works fine.

matthewlipski:
Closes #359

### DIFF
--- a/packages/react/src/SideMenu/components/DragHandleMenu/DragHandleMenuItem.tsx
+++ b/packages/react/src/SideMenu/components/DragHandleMenu/DragHandleMenuItem.tsx
@@ -2,7 +2,7 @@ import { Menu } from "@mantine/core";
 import { PolymorphicComponentProps } from "@mantine/utils";
 
 export const DragHandleMenuItem = (
-  props: PolymorphicComponentProps<"button">
+  props: PolymorphicComponentProps<"button"> & { closeMenuOnClick?: boolean }
 ) => {
   const { children, ...remainingProps } = props;
   return <Menu.Item {...remainingProps}>{children}</Menu.Item>;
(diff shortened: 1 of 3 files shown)


packages/react/src/SideMenu/components/DragHandleMenu/DragHandleMenuItem.tsx
```diff
@@ -2,7 +2,7 @@ import { Menu } from "@mantine/core";
 import { PolymorphicComponentProps } from "@mantine/utils";
 
 export const DragHandleMenuItem = (
-  props: PolymorphicComponentProps<"button">
+  props: PolymorphicComponentProps<"button"> & { closeMenuOnClick?: boolean }
```
YousefED, line 5: instead of hardcoding this, can't we reuse the Mantine props? i.e.: is there a prop type of Menu.Item we can reuse?

packages/website/docs/docs/side-menu.md
```diff
@@ -177,6 +177,10 @@ type SideMenuButtonProps = {
 }
 export const SideMenuButton = (props: SideMenuButtonProps) => ...;
 
-// Takes same props as `button` elements, e.g. onClick.
-export const DragHandleMenuItem = (props) => ...;
+// Also includes all props of regular `button` elements, e.g. onClick.
```
YousefED, line 180: this comment conflicts with the type definition below. I'd either:
a) make the type definition complete
b) put the comment inside the `= { ... }`
c) point to type definition in code

PR #376, opened 2023-10-13
fix: saving-loading meta info

danlgz:
The current saving-loading page info is not the correct

image


matthewlipski:
Nice catch and thanks for the quick fix!

PR #380, opened 2023-10-15
Fix #361 Error converting Image Block to Markdown

### LINKED ISSUE #361: Error converting Image Block to Markdown
**Describe the bug**
Newly inserted "Image Block" to the editor causes error when converting to markdown.

Error message:
```
Error: Expected node, not `undefined`
```

It happens when `editor.blocksToMarkdown()` is called with empty Image Block.

**To Reproduce**

1. Initialize editor with `blocksToMarkdown()` call in `onEditorContentChange`:

```
const editor = useBlockNote({
    onEditorContentChange: editor => {
      editor.blocksToMarkdown(editor.topLevelBlocks).then(markdown => {
        console.debug(markdown);
      });
    },
  });
```
2. Open editor in the browser & insert Image Block. Immediately `onEditorContentChange` event will fire, causing `blocksToMarkdown` to run, which throws an error.

Here is the block object with empty image causing the error:

```
{
    "id": "cdd6b0c5-fcb4-475c-a582-90a57f8639e1",
    "type": "image",
    "props": {
        "backgroundColor": "default",
        "textAlignment": "left",
        "url": "",
        "caption": "",
        "width": 512
    },
    "content": undefined,
    "children": []
}
```

**Expected behavior**

- Do not generate any markdown when image doesn't have url
- Generate markdown code (`[alt text](image_url)`) when there is an image.

jmas:
Fix for #361 Error converting Image Block to Markdown.

### DIFF
--- a/packages/core/src/api/formatConversions/simplifyBlocksRehypePlugin.ts
+++ b/packages/core/src/api/formatConversions/simplifyBlocksRehypePlugin.ts
@@ -38,6 +38,8 @@ export function simplifyBlocks(options: SimplifyBlocksOptions) {
         blockContent.properties!["dataContentType"] as string
       );
 
+      const isImageBlock = blockContent.properties!["dataContentType"] as string === 'image';
+
       const listItemBlockType = isListItemBlock
         ? options.orderedListItemBlockTypes.has(
             blockContent.properties!["dataContentType"] as string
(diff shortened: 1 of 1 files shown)


jmas:
Close, reason: need create seporate branch.

PR #408, opened 2023-11-06
Added editor option to disable nested blocks

abasse:
This PR adds an option to the editor  to disable nested blocks. (on tab command or via formatting toolbar)

example configuration "canNestBlock":

`export type BlockNoteEditorOptions = Partial<{
  editable: boolean;
  canNestBlock: boolean;
  initialContent: PartialBlock[];
  editorDOMAttributes: Record<string, string>;
  onEditorReady: (editor: BlockNoteEditor) => void;
  onEditorContentChange: (editor: BlockNoteEditor) => void;
  onTextCursorPositionChange: (editor: BlockNoteEditor) => void;
  slashMenuItems: ReactSlashMenuItem[];
  defaultStyles: boolean;
  uploadFile: (file: File) => Promise<string>
}>;`

### DIFF
--- a/packages/core/src/BlockNoteEditor.ts
+++ b/packages/core/src/BlockNoteEditor.ts
@@ -94,7 +94,7 @@ export type BlockNoteEditorOptions<BSchema extends BlockSchema> = {
   /**
    * Disables block nesting by the user if set to `false`.
    */
-  canNestBlock: boolean;
+  enableNestedBlocks: boolean;
   /**
    * The content that should be in the editor when it's created, represented as an array of partial block objects.
    */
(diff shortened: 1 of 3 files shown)


matthewlipski:
Looks good! Though these changes will still let you nest/unnest blocks programatically, e.g. using the formatting toolbar. I think we need 2 more things before we merge this:

1. Prevent nesting at the schema level - the `blockContainer` node currently has `content: "blockContent blockGroup?"`, meaning you can place a `blockContent` node in it, and optionally a `blockGroup` node for nested blocks. If the `canNestBlock` option is true, we should change this to `content: "blockContent"` and make sure everything still works.
2. Edit the definitions of `editor.canNestBlock` and `editor.canUnnestBlock` to just return false if the `canNestBlock` option is true.

abasse:
> Looks good! Though these changes will still let you nest/unnest blocks programatically, e.g. using the formatting toolbar. I think we need 2 more things before we merge this:
> 
> 1. Prevent nesting at the schema level - the `blockContainer` node currently has `content: "blockContent blockGroup?"`, meaning you can place a `blockContent` node in it, and optionally a `blockGroup` node for nested blocks. If the `canNestBlock` option is true, we should change this to `content: "blockContent"` and make sure everything still works.
> 2. Edit the definitions of `editor.canNestBlock` and `editor.canUnnestBlock` to just return false if the `canNestBlock` option is true.

Thanks for the feedback! I had a look at the code, but I'm not really sure where to start. Is that something you would be able to look at?

abasse:
The editor option is now called "enableNestedBlocks" to avoid conflicts with the existing function with the same name.

NicHaley:
I'm looking to disable nesting as well

jueinin:
@NicHaley https://github.com/TypeCellOS/BlockNote/issues/876 maybe you can view this

PR #413, opened 2023-11-16
feat: tables

YousefED:
The design / functionality of the tables follow Notion basic tables

_TODO_

**polish**
- [x] handle / prevent backspace in cells
- [x] design polish (handles)
- [x] prevent selecting entire table after inserting with /

**code**
- [x] implement insert / remove rows / columns functions
- [x] fix types / build
- [x] figure out where to put CSS

**testing**
- [x] manual testing (e.g.: test possible keyboard / command issues)
- [ ] add tests (unit tests for block handling like updateBlock and block return values, and maybe e2e)
- [ ] test copy / paste support

_Relevant research:_
- https://github.com/ProseMirror/prosemirror-tables
- https://tiptap.dev/api/nodes/table
- https://remirror.io/docs/extensions/tables-extension/
- https://remirror.io/docs/extensions/react-tables-extension
- https://atlaskit.atlassian.com/packages/editor/editor-core/example/kitchen-sink

grid:
- https://grid.glideapps.com/
- https://www.mantine-react-table.com/
- https://www.ag-grid.com/

custom elements:
- https://github.com/guardian/prosemirror-elements


packages/core/src/extensions/Blocks/nodes/BlockContainer.ts
```diff
@@ -253,8 +253,13 @@ export const BlockContainer = Node.create<{
...
                   )
                 )
+                // If the node doesn't contain editable content, we want to
+                // select the whole node. But if it does have editable content,
+                // we want to set the selection to the start of it.
                 .setSelection(
-                  new NodeSelection(state.tr.doc.resolve(startPos))
+                  state.schema.nodes[newType].spec.content === ""
+                    ? new NodeSelection(state.tr.doc.resolve(startPos))
+                    : new TextSelection(state.tr.doc.resolve(startPos))
```
YousefED, line 262: why don't we need the offset logic ( + 4 or + 1) for textselection here, but we do need it in setTextCursorPosition?

packages/core/src/extensions/SlashMenu/defaultSlashMenuItems.ts
```diff
@@ -1,13 +1,34 @@
 import { BlockNoteEditor } from "../../BlockNoteEditor";
-import { BlockSchema, PartialBlock } from "../Blocks/api/blockTypes";
+import { Block, BlockSchema, PartialBlock } from "../Blocks/api/blockTypes";
 import { defaultBlockSchema } from "../Blocks/api/defaultBlocks";
 import { imageToolbarPluginKey } from "../ImageToolbar/ImageToolbarPlugin";
 import { BaseSlashMenuItem } from "./BaseSlashMenuItem";
 
+function setSelectionToNextContentEditableBlock<BSchema extends BlockSchema>(
```
YousefED, line 7: can you add comments to explain the changes to this file? Seems like some significant logic has been changed and it's unclear to me why

packages/core/src/extensions/Blocks/nodes/BlockContainer.ts
matthewlipski: It does actually throw a warning with +1, i guess PM just shifts it to the nearest inline content but it should also be +4

packages/core/src/extensions/SlashMenu/defaultSlashMenuItems.ts
matthewlipski: Our logic when inserting/selecting an "add image button" works differently to Notion, mainly because Notion blurs the editor while we don't. So the added code here makes the text cursor go to the next block with editable content instead of selecting the button, since selecting the button doesn't open the image replacement toolbar anyway. Do you think we should change the behaviour to match Notion (probably in a different PR)?

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -8,36 +8,62 @@ import {
...
-  referencePosLeft: { top: number; left: number };
+  referencePosCell: DOMRect;
+  referencePosTable: DOMRect;
+
+  block: Block<(typeof Table)["config"]>;
   colIndex: number;
   rowIndex: number;
-  block: Block<(typeof Table)["config"]>;
+
+  isDragging:
```
YousefED, line 24: rename to `dragInfo` or `draggingState`?

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -83,56 +105,142 @@ export class TableHandlesView {
...
+      this.state.isDragging.draggedCellOrientation === "row"
+        ? boundedMouseCoords.top
+        : boundedMouseCoords.left;
+
+    // Gets the ProseMirror position corresponding to the projected mouse
+    // coordinates.
+    let proseMirrorPos = this.pmView.posAtCoords(boundedMouseCoords)!.pos;
+    let resolvedPos = this.pmView.state.doc.resolve(proseMirrorPos);
+
+    // Gets the ProseMirror node type at `proseMirrorPos`. This node will always
```
YousefED, line 191: let's assert this and throw an error if it's not the case?

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -236,4 +458,86 @@ export class TableHandlesProsemirrorPlugin<
   public onUpdate(callback: (state: TableHandlesState) => void) {
     return this.on("update", callback);
   }
+
+  colDragStart = (event: {
```
YousefED, line 462: add comments that these functions are part of the public interface / what they're used for

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -215,19 +318,138 @@ export class TableHandlesProsemirrorPlugin<
...
+                  : 0);
+              decorations.push(
+                // The widget is a small bar which spans the height of the cell.
+                Decoration.widget(decorationPos, () => {
+                  const widget = document.createElement("div");
+                  widget.className = "bn-table-drop-cursor";
+                  widget.style.top = "0";
+                  widget.style.bottom = "0";
+                  if (pluginState.newIndex > pluginState.originalIndex) {
+                    widget.style.right = "-2px";
```
YousefED, line 412: why this difference? (can you add a comment?)

PR #414, opened 2023-11-17
feat: Parsing blocks from external HTML

matthewlipski:
This PR adds a `fromExternalHTML` function that can be defined for each block type. The function provides a HTML element and a function to parse `InlineContent[]` from inside a HTML element, as attributes. The function should return a `PartialBlock` with the parsed block, or `undefined` if the element should not be parsed as that block type. Currently, it is the responsibility of the `fromExternalHTML` function to construct every aspect of the block. This means that `content` and `children` must be defined manually as the parser will not fill them. This is probably not the approach we want to go for, for reasons documented in `parseExternalHTML.ts`.


YousefED:
superseded by https://github.com/TypeCellOS/BlockNote/pull/422

PR #418, opened 2023-11-21
feat: custom styles and custom inline content

YousefED:
WIP:

This PR makes it possible to create custom styles or inline content nodes.

- Abstracted Styles and InlineContent, similar to Blocks. Refactored quite some types to accomplish this
- Refactored unit tests, we now have separated test scenarios + schemas that we can reuse across tests

**Breaking changes**
- `BlockNoteEditor.create()` instead of `new BlockNoteEditor`
- separated config / implementation for creating custom blocks / styles / inline content

**TODO**

- [x] Custom Inline Content typings collide with the built-in Link and StyledText, because Custom Inline Content nodes have a `type: string` attribute, and therefore collides with `type: "link"` and `type: "text"` when using discriminated unions. Decide on best way forward
- [ ] Decide: do we want custom styles that allow more than 1 attribute? In Prosemirror, marks can have multiple attributes, we currently don't make use of this
- [ ] switch to jsx instead of FC?
- [x] React
- [ ] Demo / test apps
- [ ] API for mentions
- [ ] "Typescript unit tests" to validate strong typing of custom editors
- [ ] Revise file structure
- [ ] Docs
- [ ] manual testing
- [ ] e2e tests
- [ ] review usage of `defaultProps.ts`


packages/react/src/FormattingToolbar/components/DefaultButtons/ToggledStyleButton.tsx
```diff
@@ -9,30 +12,35 @@ import {
...
   italic: RiItalic,
   underline: RiUnderline,
   strike: RiStrikethrough,
   code: RiCodeFill,
 };
 
-export const ToggledStyleButton = <BSchema extends BlockSchema>(props: {
-  editor: BlockNoteEditor<BSchema>;
-  toggledStyle: ToggledStyle;
+export const ToggledStyleButton = <
```
matthewlipski, line 37: We probably want to make this extensible, so you can easily make buttons for custom styles.

packages/react/src/ReactInlineContentSpec.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+      root.unmount();
+
+      return {
+        dom,
+        contentDOM: contentDOMClone || undefined,
+      };
+    },
+
+    // TODO: needed?
```
matthewlipski, line 114: I think it's a nice to have but I don't think there are many (if any) use cases that would take advantage of it. The custom inline content type would either have to reference its parent block or attach listeners to elements other than itself. I guess the only concern is that a `ReactNodeViewRenderer` is how TipTap recommends rendering a React component for your node, so might have to look inside their code to see if there's anything we're missing.

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -160,26 +187,22 @@ export function tableContentToNodes(
   return rowNodes;
 }
 
-/**
- * Converts a BlockNote block to a TipTap node.
- */
-export function blockToNode<BSchema extends BlockSchema>(
-  block: PartialBlock<BSchema>,
-  schema: Schema
+function blockOrInlineContentToContentNode(
```
matthewlipski, line 190: Why does this function also accept inline content? AFAIK it's only used once and it's with a block

packages/react/vite.config.ts
```diff
@@ -5,12 +5,21 @@ import pkg from "./package.json";
...
 
 // https://vitejs.dev/config/
-export default defineConfig({
+export default defineConfig((conf) => ({
   test: {
     environment: "jsdom",
     setupFiles: ["./vitestSetup.ts"],
   },
   plugins: [react()],
+  resolve: {
```
matthewlipski, line 14: Could you add a comment what this is for?

packages/core/src/BlockNoteEditor.ts
```diff
@@ -629,32 +783,27 @@ export class BlockNoteEditor<BSchema extends BlockSchema = DefaultBlockSchema> {
...
-    ]);
-    const colorStyles = new Set<ColorStyle>(["textColor", "backgroundColor"]);
-
     for (const mark of marks) {
-      if (toggleStyles.has(mark.type.name as ToggledStyle)) {
-        styles[mark.type.name as ToggledStyle] = true;
-      } else if (colorStyles.has(mark.type.name as ColorStyle)) {
-        styles[mark.type.name as ColorStyle] = mark.attrs.color;
+      const config = this.styleSchema[mark.type.name];
+      if (!config) {
```
matthewlipski, line 800: Shouldn't this be a warning/error?

matthewlipski (commented):
Nice! Lots of changes though😅 - still need to dive deeper into the new testing setup and changes on `nodeConversions.ts`

packages/core/src/BlockNoteEditor.ts
```diff
@@ -145,54 +195,111 @@ const blockNoteTipTapOptions = {
...
+  >;
+  public readonly tableHandles:
+    | TableHandlesProsemirrorPlugin<any, ISchema, SSchema>
+    | undefined;
 
   public readonly uploadFile: ((file: File) => Promise<string>) | undefined;
 
-  constructor(
-    private readonly options: Partial<BlockNoteEditorOptions<BSchema>> = {}
+  public static create<
```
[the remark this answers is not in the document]
YousefED: I couldn't get the types working with a regular constructor. I'm doing an `as` "cast" in the `create` function, afaik that's not possible with regular constructors

packages/core/src/extensions/Blocks/api/blocks/types.ts
```diff
@@ -77,130 +92,186 @@ export type TiptapBlockImplementation<T extends BlockConfig> = {
...
 export type BlockSchemaWithBlock<
   BType extends string,
   C extends BlockConfig
 > = {
-  [k in BType]: BlockSpec<C>;
+  [k in BType]: C;
 };
 
-export type TableContent = {
+export type TableContent<
```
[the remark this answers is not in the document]
YousefED: hmm, I think not, because this is not a type of "inline content". it's a type of "block content", which for most blocks will be "inline content" / none, except for tables, where it is "table content"

packages/core/src/extensions/Blocks/api/blocks/createSpec.ts
```diff
@@ -1,31 +1,37 @@
...
   parse,
   propsToAttributes,
   wrapInBlockStructure,
-} from "./block";
-import { Block, BlockConfig, BlockSchemaWithBlock } from "./blockTypes";
+} from "./internal";
+import { BlockConfig, BlockFromConfig, BlockSchemaWithBlock } from "./types";
 
 // restrict content to "inline" and "none" only
 export type CustomBlockConfig = BlockConfig & {
```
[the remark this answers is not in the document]
YousefED: I think it comes down to the same thing, as the `intersection` operator `&` will only match overlapping types

packages/core/src/extensions/Blocks/api/blocks/types.ts
```diff
@@ -77,130 +92,186 @@ export type TiptapBlockImplementation<T extends BlockConfig> = {
...
+  ? Blocks
+  : never;
 
 // Defines multiple block specs. Also ensures that the key of each block schema
 // is the same as name of the TipTap node in it. This should be passed in the
 // `blocks` option of the BlockNoteEditor. From a block schema, we can derive
 // both the blocks' internal implementation (as TipTap nodes) and the type
 // information for the external API.
-export type BlockSchema = NamesMatch<Record<string, BlockSpec<any>>>;
+export type BlockSchema = NamesMatch<Record<string, BlockConfig>>;
```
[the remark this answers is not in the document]
YousefED: done

packages/core/src/extensions/Blocks/api/defaultBlocks.ts
```diff
@@ -1,18 +1,75 @@
...
   paragraph: Paragraph,
   heading: Heading,
   bulletListItem: BulletListItem,
   numberedListItem: NumberedListItem,
   image: Image,
   table: Table,
-} satisfies BlockSchema;
+} satisfies BlockSpecs;
+
+export function getBlockSchemaFromSpecs<T extends BlockSpecs>(specs: T) {
```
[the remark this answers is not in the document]
YousefED: done

packages/core/src/extensions/Blocks/api/inlineContent/types.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+// Container for both the config and implementation of a block,
+// and the type of BlockImplementation is based on that of the config
+export type InlineContentSpec<T extends InlineContentConfig> = {
+  config: T;
+  implementation: InlineContentImplementation<T>;
+};
+
+export type InlineContentSchema = Record<string, InlineContentConfig>;
```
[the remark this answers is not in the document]
YousefED: added comments

packages/core/src/api/nodeConversions/nodeConversions.ts
```diff
@@ -2,47 +2,51 @@ import { Mark } from "@tiptap/pm/model";
...
-  "code",
-]);
-const colorStyles = new Set<ColorStyle>(["textColor", "backgroundColor"]);
-
 /**
  * Convert a StyledText inline element to a
  * prosemirror text node with the appropriate marks
  */
-function styledTextToNodes(styledText: StyledText, schema: Schema): Node[] {
+function styledTextToNodes<T extends StyleSchema>(
```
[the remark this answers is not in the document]
YousefED: no, because we also need the style.propschema right?

packages/core/src/BlockNoteEditor.ts
YousefED: adding a warn

packages/react/vite.config.ts
YousefED: Removed, I don't think it's necessary here

packages/core/src/api/nodeConversions/nodeConversions.ts
YousefED: we also call it in `blockToNode`

packages/core/src/extensions/SlashMenu/defaultSlashMenuItems.ts
```diff
@@ -54,52 +65,56 @@ function insertOrUpdateBlock<BSchema extends BlockSchema>(
...
+  BSchema extends BlockSchema,
+  I extends InlineContentSchema,
+  S extends StyleSchema
+>(
   // This type casting is weird, but it's the best way of doing it, as it allows
   // the schema type to be automatically inferred if it is defined, or be
   // inferred as any if it is not defined. I don't think it's possible to make it
   // infer to DefaultBlockSchema if it is not defined.
-  schema: BSchema = defaultBlockSchema as unknown as BSchema
+  schema: BSchema
```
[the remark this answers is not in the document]
YousefED: agree

packages/react/src/FormattingToolbar/components/DefaultButtons/ToggledStyleButton.tsx
YousefED: Yes, can you give this a go? I think we mainly want to move `shortcuts` and `icons` to DefaultToolbar

packages/react/vite.config.ts
YousefED: added again, was needed for vitest

Issue #366, opened 2023-10-05
Error: Could not find block in the editor with matching ID with Collaboration enabled


ursnj:
**Describe the bug**
When using editor with collaboration replacingBlocks() is throwing this error, but same code is working fine with out collaboration

const editor: any = useBlockNote({
    editable: true,
    theme: theme === THEMES.DARK ? THEMES.DARK : THEMES.LIGHT,
    collaboration: {
      provider,
      fragment: doc.getXmlFragment('document-store'),
      user: {
        name: myName,
        color: 'red'
      }
    }
  } as any);
  
  useEffect(() => {
    if (editor && !isPageLoading) {
      const getBlocks = async () => {
        const initContent = pageData?.uuid ? pageData.details : defaultPage.details;
        const blocks: any = await editor.HTMLToBlocks(initContent);
        editor.replaceBlocks(editor.topLevelBlocks, blocks); // This is causing issue
      };
      getBlocks();
    }
  }, [!!editor, !!pageData]);

Uncaught (in promise) Error: Could not find block in the editor with matching ID.
    at Z (nodeUtil.ts:31:11)
    at xt (blockManipulation.ts:27:35)
    at Qe (blockManipulation.ts:124:3)
    at Qo.replaceBlocks (BlockNoteEditor.ts:576:19)
    at getBlocks (PagesNew.tsx:78:16)
Z @ nodeUtil.ts:31
xt @ blockManipulation.ts:27
Qe @ blockManipulation.ts:124
replaceBlocks @ BlockNoteEditor.ts:576
getBlocks @ PagesNew.tsx:78
await in getBlocks (async)
(anonymous) @ PagesNew.tsx:80
commitHookEffectListMount @ react-dom.development.js:23150
commitPassiveMountOnFiber @ react-dom.development.js:24926
commitPassiveMountEffects_complete @ react-dom.development.js:24891
commitPassiveMountEffects_begin @ react-dom.development.js:24878
commitPassiveMountEffects @ react-dom.development.js:24866
flushPassiveEffectsImpl @ react-dom.development.js:27039
flushPassiveEffects @ react-dom.development.js:26984
performSyncWorkOnRoot @ react-dom.development.js:26076
flushSyncCallbacks @ react-dom.development.js:12042
commitRootImpl @ react-dom.development.js:26959
commitRoot @ react-dom.development.js:26682
finishConcurrentRender @ react-dom.development.js:25981
[...]

PR #419, opened 2023-11-22
fix: block id null when using collaboration

YousefED:
fixes https://github.com/TypeCellOS/BlockNote/issues/366

and `Uncaught TypeError: Cannot read properties of null (reading 'id')` when hovering over an empty block when using collaboration (reported via discord)

yjs calls `createAndFill` here: https://github.com/yjs/y-prosemirror/blob/e0e5e951614abe1be2295e5ab8987ab5916bcaec/src/plugins/sync-plugin.js#L194

### DIFF
--- a/examples/editor/src/main.tsx
+++ b/examples/editor/src/main.tsx
@@ -6,7 +6,8 @@ window.React = React;
 
 const root = createRoot(document.getElementById("root")!);
 root.render(
-  <React.StrictMode>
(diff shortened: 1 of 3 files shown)


ConnorDoyle:
Thanks for the fix, do you plan to cut a release with this patch soon?

YousefED:
Yes, either tomorrow or early next week

PR #422, opened 2023-11-27
feat: HTML paste handling

YousefED:
Add tests and fix pasting of nested lists and custom content from external editors


packages/react/vite.config.ts
```diff
@@ -11,6 +11,16 @@ export default defineConfig((conf) => ({
     setupFiles: ["./vitestSetup.ts"],
   },
   plugins: [react()],
+  // used so that vitest resolves the core package from the sources instead of the built version
```
matthewlipski, line 14: Is this necessary here? In PR #418 I saw you removed it, not sure if it's the same case here

packages/core/src/extensions/Blocks/api/blocks/createSpec.ts
```diff
@@ -50,8 +55,60 @@ export type CustomBlockImplementation<
...
+
+        if (block === undefined) {
+          return false;
+        }
+
+        return block.props || {};
+      },
+    });
+  }
+  //     getContent(node, schema) {
```
matthewlipski, line 93: I'm kind of wondering if this use case and API makes sense. I'm not really sure if I can see any scenarios where someone would want to override the default inline content parsing, especially since if you can provide a parse function for custom inline content.

matthewlipski (commented):
Nice! One question though - apart from headings, does this change the external parsing logic, or is it more just for refactoring?

packages/react/vite.config.ts
YousefED: yes, I readded it with the comment

packages/core/src/api/parsers/markdown/parseMarkdown.ts
```diff
@@ -0,0 +1,80 @@
...
+    .use(remarkParse)
+    .use(remarkGfm)
+    .use(remarkRehype, {
+      handlers: {
+        ...(defaultHandlers as any),
+        code,
+      },
+    })
+    .use(rehypeStringify)
+    .process(markdown);
```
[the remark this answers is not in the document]
YousefED: Let's do it, but keep the top level one on BlockNoteEditor async

packages/core/src/api/exporters/markdown/markdownExporter.ts
```diff
@@ -0,0 +1,43 @@
...
+    .use(rehypeRemark)
+    .use(remarkGfm)
+    .use(remarkStringify)
+    .processSync(cleanHTMLString);
+
+  return markdownString.value as string;
+}
+
+// TODO: add tests
+export async function blocksToMarkdown<
```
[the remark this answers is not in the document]
YousefED: ok to change

packages/core/src/extensions/Blocks/api/blocks/createSpec.ts
YousefED: Ok, let's not add it if we don't have a use-case yet

PR #425, opened 2023-11-28
Custom inline content and styles commands/copy & paste fixes

matthewlipski:


YousefED:
I accidentally didn't have all tests enabled. I did that now and fixed 1. There's an error with the "tag" tests. Although we agreed that use-case isn't supported, maybe it's an easy fix to make `addInlineContentAttributes` more similar to `wrapInBlockStructure`. 

I'll be merging some PRs because there's to much in flux, but keeping TODOs:

- [ ] review my changes
- [ ] decide on classnames (see comments)
- [ ] Fix tag tests (see above)
- [ ] Maybe add a similar tests (and / or playground) for a Custom Block "static" and "editable" content

Issue #267, opened 2023-07-07
Custom `InlineContent` types


matthewlipski:
Alongside custom block types, custom `InlineContent` types would allow for more extensibility within BlockNote. This would allow consumers to implement features like comments and spellcheck.

It would also be worth considering custom marks/styles while working on this issue.

man-shar:
Is this something that's being worked on actively already? If not, I'd love to try and implement the custom marks support.

matthewlipski:
We're currently working on fixing the clipboard & HTML/Markdown conversion, so go ahead! I would start by taking a look at the custom blocks API (`block.ts` for vanilla, `ReactBlockSpec.ts` for React) and doing something similar for marks/inline nodes, i.e. converting to the TipTap API, storing a schema, typing, etc.

man-shar:
Great, thanks! Last week, I did end up implementing this as a new Tiptap node like so:

```
import { Node, mergeAttributes } from "@tiptap/core";
import { ReactNodeViewRenderer } from "@tiptap/react";
import CustomInlineNodeView from "./CustomInlineNodeView";

export const CustomInlineNode = Node.create({
  name: "custom-inline-node",
  group: "inline",
  inline: true,
  selectable: false,
  atom: true,
  draggable: true,
  addAttributes() {
    return {
      
    };
  },
  parseHTML() {
    return [
      {
        tag: "custom-inline-node",
      },
    ];
  },
  renderHTML({ HTMLAttributes }) {
    return ["custom-inline-node", mergeAttributes(HTMLAttributes)];
  },
  addNodeView() {
    return ReactNodeViewRenderer(CustomInlineNodeView);
  },
});
```

```
const editor = useBlockNote({
    ...,
    _tiptapOptions: {
      extensions: [ReactiveVariableNode],
    },
  });
```

It seems to work fine. Though my use case was dragging and dropping these nodes from somewhere else into the text. I'm wondering if it's worth adding this as an example somewhere for the time being?

matthewlipski:
Oh awesome! I guess there are still probably some quirks when converting the node to `InlineContent[]` or converting to HTML, but good to see that it's not too much hassle to get it working. I think for now it's ok if ppl just get linked to this issue if they ask about custom inline nodes in discord, and make a documentation page once we can make custom inline content types using the BlockNote API rather than using TipTap's.

Issue #194, opened 2023-05-18
Tables support ?


AhmedRyad22:
hi everyone, by any chance can we have tables in the editor ? or is there any way to add it as a block currently ?

thanks <3

YousefED:
Hi @AhmedRyad22 ! This is on our roadmap, but we'll first focus on Custom Blocks and Images

hsingh-ds:
@matthewlipski do you have any timeline for the table block? Because we need it asap, so I was thinking to create a custom block for table. Any thoughts?

matthewlipski:
> @matthewlipski do you have any timeline for the table block? Because we need it asap, so I was thinking to create a custom block for table. Any thoughts?

We're currently working on fixing/improving HTML and Markdown conversions so table blocks are unfortunately gonna be a while still. Also you may have trouble implementing a table as a custom block as each BlockNote block has a single corresponding ProseMirror node, meaning it can only have a single piece of editable content.

metaYogeshBhutkar:
@YousefED @matthewlipski by when can we expect the tables feature to roll out? The project that we are working on requires the use of Tables right now, I thought of implementing the tables feature all by myself using custom components, but that won't be feasible I think. It's just not possible to do it that way.

If you can let us know when are you going to merge the above PR #426, I can plan my proceedings accordingly.

PR #381, opened 2023-10-15
Fix for #361 Converting image block to md

jmas:
Related issue https://github.com/TypeCellOS/BlockNote/issues/361

PR #426, opened 2023-11-29
feature: v0.10: tables, custom inline content, custom styles, better copy / paste handling

### LINKED ISSUE #221: Custom blocks: copy & paste, markdown, html
We want to write our own serializer to handle clipboard copy / paste. Possibly, we can also re-use this for the export to markdown / html functions

YousefED:
This PR is a major overhaul of BlockNote and brings in several new features.

The following PRs are part of this:
- Tables: #413 
- Better copy / paste handling: #257 #422 #425 
- Support for custom inline content, custom styles, and related refactors: #412 #418 #424

<small>(note that these PRs have been merged into each other, so the code change view will not be ideal)</small>

# Features

Some nice new stuff now possible with BlockNote!

## Tables
ezgif-2-fd6034a112

Closes https://github.com/TypeCellOS/BlockNote/issues/194

## Custom Inline Content and Custom Styles
Until now, it was only possible to create your own custom Blocks. Now, you can also create custom Styles and Inline Content. Styles are markup that can be enabled on text (think bold, italic, etc.). Custom Inline Content can be used for "inline blocks", for example for Mentions.

Closes https://github.com/TypeCellOS/BlockNote/issues/267

## Better copy paste handling

The copy / paste and Blocks to / from Markdown / HTML system has been overhauled and should be more robust. Copying and pasting of custom blocks is now also supported!

Closes #221 #69 #361 #381 #282 #226 #249

# Breaking changes:

### Custom Block API
If you're using [Custom Blocks](https://www.blocknotejs.org/docs/block-types#creating-a-custom-block-type) (`createReactBlockSpec`), there are two major changes:
- The "schema" of your block and the "implementation" are now separated
- The `<InlineContent />` element has been removed in favor of `contentRef`
- Use `content: "inline" | "none"` instead of `containsInlineContent: boolean`

Old (deprecated) code example:

```typescript
const ImageBlock = createReactBlockSpec({
  type: "image",
  propSchema: {
    src: {
      default: "https://via.placeholder.com/1000",
    },
  },
  containsInlineContent: true,
  render: ({ block }) => (
    <div
      style={{
        display: "flex",
        flexDirection: "column",
      }}>
      <img
        style={{
          width: "100%",
        }}
        src={block.props.src}
        alt={"Image"}
        contentEditable={false}
      />
      <InlineContent />
    </div>
  ),
});
```

New version:

```typescript
const ImageBlock = createReactBlockSpec({
  type: "image",
  propSchema: {
    src: {
      default: "https://via.placeholder.com/1000",
    },
  },
  content: "inline"
}, {
  render: ({ block, contentRef }) => (
    <div
      style={{
        display: "flex",
        flexDirection: "column",
      }}>
      <img
        style={{
          width: "100%",
        }}
        src={block.props.src}
        alt={"Image"}
        contentEditable={false}
      />
      <span ref={contentRef} />
    </div>
  ),
});
```

### Naming of export / parse functions
To better indicate the conversion to / from Markdown and HTML is lossy, the following functions have been renamed: 

- `editor.blocksToHTML` -> `editor.blocksToHTMLLossy`
- `editor.HTMLToBlocks` -> `editor.tryParseHTMLToBlocks`
- `editor.blocksToMarkdown` -> `editor.blocksToMarkdownLossy`
- `editor.markdownToBlocks` -> `editor.tryParseMarkdownToBlocks`

### Vanilla (non-react) constructor usage

Instead of `new BlockNoteEditor(...)`, now use `BlockNoteEditor.create(...)`

# TODO (before merging)

- [x] @YousefED : review Inline Content selection
- [x] ~~@YousefED : Disable Inline Content elements with content~~ -> @matthewlipski fix backspace behavior
- [x] Update APIs that have been changed in docs
- [x] Document breaking changes in this PR
- [x] @YousefED : Readd tests for markdown conversions
- [x] @matthewlipski, review #422 (code + testcases), #418 and #425 (+ address todos)
- [x] @matthewlipski Add more tests + playground examples of custom blocks. We only now have very basic custom blocks, but for example, blocks with both "static" and editable content are missing
- [x] Review this PR
- [x] Good moment to discuss possible naming changes
- [x] @matthewlipski  fix hardbreaks
- [x] @YousefED review slash menu item typings
- [x] Fix e2e tests / CI
- [x] @matthewlipski Add vanilla styles and non-editable custom block to playground
- [x] Add table GIF to this PR

# TODO (separately, can be done after merging)

Docs:
- [ ] Maybe it makes sense to revisit docs + examples + playground completely, but at least:
- [ ] Add custom styles and inline content to docs
- [x] Add Table to docs
- [x] Add serialization related information to docs
- [ ] Figure out what to do with theming now that the CSS classes are accessible

Tests: 
- [ ] Add test for [this Commit](https://github.com/TypeCellOS/BlockNote/pull/257/commits/755a5744a71c71ea9ae83991cad281aacafb0ab0)
- [ ] Add more tests for tables (e2e + unit)
- [ ] @YousefED : Add "typescript tests" that enforce strict typing of schemas
- [ ] Add tests for creating custom blocks / IC / Styles that can parse "external HTML"
- [x] Add copy / paste tests with examples from copying from gdocs / notion / other websites
- [ ] Add playwright test for custom blocks / ic / styles? (and keyhandling)

Other:
- [ ] Create an API for mentions (similar to slash menu)
- [ ] lowprio: make sure "external" HTML is sanitized from unnecessary classes & attributes?
- [ ] Now we've deprecated CSS Modules, we might want to document the possibility of styling via CSS directly


matthewlipski (commented):
Looks good! Reviewed changes from #422

packages/core/src/api/parsers/html/__snapshots__/paste/parse-deep-nested-content.json
```diff
@@ -0,0 +1,240 @@
...
+    "type": "paragraph",
+    "props": {
+      "textColor": "default",
+      "backgroundColor": "default",
+      "textAlignment": "left"
+    },
+    "content": [
+      {
+        "type": "text",
+        "text": " Outer 4 Div After Outer 3 Div After Outer 2 Div After Outer 1 Div After",
```
matthewlipski, line 234: Is this correct? Think the "Outer n div after"s should be in seperate blocks

packages/react/src/ReactRenderUtil.ts
```diff
@@ -20,13 +20,12 @@ export function renderToDOMSpec(
   }
 
   // clone so we can unmount the react root
-  contentDOM?.setAttribute("data-tmp-find", "true");
+  contentDOM?.setAttribute("data-editable", "");
```
YousefED, line 23: Although this change is possible, maybe it's nicer to keep concerns separate and keep the old implementation?

The responsibility for data-editable is then completely somewhere else, instead of in two locations. Unless removing it here and reverting to data-tmp-find breaks anything?

packages/core/src/api/exporters/markdown/__snapshots__/formatConversions.test.ts.snap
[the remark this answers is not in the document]
YousefED: removed in #428

YousefED:
> review Inline Content selection

Clicking a react mention results in a "hidden cursor" and selection stays the same. This is different from e.g.: https://embed.tiptap.dev/src/nodes/mention/react/

The reason the selection is different from the tiptap example is that we're using a NodeView for React inline content, and TipTap's implementation of stopEvent and ignoreMutation blocks the selection changes / click propagation. It's fixable, but atm I don't think it's super blocking.

Might be better to fix if at some point we move away from tiptap, or if a critical reason comes up
(note to self): I tested it by comparing a "vanilla" inline content mention with the react version. The vanilla version works ok, the react doesn't (because the node view blocks some things)

packages/core/src/api/parsers/html/__snapshots__/paste/parse-deep-nested-content.json
YousefED: Agree, maybe file a separate issue for this?

PR #427, opened 2023-11-29
move and update playwright tests

YousefED:
- moving pw test code to a separate "project" so that it will be typechecked and linted
- updating tests to work with refactor


YousefED:
Build works, but:

- [ ] need to fix broken tests
- [ ] I think the typings for slash menu items is not ideal. Need to make sure this is not a regression before shipping

PR #429, opened 2023-11-29
Fixed parsing error when `dom` is same as `contentDOM`

matthewlipski:
This fixes a parse rule for custom inline content and styles which threw an error if the rendered `dom` was the same element as the `contentDOM`, as it was looking for the `[data-editable]`/`contentDOM` in the children of the `dom`.


YousefED:
sweet!

PR #436, opened 2023-12-01
Playwright refactor fixes

matthewlipski:
This PR contains some fairly basic changes which fix most of the failing tests after the playwright refactor.

TODO:

- [x] Update slash menu screenshots due to table item being added
- [x] Update image copy/paste snapshot as props which are set as attributes on `blockContainer` nodes are no longer redundantly added to custom `blockContent` nodes
- [x] Fix image upload file not being found
- [x] Fix unit test cases with empty snapshots failing - I can't reproduce it locally, since the snapshot is empty maybe it's an issue with Linux vs macOS line endings or smth?
- [x] Make sure that `playwright-report` is generated as an artifact in the CI


YousefED:
Was able to reproduce with `CI=true npm run test`. It's a bug with empty snapshots, made a workaround for now

Issue #437, opened 2023-12-01
Slash / Command is not working in android chrome browser.


Kashifwahaj:
**Describe the bug**
In Android mobile devices chrome browser/command does not show dropdown, like on the web, and iOS, and it's working fine.
I also have tried in real web app as well as in the official website docs/quickstart sandbox.

**To Reproduce**
I am attaching a video, you check 
https://github.com/TypeCellOS/BlockNote/assets/30648242/9c3a84a6-ba86-4bef-8743-2a44d927f933
 https://www.blocknotejs.org/docs/quickstart,

**Misc**

- Browser: Chrome Android

PR #443, opened 2023-12-02
fix: Slash Command Android Chrome Issue  #437 

Kashifwahaj:
Hi,
I investigated the issue, The `handleKeyDown` event is not returning properly updated objects on the Android Chrome browser, so I think we need to use the `handleTextInput` method to fix this.

Thanks


matthewlipski:
Nice find! Do you think you could just extract the code into a separate function? Since seems like it's basically the same as for the `handleKeyDown` handler, so would be great if we could use the same code for both

Kashifwahaj:
> Nice find! Do you think you could just extract the code into a separate function? Since seems like it's basically the same as for the `handleKeyDown` handler, so would be great if we could use the same code for both

Thanks for the suggestion, I updated and pushed the code.

Kashifwahaj:
@matthewlipski is there anything else you suggest for this PR ?

Kashifwahaj:
@matthewlipski ?

Kashifwahaj:
@matthewlipski @YousefED Please can we proceed with this PR, let me know if any improvement or changes required

matthewlipski (approved):

YousefED:
@matthewlipski did you check if we still need both handlers?

PR #458, opened 2023-12-06
feat: Return values for block manipulation methods

### LINKED ISSUE #410: Return the data (or id atleast) of the new block created upon using `insertBlocks`
i have this use-case where i have the base64 of an image, and wanna insert it (say on click of a button).
But uploading to s3, then adding to the note doesn't seem ideal, cause of the delay(poor ux).

So i was thinking of creating a custom component, which inserts the base64 at first, then updates the url of the image once the file upload is done. 

Return either just the id or the whole data of the new block created on using `insertBlocks`.

Uploading to storage, then adding to the editor seems to be the best alternative option, but like i said thats bad ux.

**Additional context**
None

matthewlipski:
This PR adds return values for the following `BlockNoteEditor` methods:
```
// Returns the blocks after being inserted
insertBlocks(...): Block[]

// Returns the updated block
updateBlock(...): Block

// Returns the blocks that were removed
removeBlocks(...): Block[]

// Returns both the blocks that were removed and the blocks after being inserted
replaceBlocks(...): {
  insertedBlocks: Block[];
  removedBlocks: Block[];
}
```

Closes #410

### DIFF
--- a/packages/core/src/api/blockManipulation/blockManipulation.ts
+++ b/packages/core/src/api/blockManipulation/blockManipulation.ts
@@ -130,7 +130,6 @@ function removeBlocksWithCallback<
   ) => number
 ): Block<BSchema, I, S>[] {
   const ttEditor = editor._tiptapEditor;
-
   const tr = ttEditor.state.tr;
 
   const idsOfBlocksToRemove = new Set<string>(
@@ -155,6 +154,7 @@ function removeBlocksWithCallback<
       return true;
     }
 
+    // Saves the block that is being deleted.
     removedBlocks.push(
       nodeToBlock(
         node,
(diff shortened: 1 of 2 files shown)


sahilpohare:
Could we have this merged?

packages/core/src/api/blockManipulation/blockManipulation.ts
```diff
@@ -60,13 +66,31 @@ export function insertBlocks<
...
 
-      return;
+      insertionPos++;
+    } else {
+      insertionPos = posBeforeNode + node.firstChild!.nodeSize + 2;
     }
+  }
 
-    insertionPos = posBeforeNode + node.firstChild!.nodeSize + 2;
+  const insertedBlocks: Block<BSchema, I, S>[] = [];
```
YousefED, line 75: why can't we just return `nodesToInsert`? I have some guesses (i.e.: ids), but from a code readability / maintainability perspective it seems weird ("future-us" would ask: "why do we query the nodes again"?

So I think it's smart to explain this in comments

matthewlipski: Ah this is just an oversight from my part - forgot that nodes are assigned IDs before being added to the editor, so this is not needed.

YousefED (approved):

PR #479, opened 2023-12-15
chore: Mantine update

### LINKED ISSUE #383: Update Mantine & TipTap
Hi team.

Is there any chance you'd have time to look into updating Mantine and TipTap? There various interesting fixes upstream.

matthewlipski:
This PR updates Mantine to the newest version. The way themes are defined and applied has changed drastically since v7, and now uses CSS variables to store theming data instead of emotion. Currently, the API used to provide a theme is the same, but the theme is now converted into CSS variables instead of being passed as a `MantineThemeOverride`. Some styling may be slightly different after moving to CSS, but these changes are small and should be easily fixable.

NOTE: This PR breaks styling for tooltips, as they are rendered outside the `.bn-container` element due to the current implementation, which uses Tippy. #501 fixes the issue and converts them to Mantine Tooltip components, causing the tooltips to be rendered as a sibling of the target element. So recommend merging that PR before this one.

closes #383


packages/react/src/editor/styles.css
```diff
@@ -0,0 +1,414 @@
...
+    background-color: var(--bn-color-menu-background);
+    border: var(--bn-border);
+    border-radius: var(--bn-border-radius-medium);
+    box-shadow: var(--bn-shadow-medium);
+    flex-wrap: nowrap;
+    gap: 2px;
+    padding: 2px;
+    width: fit-content;
+
+    & .mantine-Button-root, .mantine-ActionIcon-root {
```
YousefED, line 245: is this nesting valid in css or only in scss?

packages/react/src/editor/BlockNoteTheme.ts
```diff
@@ -29,467 +26,181 @@ export type ColorScheme = {
...
   borderRadius: number;
   fontFamily: string;
-  componentStyles?: (theme: Theme) => ComponentStyles;
 };
 
-export const blockNoteToMantineTheme = (theme: Theme): MantineThemeOverride => {
-  const shadow = `0 4px 12px ${theme.colors.shadow}`;
-  const lightShadow = `0 2px 6px ${theme.colors.border}`;
-  const border = `1px solid ${theme.colors.border}`;
+export const themeToCSSVariables = (theme: Theme) => {
```
YousefED, line 35: should we use this manual code, or reuse sth like https://mantine.dev/styles/css-variables/#css-variables-resolver?

matthewlipski: I think it's better to do it manually in this case, since the theme we use is different to the Mantine theme that the `CSSVariablesResolver` uses. I think either way the amount of code is roughly the same, but if we use the `CSSVariablesResolver`, more of it becomes coupled to Mantine so it would be more of a pain if we were to switch UI libraries.

packages/react/src/editor/styles.css
matthewlipski: I think it's a fairly recent change, but it's valid for vanilla CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector

Though I noticed it's causing issues in docker, so will probably have to change it either way

YousefED: interesting, didn't know. But support is too limited for now indeed

packages/react/src/editor/BlockNoteTheme.ts
YousefED: makes sense. Should we also add code to remove the variables?

matthewlipski: Hmm not sure how useful that would be, or do you mean like reverting them to defaults?

packages/react/src/components/TableHandles/TableHandleMenu/TableHandleMenu.tsx
```diff
@@ -20,14 +20,10 @@ export type TableHandleMenuProps<
...
-  return (
-    <Menu.Dropdown className={classes.root} style={{ overflow: "visible" }}>
-      {props.children}
-    </Menu.Dropdown>
-  );
-};
+export const TableHandleMenu = (props: { children: ReactNode }) => (
+  <Menu.Dropdown
+    className={"bn-table-handle-menu"}
+    style={{ overflow: "visible" }}>
```
YousefED, line 26: maybe extract all styles to consts

YousefED: or to css (preferable)

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -2,21 +2,23 @@ import {
...
-import { Theme, blockNoteToMantineTheme } from "./BlockNoteTheme";
+import { Theme, themeToCSSVariables } from "./BlockNoteTheme";
 import { FormattingToolbarPositioner } from "../components/FormattingToolbar/FormattingToolbarPositioner";
 import { HyperlinkToolbarPositioner } from "../components/HyperlinkToolbar/HyperlinkToolbarPositioner";
 import { ImageToolbarPositioner } from "../components/ImageToolbar/ImageToolbarPositioner";
 import { SideMenuPositioner } from "../components/SideMenu/SideMenuPositioner";
 import { SlashMenuPositioner } from "../components/SlashMenu/SlashMenuPositioner";
 import { TableHandlesPositioner } from "../components/TableHandles/TableHandlePositioner";
 import { darkDefaultTheme, lightDefaultTheme } from "./defaultThemes";
+import "@mantine/core/styles.css";
```
YousefED, line 20: remove css imports here

(similar to https://github.com/TypeCellOS/BlockNote/pull/467/files)

packages/react/src/editor/BlockNoteTheme.ts
YousefED, line 35: Todo: naming (`applyCSSVariablesFromTheme`?)

examples/editor/examples/basic/App.tsx
```diff
@@ -1,6 +1,7 @@
 import { uploadToTmpFilesDotOrg_DEV_ONLY } from "@blocknote/core";
 import "@blocknote/core/style.css";
 import { BlockNoteView, useBlockNote } from "@blocknote/react";
+import "@blocknote/react/style.css";
```
YousefED, line 4: TBD: import core from react?

packages/react/src/editor/styles.css
```diff
@@ -0,0 +1,426 @@
+/* Default theme params, overwritten by BlockNoteView */
```
YousefED, line 1: brainstorming:

- Can we have multiple editors (with different themes?)? Should we apply vars to root / should we clean them?
- Can / should  we specify a theme completely in CSS?
- Should we reuse "base" (color) variables?

fix ->
- clearer separation of layers / concerns, let's zoom out and first create pure CSS solution (without themes from code). then build on top of that

YousefED (commented):
joint review

packages/react/src/editor/styles.css
```diff
@@ -1,144 +1,186 @@
...
-    --bn-color-highlight-text-purple: #6940a5;
-    --bn-color-highlight-background-purple: #eae4f2;
-    --bn-color-highlight-text-pink: #ad1a72;
-    --bn-color-highlight-background-pink: #f4dfeb;
+@import url("@mantine/core/styles.css");
+/* TODO: Cannot resolve file? */
+@import url("@blocknote/core/style.css");
+
+/* Default theme params */
+.bn-editor-wrapper {
```
YousefED, line 6: Instead of `.bn-editor-wrapper` I think I'd prefer `.bn-editor` or `.bn-container` or sth

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -77,32 +41,55 @@ export function BlockNoteView<
...
+      return "light";
+    }
+
+    if (theme === "dark") {
+      return "dark";
     }
-  }, [preferredTheme, theme]);
+
+    if (typeof theme === "object" && "light" in theme && "dark" in theme) {
+      if ("light" in theme && "dark" in theme) {
```
YousefED, line 58: this if will always be true, so I don't think this code works

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -77,32 +41,55 @@ export function BlockNoteView<
...
+        );
+        return systemColorScheme === "dark" ? "dark" : "light";
+      }
+
+      applyCSSVariablesFromTheme(theme, editor.domElement.parentElement!);
+      return undefined;
+    }
+
+    return systemColorScheme === "dark" ? "dark" : "light";
+  }, [systemColorScheme, editor.domElement, theme]);
```
YousefED, line 71: Did you test changing themes, including from a custom theme to a built in one, etc? I doubt this works as the css variables aren't removed

YousefED (changes requested):

examples/editor/src/style.css
```diff
@@ -1,14 +1,39 @@
...
-  margin: 0;
+.mantine-AppShell-navbar {
+  background-color: #f7f7f5;
 }
 
-/* .root {
-  height: 100%;
-  width: 100%;
-} */
+/* Example theme override using CSS */
```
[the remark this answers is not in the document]
matthewlipski: This was just an example to demo overriding the theme in CSS (changes it to the red theme in the docs), so will remove.

packages/react/src/components/TableHandles/DefaultTableHandle.tsx
```diff
@@ -9,15 +9,6 @@ export const DefaultTableHandle = <
...
-      style={{
-        display: "flex",
-        alignItems: "center",
-        justifyContent: "center",
-        marginInline: "-4px",
-        overflow: "visible",
-      }}>
-      <MdDragIndicator size={24} data-test={"tableHandle"} />
-    </div>
+    <MdDragIndicator size={24} data-test={"tableHandle"} />
```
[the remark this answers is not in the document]
matthewlipski: Yes, switching to CSS made some divs in the table handles redundant so I cleaned up the code for those

packages/react/src/editor/styles.css
```diff
@@ -1,144 +1,186 @@
...
+    --bn-colors-editor-text: #3F3F3F;
```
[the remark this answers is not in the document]
matthewlipski: Wdym by this? Where else would the primary-color and primary-background-color be used?

packages/react/src/components/SideMenu/DragHandleMenu/DragHandleMenu.tsx
```diff
@@ -18,9 +18,7 @@ export type DragHandleMenuProps<
 };
 
 export const DragHandleMenu = (props: { children: ReactNode }) => (
-  <Menu.Dropdown
-    className={"bn-drag-handle-menu"}
-    style={{ overflow: "visible" }}>
+  <Menu.Dropdown className={"bn-drag-handle-menu"}>
```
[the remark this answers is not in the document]
matthewlipski: Ah, was actually missing a comma in one of the CSS rules, thx

packages/react/src/editor/BlockNoteTheme.ts
```diff
@@ -32,133 +32,165 @@ export type Theme = {
...
-  editorDOM.style.setProperty(
-    "--bn-color-disabled-text",
-    theme.colors.disabled.text
-  );
-  editorDOM.style.setProperty(
-    "--bn-color-disabled-background",
-    theme.colors.disabled.background
-  );
+  // Colors
+  for (const [colorSchemeKey, colorSchemeValue] of Object.entries(
```
YousefED, line 44: why are colors a separate case? Why can't we make this code completely generic where `{ a: { b: {c: 4 }}}` is translated to `setProperty("--bn-a-b-c", 4);`? Maybe I'm missing something?

packages/react/src/editor/styles.css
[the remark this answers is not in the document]
YousefED: it would just be a way to simplify the code (basically, a refactor that extracts variables for reused values), so for example to create the dark theme, we would only need to set primary color, background color, and not all of these, because by default they could be derived of those two:

```
--bn-colors-editor-text: #CFCFCF;
    --bn-colors-editor-background: #1F1F1F;
    --bn-colors-menu-text: #CFCFCF;
    --bn-colors-menu-background: #1F1F1F;
    --bn-colors-tooltip-text: #CFCFCF;
    --bn-colors-hovered-text: #CFCFCF;
    --bn-colors-selected-text: #CFCFCF;
```

not sure if the result will be nice

packages/react/src/editor/styles.css
```diff
@@ -53,15 +58,12 @@
...
-    --bn-colors-menu-background: #1F1F1F;
-    --bn-colors-tooltip-text: #CFCFCF;
+    --bn-colors-text: #CFCFCF;
+    --bn-colors-background: #1F1F1F;
+
     --bn-colors-tooltip-background: #161616;
-    --bn-colors-hovered-text: #CFCFCF;
     --bn-colors-hovered-background: #161616;
-    --bn-colors-selected-text: #CFCFCF;
+    --bn-colors-selected-text: var(--bn-colors-text);
```
YousefED, line 66: I suppose it's correct that this pattern is different than the light theme?

packages/react/src/editor/styles.css
```diff
@@ -4,16 +4,21 @@
 
 /* Default theme params */
 .bn-container {
-    --bn-colors-editor-text: #3F3F3F;
-    --bn-colors-editor-background: #FFFFFF;
-    --bn-colors-menu-text: #3F3F3F;
-    --bn-colors-menu-background: #FFFFFF;
-    --bn-colors-tooltip-text: #3F3F3F;
+    /* TODO: Do we want this? If yes, we should probably have a similar
```
YousefED, line 7: comment for code below, should we prefix all mantine related stiles with .bn-container? (i.e.: `.mantine-Badge-root`)

packages/react/src/editor/styles.css
matthewlipski: Kind of - at least in the default theme, the selected text has the same color as the primary text color. Though come to think of it, this is probably more like a coincidence with the particular shades that the default light theme uses. Think it's probably better to just revert it tbh, so that the light and dark theme work the same way, I think that's more important.

YousefED: ok, leave you in charge of this!

YousefED (approved):

PR #483, opened 2023-12-19
fix tippy -> floatingui

YousefED:
continue of #306 

(currently branched of playground, might not be ideal)


packages/react/src/components/ImageToolbar/ImageToolbarPositioner.tsx
```diff
@@ -32,6 +33,14 @@ export const ImageToolbarPositioner = <
 
   const referencePos = useRef<DOMRect>();
 
+  const { refs, update, context, floatingStyles } = useFloating({
+    open: show,
+    placement: "left",
```
matthewlipski, line 38: Should be `"bottom"`

packages/react/src/components/SideMenu/SideMenuPositioner.tsx
```diff
@@ -77,40 +83,26 @@ export const SideMenuPositioner = <
...
-    {
-      name: "preventOverflow",
-      options: {
-        mainAxis: false,
-        altAxis: false,
-      },
-    },
-  ],
-};
+// TODO: anything from below needed?
```
matthewlipski, line 90: This is just to ensure the side menu doesn't change position if it overflows. In Floating UI I think the way to do this is to just omit `flip()` from the middleware.

packages/react/src/components/SlashMenu/SlashMenuPositioner.tsx
```diff
@@ -33,64 +32,57 @@ export const SlashMenuPositioner = <
...
-    () => {
-      if (!referencePos.current) {
-        return undefined;
-      }
-
-      return () => referencePos.current!;
-    },
-    [referencePos.current] // eslint-disable-line
-  );
+  // const { ref, updateMaxHeight } = usePreventMenuOverflow(); // TODO, needed?
```
matthewlipski, line 54: This is indeed still needed - it's basically a fallback for handling overflow. Normally, if there's no space to render the menu in the default `"bottom"` position, it will get flipped to the top. But if the viewport itself is too small to see the whole menu, this hook constrains the height of the menu to fit in the viewport.

matthewlipski (commented):
Looks good! I think the default offsets for Floating UI are a bit smaller than Tippy, so might be worth adjusting those if we don't want to update all the screenshots. Aside from that, left a few other comments.

PR #487, opened 2023-12-21
fix: Bug in `replaceBlocks`

### LINKED ISSUE #475: InitialBlock ID is 'initialBlockId'.It will replace the initialEditorContent that contains the same id.
**Describe the bug**
Editor has a InitialBlock  and the id is "initialBlockId". If my initial data has the same block(id is 'initialBlockId'). It will be replace by empty block.

**To Reproduce**
1. Input some text in the Initial block.
2. Put the data as the initial content. 

image
image
https://codesandbox.io/p/sandbox/block-editor-alert-h5c5nk

matthewlipski:
Closes #475

### DIFF
--- a/packages/core/src/api/blockManipulation/blockManipulation.ts
+++ b/packages/core/src/api/blockManipulation/blockManipulation.ts
@@ -11,6 +11,7 @@ import {
 } from "../../schema";
 import { blockToNode } from "../nodeConversions/nodeConversions";
 import { getNodeById } from "../nodeUtil";
+import { Transaction } from "prosemirror-state";
 
 export function insertBlocks<
   BSchema extends BlockSchema,
@@ -85,10 +86,19 @@ export function updateBlock<
   editor.commands.BNUpdateBlock(posBeforeNode + 1, update);
 }
 
-export function removeBlocks(
+function removeBlocksWithCallback(
   blocksToRemove: BlockIdentifier[],
-  editor: Editor
+  editor: Editor,
+  // Should return new removedSize.
+  callback?: (
+    node: Node,
+    pos: number,
+    tr: Transaction,
+    removedSize: number
+  ) => number
 ) {
+  const tr = editor.state.tr;
+
   const idsOfBlocksToRemove = new Set<string>(
     blocksToRemove.map((block) =>
       typeof block === "string" ? block : block.id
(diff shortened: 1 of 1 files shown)


packages/core/src/api/blockManipulation/blockManipulation.ts
```diff
@@ -141,6 +141,68 @@ export function replaceBlocks<
...
+      const oldDocSize = tr.doc.nodeSize;
+      tr.insert(pos, nodesToInsert);
+      const newDocSize = tr.doc.nodeSize;
+      removedSize += oldDocSize - newDocSize;
+    }
+
+    idsOfBlocksToRemove.delete(node.attrs.id);
+
+    const oldDocSize = tr.doc.nodeSize;
+    tr.delete(pos - removedSize, pos - removedSize + node.nodeSize);
```
YousefED, line 191: is there a reason that for `removeBlocks` we use `BNDeleteBlock` instead of this?

packages/core/src/api/blockManipulation/blockManipulation.ts
```diff
@@ -141,6 +141,68 @@ export function replaceBlocks<
...
+    if (
+      node.type.name !== "blockContainer" ||
+      !idsOfBlocksToRemove.has(node.attrs.id)
+    ) {
+      return true;
+    }
+
+    if (
+      node.attrs.id ===
+      (typeof blocksToRemove[0] === "string"
```
YousefED, line 178: more readable and better reuse to just use `const idOfFirstBlock = idsOfBlocksToRemove[0]` before the call to `descendants`

packages/core/src/api/blockManipulation/blockManipulation.ts
```diff
@@ -141,6 +141,68 @@ export function replaceBlocks<
   blocksToInsert: PartialBlock<BSchema, I, S>[],
   editor: BlockNoteEditor<BSchema, I, S>
 ) {
-  insertBlocks(blocksToInsert, blocksToRemove[0], "before", editor);
-  removeBlocks(blocksToRemove, editor._tiptapEditor);
+  const ttEditor = editor._tiptapEditor;
```
YousefED, line 144: this code is now pretty similar to `removeBlocks`. Can / should we reuse the same code?

packages/core/src/api/blockManipulation/blockManipulation.ts
matthewlipski: Just so we can add the deletion and insertion operations into a single transaction. We don't have a `BNInsertBlocks` function that we can use for the insertion (would need to use `BNCreateBlock` and `BNUpdateBlock`), so it's cleaner to just use PM transactions for both instead of TipTap commands.

YousefED (approved):

Issue #488, opened 2023-12-23
Custom search function for the slash menu


sahilpohare:
**Is your feature request related to a problem? Please describe.**
I am using the editor for creating an editor for a niche. Where I need to be able to add tags, custom blocks etc. I want to use slash menu to execute commands for this. like /addgears or /xyz. Now there are about 200k items and I need a SlashMenu item in the list for each of them, to search and execute essentially the same function with a bit different arguments. I cannot store the data on the frontend. I have an api for searching. 

I looked up features where I can modify the search function (give results based on api + default blocks) 

**Describe the solution you'd like**
I would like a way to override the search blocks feature. Where I can take the query and return my own Slash menu item and execute it.

**Describe alternatives you've considered**
I tried mapping through the list in a json (dumb decision) and the app became a memory hog as it was 200k instances of the slash menu item + the execute function and no way to limit the number of results, rendereing everything at once.

YousefED:
I think this makes sense, and we need this for Mentions as well.

sahilpohare:
@YousefED I am working on this

sahilpohare:
@YousefED Need a bit of help and review on the PR.

YousefED:
See https://www.blocknotejs.org/docs/ui-components/suggestion-menus for the
new docs :)

PR #495, opened 2023-12-29
Custom Search for Slash Menu

sahilpohare:
Working on custom search function for slash menu
This is a in response to Issue #488 
impl
```ts
class CustomQueryManager extends SlashMenuQuery {
         async query(query : string, items) {}
         async execute({item, editor}) {}
}
```

```ts
<Blocknote slashMenuQueryManager={new CustomQueryManager()}/>
```

### DIFF
--- a/examples/editor/examples/basic/App.tsx
+++ b/examples/editor/examples/basic/App.tsx
@@ -1,10 +1,28 @@
-import { uploadToTmpFilesDotOrg_DEV_ONLY } from "@blocknote/core";
+import {
+  BaseSlashMenuItem,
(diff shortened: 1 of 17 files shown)


sahilpohare:
Still have some code to figure out. For example the notFoundCount is updated in the apply function but we no longer have access to the items array length as it is a promise not and I am awaiting it in the update function

sahilpohare:
This also adds the ability to load items in the slash menu dynamically

YousefED:
See https://www.blocknotejs.org/docs/ui-components/suggestion-menus for the
new docs :)

PR #501, opened 2024-01-06
feat: Replace Tippy with Floating UI

matthewlipski:


packages/react/src/components/FormattingToolbar/DefaultButtons/NestBlockButtons.tsx
```diff
@@ -9,9 +9,12 @@ import { formatKeyboardShortcut } from "@blocknote/core";
 export const NestBlockButton = <BSchema extends BlockSchema>(props: {
   editor: BlockNoteEditor<BSchema>;
 }) => {
-  const [canNestBlock, setCanNestBlock] = useState<boolean>();
+  const [canNestBlock, setCanNestBlock] = useState<boolean>(
+    props.editor.canNestBlock()
```
YousefED, line 13: I think we should use an initializer here (`() => ...`) otherwise this gets called every render

packages/react/src/components/FormattingToolbar/DefaultButtons/NestBlockButtons.tsx
```diff
@@ -34,7 +37,9 @@ export const NestBlockButton = <BSchema extends BlockSchema>(props: {
 export const UnnestBlockButton = <BSchema extends BlockSchema>(props: {
   editor: BlockNoteEditor<BSchema>;
 }) => {
-  const [canUnnestBlock, setCanUnnestBlock] = useState<boolean>();
+  const [canUnnestBlock, setCanUnnestBlock] = useState<boolean>(
+    props.editor.canUnnestBlock()
```
YousefED, line 41: same

packages/react/src/components-shared/Toolbar/ToolbarInputDropdownButton.tsx
```diff
@@ -26,22 +25,20 @@ export const ToolbarInputDropdownButton = (
...
-      content={renderDropdown ? props.children[1] : null}
-      trigger={props.visible === undefined ? "click" : undefined}
-      interactive={true}
-      maxWidth={500}
-      zIndex={9000}
+      zIndex={10000}
       {...props}>
-      {props.children[0]}
-    </Tippy>
+      <Popover.Target>{props.children[0]}</Popover.Target>
```
YousefED, line 38: not necessarily related to this PR, but the pattern children[0] and children[1] is not very nice imo

packages/react/src/components/SideMenu/SideMenuPositioner.tsx
```diff
@@ -77,41 +84,9 @@ export const SideMenuPositioner = <
...
-    {
-      name: "preventOverflow",
-      options: {
-        mainAxis: false,
-        altAxis: false,
-      },
-    },
-  ],
-};
+// TODO: anything from below needed?
```
YousefED, line 91: remove if not needed

packages/react/src/components/SlashMenu/SlashMenuPositioner.tsx
```diff
@@ -33,65 +38,65 @@ export const SlashMenuPositioner = <
...
+        apply({ availableHeight, elements }) {
+          Object.assign(elements.floating.style, {
+            // Minimum acceptable height is 200px.
+            // `flip` will then take over.
+            height: `${Math.max(200, availableHeight - 10)}px`,
+          });
+        },
+      }),
+      flip({
+        fallbackStrategy: "initialPlacement",
```
YousefED, line 56: Nice you were able to fix all this. maybe add an extra comment to explain the scenarios / why `initialPlacement` is needed

packages/react/src/editor/styles.css
```diff
@@ -159,6 +160,12 @@
     color: var(--bn-colors-hovered-text);
 }
 
+/* Mantine Popover component base styles */
+.mantine-Popover-dropdown {
+    all: revert;
```
YousefED, line 165: why are these `all: revert` necessary? In general using `all` is not desirable I think (I know I did this somewhere else too, we should probably get rid of that one as well at some point)

YousefED (approved):
sweet

PR #503, opened 2024-01-08
feat: Editor option to customize placeholders

### LINKED ISSUE #320: Perf issues with a large document
**Describe the bug**
When we create a large document, editing is becoming very slow, we ran some profiling and found out that placeholder extension causes it, so can we expose an option to disable this

### LINKED ISSUE #403: Default placeholder changing problem
Hello, 

I don't use the slash menu, but in the placeholder of each block it says ```Enter text or type '/' for commands. ``` I couldn't change this placeholder no matter what I did.

Is there any way to change it?

matthewlipski:
This PR adds an easier way to customize placeholders, using an editor option. The option currently has the following definition:

```ts
type BlockNoteEditorOptions = {
  /**
   * Determines if/what placeholder should be rendered for a given block.
   * @param block The block to render a placeholder for.
   * @param containsCursor Whether the text cursor is inside the block.
   * @param isFilter Whether the block was created using the add block button.
   * @return The placeholder string, or `undefined` if one should not be rendered.
   */
  placeholder: (
    block: Block<any, any, any>,
    containsCursor: boolean,
    isFilter: boolean
  ) => string | undefined;
  ...
}
```

This is both flexible while being simple, e.g. setting a single placeholder for everything goes like this:

```ts
const options = {
  placeholder: (
    block: Block<any, any, any>,
    containsCursor: boolean,
    isFilter: boolean
  ): string | undefined => {
    if(containsCursor) {
      return "Placeholder";
    }
    
    return undefined;
  };
  ...
}
```

I'm not sure if this is necessarily the best way of doing things vs e.g. using CSS variables, and also the function arguments can probably be streamlined a bit, but would be great to get feedback on this.

Closes #403
Closes #320


grutt:
Is there a plan to integrate this change in the near term? This is an open requirement for a project I'm working on.

YousefED:
> Is there a plan to integrate this change in the near term? This is an open requirement for a project I'm working on.

Yes, we want to make it possible to customize all strings soon.

Whats your exact use case?

packages/core/src/extensions/Placeholder/PlaceholderExtension.ts
```diff
@@ -13,54 +12,42 @@ const PLUGIN_KEY = new PluginKey(`blocknote-placeholder`);
...
+        numberedListItem: "List",
       },
     };
   },
 
-  addProseMirrorPlugins() {
-    const styleEl = document.createElement("style");
-
-    // Append <style> element to <head>
+  onCreate() {
```
YousefED, line 41: can't we use the PM API for this instead of TipTap API? (we're trying this as a general rule right?)

YousefED (approved):
nice, one small question inline.

All tested well including nested and custom blocks?

PR #506, opened 2024-01-10
feat: Asynchronous slash menu item fetching

### LINKED ISSUE #488: Custom search function for the slash menu

matthewlipski:
This PR refactors the suggestion plugin to support fetching items asynchronously. To do this, some of the existing responsibility of the suggestion plugin has to be move. Specifically, the keyboard handling is moved to the slash menu component and item filtering is moved to the function provided in the `slashMenuItems` option.

I think we can move the plugin state to the plugin view like it is in the table handles plugin, though this is still TODO.

Closes #488

### DIFF
--- a/packages/core/src/extensions-shared/suggestion/SuggestionPlugin.ts
+++ b/packages/core/src/extensions-shared/suggestion/SuggestionPlugin.ts
@@ -8,33 +8,30 @@ import { SuggestionItem } from "./SuggestionItem";
 
 const findBlock = findParentNode((node) => node.type.name === "blockContainer");
 
-export type SuggestionsMenuState<T extends SuggestionItem> =
-  BaseUiElementState & {
-    // The items that should be shown in the menu.
-    items: Promise<T[]>;
-  };
+export type SuggestionsMenuState = BaseUiElementState & {
+  query: string;
+};
 
 class SuggestionsMenuView<
-  T extends SuggestionItem,
   BSchema extends BlockSchema,
   I extends InlineContentSchema,
   S extends StyleSchema
 > {
-  private suggestionsMenuState?: SuggestionsMenuState<T>;
+  private suggestionsMenuState?: SuggestionsMenuState;
   public updateSuggestionsMenu: () => void;
 
-  pluginState: SuggestionPluginState<T>;
+  pluginState: SuggestionPluginState;
 
   constructor(
     private readonly editor: BlockNoteEditor<BSchema, I, S>,
     private readonly pluginKey: PluginKey,
     updateSuggestionsMenu: (
-      suggestionsMenuState: SuggestionsMenuState<T>
+      suggestionsMenuState: SuggestionsMenuState
     ) => void = () => {
       // noop
     }
   ) {
-    this.pluginState = getDefaultPluginState<T>();
+    this.pluginState = getDefaultPluginState();
 
     this.updateSuggestionsMenu = () => {
       if (!this.suggestionsMenuState) {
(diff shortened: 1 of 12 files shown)


packages/core/src/extensions-shared/suggestion/SuggestionPlugin.ts
```diff
@@ -215,11 +211,7 @@ export const setupSuggestionsMenu = <
               triggerCharacter:
                 transaction.getMeta(pluginKey)?.triggerCharacter || "",
               queryStartPos: newState.selection.from,
-              items: items(""),
-              keyboardHoveredItemIndex: 0,
-              // TODO: Maybe should be 1 if the menu has no possible items? Probably redundant since a menu with no items
-              //  is useless in practice.
-              notFoundCount: 0,
+              items: getItems(""),
```
YousefED, line 214: - what happens with notFoundCount?
- can we cancel promises?
- how do we show loading state?

packages/react/src/components/SlashMenu/DefaultSlashMenu.tsx
```diff
@@ -5,37 +5,137 @@ import groupBy from "lodash.groupby";
...
+          props.closeMenu();
+        } else {
+          notFoundCount.current++;
+        }
+      }
+    });
+  }, [props, selectedIndex]);
+
+  // Handles keyboard navigation.
+  useEffect(() => {
```
YousefED, line 77: TODO: discuss

packages/react/src/components/SlashMenu/DefaultSlashMenu.tsx
```diff
@@ -5,37 +5,137 @@ import groupBy from "lodash.groupby";
...
+            />
+          );
+          itemIndex++;
+        }
+      });
+
+      setOrderedItems(orderedItems);
+      setRenderedItems(renderedItems);
+
+      // Closes menu if query does not match any items after 3 tries.
```
YousefED, line 63: TODO: discuss

packages/react/src/components/SlashMenu/DefaultSlashMenu.tsx
```diff
@@ -14,69 +14,106 @@ export function DefaultSlashMenu<BSchema extends BlockSchema>(
...
   >(undefined);
-  const [renderedItems, setRenderedItems] = useState<JSX.Element[] | undefined>(
-    undefined
-  );
+  const [loader, setLoader] = useState<JSX.Element | null>(null);
   const [selectedIndex, setSelectedIndex] = useState<number>(0);
-  const notFoundCount = useRef(0);
 
-  // Sets the items to render.
+  const prevQuery = useRef<string | undefined>(undefined);
```
YousefED, line 20: I think the code is getting a bit complicated now. Maybe the easiest way to do this is to ignore the result when the current query doesn't match the query of the incoming result?

YousefED:
See https://www.blocknotejs.org/docs/ui-components/suggestion-menus for the
new docs :)

PR #516, opened 2024-01-13
feat: CSS variables for placeholders

matthewlipski:
This PR allows for setting CSS variables to customize the default placeholders. It contains only basic functionality as it's meant as a temporary solution before we fix performance issues in #503:
- Can change default placeholder which appears in the block which contains the text cursor
- Can change placeholder which is appears after pressing the add block button
- Can change placeholders for default heading and list item blocks, which are always visible and overwrite the default placeholder
- No support for custom blocks or more advanced filtering based on block

Default placeholders:
```css
.bn-container {
  --bn-default-placeholder: "Enter text or type '/' for commands";
  --bn-filter-placeholder: "Type to filter";
  --bn-heading-placeholder: "Heading";
  --bn-list-placeholder: "List";
}
```

### DIFF
--- a/packages/core/src/editor/Block.css
+++ b/packages/core/src/editor/Block.css
@@ -314,31 +314,31 @@ NESTED BLOCKS
 }
 
 /* TODO: would be nicer if defined from code */
+.bn-container {
+  --bn-default-placeholder: "Enter text or type '/' for commands";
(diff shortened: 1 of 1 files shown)


YousefED:
will be supserseded by #503

PR #525, opened 2024-01-18
fix: Mantine setting global styles

### LINKED ISSUE #523: Importing react style.css is breaking our own CSS
**Describe the bug**
We are using your awesome BlockNote on Twenty and wanted to migrate to the new version: https://github.com/twentyhq/twenty/pull/3517/files

From the release notes:
```
Breaking Changes
Change import "@blocknote/core/style.css"; to import "@blocknote/react/style.css"; if you're using @blocknote/react

This update changes how themes and styles are applied to the editor. This means if you were previously overriding CSS using the componentStyles field in a Theme object, you should convert those styles to CSS rules. See the new theming and styling documentation [here](https://www.blocknotejs.org/docs/theming).
```

However, react/style.css does change the style of non block-note related class, for example: 
`body,html{height:100%}...`

This is an issue as we import it in our project using @blocknote/react/style.css, it completely breaks our page layout.

Before:
image

After:
image

(see that the Sort button for example is broken)

matthewlipski:
Due to the Mantine update from 0.11, the way in which Mantine component styling is now different. Mantine now sets styles in CSS, some of which are set globally. This can cause issues by overwriting existing styles, when in reality they should be scoped to only elements within the editor.

This PR unsets global styles set by Mantine and moves them inside the `.bn-container` element (excluding setting CSS variables). Unfortunately, for Mantine styling to work at all, you need to import `@mantine/core/styles/global.css`, which includes said global styles. This means that AFAIK, there is no way to skip importing them, and the best we can do is to unset them. This feels really hacky, so any suggestions for a nicer fix are welcome.

Closes #523

### DIFF
--- a/packages/react/src/editor/styles.css
+++ b/packages/react/src/editor/styles.css
@@ -1,6 +1,59 @@
 @import url("@mantine/core/styles.css");
 @import url("@blocknote/core/style.css");
 
+/* Move Mantine global styles to the scope of bn-container */
+body, html {
+    height: unset;
+}
+
+*, :after, :before {
+    box-sizing: unset;
+}
+.bn-container *, .bn-container :after, .bn-container :before {
+    box-sizing: border-box;
+}
+
+button, input ,select ,textarea {
+    font: unset;
+}
+
+button, select {
+    text-transform: unset;
+}
+.bn-container button, .bn-container select {
+    text-transform: none;
+}
+
+body {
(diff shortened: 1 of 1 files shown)


YousefED:
closing, main issue is fixed by upgrading mantine (#535)
For other styles, using "unset" is probably not an option because it'd also unset user added styles

PR #529, opened 2024-01-18
fix: Formatting toolbar not hiding while dragging blocks

### LINKED ISSUE #500: Formatting Toolbar become visible when dragging a block
When we dragging any block,  the formatting toolbar is always there.

We can also reproduce it in our site https://www.blocknotejs.org/

image

matthewlipski:
Closes #500


packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
```diff
@@ -52,7 +52,8 @@ export class FormattingToolbarView {
 
     pmView.dom.addEventListener("mousedown", this.viewMousedownHandler);
     pmView.dom.addEventListener("mouseup", this.viewMouseupHandler);
-    pmView.dom.addEventListener("dragstart", this.dragstartHandler);
+    pmView.dom.addEventListener("dragstart", this.dragHandler);
+    pmView.dom.addEventListener("dragover", this.dragHandler);
```
YousefED, line 56: this never gets removed?

PR #534, opened 2024-01-18
refactor: Mentions & custom suggestion menus

matthewlipski:
This PR refactors the mentions & custom suggestion menus code to support creating menus using React hooks. There are 3 parts to this refactor:

- There is now a single suggestion menu plugin which controls all suggestion menus. This means that menus can be added dynamically after editor initialization (and removed, if that's at all useful). This also brings the UX closer to Notion, as only a single suggestion menu can be active at a time.
- You can now add a suggestion menu to the editor using the `createSuggestionMenu` function, or `useSuggestionMenu` hook. The hook returns both props for the menu, and props for positioning the menu if they are set on a parent element.
- Any item type can now be used in the suggestion menus. It is now entirely the UI layer's responsibility to figure out what to do with those items. However, since the UI layer also sets the items via the `useSuggestionMenu` hook, it should already know what type those items are.

I think this pattern makes a lot of sense, but the `useSuggestionMenu` return data and the general idea behind positioners needs to be reviewed. @YousefED I think it's best to jump on a call to discuss these, also for other UI elements because using hooks for positioning is imo more intuitive and it's how Floating UI does it in the first place.


packages/react/src/components/SuggestionMenu/DefaultRenderItems.tsx
```diff
@@ -0,0 +1,37 @@
...
+  loadingState?: "initial" | "loading-initial" | "loading" | "loaded";
+  selectedIndex?: number;
+};
+
+export function DefaultRenderItems(props: SuggestionMenuItemsProps) {
+  let numLabels = 0;
+
+  return (
+    <SuggestionMenu loadingState={props.loadingState}>
+      {props.items.map((item, index) => {
```
YousefED, line 22: we need to add a `key` to children, pretty sure you'll get a react warning otherwise (see console)

packages/react/src/components/SuggestionMenu/DefaultSuggestionMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+  useCloseSuggestionMenuNoItems(items, usedQuery, closeMenu);
+
+  const selectedIndex = useSuggestionMenuKeyboardNavigation(
+    editor,
+    items.filter((item) => !("label" in item)),
+    (item) => (item as SuggestionMenuItemProps).executeItem(),
+    closeMenu
+  );
+
+  const SuggestionMenuItems = renderItems || DefaultRenderItems;
```
YousefED, line 106: Some of the namings are still a bit confusing / would benefit from streamlining.

Thoughts:
- now that a lot of Mantine stuff is extracted, we could rename `SuggestionMenu` to `MantineSuggestionMenu`
- is `DefaultRenderItems` / `renderItems` still useful? or should that just be `SuggestionMenuComponent` and we get sth like: `SuggestionMenuComponent = MantineSuggestionMenu` as default? (and remove the renderItems layer)?
-

packages/react/src/hooks/useCloseSuggestionMenuNoItems.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useRef } from "react";
+
+export function useCloseSuggestionMenuNoItems<T>(
```
YousefED, line 3: add description

packages/react/src/components/SuggestionMenu/DefaultSuggestionMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+  I extends InlineContentSchema,
+  S extends StyleSchema
+>(props: {
+  editor: BlockNoteEditor<BSchema, I, S>;
+  query: string;
+  closeMenu: () => void;
+  clearQuery: () => void;
+  getItems?: (
+    query: string,
+    closeMenu: () => void,
```
YousefED, line 77: I'm not convinced closemenu / clearquery should be part of `getItems`

packages/react/src/components/SuggestionMenu/DefaultSuggestionMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const { items, usedQuery, loadingState } = useLoadSuggestionMenuItems(
+    query,
+    getItemsForLoading
+  );
+
+  useCloseSuggestionMenuNoItems(items, usedQuery, closeMenu);
+
+  const selectedIndex = useSuggestionMenuKeyboardNavigation(
+    editor,
+    items.filter((item) => !("label" in item)),
```
YousefED, line 101: I think you'll run into this when splitting the type of `DefaultPositionedSuggestionMenu`, but I don't think the labels should be returned from `getItems`. Maybe the rendering of the labels should be part of the UI layer?

examples/editor/examples/basic/App.tsx
```diff
@@ -153,14 +108,19 @@ export function App() {
     <BlockNoteView className="root" editor={editor}>
       <FormattingToolbarPositioner editor={editor} />
       <HyperlinkToolbarPositioner editor={editor} />
-      <SlashMenuPositioner editor={editor} />
       <SideMenuPositioner editor={editor} />
+      <DefaultPositionedSuggestionMenu editor={editor} />
```
YousefED, line 112: What do you think about the following?

```
<BlockNoteView ... >
    <SuggestionMenu
        editor={editor}
        triggerCharacter={"@"}
        getItems={(query, closeMenu, clearQuery) =>
          getMentionMenuItems(editor, query, closeMenu, clearQuery)
        }
      />
   <DefaultBlockNoteUIElements formattingToolbar={false} slashMenu={true} ...etc... /> // props default to true
</BlockNoteView>
```

This way, `DefaultBlockNoteUIElements` can be used to toggle on / off built-in UI elements

packages/react/src/components/SuggestionMenu/DefaultSuggestionMenu.tsx
matthewlipski: That's one thing I forgot to discuss in the call - there are trade offs with both options:

Including labels in `getItems` allows us to use the same type for items and component props, but means we have to introduce more logic for keyboard navigation.

Not including them means we have to add smth like the old `group` attribute to the default return type of `getItems`, but it does simplify the keyboard navigation since we don't have to account for the labels. It also means we have to sort the items before returning them.

Right now, we also include `icon` in the default item type, so the separation of data/UI is already iffy, so imo it's better to include labels in `getItems`.

examples/editor/examples/basic/App.tsx
matthewlipski: Yeah I think this is way better, shortens basically every use case except for removing most elements (which I think is pretty uncommon).

PR #540, opened 2024-01-23
Add Liveblocks info and BlockNote download command

CTNicholas:
This adds info on how to download a BlockNote example with Liveblocks in one command! It also shows a screenshot and mentions our provider.

~_These links and the command won't work until we've merged a couple PRs on our side._ They'll be online within a couple of days.~ This PR is now ready to go on my end!

- https://github.com/liveblocks/liveblocks/pull/1185


CTNicholas:
All the links work now, we're online 🥳 
- https://liveblocks.io/docs/get-started/yjs-blocknote-react
- https://liveblocks.io/examples/collaborative-text-editor/nextjs-yjs-blocknote
- https://liveblocks.io/examples/collaborative-text-editor-advanced/nextjs-yjs-blocknote-advanced

YousefED:
Thanks! 

I think the image doesn't add a lot, imo the video would be better, although I can't guarantee it will stay up forever (e.g.: when we add more providers etc). Could you either remove the image or switch to the video?

CTNicholas:
> I think the image doesn't add a lot, imo the video would be better, although I can't guarantee it will stay up forever (e.g.: when we add more providers etc). Could you either remove the image or switch to the video?

Sure thing! I've compressed and shrunk the video to save on bandwidth, and replaced the image.

Issue #231, opened 2023-06-02
Input rule API support


matthewlipski:
TipTap/ProseMirror have customizable input rules which can be used to execute commands when a certain string or expression is input somewhere in the document. BlockNote already uses these (e.g. "### " converts the selected block to a H3, "- " converts it to a list, etc), but it would be nice to allow custom ones too and have an easy way of doing so in the API.

PR #541, opened 2024-01-24
Added inputRule to CustomBlockConfig

hyuckkim:
hello.

I need simple custom input rule, then I made one.
https://github.com/TypeCellOS/BlockNote/issues/231

some RegExp that I used is:  [stackoverflow Link](https://stackoverflow.com/questions/5663987/how-to-properly-escape-characters-in-regexp)

Then, works will like that:
tabtab

### DIFF
--- a/examples/editor/examples/react-custom-blocks/App.tsx
+++ b/examples/editor/examples/react-custom-blocks/App.tsx
@@ -44,6 +44,12 @@ export const alertBlock = createReactBlockSpec(
         values: ["warning", "error", "info", "success"] as const,
       },
     },
+    inputRule: [
+      {input: "[]w", prop: { type: "warning"} },
+      {input: "[]e", prop: { type: "error"} },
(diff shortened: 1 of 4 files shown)


YousefED:
Awesome, thanks!

From an API-design perspective, I'm a bit in doubt whether this should be a property of a custom block. An input rule is basically a rule saying "In my editor, when user types X, insert Y". Making it a part of the block really couples X and Y, where I think conceptually X should be part of the editor setup / specification instead of custom block Y

What do you think?

hyuckkim:
I agree with that.
When I thought about managing all inputRules in useBlockNote, I think it would be much cleaner.
However, it seemed too difficult to implement to me.
It seemed like the only way to change the inputRule was createBlockSpec, so I just used it.

Refactoring inputRules to be specified in useBlockNote is beyond my scope.If the implementation has to be that way, there's nothing I can do.

PR #548, opened 2024-01-26
docs: Dark mode styling fix

matthewlipski:


packages/website/docs/demoUtils.ts
```diff
@@ -14,5 +14,16 @@ export const getStyles = (isDark: Ref<boolean>): string => `body {
   -webkit-tap-highlight-color: transparent;
   -webkit-touch-callout: none;
   
-  background-color: ${isDark ? "#151515" : "white"};
+  background-color: ${isDark ? "#151515" : "white"} !important;
 }`;
+
+export const index = `import React, { StrictMode } from "react";
```
YousefED, line 20: why was this added / needed?

matthewlipski: Forgot to remove

matthewlipski: Forgot to remove

YousefED: ok, so will be fixed / removed?

Issue #546, opened 2024-01-25
Any way not to have Mantine set styles globally?


vincaslt:
**Describe the bug**

BlockNote styles are importing Mantine styles which set a bunch of global styles. Maybe makes sense if your whole project is using Mantine, but mine doesn't and these things interfere with existing custom styles in my project (see below).

Additionally Mantine sets a bunch of color variables that I don't know if even BlockNote is using, so it's a bunch of extra useless css in the bundle.

```
*,:after,:before {
    box-sizing: border-box
}

button,input,select,textarea {
    font: inherit
}

button,select {
    text-transform: none
}

body {
    -webkit-font-smoothing: var(--mantine-webkit-font-smoothing);
    -moz-osx-font-smoothing: var(--mantine-moz-font-smoothing);
    background-color: var(--mantine-color-body);
    color: var(--mantine-color-text);
    font-family: var(--mantine-font-family);
    font-size: var(--mantine-font-size-md);
    line-height: var(--mantine-line-height);
    margin: 0
}

@media screen and (max-device-width: 500px) {
    body {
        -webkit-text-size-adjust:100%
    }
}

@media (prefers-reduced-motion:reduce) {
    [data-respect-reduced-motion] [data-reduce-motion] {
        animation: none;
        transition: none
    }
}

[data-mantine-color-scheme=dark] .mantine-dark-hidden,[data-mantine-color-scheme=light] .mantine-light-hidden {
    display: none
}
```

**To Reproduce**
<links to a sandbox or clear steps to reproduce are super helpful!>

**Misc**
- Node version:
- Package manager:
- Browser: 
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

rockiger:
Yes, Mantine seems to be a problematic dependency. It also messes the Blocknote website up in dark-mode:

![image](https://github.com/TypeCellOS/BlockNote/assets/458677/20e7e1e7-714a-4bef-85be-1eeb022132ea)
Source: https://www.blocknotejs.org/examples/custom-ui

leighton-tidwell:
I was experiencing this as well using tailwind, it was causing my theme to always default to dark mode. I have since fixed it by hardcoding the page to be light (I don't need a dark mode for my application).

```css
html {
  color-scheme: light !important;
}
```

matthewlipski:
[...]

PR #553, opened 2024-01-31
fix: Mantine styles

### LINKED ISSUE #549: Spinner showing in all icons
**Describe the bug**
<what's going wrong!?>
Spinner icon is showing up beside all icon buttons, see image below:
Screenshot 2024-01-27 at 6 19 26 PM

Screenshot 2024-01-27 at 6 19 32 PM

Environment: Nextjs

Edit:
I added a CSS style to the editor to hide the spinner, the icon buttons render correctly now. I do hope this isn't masking an issue.
.mantine-ActionIcon-loader {display: none}

**To Reproduce**
I followed the quick start instructions in https://www.blocknotejs.org/docs/quickstart

**Misc**
- Node version: 18.18.0
- Package manager: npm vs 9.8.1
- Browser: Chrome, Firefox

matthewlipski:
Updating Mantine did not get rid of all of the global styles it sets, which are still causing issues. So it seems like it is a better alternative to just reset them, as it doesn't seem that the Mantine team is going to change this anytime soon (https://discord.com/channels/854810300876062770/1193507663020032060).

This PR also fixes loaders being rendered inside `ActionIcon` components:
> This was kind of a pain to diagnose, but bottom line is that it should be fixed by updating Mantine from 7.4.2 to 7.5.0. I'm not sure exactly what's causing the issue, but there seem to be styles missing in `@mantine/core/styles.css` for the `.m-302b9fb1` selector that's used in the loader in 7.4.2. Using 7.5.0 and disabling these additional styles causes icons to have the loader visible next to them as in the screenshot.

Closes #549 #546


YousefED (approved):
Approve, but I don't think it closes #546 definitively. We should open an issue with Mantine and:
- explain clearly what the issue is and what the impact is (many people running into it)
- propose a solution (quick brainstorm, there might be better options; like splitting the css to an optional file, or making the root container where it's applied to configurable?)

rockiger (approved):
Works as a workaround. But why not override all global styles from Mantine?

matthewlipski:
Changed it to this solution: https://github.com/orgs/mantinedev/discussions/5685
This avoids having to import Mantine global styles at all, and sets them on `.bn-container` instead, effectively scoping them to the editor only.

rockiger (approved):

packages/core/package.json
```diff
@@ -86,7 +86,7 @@
     "remark-stringify": "^10.0.2",
     "unified": "^10.1.2",
     "uuid": "^8.3.2",
-    "y-prosemirror": "1.0.20",
+    "y-prosemirror": "1.2.1",
```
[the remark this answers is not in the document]
matthewlipski: I also re-generated `package-lock.json` since it seemed to be stuck on Mantine 7.4.2. This gave me an error that there was a dependency mismatch (with `@tiptap/extension-collaboration`) between 1.0.20 and 1.2.1, so this change resolved it.

packages/react/src/editor/mantineStyles.css
```diff
@@ -79,6 +84,8 @@
 @import url("@mantine/core/styles/TypographyStylesProvider.css");
 
 /* Mantine global styles, scoped to bn-container */
+/* Based on @mantine/core/styles/global.css but with styles set on `body` and
```
YousefED, line 87: can you add a link to the file on github?

packages/react/src/editor/mantineStyles.css
```diff
@@ -1,3 +1,8 @@
+/* Based on https://github.com/orgs/mantinedev/discussions/5685 */
+
+/* We need all the Mantine styles except the global styles, so unfortunately our
```
YousefED, line 3: can you add a direct link to where this list comes from?

Crsk:
This solution was merged into v0.11.2 but in that version I'm still having the issue. I just tried BlockNote first time (v0.11.2) and now my styles are messed up. I also tried updating my app mantine version from v7.3.2 to v7.5.1 with no luck

Issue #471, opened 2023-12-08
Real-time collaboration causes an infinite stream of updates when focusing the editor on a standard Next.js app


rodobre:
**Describe the bug**
When using a basic Next.js app, the real-time collaboration feature shown in the documentation causes an infinite stream of updates when focusing the editor, leading to constant flickering and re-renders.

**To Reproduce**
https://github.com/rodobre/blocknote-collaboration-issue

**Misc**
- Node version: 18.0.0 to 21.4.0
- Package manager: NPM
- Browser: Google Chrome Version 119.0.6045.199 (Official Build) (arm64)
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

Video:

https://cdn.discordapp.com/attachments/1015169282444894219/1182540721383821412/Screen_Recording_2023-12-08_at_05.31.10.mov?ex=658511de&is=65729cde&hm=5584f4d456219294547853f0f8fa4920f6df7120c0cb429ee0fb7e1212733eef&

rodobre:
Related to this:

https://github.com/ueberdosis/tiptap/issues/4482

Bug only occurs in dev due to React strict mode. Feel free to close the issue if you think no fix can be done on BlockNote's end.

matthewlipski:
Thanks for reporting this! Seems like the TipTap guys are still working on this, hopefully we see a fix from them relatively soon

YousefED:
StrictMode + collab now works again in the new version! We improved our React architecture

Issue #342, opened 2023-09-20
Better error handling for `useBlockNote`


infinite-persistence:
**Describe the bug**
`useBlockNote` currently throws an error when the given `initialContent` is not kosher.
Handling the error is cumbersome since effects can't be called conditionally. However, it seems like an `ErrorBoundary` HOC can be a workaround to suppress the error.

![image](https://github.com/TypeCellOS/BlockNote/assets/64950861/a47cd45d-c5c6-4a92-97c0-f7b5ef296a94)

**To Reproduce**
Pass a non-json value to `initialContent`, e.g. `initialContent: "abc"`.
This could come from corrupted saved data.

PR #567, opened 2024-02-10
Refactor/cleanup mount system

### LINKED ISSUE #497: Type Error: Generic type 'Block' requires 3 type arguments
**Describe the bug**
Facing error of typescript regarding Block[ ] 

image

**Misc**
- Node version: latest
- Package manager: npm
- Browser: Google Chrome

YousefED:
This PR:
- Revisits how we mount the React component / editor
- Improves the React API to be more idiomatic React (as part of this, remove the event handlers passed to `options`)

Smaller fixes:
- Cleans some docs / examples
- Add theme switch support for website demos / examples

closes https://github.com/TypeCellOS/BlockNote/issues/471
closes https://github.com/TypeCellOS/BlockNote/issues/342
closes #349 
closes #497


examples/01-basic/04-saving-loading/App.tsx
```diff
@@ -0,0 +1,53 @@
...
+    });
+  }, []);
+
+  // Creates a new editor instance.
+  // We use useMemo + createBlockNoteEditor instead of useBlockNote so we can delay the creation of the editor until the initial content is loaded.
+  const editor = useMemo(() => {
+    if (initialContent === "loading") {
+      return undefined;
+    }
+    return createBlockNoteEditor({ initialContent });
```
matthewlipski, line 37: This seems like a bit of a weird pattern - wouldn't it make more sense for you to be able to e.g. pass editor options in an async function? Then BlockNote can wait to init the editor instead of the consumer having to do it.

examples/01-basic/06-keyboard-shortcuts/App.tsx
```diff
@@ -32,13 +35,11 @@ const cycleBlocksShortcut = (event: KeyboardEvent, editor: BlockNoteEditor) => {
...
-      ),
-  });
+  const editor = useBlockNote({});
 
-  return <BlockNoteView editor={editor} theme={"light"} />;
+  const onKeyDown = (event: React.KeyboardEvent) => {
+    cycleBlocksShortcut(event, editor);
+  };
+
+  return <BlockNoteView editor={editor} onKeyDown={onKeyDown} />;
```
matthewlipski, line 44: Nice, this is way more intuitive👍

examples/01-basic/06-keyboard-shortcuts/README.md
```diff
@@ -1,9 +1,11 @@
 # Keyboard Shortcuts
 
+TODO: useful example? Non-basic?
```
matthewlipski, line 3: I would keep it, since in the docs ig we can just say that `BlockNoteView` takes the same props as a `div` would. I think the example is useful to show people that this is indeed the preferred method of setting things like keyboard shortcuts.

packages/react/src/hooks/useEditorContentOrSelectionChange.ts
matthewlipski, line 1: Not 100% sure on the naming here. I think if we have `useEditorContentOrSelectionChange` it kind of implies that `useEditorContentChange` and `useEditorSelectionChange` are both separate hooks, but we have `useEditorChange` and `useEditorSelectionChange` instead. Also the name is just a little too long imo.

examples/05-cursor-selections/selection-blocks/README.md
```diff
@@ -1 +1,3 @@
-# Highlighting Blocks in Selection
\ No newline at end of file
+# Highlighting Blocks in Selection
+
+TODO: same as text-cursor block. Perhaps replace both by 1 single example
```
matthewlipski, line 3: Good idea, can just use a different background color for block with text cursor vs blocks in selection.

tests/src/utils/const.ts
```diff
@@ -5,7 +5,7 @@ export const BASE_URL = !process.env.RUN_IN_DOCKER
 
 export const PASTE_ZONE_SELECTOR = "#pasteZone";
 
-export const EDITOR_SELECTOR = `[data-test="editor"]`;
+export const EDITOR_SELECTOR = `.ProseMirror`;
```
matthewlipski, line 8: Maybe use `.bn-editor` instead?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -435,7 +365,7 @@ export class BlockNoteEditor<
           ...newOptions._tiptapOptions?.editorProps?.attributes,
           ...newOptions.domAttributes?.editor,
           class: mergeCSSClasses(
-            "bn-root",
+            "bn-root", // TODO: remove this class?
```
matthewlipski, line 368: 👍

packages/core/src/editor/BlockNoteTipTapEditor.ts
matthewlipski, line 1: TODO: discuss

packages/react/src/editor/EditorContent.tsx
matthewlipski, line 1: TODO: discuss

docs/components/example/ExampleWrapper.tsx
matthewlipski, line 1: Would rename to `ThemedExample`, more explicit about what it does

matthewlipski (commented):
Looks good! Added a few suggestions and things to discuss

examples/01-basic/04-saving-loading/App.tsx
YousefED: @YousefED tbd

packages/react/src/hooks/useEditorContentOrSelectionChange.ts
YousefED: better proposal?

tests/src/utils/const.ts
YousefED: good idea, can you update?

docs/components/example/ExampleWrapper.tsx
YousefED: Good idea, can you update?

examples/05-cursor-selections/selection-blocks/README.md
YousefED: Ah, not what I meant, see comments on the other related example

packages/react/src/hooks/useEditorContentOrSelectionChange.ts
matthewlipski: I think better to just keep it as it was, i.e. `useEditorSelectionChange`, `useEditorContentChange`, and `useEditorChange`.

YousefED: The reason I changed it originally was because I added a handler to <BlockNoteView> that is called `onChange`. I do think that's clean from a consumer perspective (similar to native React textarea / input `onChange` handlers).

But also agree with your earlier feedback. Not sure what best way to go is

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -365,7 +365,6 @@ export class BlockNoteEditor<
           ...newOptions._tiptapOptions?.editorProps?.attributes,
           ...newOptions.domAttributes?.editor,
           class: mergeCSSClasses(
-            "bn-root", // TODO: remove this class?
```
YousefED: @matthewlipski was this safe to remove like this? because it is used in editor.css

PR #571, opened 2024-02-12
fix: List item enter handler applying to custom blocks

### LINKED ISSUE #558: "ListItem" is hardcoded.
https://github.com/TypeCellOS/BlockNote/blob/6ce6e57efa32372522dcf334691ed1dfa0e5ec4a/packages/core/src/blocks/ListItemBlockContent/ListItemKeyboardShortcuts.ts#L13

I cannot how it works, but It works in custom block also.
I changed my custom block's type to JohnDoeListItem and use their feature.

I think it need to allow modification to boolean values in blockspec.

matthewlipski:
Closes #558


troglodytto:
@matthewlipski What if we **do** want this behaviour to persist in a custom Block?

matthewlipski:
> @matthewlipski What if we **do** want this behaviour to persist in a custom Block? https://discord.com/channels/928190961455087667/1209559379095855154

That's fair, though it should still be possible to add that functionality even with this PR - see the replies in the thread. However, I don't think that having your block end in `listItem` is a good API for it, so I think we should still go ahead with this PR.

YousefED (approved):

PR #576, opened 2024-02-13
feat: UI element creation API

matthewlipski:
Based on the findings from implementing the custom suggestion menu API, this PR changes how UI elements and their positioners are created. There are now 2 hooks for each UI element:

`useUiElementData` - Provides the data associated with the UI element, e.g. `block`, `dragStart`, etc for the side menu.
`useUiElementPosition` - Provides attributes which, when added to the UI element's parent, control and position it.

This should make it easier and more intuitive to create custom menus, since menu data and position props are separated, making each of them easier to explain.


packages/react/src/components/FormattingToolbar/hooks/useFormattingToolbarPosition.ts
```diff
@@ -97,17 +92,14 @@ export const FormattingToolbarPositioner = <
...
-      ref={refs.setFloating}
-      style={{ ...styles, ...floatingStyles, zIndex: 3000 }}>
-      <FormattingToolbar editor={props.editor} />
-    </div>
-  );
-};
+  return {
+    isMounted: isMounted,
+    ref: refs.setFloating,
+    style: {
```
YousefED, line 98: might be good to put this in a useMemo(..., [styles, floatingStyles])

packages/react/src/components/HyperlinkToolbar/hooks/useHyperlinkToolbarData.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    return editor.hyperlinkToolbar.onDataUpdate((data) => {
+      setText(data.text);
+      setUrl(data.url);
+    });
+  }, [editor]);
+
+  return {
+    text: text!,
```
YousefED, line 32: this seems unsafe. Can we prevent the `!`?

packages/react/src/components/TableHandles/hooks/useTableHandlesPosition.ts
```diff
@@ -0,0 +1,139 @@
...
+  });
+
+  const rowTransitionStyles = useTransitionStyles(rowFloating.context);
+  const colTransitionStyles = useTransitionStyles(colFloating.context);
+
+  useEffect(() => {
+    return editor.tableHandles!.onPositionUpdate((position) => {
+      // console.log("update", state.draggingState);
+      setShow(position.show);
+      setForceUpdate(Math.random());
```
YousefED, line 54: why do we need this here? maybe needs a comment (might have been my code ofc)

YousefED (commented):
Awesome, only some small comments and these things we should discuss together:

- was thinking there's some repetitive code, maybe we can simplify this? Although then I stumbled upon the Table + Suggestion implementations which do differ, so maybe it's not that useful
- Discuss naming of `Default****`
- Discuss API of `BlockNoteDefaultUI`
- I think I still need to review the "final state" of the mentions implementation, right? e.g.: https://github.com/TypeCellOS/BlockNote/pull/534#discussion_r1470014425

packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
```diff
@@ -221,15 +212,19 @@ export class FormattingToolbarProsemirrorPlugin extends EventEmitter<any> {
...
       key: formattingToolbarPluginKey,
       view: (editorView) => {
-        this.view = new FormattingToolbarView(editor, editorView, (state) => {
-          this.emit("update", state);
-        });
+        this.view = new FormattingToolbarView(
+          editor,
+          editorView,
+          (position) => {
+            this.emit("update position", position);
```
YousefED, line 219: To discuss on call:

naming (e.g.:  `updatePosition`), and theory about splitting all in 2 events

packages/react/src/components-shared/DefaultPositionedUiElement.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const DefaultPositionedUiElement = <
+  State extends UiElementPosition,
+  BSchema extends BlockSchema = DefaultBlockSchema,
+  I extends InlineContentSchema = DefaultInlineContentSchema,
+  S extends StyleSchema = DefaultStyleSchema
+>(props: {
+  editor: BlockNoteEditor<BSchema, I, S>;
+  onUpdate: (callback: (state: State) => void) => void;
+  zIndex: number;
+  options?: Partial<UseFloatingOptions>;
```
YousefED, line 25: options is not used

packages/react/src/hooks/useUiElement.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from "react";
+
+export function useUiElement<State>(
```
YousefED, line 3: maybe sth like `useProsemirrorPluginEventState`?

packages/react/src/hooks/useUiElementPosition.ts
```diff
@@ -0,0 +1,45 @@
...
+      refs.setReference({
+        getBoundingClientRect: () => referencePos,
+      });
+    }
+  }, [referencePos, refs]);
+
+  return {
+    isMounted: isMounted,
+    ref: refs.setFloating,
+    style: {
```
YousefED, line 38: let's memo this one

packages/react/src/hooks/useUiElementPosition.ts
```diff
@@ -0,0 +1,45 @@
+import {
+  useFloating,
+  UseFloatingOptions,
+  useTransitionStyles,
+} from "@floating-ui/react";
+import { useEffect } from "react";
+import { UiComponentPosition } from "../components-shared/UiComponentTypes";
+
+export function useUiElementPosition(
```
YousefED, line 9: `useFloatingUIElement`?

packages/react/src/components-shared/DefaultPositionedUiElement.tsx
matthewlipski: This component is the generic version of the menu/toolbar specific `DefaultPositioned...` components. It's not finished bc we still need to discuss whether adding another level of abstraction here is a good idea.

packages/react/src/hooks/useUiElement.ts
matthewlipski: Hmm I think it's not gonna be super obvious what it does if we look at it in a few months time, since we use other prosemirror plugins which aren't relevant for this hook. How about just `useUiElementState`?

packages/react/src/hooks/useUiElementPosition.ts
matthewlipski: I think the fact that we use FloatingUI is more of an implementation detail, so not sure if that should be reflected in the name (could say the same for `MantineSuggestionMenu`). I would just do smth like `useUiElementPositioning` or `useDefaultUiElementPositioning`.

packages/react/src/components/TableHandles/hooks/useTableHandlesPosition.ts
matthewlipski: I think it's something you added, after removing it everything still seems to work just fine so no clue what it's for. Any ideas?

Issue #530, opened 2024-01-18
Hyperlink toolbar button is not working on Safari.


hsingh-ds:
**Describe the bug**
I am trying to create a hyperlink on safari browser but the button is not working.

**To Reproduce**
Open editor app in Safari browser and try to add hyperlink.

v-porytskyi:
And I see the same behavior. But the paragraph colors and buttons are also not working in my case.

eatmoarrice:
0.10.1 works btw. It breaks on 0.11.0.

eatmoarrice:
This seems to be fixed in 0.11.1.

artwist-polyakov:
Hello, can you tell me why you are closing this Issue? After all, the bug persists in safari until now.

YousefED:
> Hello, can you tell me why you are closing this Issue? After all, the bug persists in safari until now.

Hi! Can you confirm you're still having this issue on the latest version? (@eatmoarrice reported it was fixed).

On a side note, we can always prioritize issues for sponsors and / or take contributions, these largely drive our roadmap / prioritization

artwist-polyakov:
> Hi @artwist-polyakov ! Can you confirm you're still having this issue on the latest version? (@eatmoarrice reported it was fixed).

Hello,

Yes, I can confirm that I am still experiencing the issue with hyperlinks not working in Safari even after checking the latest version available at [BlockNote.js](https://www.blocknotejs.org). If there has been a more recent update that is not reflected on the website, could you please point me to where I can access it? Otherwise, it seems the problem persists in the version currently provided on the official site.

eatmoarrice:
Sorry it is indeed not working. There are several conditions where it works, and my testing happened to be under one such:
- When you interact with the menu that appears after rearranging the blocks.
- When you interact with the menu that appears after a long press on the hyperlink button (the menu will disappear for a sec, then re-appear).
![hyperlink](https://github.com/TypeCellOS/BlockNote/assets/63140320/d57a1626-42b3-4616-a59a-e683cc57c359)

artwist-polyakov:
Could we reopen this issue (as not fixed)?

PR #599, opened 2024-03-03
fix: Safari toolbar buttons & dropdowns

matthewlipski:
Apparently, buttons are not focused on click in Safari (https://stackoverflow.com/questions/61245883/why-blur-and-focus-doesnt-work-on-safari). Since we use a blur event handler to hide the formatting toolbar sometimes, this wasn't working which caused the link button and block type dropdown to close the formatting toolbar on click. This PR forces toolbar buttons to become focused on mouse down, which fixes the issue. Doesn't seem to be a problem for other UI elements.

closes https://github.com/TypeCellOS/BlockNote/issues/530

### DIFF
--- a/packages/react/src/components/mantine-shared/Toolbar/ToolbarButton.tsx
+++ b/packages/react/src/components/mantine-shared/Toolbar/ToolbarButton.tsx
@@ -1,5 +1,5 @@
 import { ActionIcon, Button, Tooltip } from "@mantine/core";
-import { ForwardedRef, MouseEvent, forwardRef } from "react";
+import { MouseEvent, forwardRef, useRef } from "react";
 import type { IconType } from "react-icons";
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/mantine-shared/Toolbar/ToolbarButton.tsx
```diff
@@ -33,6 +34,13 @@ export const ToolbarButton = forwardRef(
         {/*Creates an ActionIcon instead of a Button if only an icon is provided as content.*/}
         {props.children ? (
           <Button
+            // Needed as Safari doesn't focus button elements on mouse down
+            // unlike other browsers.
+            onMouseDown={() => {
+              if (buttonRef.current !== null) {
```
YousefED, line 40: Nice find!

I think you can get the element from the event (first argument of event handler). That way we don't need the ref / hacks and the code gets more succint

Issue #603, opened 2024-03-05
Refresh pages content will cause SideMenu false trigger


tcper:
**Describe the bug**
I have a notes list, select list item will refresh blocknote content,
The scenario is
`request content from server -> remove all blocks -> insert all blocks`

`
Pseudocode:

editor.removeBlocks(editor.topLevelBlocks.map(_ => _.id))
editor.insertBlocks(blockList, id, 'before')
`

After that, when you open Color menu, the whole sidemenu will always follow the mouse, changing hover blocks, without correct frozen.
![20240305-135220](https://github.com/TypeCellOS/BlockNote/assets/363082/d990fbda-3ed5-424f-90e6-fd4839876cf2)

**To Reproduce**
1. init a content
2. Clear all blocks
3. init a new content
4. repeat more than 2 times

And I find it, when you refresh content, SideMenuPlugin will bind more than 1 times, it will cause `mousemove` event bind many times.

So, I guess, is that a proper way to refresh blocknote content? Or my way is not correct?

**Misc**
- Node version: v18.17.0
- Package manager: 9.8.0
- Browser: 120.0.6099.109
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

PR #604, opened 2024-03-05
Fix remove 'mousemove' event listener

tcper:
L 290 shows mousemove event is use capure, but destroy not use it, it makes remove not work.

Issue 603 is caused by this bug.
closes https://github.com/TypeCellOS/BlockNote/issues/603


YousefED:
thanks!

PR #611, opened 2024-03-05
build(deps): bump prosemirror-trailing-node to 2.0.8

LoneRifle:
Bump prosemirror-trailing-node to 2.0.8 to pick up remirror/remirror#2245, which drops @remirror/core-helpers.
The latter package was heavy, bringing in several additional dependencies, but was only used to create
an array of unique elements, and to check for the presence of elements in arrays.

Accordingly, they have been replaced in 2.0.8 with equivalent code that relies on standard built-in objects.

_This PR has been submitted through the course of work for @opengovsg_

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4574,43 +4574,8 @@
     },
     "node_modules/@remirror/core-constants": {
       "version": "2.0.2",
-      "license": "MIT"
-    },
-    "node_modules/@remirror/core-helpers": {
-      "version": "3.0.0",
-      "license": "MIT",
-      "dependencies": {
-        "@remirror/core-constants": "^2.0.2",
-        "@remirror/types": "^1.0.1",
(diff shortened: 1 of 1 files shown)


YousefED:
Thanks!

@LoneRifle I would love to hear more about how @opengovsg is using BlockNote. Exciting! Mind shooting a message on discord or to yousef@blocknotejs.org?

YousefED:
> > Thanks!
> > @LoneRifle I would love to hear more about how @opengovsg is using BlockNote. Exciting! Mind shooting a message on discord or to [yousef@blocknotejs.org](mailto:yousef@blocknotejs.org)?
> 
> We're not - a patch was made to prosemirror-trailing-node as we were dependent on it downstream. This PR is part of follow-up work to help other projects pick up the changes we have made.

Got it, thanks. Very kind!

PR #617, opened 2024-03-05
chore: Replace link to file of Architecture

NaotoCoding:
https://github.com/TypeCellOS/BlockNote/blob/main/packages/core/ARCHITECTURE.md

is now already replaced by

https://github.com/TypeCellOS/BlockNote/blob/main/packages/core/src/pm-nodes/README.md

and `ARCHITECTURE.md` is already deleted.


YousefED (approved):
Thanks!

PR #625, opened 2024-03-07
feat: Separate image, hyperlink, & table components for export

### LINKED ISSUE #608: Hyperlink, table, and image menus/toolbars default component exports
There isn't currently a good way to customize the hyperlink, table, and image menus/toolbars as we don't export the default components that are used in them. These should be refactored and exported so consumers can use them in their own custom components.

matthewlipski:
This PR separates the components for the hyperlink toolbar, image toolbar, and table handles so that they can be exported and used by consumers to customize those UI elements. Closes #608 

It also cleans up code for children of the Toolbar component, as well as some other small/misc cleanups for components. See #637 for that.


packages/react/src/components/HyperlinkToolbar/mantine/DefaultButtons/EditLinkButton.tsx
```diff
@@ -1,31 +1,22 @@
...
-
 import { ToolbarInputDropdown } from "../../../mantine-shared/Toolbar/ToolbarInputDropdown";
-import { ToolbarInputDropdownItem } from "../../../mantine-shared/Toolbar/ToolbarInputDropdownItem";
 
-export type EditHyperlinkMenuProps = {
-  url: string;
-  text: string;
-  update: (url: string, text: string) => void;
-};
+export const EditLinkMenu = (
```
YousefED, line 15: separate file per component?

packages/react/src/components/ImageToolbar/mantine/ImageToolbar.tsx
```diff
@@ -5,169 +5,100 @@ import {
...
+ * can use the default tab panels in the `DefaultTabPanels` directory or make
+ * your own using the `ImageToolbarPanel` component.
+ */
 export const ImageToolbar = <
   I extends InlineContentSchema = DefaultInlineContentSchema,
   S extends StyleSchema = DefaultStyleSchema
 >(
-  props: ImageToolbarProps<I, S>
+  props: ImageToolbarProps<I, S> & {
+    tabs?: {
```
YousefED, line 28: I'm not sure if providing a `tabs` prop as API is useful / maybe too much to maintain? IF we don't make it configurable like that, the code to add a panel is fairly easy because you re-use the default panels, right?

packages/react/src/components/ImageToolbar/mantine/ImageToolbar.tsx
```diff
@@ -5,169 +5,100 @@ import {
...
 import { ImageToolbarProps } from "../ImageToolbarProps";
+import { UploadPanel } from "./DefaultTabPanels/UploadPanel";
+import { EmbedPanel } from "./DefaultTabPanels/EmbedPanel";
 
+/**
+ * By default, the ImageToolbar component will render with default tabs.
+ * However, you can override the tabs to render by passing the `tabs` prop. You
+ * can use the default tab panels in the `DefaultTabPanels` directory or make
+ * your own using the `ImageToolbarPanel` component.
+ */
```
YousefED, line 22: Is `ImageToolbar` the right name? It doesn't really feel like a toolbar, right? (as it's a lot larger)

packages/react/src/components/ImageToolbar/mantine/ImageToolbarPanel.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const ImageToolbarPanel = forwardRef<
+  HTMLDivElement,
+  ComponentPropsWithoutRef<"div">
+>((props, ref) => {
+  const { className, children, ...rest } = props;
+
+  return (
+    <div
+      className={`bn-image-toolbar-panel${
```
YousefED, line 11: we have a function for this, right?

packages/react/src/components/TableHandles/mantine/TableHandle.tsx
```diff
@@ -2,20 +2,75 @@ import {
...
-import { MdDragIndicator } from "react-icons/md";
+import { Menu } from "@mantine/core";
+import { ReactNode, useState } from "react";
 
 import { TableHandleProps } from "../TableHandleProps";
-import { TableHandleWrapper } from "./TableHandleWrapper";
+import { MdDragIndicator } from "react-icons/md";
+import { TableHandleMenu } from "../TableHandleMenu/mantine/TableHandleMenu";
 
+// TODO: props.tableHandleMenu should only be available if no children are
```
YousefED, line 15: why?

YousefED (changes requested):
Nice. Left some comments, also, I think we wanted to streamline more inside the shared/Toolbar directory. iirc this was related to ToolbarDropdown* and ToolbarInputDropdown* etc. right? (I think at least the naming isn't clear, but I think you wanted to combine more code)

packages/react/src/components/TableHandles/mantine/TableHandle.tsx
matthewlipski: Same as side menu - if consumer implements their own drag handle (custom or using default component) they will already define the menu there so it's useless to pass it in as an additional prop.

packages/react/src/components/ImageToolbar/mantine/ImageToolbar.tsx
matthewlipski: True, maybe `ImagePanel`?

packages/react/src/components/ImageToolbar/mantine/ImageToolbar.tsx
matthewlipski: I wanted to make the API the same as other UI components, i.e. that you can pass `children` to replace the default content. But tabs are a little different since each tab panel you add needs to correspond to a tab name. So I'm not sure there's a better way of making tabs configurable.

packages/react/src/components/ImageToolbar/ImagePanelController.tsx
```diff
@@ -11,14 +11,14 @@ import { FC } from "react";
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor";
 import { useUIElementPositioning } from "../../hooks/useUIElementPositioning";
 import { useUIPluginState } from "../../hooks/useUIPluginState";
-import { ImageToolbarProps } from "./ImageToolbarProps";
-import { ImageToolbar } from "./mantine/ImageToolbar";
+import { ImagePanelProps } from "./ImagePanelProps";
+import { ImagePanel } from "./mantine/ImagePanel";
 
-export const ImageToolbarController = <
+export const ImagePanelController = <
```
YousefED, line 17: Shouldn't this still be ImageToolbarController? It controls the toolbar right, not the panel? Maybe discuss on call

YousefED (approved):

PR #629, opened 2024-03-08
LiveBlocks → Liveblocks

CTNicholas:
Hey there, nice work on the new docs, they're looking good! Just a quick update, we don't use the upper case "B" in our branding.


YousefED (approved):

YousefED:
Thanks!

PR #640, opened 2024-03-12
chore: Playwright update

matthewlipski:
Seems like updating the dependencies in 0.12.0 caused Playwright to get updated to 1.40.1. This introduced a bug where placeholders were not being applied in Firefox, causing test screenshots to be incorrect. Updating playwright to 1.42.1 seems to revert this issue.


YousefED:
the package lock change updated a lot of dependencies. I think we should do that separately for a different release. Added a commit to revert this and just updated pw

PR #641, opened 2024-03-13
fix: serialize React content with existing React tree

### LINKED ISSUE #592: Copying to clipboard of react custom blocks that use context
Copying the React Alert demo block doesn't work, because it's render function relies on React Context (for dropdown menu).

Overriding `toExternalHtml` doesn't suffice, because we also call `toInternalHtml` to set the blocknote data on the clipboard. However, `toInternalHTML` always calls `render()`, which depends on the React Context.

We need to review to usage of toInternalHTML and see if it needs to be customizable

### LINKED ISSUE #623: Custom block is not draggable
**Describe the bug**
<what's going wrong!?>
Can't change the position of the custom block using the drag handle button.

**To Reproduce**

https://www.blocknotejs.org/docs/custom-schemas/custom-blocks#:~:text=Custom%20Blocks-,Custom%20Block%20Types,-In%20addition%20to

open the link, under Custom Block Types, inside the demo box there is a custom block that is not draggable, I have this problem on my project too.

YousefED:
closes #623
closes #592

Todo:

- [x] Add unit test, probably to `htmlConversion.test.ts`


YousefED (commented):
Thanks @matthewlipski ! Ready to be merged imo

PR #642, opened 2024-03-13
fix: Formatting toolbar appearance on fade out

matthewlipski:
Because formatting toolbar buttons manage their own state, they check whether they should be shown based on the block that the text cursor is in. This can be a problem if the user clicks a block of a different type to close the toolbar, as it can cause the toolbar state to change as it's fading out, which looks bad. This PR fixes that by making the plugin figure out which block(s) are selected instead, and including them in the plugin state for the toolbar buttons.

The difference between this approach and using `useSelectedBlocks` is that the plugin will NOT update the selected blocks if `state.show` is `false`, so the selected blocks are only updated when the formatting toolbar itself is shown/updated.

This is similar to how & why we pass the hovered block in the side menu/drag handle menu, so I think this pattern makes sense.


matthewlipski:
Fixed with #645

PR #645, opened 2024-03-14
fix: Formatting toolbar appearance on fade out using innerHTML

YousefED:
Alternative to #642.

@matthewlipski curious to here your thoughts + can you try to find any bugs in this approach?


matthewlipski:
> Alternative to #642.
> 
> @matthewlipski curious to here your thoughts + can you try to find any bugs in this approach?

Looks good! Couldn't find any situations where it breaks. However, I'm still a bit hesitant to merge this vs #642 since using `dangerouslySetInnerHTML` seems like a bit of a hack. #642 on the other hand introduces a lot of extra code just to deal with animations, but also fixes the issue at the root imo. Not sure which one is more sensible to use tbh

PR #665, opened 2024-03-27
feat: add option to give control over trailing block

### LINKED ISSUE #631: Empty block is unnecessarily created
**Describe the bug**
Empty block is unnecessarily created. I'm wondering why is this and whether there is a way to avoid this?

**To Reproduce**
1. Visit https://www.blocknotejs.org/examples/basic/minimal
2. Focus the editor
3. Press "a"
You'll see that a new empty block is added under the "a" (you can press the Down arrow to go there)

**Misc**
- Node version: 20
- Package manager: bun
- Browser: Chrome

35C4n0r:
closes #631 
/claim #631


YousefED (commented):
I don't see why this should be part of the blockschema. For now #679 seems like a better approach

YousefED:
closed in favor of #679

PR #669, opened 2024-03-29
Fix typescript build

otang:
We were getting build errors in our project, I think due to syntax error in the generic types

image


otang:
Closing because I see this is solved by upgrading to typescript 5.0+

vincenzodomina:
Can confirm that changing typescript from v4.9.5 to v5.4 solved the issue for me.

I had similar build issues with the types mentioned above:

> next build

  ▲ Next.js 14.2.3
  - Environments: .env.local

   Creating an optimized production build ...
 ✓ Compiled successfully
   Linting and checking validity of types  ..Failed to compile.

./node_modules/@blocknote/react/types/src/schema/ReactBlockSpec.d.ts:21:46
Type error: Type parameter declaration expected.

  19 |     children: ReactNode;
  20 | }): import("react/jsx-runtime").JSX.Element;
> 21 | export declare function createReactBlockSpec<const T extends CustomBlockConfig, const I extends InlineContentSchema, const S extends StyleSchema>(blockConfig: T, blockImplementation: ReactCustomBlockImplementation<T, I, S>): {
     |                                              ^
  22 |     config: T;
  23 |     implementation: import("@blocknote/core").TiptapBlockImplementation<T, any, InlineContentSchema, StyleSchema>;
  24 | };

PR #671, opened 2024-03-29
Improve type compatability with TypeScript < 5.0

otang:
image


YousefED:
The `const` is there to improve type inference. For now, we recommend upgrading typescript instead

otang:
@YousefED tha you for the response. Unfortunately upgrading TypeScript results in many conflicts with our existing packages so it's not a viable option at the moment. Sadly this prevents us from using BlockNote.

PR #678, opened 2024-04-02
chore: Update Mantine

matthewlipski:


YousefED (changes requested):
I think there are more references to mantine in package.json files, please check all

YousefED (approved):

PR #679, opened 2024-04-02
feat: option to disable trailing block

Kashifwahaj:
```js
//Creates a new editor instance.
    trailingBlock: true,
  });

  // Renders the editor instance using a React component.
  return <BlockNoteView editor={editor} />;
```
  
  So user need to pass `trailingBlock` boolean option to enable trailing Node.

/claim #631

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -133,6 +133,8 @@ export type BlockNoteEditorOptions<
 
   // tiptap options, undocumented
   _tiptapOptions: Partial<EditorOptions>;
+
+  trailingBlock?: boolean;
 };
 
 const blockNoteTipTapOptions = {
@@ -179,6 +181,8 @@ export class BlockNoteEditor<
   >;
 
   public readonly uploadFile: ((file: File) => Promise<string>) | undefined;
+
+  public trailingBlock?: boolean = false;
 
   public static create<
     BSchema extends BlockSchema = DefaultBlockSchema,
@@ -228,6 +232,7 @@ export class BlockNoteEditor<
     this.blockImplementations = newOptions.schema.blockSpecs;
     this.inlineContentImplementations = newOptions.schema.inlineContentSpecs;
     this.styleImplementations = newOptions.schema.styleSpecs;
+    this.trailingBlock = newOptions.trailingBlock;
 
     this.formattingToolbar = new FormattingToolbarProsemirrorPlugin(this);
     this.linkToolbar = new LinkToolbarProsemirrorPlugin(this);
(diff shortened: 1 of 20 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -180,6 +182,8 @@ export class BlockNoteEditor<
 
   public readonly uploadFile: ((file: File) => Promise<string>) | undefined;
 
+  public trailingBlock?: boolean = false;
```
YousefED, line 185: Can we remove this member, and just pass the option to `getBlockNoteExtensions` on line 249, similar to the other options?

packages/core/src/editor/BlockNoteSchema.ts
```diff
@@ -66,6 +66,8 @@ export class BlockNoteSchema<
      * A list of custom Styles that should be available in the editor.
      */
     styleSpecs?: SSpecs;
+
+    trailingNode?: boolean;
```
YousefED, line 70: why is this added here?

YousefED (commented):
- can you make the default `true`? (inline with current behaviour)
- can you also update the documentation where the other editor options are explained?

Kashifwahaj:
Got it, I will make the default value `true` and update the documentation also , -I am on it.

packages/core/src/editor/BlockNoteEditor.ts
Kashifwahaj: Yes, it's better and resulted in less code. Thanks! 👍

packages/core/src/editor/BlockNoteSchema.ts
Kashifwahaj: It was necessary for the 'typescript' warning, but with 'getBlockNoteExtensions', it's no longer needed, so I removed it in the new commits.

Kashifwahaj:
@YousefED 

- [x] Now, the default behavior is `true`. Therefore, if a user explicitly passes `trailingBlock: false`, it will disable the automatic creation of a trailing new block.
- [x] Updated documentation in `/docs/editor-basics/setup#usecreateblocknote-hook` to include the explanation of the new option.

Screenshot 2024-04-09 at 7 28 23 AM


Please let me know if any change is needed.

Thanks

Kashifwahaj:
@YousefED Hi, Please Let me know if any improvement or changes required for this PR.

YousefED:
Looks good, awesome!

hampak:
Hello, @YousefED 

Is #679 going to be released in version 0.12.5?

Thank you for your hard work :)

hampak:
Thank you for this in the latest release :D

PR #696, opened 2024-04-08
fix: ShadCN color styles button & form components

matthewlipski:


packages/shadcn/src/components/ui/dropdown-menu.tsx
```diff
@@ -57,17 +57,15 @@ const DropdownMenuContent = React.forwardRef<
...
-      ref={ref}
-      sideOffset={sideOffset}
-      className={cn(
-        "z-50 min-w-[8rem] overflow-hidden rounded-md border bg-popover p-1 text-popover-foreground shadow-md data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
-        className
-      )}
-      {...props}
-    />
-  </DropdownMenuPrimitive.Portal>
+  <DropdownMenuPrimitive.Content
```
YousefED, line 60: why was this changed?

packages/shadcn/src/components/ui/dropdown-menu.tsx
```diff
@@ -96,17 +94,17 @@ const DropdownMenuCheckboxItem = React.forwardRef<
   <DropdownMenuPrimitive.CheckboxItem
     ref={ref}
     className={cn(
-      "relative flex cursor-default select-none items-center rounded-sm py-1.5 pl-8 pr-2 text-sm outline-none transition-colors focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
+      "relative flex cursor-default select-none items-center justify-between rounded-sm gap-2 py-1.5 px-2 w-full text-sm outline-none transition-colors focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
```
YousefED, line 97: similarly, do we need these changes? that will make it more difficult for others to pass in their own shadcn components

packages/shadcn/src/input/TextInput.tsx
```diff
@@ -0,0 +1,36 @@
+import { TextInputProps } from "@blocknote/react";
+
+import { Input } from "../components/ui/input";
+import { Label } from "../components/ui/label";
+
+export const TextInput = (props: TextInputProps) => {
+  if (!props.label) {
+    return (
+      <Input
+        name={props.name}
```
YousefED, line 10: add aria-label for this case

packages/shadcn/src/components/ui/form.tsx
```diff
@@ -0,0 +1,176 @@
+import * as React from "react";
```
YousefED, line 1: for the new components like form, similar to my questions above; did you had to make any changes to the code of the components? We should keep that to a minimum where possible, or at least add comments while doing so, so that we can enable the functionality for people to add in there own components

(I should also check which changes I made and comment those, btw. Were quite little, but had to do some)

YousefED (commented):
nice! Main question is ho

PR #689, opened 2024-04-05
refactor: Extract Mantine & Ariakit packages

matthewlipski:

PR #684, opened 2024-04-03
refactor: Extract image panel components

matthewlipski:

PR #698, opened 2024-04-11
refactor: UI libs CSS+API refactor

matthewlipski:
TODO:

- [X] Suggestion Menu, Side Menu, Table Handles in component context
- [x] Fix Mantine Formatting Toolbar color styles button dropdown
- [x] Ariakit remaining styling (image replace/caption buttons, link create button)
- [x] ShadCN styling
- [x] Set base `bn-` classes in a more elegant way
- [x] Fix `BlockNoteViewRaw` not having schema type args

Thoughts:

- I know that using a select for the block type dropdown is most semantically correct, but I think it's a bit redundant.
    - Adds an additional field to the API.
    - Functionally the same as color style button, i.e. items run editor commands.
    - Cannot pass a class name to target and dropdown separately as the select is treated as 1 element.
- Since we want to minimize any changes made to the base ShadCN components, it means that some parts of the UI don't look quite right. These are fairly minor changes, just adjusting some element spacing, but we have 3 options for handling this:
    1. We leave the styling as-is, since we expect people to provide their own pre-styled ShadCN components anyway.
    2. We change the Tailwind styles within the base ShadCN components.
    3. We add an additional CSS file to with said styling fixes.
- How should we handle icons? Currently these are set by the `react` package without possibility of overriding them. This also means that `react-icons` is a dependency in the `react` package.
- We probably want to make the components in the context as generic as possible, as right now there's an ugly mix of specific and generic ones.


packages/shadcn/src/toolbar/Toolbar.tsx
```diff
@@ -16,55 +15,52 @@ import {
...
       ) : (
         <Toggle
-          onPressedChange={onClick}
-          pressed={isSelected}
-          disabled={isDisabled}
-          ref={ref}
-          data-state={isSelected ? "on" : "off"}
-          data-disabled={isDisabled}
-          {...rest}>
+          onClick={props.onClick}
```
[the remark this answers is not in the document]
matthewlipski: Actually I think here it's better to use `onClick`. Seems like `onPressedChange` is more meant to be used to update a `pressed` state when using the Toggle as a controlled component. In our case, the state is already managed by a parent component which passes it using `props.isSelected`, so `onClick` is more applicable.

matthewlipski:
To make things easier, I'm implementing feedback from #684, #689, and #696 here since this PR has the biggest changes, so everything can be in one place.

packages/ariakit/src/ariakitStyles.css
```diff
@@ -0,0 +1,813 @@
+/* This file contains styles copied directly from Ariakit's website, and is
+ responsible for the majority of the styling. */
+
+.bn-container {
```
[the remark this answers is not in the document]
YousefED: also, let's discuss if we want `button`, `bn-ak-button` or css module style generated clasnames

packages/mantine/src/suggestionMenu/SuggestionMenu.tsx
```diff
@@ -1,7 +1,10 @@
 import { Children, useMemo } from "react";
 
-import { useComponentsContext } from "../../../editor/ComponentsContext";
-import { DefaultReactSuggestionItem, SuggestionMenuProps } from "../types";
+import { useComponentsContext } from "../../../react/src/editor/ComponentsContext";
```
YousefED, line 3: let's discuss next steps re. suggestionmenu

packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
```diff
@@ -116,7 +115,7 @@ export function SuggestionMenuController<
         closeMenu={callbacks.closeMenu}
         clearQuery={callbacks.clearQuery}
         getItems={getItems!}
-        suggestionMenuComponent={suggestionMenuComponent || SuggestionMenu}
+        suggestionMenuComponent={suggestionMenuComponent || (() => <div />)}
```
YousefED, line 118: why the empty div?

packages/react/src/editor/ComponentsContext.tsx
```diff
@@ -0,0 +1,564 @@
...
+};
+
+export type Components = {
+  [Components in keyof Omit<ComponentProps, "Generic">]: {
+    [Component in keyof ComponentProps[Components]]: ComponentType<
+      ComponentProps[Components][Component]
+    >;
+  };
+} & {
+  Generic: {
```
YousefED, line 361: why are these defined separately (generic ones)?

packages/react/src/editor/ComponentsContext.tsx
```diff
@@ -0,0 +1,564 @@
...
+      >;
+    };
+  };
+};
+
+export const ComponentsContext = createContext<Components | undefined>(
+  undefined
+);
+
+// TODO: We should be able to apply the class names in a smarter way
```
YousefED, line 374: seems a bit cumbersome indeed, let's discuss

packages/shadcn/src/index.tsx
```diff
@@ -24,41 +36,263 @@ import {
...
+  TabsList: typeof ShadCNTabs.TabsList;
+  TabsTrigger: typeof ShadCNTabs.TabsTrigger;
+  Toggle: typeof ShadCNToggle.Toggle;
+  Tooltip: typeof ShadCNTooltip.Tooltip;
+  TooltipContent: typeof ShadCNTooltip.TooltipContent;
+  TooltipProvider: typeof ShadCNTooltip.TooltipProvider;
+  TooltipTrigger: typeof ShadCNTooltip.TooltipTrigger;
 };
 
+export const createShadCNComponentContext = (
```
YousefED, line 113: this seems a but cumbersome (all the prop passing), isn't this easier with a separate context?

examples/01-basic/01-minimal/App.tsx
```diff
@@ -1,16 +1,26 @@
...
 import { useCreateBlockNote } from "@blocknote/react";
-import "@blocknote/react/style.css";
 import { uploadToTmpFilesDotOrg_DEV_ONLY } from "@blocknote/core";
 
-import { BlockNoteView } from "@blocknote/mantine";
-// import { BlockNoteView } from "@blocknote/ariakit";
+// import { BlockNoteView } from "@blocknote/mantine";
+// import "@blocknote/mantine/styles.css";
+import { BlockNoteView } from "@blocknote/ariakit";
+import "@blocknote/ariakit/style.css";
```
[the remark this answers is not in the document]
matthewlipski: TODO: Add playwright test w screenshots for ShadCN/Ariakit

PR #718, opened 2024-05-02
feat: accessibility improvements

YousefED:
After a consultation with accessibility experts, this PR implements some important improvements in terms of accessibility:

- Better keyboard navigation:
-- the editor can be exited with "esc"
-- menus (formatting toolbar, link menu, image panel) can be tabbed into with "tab", exited with "esc"
- the suggestion menu is now a combobox / listbox aria element of the editor. The active item is marked using virtual focus (activedescendent), which should bring better support for screenreaders. (we need to test this a bit more extensively)
- All buttons have the right `aria-label` attributes, toggle buttons have `aria-pressed`

As part of this PR, we also:
- moved responsibility of closing menus from the plugins to Floating-UI, with the `useDismiss` hook
- cleaned up some CSS
- added stronger typescript support which enforces all props to be used (or at least destructured) when implementing a UI layer. This makes sure we get compile errors when new props are being added, but not implemented by all UI layers

This should be a significant improvement in terms of accessibility! However, we should also recognize this is a subject that's never "done" and continuously improve here. Next steps could include:

- Extensive manual testing using screen readers and / or a test group
- Automated testing, ideally integrated into CI workflow
- Implement keyboard navigation for drag handle menu
- Add a "hint sheet" that shows keyboard shortcuts / accessibility tips to end users
- Validate / improve color contrasts
- Improve accessibility / keyboard nav for lists, tables, and images


examples/03-theming/03-theming-css/styles.css
```diff
@@ -4,6 +4,13 @@
...
-.bn-container[data-theming-css-demo] .bn-slash-menu .mantine-Menu-item[data-hovered] {
+.bn-container[data-theming-css-demo]
+  .bn-slash-menu
+  .bn-container
+  .mantine-Menu-item[aria-selected="true"],
+.bn-container[data-theming-css-demo]
+  .bn-slash-menu
+  .bn-container
+  .mantine-Menu-item:hover {
   background-color: blue;
```
matthewlipski, line 15: Doesn't seem to be working atm

packages/mantine/src/suggestionMenu/SuggestionMenuItem.tsx
```diff
@@ -1,65 +1,45 @@
...
-    badge,
-    // aliases,
-    // onItemClick,
-    isSelected,
-    setSelected,
-    onClick,
-  } = props;
+  const { className, isSelected, onClick, item, id, ...rest } = props;
 
-  const handleMouseLeave = useCallback(() => {
```
matthewlipski, line 24: How come we don't need this anymore?

packages/react/src/components/FormattingToolbar/FormattingToolbarController.tsx
```diff
@@ -72,13 +72,21 @@ export const FormattingToolbarController = (props: {
   const state = useUIPluginState(
     editor.formattingToolbar.onUpdate.bind(editor.formattingToolbar)
   );
-  const { isMounted, ref, style } = useUIElementPositioning(
+
+  const { isMounted, ref, style, getFloatingProps } = useUIElementPositioning(
```
matthewlipski, line 76: What are the `getFloatingProps` for?

packages/react/src/editor/BlockNoteContext.ts
```diff
@@ -8,13 +8,14 @@ import {
...
 } from "@blocknote/core";
-import { createContext, useContext } from "react";
+import { createContext, useContext, useState } from "react";
 
 type BlockNoteContextValue<
   BSchema extends BlockSchema = DefaultBlockSchema,
   ISchema extends InlineContentSchema = DefaultInlineContentSchema,
   SSchema extends StyleSchema = DefaultStyleSchema
 > = {
+  setContentEditableProps?: ReturnType<typeof useState<Record<string, any>>>[1];
```
matthewlipski, line 18: This typing looks a bit hacky

packages/react/src/hooks/useUIElementPositioning.ts
```diff
@@ -1,25 +1,29 @@
...
-): UiComponentPosition {
+) {
   const { refs, update, context, floatingStyles } = useFloating({
     open: show,
     ...options,
   });
 
   const { isMounted, styles } = useTransitionStyles(context);
 
+  const dismiss = useDismiss(context, {});
```
matthewlipski, line 23: Could use some comments to briefly explain these I think

matthewlipski (commented):
Looks good! While doing some manual testing I found a few behaviours that are still not quite there:
- Formatting toolbar loses tab index after pressing enter (maybe intentional?)
- Pressing escape in link creation & caption menus closes whole formatting toolbar instead of just those menus
- Pressing enter on colors drag handle menu button closes menu (mantine)
- Tabbing order in image toolbar not working (mantine, though it has some quirks in the others too)
- Tooltips don’t show up when hovering using keyboard (mantine)
- Formatting toolbar tabbing not working (ariakit)

packages/react/src/components/FormattingToolbar/FormattingToolbarController.tsx
YousefED: will add a comment

packages/react/src/editor/BlockNoteContext.ts
YousefED: added comment

examples/03-theming/03-theming-css/styles.css
YousefED: fixed, nice spot

packages/mantine/src/suggestionMenu/SuggestionMenuItem.tsx
YousefED: I've disabled this because I figured copying the ariakit combobox behavior (separate focus and hover states) would be fine.

But then I noticed for menus they don't have this behavior, so I started a conversation with the ariakit maintainer:

image

Perhaps bringing it back is a good idea, but let's do this as a separate PR or if more people experience this as an issue, as we're a bit limited on time atm

PR #719, opened 2024-05-02
fix: match query anywhere to filter slash menu items

shahvicky:
Notion menu item filtering 
image

Current BlockNote filtering only matches start of the word. This PR fixes it to match query anywhere in the item or alias text


YousefED:
thanks!

PR #726, opened 2024-05-06
feat: Add Chinese i18n support

lujdong:
I think that i18n should be configured by the developers themselves according to the documentation examples, but for those who wish to use it directly, they undoubtedly have one more option.


YousefED:
Thanks @lujdong ! Curious, what are you building with BlockNote?

linxianxi:
There is a warning in the console, can you take a look?
https://github.com/TypeCellOS/BlockNote/issues/779

lujdong:
> Thanks @lujdong ! Curious, what are you building with BlockNote?

I am using Blocknote to build an AI note similar to Notion, and it has given me a great deal of help.

PR #729, opened 2024-05-06
feat: Checkbox/TODO list item block

### LINKED ISSUE #18: TODO items
Add support for TODO items:

- TODO items should show a checkbox on the side
- The checked state should be stored in the prosemirror state
- If possible, content of checked todo items should appear strikethrough (I think we can do this directly from CSS)
- show "To-do" as placeholder

We currently store a listType attribute on Blocks. On first thought, we could:
a. use `listType="todo-checked"` and `listType="todo-unchecked"`
b. use `listType="todo"` and `todo-checked="true"` as attributes

Option B seems cleaner, but option A has other benefits (we don't need to clean for "invalid" Prosemirror states, for example, what does `listType="li"` and `todo-checked` mean? In that case `todo-checked` should be removed. What do you think is the best architecture for this?

matthewlipski:
This PR adds a TODO list item style block to the editor. Also includes items for the block type dropdown and slash menu.

Closes #18 

TODO: We should merge the `releases` branch into main first
TODO: Figure out what we should render to external HTML, since there aren't set standards and different platforms do it in different ways.


packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -0,0 +1,208 @@
...
+          }
+
+          chain()
+            .BNUpdateBlock(state.selection.from, {
+              type: "checkListItem",
+              props: {
+                checked: false as any,
+              },
+            })
+            // Removes the "-", "+", or "*" character used to set the list.
```
YousefED, line 55: todo

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -0,0 +1,208 @@
...
+          }
+
+          chain()
+            .BNUpdateBlock(state.selection.from, {
+              type: "checkListItem",
+              props: {
+                checked: true as any,
+              },
+            })
+            // Removes the "-", "+", or "*" character used to set the list.
```
YousefED, line 73: todo

packages/core/src/editor/Block.css
```diff
@@ -184,6 +184,24 @@ NESTED BLOCKS
   gap: 1.2em;
 }
 
+/* Checked */
```
YousefED, line 187: makes me think, would it even be possible to use a label for the inline content + checkbox input? That would be the most "semantic" / accessible way, but not sure if that only introduces more complexity in terms of keyboard handling etc

YousefED:
Other todos:
[] Placeholder text
[] make UI same as notion (strike when completed, etc)

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -0,0 +1,208 @@
...
+      // Checkbox only.
+      {
+        tag: "input",
+        getAttrs: (element) => {
+          if (typeof element === "string") {
+            return false;
+          }
+
+          if ((element as HTMLInputElement).type === "checkbox") {
+            return { checked: (element as HTMLInputElement).checked };
```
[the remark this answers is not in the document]
matthewlipski: I suppose we don't, but we render out all props to `data-*` attributes since they can come in handy when styling.

packages/core/src/editor/Block.css
matthewlipski: Tested this and while keyboard navigation is no problem, clicking the text (which is now a label) toggles the checkbox, which is default browser behaviour so I don't think there's an easy workaround (assuming the checkbox `id` and label `for` attributes match). Also not sure if the semantics are correct actually, since the `contentDOM` is more like a second input field rather than a label.

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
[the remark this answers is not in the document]
matthewlipski: Also worth noting that checkbox `input`s don't render their checked state to any attributes either.

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -0,0 +1,208 @@
...
+          {
+            type: "checkListItem",
+            props: {},
+          }
+        );
+      },
+    };
+  },
+
+  parseHTML() {
```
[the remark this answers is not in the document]
matthewlipski: Added tests, but the intended output is a point of contention. See the TODO in `CheckListItemBlockContent.ts`.

packages/core/src/editor/Block.css
roperrone: @matthewlipski you can check how Tiptap handles that in their own extension: https://tiptap.dev/docs/editor/api/nodes/task-item

Clicking the text shouldn't trigger the checkbox, as this can become annoying for the user

YousefED:
[] make sure we can't change checked status when doc is not editable

packages/core/src/editor/Block.css
aoztanir: @matthewlipski Is there any way we can use the mantine checkbox with the indigo colorway. I think it looks a lot better than the one in the current preview. Would really appreciate this change. I am currently using a custom schema block that does the todo list with the mantine checkbox item. Here's what it looks like:

Screenshot 2024-05-14 at 10 29 29 PM

aoztanir: I'm also happy to paste my custom code even though most if it is a crummy fix to the block not continuing on a new line. 



`'use client';
import { defaultProps, insertOrUpdateBlock } from '@blocknote/core';
import { createReactBlockSpec } from '@blocknote/react';
import { Box, Checkbox, Flex } from '@mantine/core';
import { IconCheck } from '@tabler/icons-react';
import { useEffect } from 'react';
import CategorySelect from './CategorySelect';
import './styles.css';

// The types of alerts that users can choose from.

export const insertCheckBox = (editor) => ({
  title: 'Todo',
  onItemClick: () => {
    insertOrUpdateBlock(editor, {
      type: 'checkbox',
    });
  },
  aliases: ['checkbox', 'todo', 'list', 'bullets', 'to-do', 'tasks', 'taskboard', 'board'],
  subtext: 'Plan out your tasks using a checkbox todo list',
  //   group: 'Other',
  icon: <IconCheck size="20" />,
});
// The Alert block.
export const CheckBoxBlock = createReactBlockSpec(
  {
    type: 'checkbox',
    propSchema: {
      type: {
        default: false,
        values: [true, false],
      },
    },
    content: 'inline',
  },
  {
    render: (props) => {
      useEffect(() => {
        //LOGIC FOR NEW CHECKBOX ON ENTER
        // console.log('hello');
        const handleKeyDown = (event) => {
          if (event.key == 'Tab') {
            event.preventDefault();
            return;
          }
          if (event.key === 'Enter') {
            if (props?.editor?.getTextCursorPosition()?.prevBlock?.id != props.block.id) {
              return;
            }

            //!OLD STUFF BELOW
            var thisFunctionalComponentBlock = props.editor.getTextCursorPosition().prevBlock;

            if (thisFunctionalComponentBlock.content.length == 0) {
              props.editor.updateBlock(props.editor.getTextCursorPosition().prevBlock, {
                type: 'paragraph',
                // props: { type: event.currentTarget.checked },
              });

              props.editor.setTextCursorPosition(
                props.editor.getTextCursorPosition().prevBlock,
                'start'
              );
              if (props.editor.getTextCursorPosition().nextBlock.content.length == 0) {
                props.editor.removeBlocks([props.editor.getTextCursorPosition().nextBlock]);
              }
              return;
            }

            props.editor.updateBlock(props.editor.getTextCursorPosition().block, {
              type: 'checkbox',
              // props: { type: event.currentTarget.checked },
            });
          }
        };

        document.addEventListener('keydown', handleKeyDown);

        return () => {
          document.removeEventListener('keydown', handleKeyDown);
        };
      }, []); // empty dependency array to run effect only once

      return (
        <Flex
          // align="center"

          style={{ textWrap: 'wrap', wordBreak: 'break-word' }}
          gap="xs"
          data-alert-type={props.block.props.type}
        >
          {/* <div style={{ textWrap: 'wrap', wordBreak: 'break-word' }} contentEditable={false}> */}
          <div contentEditable={false}>
            <Checkbox
              checked={props.block.props.type}
              // value={props.block.props.type}
              mt="2"
              ml="2"
              onChange={(event) =>
                props.editor.updateBlock(props.block, {
                  type: 'checkbox',
                  props: { type: event.currentTarget.checked },
                })
              }
              radius="1"
              color="indigo"
            />
          </div>

          {/* <Icon
              className={'alert-icon'}
              data-alert-icon-type={props.block.props.type}
              size={32}
            /> */}
          {/* </div> */}
          {/*Rich text field for user to type in*/}
          <Box
            mt="1"
            className={'inline-content'}
            style={{ textDecoration: props.block.props.type ? 'line-through' : '' }}
            ref={props.contentRef}
          />

          {/* <Box contentEditable={false} ml="auto" mr="0">
            <CategorySelect />
          </Box> */}
        </Flex>
      );
    },
  }
);
`

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -232,6 +228,13 @@ const checkListItemBlockContent = createStronglyTypedTiptapNode({
         this.options.domAttributes?.inlineContent || {}
       );
 
+      if (typeof getPos !== "boolean") {
+        const label =
+          "label-" + this.editor.state.doc.resolve(getPos()).node().attrs.id;
```
YousefED, line 233: can you comment (in code) or fix, why we can't just use node.attrs.id? (but need resolve + getpos)

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -193,35 +201,23 @@ const checkListItemBlockContent = createStronglyTypedTiptapNode({
       if (node.attrs.checked) {
         checkbox.setAttribute("checked", "");
       }
-      checkbox.disabled = !editor.isEditable;
 
       const changeHandler = () => {
-        if (typeof getPos === "boolean") {
-          return;
+        if (!editor.isEditable) {
+          checkbox.checked = !checkbox.checked;
```
YousefED, line 207: Is this necessary or can we also use preventDefault or return false to prevent the change?

if it's necessary, please comment in code why we need this (it's a typical piece of code that might raise questions / confuse ourselves later!)

packages/core/src/blocks/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
```diff
@@ -17,6 +17,7 @@ const BulletListItemBlockContent = createStronglyTypedTiptapNode({
   name: "bulletListItem",
   content: "inline*",
   group: "blockContent",
+  priority: 90,
```
YousefED, line 20: still needed?

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -0,0 +1,258 @@
...
+
+const checkListItemBlockContent = createStronglyTypedTiptapNode({
+  name: "checkListItem",
+  content: "inline*",
+  group: "blockContent",
+  addAttributes() {
+    return {
+      checked: {
+        default: false,
+        // instead of "level" attributes, use "data-level"
```
YousefED, line 27: fix

packages/core/src/pm-nodes/BlockContainer.ts
```diff
@@ -671,7 +675,11 @@ export const BlockContainer = Node.create<{
             if (!blockEmpty) {
               chain()
                 .deleteSelection()
-                .BNSplitBlock(state.selection.from, selectionAtBlockStart)
+                .BNSplitBlock(
+                  state.selection.from,
+                  selectionAtBlockStart,
+                  selectionAtBlockStart
```
YousefED, line 681: why was this added? (the keepprops system)? (maybe comment as well?)

YousefED (commented):
awesome ❤️

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -204,6 +204,9 @@ const checkListItemBlockContent = createStronglyTypedTiptapNode({
 
       const changeHandler = () => {
         if (!editor.isEditable) {
+          // This seems like the most effective way of blocking the checkbox
+          // from being toggled, as event.preventDefault() does not stop it for
+          // "click" or "change" events.
```
YousefED, line 209: shouldn't we return here as well? because we still call updateBlock now?

YousefED (approved):
one q, rest looks good to merge!

PR #730, opened 2024-05-06
fix: Key inputs in blocks without content

matthewlipski:
This PR improves how keystrokes are handled in non-editable blocks like images, so that more edge cases are covered (including special characters like ś and Enter).

A more robust approach might be to filter out transactions in which a no content block is selected before, and replaced with a paragraph block after (will likely need additional filtering e.g. to not filter out `updateBlock` calls). However, I think the current implementation gets us 90% of the way there with much less complexity so I don't think it's worth looking into.

TODO: We should merge the `releases` branch into main first

### DIFF
--- a/packages/core/src/editor/BlockNoteExtensions.ts
+++ b/packages/core/src/editor/BlockNoteExtensions.ts
@@ -76,7 +76,7 @@ export const getBlockNoteExtensions = <
     UniqueID.configure({
       types: ["blockContainer"],
     }),
-    HardBreak,
+    HardBreak.extend({ priority: 10 }),
     // Comments,
 
     // basics:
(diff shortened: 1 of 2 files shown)


YousefED:
What are the events that we do want to allow? I suppose things like tab, arrows, etc.?

For the alternative solution, perhaps, we could use an approach like this? https://github.com/ueberdosis/tiptap/issues/181#issuecomment-1833401187, and make sure to only do this when a NodeSelection is active. However, indeed we would need to be able to distinguish from keyboard vs programmatic / collaboration events

matthewlipski:
> What are the events that we do want to allow? I suppose things like tab, arrows, etc.?
> 
> For the alternative solution, perhaps, we could use an approach like this? [ueberdosis/tiptap#181 (comment)](https://github.com/ueberdosis/tiptap/issues/181#issuecomment-1833401187), and make sure to only do this when a NodeSelection is active. However, indeed we would need to be able to distinguish from keyboard vs programmatic / collaboration events

I've updated the code to handle more edge cases and when comparing the UX to Notion's I think it's quite close. Don't think it's worth the hassle of figuring out which transaction comes from where tbh, since the keyboard input approach gets us the majority of the way there while adding very little complexity.

YousefED (approved):

PR #752, opened 2024-05-18
feat: add i18n support for pl-PL and is-IS

Dmkk01:
- Added `is.ts` and `pl.ts` i18n files.


YousefED:
Thanks @Dmkk01 ! We'll get this in the next release. What are you building with BlockNote?

Dmkk01:
I am considering adding a notes-like system to some of my existing applications. I previously tried using Editor.js, but found the experience with React to be cumbersome, particularly in terms of styling and editor initialization.

PR #758, opened 2024-05-21
fix: Add file button accessibility

matthewlipski:
This PR changes the element type of file blocks' add block button to a `button` from a `div` and makes the file panel open on Enter. It also removes the blue border from selected blocks when the editor is not in focus.


packages/core/src/extensions/NonEditableBlocks/NonEditableBlockPlugin.ts
```diff
@@ -34,6 +34,30 @@ export const NonEditableBlockPlugin = () => {
           }
           // Checks if key press is Enter
           if (event.key === "Enter") {
+            // Kind of a hacky way to ensure that pressing Enter when a file
```
YousefED, line 37: Hmm, this seems a bit hacky. I'm not a proponent of merging this as it makes the codebase worse for a minor win. We should invest more time in finding a cleaner solution (if we want to prioritize this(

YousefED (changes requested):

PR #772, opened 2024-05-28
chore: upgrade deps

YousefED:
upgrades tiptap, prosemirror, mantine and other deps. This should also fix a compatibility issue with later versions of prosemirror-model (types changed at https://github.com/ProseMirror/prosemirror-model/blob/master/CHANGELOG.md#1200-2024-04-08)


sradu:
I'm irrationally excited for NodePos support.

PR #776, opened 2024-05-28
feature: Added Configuration for Link Extension

EvanParden:
This pull request enhances the getBlockNoteExtensions function to provide more flexibility in configuring the Link extension based on the linkOptions provided in the editor options.


YousefED:
Thanks @EvanParden. I might be more in favor of an approach where we can easily, in a generic way, replace the built in extensions with your own / customized ones. In this case, the ability to remove the default Link extension from being added and add your own one with a custom configuration.

For now this works though and we can get it merged, but if you want to continue on this work I'd encourage the above approach. wdyt?

YousefED (approved):

PR #782, opened 2024-05-29
fix: Link keyboard shortcut and remove warning

### LINKED ISSUE #627: Error: mark not found in styleschema link
**Describe the bug**
When link is added, error is shown in the console and link is not highlighted.

Screenshot 2024-03-08 at 10 40 34 am

**To Reproduce**
See readme here: https://github.com/moroshko/blocknote-issue

**Misc**
- Node version: 20
- Package manager: bun
- Browser: Chrome

matthewlipski:
Closes #627


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -777,7 +777,6 @@ export class BlockNoteEditor<
     for (const mark of marks) {
       const config = this.schema.styleSchema[mark.type.name];
       if (!config) {
-        console.warn("mark not found in styleschema", mark.type.name);
```
YousefED, line 780: This warning is still valid I think? Maybe we need to just exclude the case for Links? (because links don't appear in the styleschema but are an IC node in blocknote)

YousefED (changes requested):

YousefED (approved):

PR #783, opened 2024-05-29
fix: Issues when editor is non-editable

### LINKED ISSUE #707: Image block editable bug
**Describe the bug**
BlockNoteView is configured as `editable={false}`. The image block with no embedded link, which is presented as an `Add Image` button, is still accessible and allowing user to input a link.
 
**To Reproduce**
1. Add a image block when BlockNoteView  is  `editable={true}`
2. Set `editable={false}`
3. Refresh and click on  image block, which is in `Add Image` button state.
4. Able to embed link, which is not an allowed and expected behavior.

**Misc**
- Node version: v20.12.2
- BlockNote version: 0.12.4
- Package manager: pnpm
- Browser: chrome 123.0.6312.107

### LINKED ISSUE #766: Even if editable=false, copy and paste will paste strings
**Describe the bug**

- BlockNoteView in editable=false state
- Selecting a string and performing a copy and paste causes the string to be pasted.

**To Reproduce**
Also reproduced in StackBlitz with editable=false

**Misc**
- Node version: v18.17.1
- Package manager: npm
- Browser: Chrome 

https://github.com/TypeCellOS/BlockNote/assets/75936/c9ad2ce7-4d67-46c1-bca6-6ccc0f0c1989

### LINKED ISSUE #770: Can't download file or get file link when editor not editable
image

I solved it by using useEffect and interacting with the dom, but it doesn't seem to be the right way to do it.
```
const divElements = document.querySelectorAll('div[data-url]');
        divElements.forEach((element) => {
          const dataUrl = element.getAttribute('data-url');

          if(dataUrl) {
            element.addEventListener('click', () => handleDownload(dataUrl));
          }
        });

  const handleDownload = (url: string): void => {
      const link = document.createElement('a');
      link.href = url;
      link.download = url.split('/').pop() || 'download';
      document.body.appendChild(link);
      link.click();
      document.body.removeChild(link);
  }
  ```
  Thread on [Discord](https://discord.com/channels/928190961455087667/1242796833781186581)

### PR DESCRIPTION
**Pull request by matthewlipski**
Closes #707
Closes #770
Closes #766

### DIFF
--- a/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
+++ b/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
@@ -110,6 +110,11 @@ export class FormattingToolbarView implements PluginView {
 
     // Checks if menu should be shown/updated.
     if (!this.preventShow && (shouldShow || this.preventHide)) {
+      // Unlike other UI elements, we don't prevent the formatting toolbar from
+      // showing when the editor is not editable. This is because some buttons,
+      // e.g. the download file button, should still be accessible. Therefore,
+      // logic for hiding when the editor is non-editable is handled
+      // individually in each button.
       this.state = {
         show: true,
         referencePos: this.getSelectionBoundingBox(),
--- a/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
+++ b/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
@@ -15,7 +15,6 @@ export class FormattingToolbarView implements PluginView {
 
   public preventHide = false;
   public preventShow = false;
-  public prevWasEditable: boolean | null = null;
 
   public shouldShow: (props: {
     view: EditorView;
@@ -93,15 +92,9 @@ export class FormattingToolbarView implements PluginView {
     const isSame =
       oldState && oldState.doc.eq(doc) && oldState.selection.eq(selection);
 
-    if (
-      (this.prevWasEditable === null ||
-        this.prevWasEditable === this.editor.isEditable) &&
-      (composing || isSame)
-    ) {
+    if (composing || isSame) {
       return;
     }
-
-    this.prevWasEditable = this.editor.isEditable;
 
     // support for CellSelections
     const { ranges } = selection;
@@ -116,11 +109,7 @@ export class FormattingToolbarView implements PluginView {
     });
 
     // Checks if menu should be shown/updated.
-    if (
-      this.editor.isEditable &&
-      !this.preventShow &&
-      (shouldShow || this.preventHide)
-    ) {
+    if (!this.preventShow && (shouldShow || this.preventHide)) {
       this.state = {
         show: true,
         referencePos: this.getSelectionBoundingBox(),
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts [hunk 1]
```diff
@@ -93,16 +92,10 @@ export class FormattingToolbarView implements PluginView {
     const isSame =
       oldState && oldState.doc.eq(doc) && oldState.selection.eq(selection);
 
-    if (
-      (this.prevWasEditable === null ||
-        this.prevWasEditable === this.editor.isEditable) &&
-      (composing || isSame)
-    ) {
+    if (composing || isSame) {
```
YousefED, line 95: why do we not need the editable checks here anymore?

packages/react/src/components/FormattingToolbar/DefaultButtons/BasicTextStyleButton.tsx
```diff
@@ -101,7 +101,7 @@ export const BasicTextStyleButton = <Style extends BasicTextStyle>(props: {
     return !!selectedBlocks.find((block) => block.content !== undefined);
   }, [basicTextStyleInSchema, selectedBlocks]);
 
-  if (!show) {
+  if (!show || !editor.isEditable) {
```
YousefED, line 104: why do we move the editable checks here? Doesn't it make more sense to make sure the entire formattingtoolbar is not shown?

YousefED (changes requested):

packages/react/src/components/FormattingToolbar/DefaultButtons/BasicTextStyleButton.tsx
matthewlipski: Unfortunately not for the tiny edge case of being able to download files while the editor is uneditable, which is done via the formatting toolbar

YousefED: Hmm, wouldn't it be cleaner then to make sure clicking the button downloads files in non-editable state? That saves a lot of logic in the buttons, as we can then just disable the formatting toolbar when non editable.

I also think that's semantically better, as for a "formatting toolbar" I'd expect I'd only be able to use it when I can actually edit / format the text. 

wdyt?

matthewlipski: Also r.e. semantics I agree, but I think the root issue of that is really having the download button in the formatting toolbar to begin with. I reckon at some point we will probably change to a file-specific menu for file blocks like Notion has.

packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
matthewlipski: This is just because the logic for checking that has been moved to the toolbar buttons, since the download file button, unlike all the others, needs to be rendered even when the editor isn't editable.

packages/react/src/components/FormattingToolbar/DefaultButtons/BasicTextStyleButton.tsx
YousefED: Ok. I still don't really like it, but agree there's not a better solution without significantly more effort. Let's go for this option for now. Can you add the reasoning (that and why we allow the formatting toolbar to show even when the editor is not editable, and that we instead do a per-button check) as a comment?

YousefED (approved):
ok, one question then good to go

PR #784, opened 2024-05-29
fix: Custom inline content placeholder bug and enter handling

### LINKED ISSUE #643: List placeholder shows on top of pasted custom inline element
**Describe the bug**
image

See that there's a gray "List" over Steve in this example

**To Reproduce**

https://www.blocknotejs.org/examples/custom-schema/suggestion-menus-mentions

- Create a list
- Add a mention
- Copy that mention and paste it

matthewlipski:
Closes #643 

Also fixes edge cases where a block contains only a piece of non-editable inline content and Enter is pressed. Previously this inline content was ignored as the enter handlers only checked for text content.

### DIFF
--- a/packages/core/src/blocks/ListItemBlockContent/ListItemKeyboardShortcuts.ts
+++ b/packages/core/src/blocks/ListItemBlockContent/ListItemKeyboardShortcuts.ts
@@ -23,9 +23,9 @@ export const handleEnter = (editor: Editor) => {
 
   return editor.commands.first(({ state, chain, commands }) => [
     () =>
-      // Changes list item block to a text block if the content is empty.
+      // Changes list item block to a paragraph block if the content is empty.
       commands.command(() => {
-        if (node.textContent.length === 0) {
+        if (node.firstChild!.nodeSize === 2) {
           return commands.BNUpdateBlock(state.selection.from, {
             type: "paragraph",
             props: {},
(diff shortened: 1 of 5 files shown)


packages/core/src/blocks/ListItemBlockContent/ListItemKeyboardShortcuts.ts
```diff
@@ -23,9 +23,9 @@ export const handleEnter = (editor: Editor) => {
 
   return editor.commands.first(({ state, chain, commands }) => [
     () =>
-      // Changes list item block to a text block if the content is empty.
+      // Changes list item block to a paragraph block if the content is empty.
       commands.command(() => {
-        if (node.textContent.length === 0) {
+        if (node.firstChild!.nodeSize === 2) {
```
YousefED, line 28: Can you explain why this works and previously the check with textContent didn't?

YousefED (commented):
Nice! Didn't think of the `only-child` fix!

The other fix is not entirely clear for me (it also doesn't seem super nice to check for `nodeSize > 2`, so would like to learn a little bit more about that

packages/core/src/blocks/ListItemBlockContent/ListItemKeyboardShortcuts.ts
matthewlipski: Yeah, so a block can have content that is not text content, i.e. non-editable inline content such as mentions. While they might look like they contain text content, the text is not user editable and so there is no text content as far as the PM state is concerned.

Previously, we checked `node.textContent.length === 0` to determine if a block is empty. This means that in the above case, we think the block is empty despite there being a mention inside it, so pressing Enter is handled incorrectly. Nodes which really have no content have `node.firstChild!.nodeSize === 2`, so we use that instead.

Also, `node.textContent.length === 0` checks if there's text content in the whole block, including children, which is just incorrect. It should be `node.firstChild!.textContent.length === 0`.

I will change it to use `node.firstChild!.childCount === 0` though as it's effectively the same and doesnt have the magic `2`, and also clean this up further by using `contentNode` instead of `node.firstChild!`.

YousefED: Great explanation and improvement!

YousefED (approved):

PR #785, opened 2024-05-29
fix: PDF block example styling

### LINKED ISSUE #767: Pdf style [Bug]
Is there anyway to make the pdf preview responsive  

Screenshot 2024-05-27 104130

here's the PDF Block docs : https://www.blocknotejs.org/examples/custom-schema/pdf-file-block

matthewlipski:
Closes #767


YousefED (commented):
Nice. Should we fix this only for the PDF example, or is this styling that we should add to the overall style file?

Issue #750, opened 2024-05-17
Cannot read properties of undefined (reading 'renderToElement')


boyanhtrai:
**Describe the bug**
Cannot read properties of undefined (reading 'renderToElement')

**To Reproduce**
I just test on example link on https: //www.blocknotejs.org/docs/custom-schemas/custom-inline-content
Link bug example: https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/05-custom-schema/02-suggestion-menus-mentions?file=App.tsx

![image](https://github.com/TypeCellOS/BlockNote/assets/29662022/42ba55b7-878b-4eaf-b190-1bc018fdf267)

version: ^0.13.2
**Misc**
- Node version: v20.11.0
- Package manager:
- Browser: chrome

conr:
I'm also running into this one

![image](https://github.com/TypeCellOS/BlockNote/assets/5826063/ce8ac899-48a7-47bc-87d8-7d19ceb286b5)

YousefED:
I'm keeping this open as #788 was a temp workaround

PR #788, opened 2024-05-31
fix: Custom inline content crash

conr:
Fixes https://github.com/TypeCellOS/BlockNote/issues/750

**Before**

image

**After**

image


YousefED:
Thanks @conr. This looks like a good temporary fix, but I still want to figure out the root cause (why is `contentComponent` undefined). If the CI passes I'd be happy to merge this regardless

conr:
@YousefED That's very fair!

Looks like CI is passing now

Dhruvagwal (commented):
it seems to be woking for now.

PR #789, opened 2024-05-31
refactor: `contentRef` prop into `useContent` hook

### LINKED ISSUE #659: Bug: Cannot add space at the end of a custom block's content
**Describe the bug**
I have implemented several custom blocks by following the documentation. In Firefox, when typing content in any custom block there is no way to leave a space. The precise behavior is this: 

- When writing at the very end of the block you can only input one space.
- Anything you type after that final space simply replaces that space. This includes pressing space again.
- The end result is that all your text is strung together in a single string with no spaces.
- This can be circumvented by placing the cursor in any other position except the very end of the block. Then you can input spaces correctly and type normally.
- This behavior can be reproduced on the docs at https://www.blocknotejs.org/docs/custom-schemas/custom-blocks.
- This seems to affect only Firefox. Chrome and Edge work normally.

**To Reproduce**
- Go to https://www.blocknotejs.org/docs/custom-schemas/custom-blocks in Firefox.
- In the editor create a new alert block or click on the existing one. If you go for the existing one, make sure the cursor is at the very end.
- Type away.

**Misc**
- Node version: 20.11.1
- Package manager: npm
- Browser: Firefox
- [x] I work for a [sponsor](https://github.com/worldanvil) and would appreciate if you could look into this sooner than later 💖

matthewlipski:
This PR removes the hacky fix used for the `contentRef` prop by refactoring it into a `useContent` prop. Now setting the editable element in a React block is done via this hook instead of the `contentRef` prop.

Closes #659

TODO: There are currently a pretty big problem in that because this uses a hook which runs asynchronously, exporting to HTML does not properly serialize the content.

### DIFF
--- a/examples/02-ui-components/03-formatting-toolbar-block-type-items/Alert.tsx
+++ b/examples/02-ui-components/03-formatting-toolbar-block-type-items/Alert.tsx
@@ -1,5 +1,14 @@
-import { defaultProps } from "@blocknote/core";
-import { createReactBlockSpec } from "@blocknote/react";
+import {
+  BlockConfig,
+  DefaultInlineContentSchema,
+  defaultProps,
+  DefaultStyleSchema,
+} from "@blocknote/core";
+import {
+  createReactBlockSpec,
+  ReactCustomBlockRenderProps,
+  useContent,
+} from "@blocknote/react";
 import { Menu } from "@mantine/core";
 import { MdCancel, MdCheckCircle, MdError, MdInfo } from "react-icons/md";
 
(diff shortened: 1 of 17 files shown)


YousefED (commented):
Couple of questions (I think we should leave this improvement for a later cycle regardless though):

- wdym with  `contentRef` feels hacky? A hook is a different API but I wouldn't consider one to be more "hacky" than the other? Especially since the hook just returns a ref, so practically it comes down to the same thing right?
- I suppose the benefit of this approach is that we can return more props (style and data-node-view-content) to the element?
- Why do we need the data-node-view-content again / why didn't we need this before?
- "hooks run async" -> this isn't the case. A hook is just the same as a regular function call. So the bug must be somewhere else

PR #790, opened 2024-05-31
fix: Remove `data-mantine-color-scheme` attribute

### LINKED ISSUE #523: Importing react style.css is breaking our own CSS

### LINKED ISSUE #651: Mantine adding data-mantine-color-scheme to the html tag
**Describe the bug**
Mantine adding data-mantine-color-scheme to the html tag

**To Reproduce**
Setup with the basic setup documentation

matthewlipski:
Closes #651
Closes #523

### DIFF
--- a/packages/mantine/src/index.tsx
+++ b/packages/mantine/src/index.tsx
@@ -160,7 +160,11 @@ export const BlockNoteView = <
       {/* as proposed here:  https://github.com/orgs/mantinedev/discussions/5685 */}
       <MantineProvider
         theme={mantineTheme}
-        cssVariablesSelector=".bn-container">
+        cssVariablesSelector=".bn-container"
+        // This gets the element to set `data-mantine-color-scheme` on. Since we
+        // don't need this attribute (we use our own theming API), we return
+        // undefined here.
+        getRootElement={() => undefined}>
         <BlockNoteViewRaw
           theme={typeof theme === "object" ? undefined : theme}
           {...rest}


YousefED (commented):
We refer to `data-mantine-color-scheme` in our CSS. is this not an issue?

matthewlipski:
> We refer to `data-mantine-color-scheme` in our CSS. is this not an issue?

Ah yeah we do indeed - it's because `mantineStyles.css` contains Mantine global styles that are scoped to the editor, and so includes some things used by Mantine themes which we don't use since we have our own theme API. I think it's ok to keep it as is for maintainability, since we can just copy the Mantine global styles and change CSS rules for `html`/`body`/`#root` to `.bn-container`, kind of how we do with Ariakit.

YousefED:
> Ah yeah we do indeed - it's because `mantineStyles.css` contains Mantine global styles that are scoped to the editor, and so includes some things used by Mantine themes which we don't use since we have our own theme API. I think it's ok to keep it as is for maintainability, since we can just copy the Mantine global styles and change CSS rules for `html`/`body`/`#root` to `.bn-container`, kind of how we do with Ariakit.

Got it! Sounds good unless `mantine-dark-hidden` (or light-hidden) is used in the default components that we use, but I doubt that. Ok to merge if you can do some testing with both themes :)

PR #791, opened 2024-05-31
fix: Side menu not updating in edge cases

### LINKED ISSUE #580: Failed Block Deletion with custom delete handler in Side Menu
**Describe the bug**
adding delete button to side menu causes failed deletes occasionally

**To Reproduce**
can use the example from docs with custom delete handle:
https://www.blocknotejs.org/docs/side-menu#custom-side-menu

add five or so blocks. Start deleting from the top in succession. Block Id references get lost and app crashes. Something with the block ids gets out of sync.

error seen:
`/App.tsx: Blocks with the following IDs could not be found in the editor: fba2f81f-8bb7-43c5-8f90-36d2701eba5a (22:36)`

**Misc**
 A custom delete handle at top level is preferable UX than using the delete inside the drag handle, but the inconsistent behavior is a blocker.

matthewlipski:
Almost always, the side menu only needs to update when the mouse cursor moves. However, in some edge cases, such as the delete side menu button, the block or block position changes without the mouse moving. This PR makes the side menu also update after editor state updates if needed.

Closes #580

### DIFF
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -255,6 +255,9 @@ export class SideMenuView<
   private state?: SideMenuState<BSchema, I, S>;
   private readonly emitUpdate: (state: SideMenuState<BSchema, I, S>) => void;
 
+  private needUpdate = false;
+  private mousePos: { x: number; y: number } | undefined;
+
   // When true, the drag handle with be anchored at the same level as root elements
   // When false, the drag handle with be just to the left of the element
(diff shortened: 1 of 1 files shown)


YousefED (approved):

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -511,6 +525,17 @@ export class SideMenuView<
     }
   };
 
+  update() {
```
YousefED, line 528: Let's add a comment why (including an example) why this is necessary, so we don't forget in the future

PR #792, opened 2024-05-31
use nonce value is supplied

instanceofnull:
This fixes an issue with dynamically injected style tags with a strict content security policy.


YousefED:
Thanks @instanceofnull , we'll make sure this gets merged. What are you currently using blocknote for?

YousefED (approved):

Issue #742, opened 2024-05-14
Cannot have \n in custom block content.


colinstantos:
**Describe the bug**
      {
        type: 'alert',
        content: 'This is an \n example \n alert!!',
      }

the above content is in a single line instead of 3.

**To Reproduce**
https://stackblitz.com/edit/github-qx9fhr?file=App.tsx

PR #794, opened 2024-06-02
Issue_Fixed : #742 

NotVishesh:
Fixed_Issue #742
/claim #742


YousefED:
Unfortunately, this is not the desired fix. We should make sure the \n's are parsed correctly internally. It might require some prosemirror knowledge to find the exact place to implement this

Issue #778, opened 2024-05-29
The suggestion popover doesn't disappear


trinhle96:
**Describe the bug**
I implemented a mentions menu that is triggered by the "@" character in my React project, and it's working great. However, I'm facing an issue in my Next.js project: when I type something that starts with "@", but isn't a mention item, the suggestion popover displaying "No items found" doesn't disappear.
![Screenshot 2024-05-28 143717](https://github.com/TypeCellOS/BlockNote/assets/97581320/765c5f19-19e0-4b85-a8a4-7a6f537f9eed)

**To Reproduce**
stackblitz: https://stackblitz.com/edit/stackblitz-starters-qy4jp9?file=README.md

**Misc**
- Next js version: 14

kedarvyas084:
Hello @trinhle96, In your stackblitz, I was able to get suggestions?
would it be your system's problem?

![image](https://github.com/TypeCellOS/BlockNote/assets/138874486/a57d6318-44b7-455d-aecc-6fbbfc8d7219)

trinhle96:
Hi @kedarvyas084 , Please type until no items are found. Then, add spaces or anything, and the popover won't disappear.

trinhle96:
Hi @YousefED, I tried your fix, and it works great. Thank you so much for your support.

PR #796, opened 2024-06-03
fix: add memoization in suggestion menu

YousefED:
Maybe related: https://github.com/TypeCellOS/BlockNote/issues/778

My assumption is there is a race condition because the event handler (useUIstate) keeps getting attached / detached. This should improve that, but regardless it's a nice cleanup anyway


matthewlipski:
Tried it out in a NextJS 14 BlockNote project I had set up and unfortunately #778 doesn't get fixed with these changes:( The mentions menu remains open no matter how many characters you add to the query. However, think it's still worth merging this since it's a nice cleanup, just edited your comment so it doesn't close the issue.

PR #797, opened 2024-06-03
fix: RangeError when un-nesting certain blocks

### LINKED ISSUE #481: RangeError: Invalid content for node blockcontainer
**Describe the bug**
Whenever you delete the parent list item and then try and delete certain things in the child

**To Reproduce**

Screenshot 2023-12-15 at 10 38 30 PM

**Misc**
- Node version:
- Package manager:
- Browser: 
- [ ] I'm a [sponsor](https://gith

ub.com/sponsors/YousefED) and would appreciate if you

 could look into this sooner than later 💖

### LINKED ISSUE #598: Getting RangeError: Invalid content for node blockContainer: && Nested block deletion bug
**Describe the bug**
Getting RangeError
Invalid content for node blockContainer: <paragraph.

On production it is deleting whole block.

**To Reproduce**
Sandbox to reproduce: https://codesandbox.io/p/sandbox/sleepy-montalcini-vnhk7c

**Misc**
- Node version:
- Package manager:
- Browser: 

matthewlipski:
This PR fixes an error caused when trying to un-nest blocks which have children.

Closes #481
Closes #598

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -917,7 +917,12 @@ export class BlockNoteEditor<
    * Lifts the block containing the text cursor out of its parent.
    */
   public unnestBlock() {
-    this._tiptapEditor.commands.liftListItem("blockContainer");
+    const { startPos } = getBlockInfoFromPos(
+      this._tiptapEditor.state.doc,
+      this._tiptapEditor.state.selection.from
+    )!;
+
+    this._tiptapEditor.commands.BNUnnestBlock(startPos);
   }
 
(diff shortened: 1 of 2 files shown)


YousefED (changes requested):
I think this introduces some bugs:

bug nesting

sb8244:
May be unrelated, I'm attempting to use this function to fix `liftListItem` in a function I'm creating.

The result when calling it is that all indentation is removed. So instead of going up 1 level, it goes to the root.

It may be a side effect of the function I'm writing (trying to have `sinkListItem` only affect the single node that's selected) though.

PR #800, opened 2024-06-04
fix: node conversion

### LINKED ISSUE #742: Cannot have \n in custom block content.

jkcs:
close #742
/claim #742


YousefED:
Thanks @jkcs. Your solution looks great, I think you've found the right place to fix it.

> https://github.com/TypeCellOS/BlockNote/blob/c9a53bb9bd79c3b1990f74d2c87956c3cf5949b0/packages/core/src/pm-nodes/BlockContainer.ts#L221-L224
> 
> It seems like there is the same issue here?

Great find! I think you're right. Could you fix this as well and create unit tests for both of them? I can double the bounty for that

YousefED:
@jkcs the `updateBlock` function in https://github.com/TypeCellOS/BlockNote/blob/4a197733e46cf2d53b888e0de5db5a82792ae796/packages/core/src/api/blockManipulation/blockManipulation.test.ts indirectly call BNUpdateBlock. You can add the test there!

Can you also confirm that the test wasn't working before your fix, but now is? (to make sure the tests really work)?

YousefED:
> @YousefED Thank you for providing more information. I just tried adding tests in `blockManipulation.test.ts` and found that the update tests for `BlockContainer.ts` are not effective. The output of `typeof block.content === "string"` consistently shows the `editor.document`. I believe this test does not verify this point correctly. In the view, it should be converted to `<br>`.

I don't understand what you mean with `typeof block.content === "string"` consistently shows the `editor.document`, can you explain?

YousefED:
Great! This latest test also failed without your fix I suppose?

@matthewlipski can you do a final review and merge if ok?

PR #802, opened 2024-06-04
fix: Mantine sub-menu rendering in the DOM

matthewlipski:
Right now, Mantine sub-menu dropdowns are being rendered within the menu item that opens them. This throws an error as the menu item is a button, and the dropdown contains more menu items which are also buttons (`button` elements are not valid descendants of other `button` elements).

This PR makes sub menu dropdowns get rendered in the same dropdown as the menu items which trigger them, same as how Ariakit and ShadCN do it.

### DIFF
--- a/packages/mantine/src/menu/Menu.tsx
+++ b/packages/mantine/src/menu/Menu.tsx
@@ -81,6 +81,8 @@ const SubMenu = forwardRef<
 
   const [opened, setOpened] = useState(false);
 
+  const itemRef = useRef<HTMLButtonElement | null>(null);
+
   const menuCloseTimer = useRef<ReturnType<typeof setTimeout> | undefined>();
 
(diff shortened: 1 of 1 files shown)


packages/mantine/src/menu/Menu.tsx
```diff
@@ -107,11 +109,22 @@ const SubMenu = forwardRef<
       }}>
       <Mantine.Menu.Item
         className="bn-menu-item bn-mt-sub-menu-item"
-        ref={ref}
+        ref={(node) => {
```
YousefED, line 112: you can use `const combinedRef = mergeRefs(ref, itemRef);` for this (from `@mantine/hooks`)

YousefED (commented):
nice! 1 comment

PR #812, opened 2024-06-05
fix: More Notion-like side menu positioning

### LINKED ISSUE #386: Side menu is incorrectly positioned in the middle of a block
**Describe the bug**

On Notion, the position of the side menu is positioned at the start of each block, as shown below.

Screenshot 2023-10-19 at 16 27 33

In BlockNote, the side menu is positioned in the _middle_ of each block, which is inconsistent with the notion-like behaviour it aims to be similar to.

incorrectpos

**To Reproduce**

Just create a multi-line block and hover over the paragraph.

matthewlipski:
This PR changes the side menu to be positioned centered on the first line of a block. Blocks with no text content usually assume the line height of a paragraph block.

This solution is a bit hacky in that the line height values for different blocks are hard coded instead of taking them from a bounding box, though fixing this would add a lot of complexity. Fortunately, it's easy for consumers to change those values via CSS.

Also, Notion does not center the side menu for things like file blocks, instead using the paragraph block line height, as you can see below:
Screenshot 2024-06-05 at 22 45 12
I think this looks really ugly since the block is short enough that I think it should be centered, so I changed this for file & audio blocks also, as well as for when the add file button is showing in all types of file blocks.

Closes #386

### DIFF
--- a/packages/ariakit/src/menu/Menu.tsx
+++ b/packages/ariakit/src/menu/Menu.tsx
@@ -40,6 +40,7 @@ export const MenuDropdown = forwardRef<
 
   return (
     <Ariakit.Menu
+      unmountOnHide={true}
       className={mergeCSSClasses("bn-ak-menu", className || "")}
       ref={ref}>
       {children}
--- a/packages/ariakit/src/sideMenu/SideMenu.tsx
+++ b/packages/ariakit/src/sideMenu/SideMenu.tsx
@@ -10,10 +10,10 @@ export const SideMenu = forwardRef<
 >((props, ref) => {
   const { className, children, ...rest } = props;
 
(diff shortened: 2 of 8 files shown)


YousefED (changes requested):
I think it's really smart to do it this way by adding an attribute and then use CSS for the exact positioning! Also great that you tested with the different frameworks.

My only suggestion would be to do this a little bit more "semantically" / pay more attention to naming (naming is always key!).

You now use `data-block-height`, but actually don't pass a height, but some sort of type. Why don't we just use `data-block-type` and pass the block type directly? (and with a bit more extra info for files / headings)

YousefED (approved):
nice, ok when passing CI!

PR #814, opened 2024-06-06
fix: Side menu block finding logic

matthewlipski:
This PR simplifies & fixes logic in the side menu used to find the hovered block. Specifically, this fixes an error that gets thrown in the built environment when the hovered block isn't found, and makes the behaviour consistent for both the dev and built envs.


YousefED (approved):
looks great!

PR #815, opened 2024-06-06
fix: Removed ShadCN focus global CSS

### LINKED ISSUE #787: Shadcn CSS has global side effects (outline is lost)
**Describe the bug**

The styles applied with `shadcn/style.css` have global effects and the outline is getting disabled for all components

Lines that have the side effect are:
https://github.com/TypeCellOS/BlockNote/blob/bea361e39846c26d9729a777a5cd3de1f48e8c3e/packages/shadcn/src/style.css#L76-L78

When we import `react/style.css` instead of `shadcn/style.css` the outline of inputs is visible.

**To Reproduce**
https://stackblitz.com/edit/github-jywdw6?file=App.tsx

Import the shadcn.css
```tsx
import '@blocknote/shadcn/style.css';
```

**Misc**
- Node version: 20
- Package manager: pnpm
- Browser: chrome

matthewlipski:
We have a CSS rule to remove the focus ring from focused elements in ShadCN that is scoped globally instead of to the editor, which this PR fixes.

Closes #787

### DIFF
--- a/packages/shadcn/src/style.css
+++ b/packages/shadcn/src/style.css
@@ -77,6 +77,10 @@
   z-index: 99999 !important;
 }
 
+.bn-editor:focus-visible {
+  outline: none;
+}
+
 .bn-side-menu {
   display: flex;
   justify-content: center;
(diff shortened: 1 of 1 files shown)


YousefED (approved):
ok when passing CI!

PR #818, opened 2024-06-07
simple latex block

### LINKED ISSUE #741: Custom Block for Katex

**Additional context**

jkcs:
close #741
/claim #741

### DIFF
--- a/packages/core/src/api/blockManipulation/__snapshots__/blockManipulation.test.ts.snap
+++ b/packages/core/src/api/blockManipulation/__snapshots__/blockManipulation.test.ts.snap
@@ -640,8 +640,8 @@ Line2",
     "content": [
       {
         "styles": {},
-        "text": "Updated Custom Block with
-line
+        "text": "Updated Custom Block with 
+line 
 break",
         "type": "text",
       },
@@ -671,8 +671,8 @@ exports[`Update Line Breaks > Update paragraph with line break 1`] = `
     "content": [
       {
         "styles": {},
-        "text": "Updated Custom Block with
-line
+        "text": "Updated Custom Block with 
+line 
 break",
         "type": "text",
       },


examples/05-custom-schema/05-latex-block/LaTex.tsx
```diff
@@ -0,0 +1,233 @@
...
+  NodeViewRenderer,
+  NodeViewWrapper,
+  ReactNodeViewRenderer,
+} from "@tiptap/react";
+import { createStronglyTypedTiptapNode, createInternalInlineContentSpec } from "@blocknote/core";
+import { mergeAttributes } from "@tiptap/core";
+import { ChangeEvent, useEffect, useState, useRef } from "react";
+import "./styles.css";
+
+function loadKaTex(callback: () => void) {
```
YousefED, line 17: I'd prefer to just have this as a dependency of the package, that makes the code a lot cleaner imo

packages/dev-scripts/examples/util.ts
```diff
@@ -139,7 +140,11 @@ export function getProjectFiles(project: Project): Files {
  */
 export function getExampleProjects(): Project[] {
   const examples: Project[] = glob
-    .globSync(path.join(dir, "../../../examples/**/*/.bnexample.json"))
+    .globSync(
```
YousefED, line 143: I suppose this is needed for windows?

Cleaner might be: `path
        .join(dir, "..","..","..","examples","**",".*","..bnexample.json")`?

examples/05-custom-schema/05-latex-block/LaTex.tsx
```diff
@@ -0,0 +1,233 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        return oldUpdated(node, decorations, undefined as any);
+      };
+      return ret;
+    };
+  };
+  return nodeView;
+}
+
+const propSchema = {
```
YousefED, line 182: I don't think `open` should be part of the blockschema. It's a state of the UI, not of the document.

For example, this will mean that the open state will be stored in databases or synced over collaboration, I don't think that's what we want

YousefED (changes requested):
This already looks pretty good, awesome!

I've left some comments inline, and have some additional questions:

- Why did you need to use `createInternalInlineContentSpec` instead of the regular content spec?
- It would be very nice to have this keyboard accessible (compare with the behavior in Notion when using the keyboards to navigate).
- Should we create a separate package (e.g.: `packages/react-extension-equations`) for this so users can easily add it to their editor without having to copy all the code? I think it should be an extension because not all users might want to pull in the katex dependency

It's great to see your contributions; it seems you're getting up to speed nicely 💪  did you find it ok to get started and / or have any feedback about the codebase?

YousefED:
Ok!

> The reason for using createInternalInlineContentSpec is that it allows customization of addNodeView. (Not sure if createInlineContentSpec would work)

What do you need for your manual NodeView? If there is anything missing in our higher level `createReactInlineContentSpec`, it'd be great to know so we can consider implementing it there. This way we would  improve our API surface + not have the Latex block dive into "internals"

YousefED:
> The implementation of keyboard accessibility is more challenging than I anticipated. I'll close this for now and will provide a new PR once I have it implemented

Hi @jkcs! Any progress on this? I think you were doing great, would be awesome to be able to merge this at some point :) How did it go with the keyboard approach?

hunxjunedo:
Hey @jkcs how is it going ? I would like to collaborate, recently worked on a similar issue (emoji picker)

PR #822, opened 2024-06-07
feat(i18n): add Arabic translation

zaaakher:
Salam 👋,

This PR simply adds Arabic translations.

Handling RTL isn't yet implemented and I will have to study the code a little more to know how to add that feature. If mantine is used then we could perhaps use the mantine [`DirectionProvider`](https://mantine.dev/styles/rtl/#directionprovider)

### DIFF
--- a/packages/core/src/i18n/locales/ar.ts
+++ b/packages/core/src/i18n/locales/ar.ts
@@ -1,0 +1,329 @@
+import type { Dictionary } from "../dictionary";
+
+export const ar: Dictionary = {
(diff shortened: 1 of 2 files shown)


YousefED:
Salam!

Great, would be nice to have arabic translations and RTL support!

zaaakher:
> would be nice to have arabic translations and RTL support!

Definitely! 👌 This PR only adds Arabic translation and I'll try to add RTL support in another PR whenever I get around it inshallah.

PR #825, opened 2024-06-07
chore(docs): include comments about dynamic locales in TS

zaaakher:
I couldn't figure out how to add additional text under the preview block so I just added comments to the actual preview block.

The comments I added simply show how to use dynamic language in TS with 

```ts
 dictionary: locales[lang as keyof typeof locales]
```


YousefED:
Nice. Might also be nice to for example select a different locale from a dropdown. Do you think that would have helped you, or was your issue specifically about integrating an i18n lib?

zaaakher:
It worked for me from the first try honesty. But Typescript kept screaming at me for using `lang` as an index of `locales` and I had to ask AI for help 😅. `lang as keyof typeof locales` is the part I got stuck with and figured maybe we should update the docs to include that scenario

PR #827, opened 2024-06-07
fix(ui): missing margin in checklist while in RTL

zaaakher:
This tiny PR fixes a missing space next to the checkbox in checklists while in an RTL user interface.

## Before

image

## After 

image

---

Althought on a side note,  `1.2em` feels a little too large of a space between the checkbox and the text. I suggest `0.5em`. 

### 1.2 em

image

### 0.5 em

image

If you agree `0.5em` is better spacing, let me know and I will add a commit changing it from 1.2 to 0.5


YousefED (approved):
great!

zaaakher:
> 1.2em feels a little too large of a space between the checkbox and the text. I suggest 0.5em.

@YousefED Any thoughts on this part? I couldn't quickly find any resources online to support my suggestion about the standard space between checkboxes and their title so I asked chatGPT 👇

image

YousefED:
I agree a smaller margin looks better and is more inline with Notion. Looping in @matthewlipski for this as he made the original check box

zaaakher:
Ok I just decreased it to `0.5em` 👍

matthewlipski:
> I agree a smaller margin looks better and is more inline with Notion. Looping in @matthewlipski for this as he made the original check box

Agreed, does look closer to Notion so I changed the other list item blocks to use the same margin also

matthewlipski (approved):

PR #837, opened 2024-06-12
chore: Update Mantine

matthewlipski:
After merging PR #652 and releasing v0.13.0, we ended up having a Mantine version regression from `7.7.1` to `7.5.0`. This is important as `7.7.1` updated [hash-css-selector](https://github.com/rtivital/hash-css-selector) to `1.1.0`, which changes CSS module classes to use underscores instead of hyphens.

This fixes bugs which can occur when importing BlockNote CSS files, as CSS processors and build tools sometimes change hyphens in the class names to underscores (I think this is because accessing them requires using `styles["class-name"]` instead of `styles.class_name`). This in turn means that components have mismatched class names with the stylesheets and the styles don't get applied.

This PR updates Mantine to the newest version, `7.10.1`.


YousefED (approved):
Nice. Did you also check if there were updates to the mantine css we need to copy?

matthewlipski:
> Nice. Did you also check if there were updates to the mantine css we need to copy?

Good question, checked and the `global.css` file is unchanged between `7.5.0` and `7.10.1`

PR #838, opened 2024-06-12
fix: Suggestion menu scrolling to selected item

### LINKED ISSUE #773: Suggestion menu does not scroll when using arrow keys
**Describe the bug**
when using up/down arrow keys to select suggestion menu item, the highlighted item does not scroll into view as you move down the list.

**To Reproduce**
reproduced in latest 0.13.4 and homepage demo editor - open suggestion, down arrow to navigate to last item

**Misc**
- Node version:
- Package manager:
- Browser: 

matthewlipski:
Closes #773


packages/ariakit/src/util/util.tsx
```diff
@@ -0,0 +1,17 @@
+import { ForwardedRef, MutableRefObject } from "react";
+
+export function mergeRefs<ElementType extends HTMLElement>(
```
YousefED, line 3: I think we already have this function in the core or react package

YousefED (changes requested):
nice! small comments

YousefED (approved):
great

PR #841, opened 2024-06-13
fix: update `remarkStringify` config

### LINKED ISSUE #834: `replaceBlocks` is removing the space between brackets for checklists
**Describe the bug**
the method `replaceBlocks` is removing the space between the brackets of a checklist which should cause it to render an actual interactive checkbox.

**To Reproduce**
This issue is broken down in this thread https://github.com/TypeCellOS/BlockNote/issues/826#issuecomment-2162085998

**Misc**
- Node version: 18.19.3
- Package manager: pnpm
- Browser: chrome

zaaakher:
Fixes #834 

---
For some reason, returning the text node within the `remarkStringify` handler option solved the issue with an extra backslash before the first bracket. 
```ts
.use(remarkStringify, { handlers: { text: (node) => node.value } })
```

# Before
339166596-baf435ae-a584-4e66-be80-64267319d125

# After 

blocknote-escape-checklist-solved

### DIFF
--- a/packages/core/src/api/exporters/markdown/__snapshots__/lists/basic/markdown.md
+++ b/packages/core/src/api/exporters/markdown/__snapshots__/lists/basic/markdown.md
@@ -4,5 +4,5 @@
 1.  Numbered List Item 1
 2.  Numbered List Item 2
 
-*   \[ ] Check List Item 1
-*   \[x] Check List Item 2
+*   [ ] Check List Item 1
+*   [x] Check List Item 2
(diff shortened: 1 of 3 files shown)


matthewlipski:
Oh nice, this looks like an elegant fix:) I'll update the unit tests and then I think we can merge, thanks for your contribution!

zaaakher:
Thanks @matthewlipski . I will remember to run `npm run test` next time 😁

TheMikeyRoss:
Nice thanks @zaaakher , would this be included in 0.14.2 ?

zaaakher:
> Nice thanks @zaaakher , would this be included in 0.14.2 ?

Yes I believe so. Although it's up to the maintainers.

PR #842, opened 2024-06-13
feat: added support for uppy

### LINKED ISSUE #807: Add support for Uppy for file uploading
We'd like to add support for https://uppy.io/ for file uploading. This would require:

Initially, creating an example (similar to other blocknote examples in the `example` directory would suffice (as part of follow up work, later we could see what we can extract to a reusable package).

The main requirements are:

- [ ] Most important: customize the BlockNote FilePanel to use the Uppy React library (use a `FilePanelController` and the Uppy Dashboard for this)
- [ ] Make sure the `uploadFile` [function](https://www.blocknotejs.org/docs/editor-basics/setup) also uses the uppy pipeline
- [ ] It should be clear in the code where you can configure additional Uppy options such as a companion URL or transloadit server

Of course, it must work for selecting / uploading resources for different blocks (e.g.: File / Image / Video)

ezhil56x:
Added support for uppy with an example

Closes #807
/claim #807


ezhil56x:
@YousefED 
<details>
  <summary>Click to expand</summary>
  
  [https://github.com/TypeCellOS/BlockNote/assets/103899034/fd6287a5-2f5d-4378-ab04-cdd154e0e659](https://github.com/TypeCellOS/BlockNote/assets/103899034/fd6287a5-2f5d-4378-ab04-cdd154e0e659)
</details>

packages/react/src/components/FilePanel/DefaultTabs/UploadTabUppy.tsx
```diff
@@ -0,0 +1,79 @@
...
+    useEffect(() => {
+      if (uppy !== undefined) {
+        const handleUploadSuccess = (file: any, response: any) => {
+          if (response.status === 200) {
+            uppy.removeFile(file.id);
+          }
+
+          let fileURL = response.body.data.url;
+
+          if (fileURL.includes("tmpfiles.org")) {
```
YousefED, line 49: why do we have this?

packages/react/src/components/FilePanel/DefaultTabs/UploadTabUppy.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const { block, setLoading } = props;
+    const editor = useBlockNoteEditor<B, I, S>();
+
+    const uppy = useMemo(() => editor.uploadFileUppy, [editor.uploadFileUppy]);
+
+    useEffect(() => {
+      if (uppy !== undefined) {
+        const handleUploadSuccess = (file: any, response: any) => {
+          if (response.status === 200) {
+            uppy.removeFile(file.id);
```
YousefED, line 44: is this standard practice, to call removefile? Maybe add comment why

packages/react/src/components/FilePanel/DefaultTabs/UploadTabUppy.tsx
```diff
@@ -0,0 +1,79 @@
...
+              name: file.name,
+              url: fileURL,
+            } as Partial<Props<B[keyof B]["propSchema"]>>,
+          };
+
+          editor.updateBlock(block, updateData);
+          setLoading(false);
+        };
+
+        uppy.on("upload-success", handleUploadSuccess);
```
YousefED, line 64: What happens on error? Does Dashboard automatically handle that and show that to the user?

packages/react/src/components/FilePanel/DefaultTabs/UploadTabUppy.tsx
```diff
@@ -0,0 +1,79 @@
...
+        return () => {
+          uppy.off("upload-success", handleUploadSuccess);
+        };
+      }
+      return;
+    }, [uppy, editor, block, setLoading]);
+
+    return (
+      <Components.FilePanel.TabPanel className="bn-panel-tab">
+        <Dashboard id="dashboard" uppy={uppy ?? new Uppy()} />
```
YousefED, line 75: Two qs:
- why the `?? new Uppy()`? I don't think that should happen right?
- We should make sure the darkmode Dashboard is used when BlockNote is set to dark

rajeshj11:
> This looks great already. I've left some comments inline but also have one main architectural request:
> 
> I don't think we want to add the uppy dependencies by default - otherwise users who don't need uppy will end up adding a lot of dependencies. This means all Uppy related code should live in the Example (or in a separate package (`blocknote-uppy` or sth), but I don't think that's necessary for now). The way to do this would be to in the example use a `FilePanelController` and pass in the Uppy tab as part of it's props. (the uppy tab should then also live in the configuration).
> 
> Secondly, I don't think we then need `uploadFileUppy` anymore in the configuration. We do need (see original requirements) to make sure that `uploadFile` uploads via Uppy. The reason for this is that while at this moment, it's true that `uploadFile` is only called from the upload tab (which we'll replace by the dashboard), but at a later point it's likely we'll call `uploadFile` from other places (like dropping files into the editor) as well.

@ezhil56x I have already made changes related to this. If needed any reference you can view my closed pr.
https://github.com/TypeCellOS/BlockNote/pull/843

ezhil56x:
Will update my PR soon and let you know

examples/01-basic/11-file-uploading-uppy/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+    endpoint: "https://tmpfiles.org/api/v1/upload",
+  });
+
+  uppy.on("upload-success", (file: any, response: any) => {
+    // File should be removed from the uppy instance after successful upload.
+    if (response.status === 200) {
+      uppy.removeFile(file.id);
+    }
+
+    const block = editor.document[2];
```
YousefED, line 42: why editor.document[2]? This seems hardcoded?

examples/01-basic/11-file-uploading-uppy/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  uppy.on("upload-error", (file: any, error: any) => {
+    console.error("Upload error:", error);
+  });
+
+  // Renders the editor instance using a React component.
+  return (
+    <BlockNoteView editor={editor} filePanel={false}>
+      <FilePanelController
+        filePanel={() => {
```
YousefED, line 64: I think this should be a `\<FilePanel>` with the dasboard passed as a tab via the `tabs` prop

examples/01-basic/11-file-uploading-uppy/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+  uppy.on("upload-success", (file: any, response: any) => {
+    // File should be removed from the uppy instance after successful upload.
+    if (response.status === 200) {
+      uppy.removeFile(file.id);
+    }
+
+    const block = editor.document[2];
+
+    let fileURL = response.body.data.url;
+    fileURL = fileURL.replace("tmpfiles.org/", "tmpfiles.org/dl/");
```
YousefED, line 45: why this? can you add a comment?

examples/01-basic/11-file-uploading-uppy/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+import Uppy from "@uppy/core";
+import XHR from "@uppy/xhr-upload";
+import { Dashboard } from "@uppy/react";
+
+import "@uppy/core/dist/style.min.css";
+import "@uppy/dashboard/dist/style.min.css";
+
+export default function App() {
+  // Creates a new editor instance.
+  const editor = useCreateBlockNote({
```
YousefED, line 16: could you also implement `uploadFile` as mentioned earlier?

examples/01-basic/11-file-uploading-uppy/App.tsx
ezhil56x: Manipulating blocks example also have hard coded values for updating a block

examples/01-basic/11-file-uploading-uppy/App.tsx
ezhil56x: File Panel needs block as prop, I'm not sure how to get block value

examples/01-basic/11-file-uploading-uppy/App.tsx
ezhil56x: Added comment

examples/01-basic/11-file-uploading-uppy/App.tsx
ezhil56x: Uppy doesn't seem to support uploading files with custom function. So I don't think so we can make changes to uploadFile

ezhil56x: On it

ezhil56x: Seems like uppy don't support custom functions for file upload. I couldn't find any walkarounds

ezhil56x:
@YousefED If we need uploadFile then we should create a similar uploadFileUppy for handing uppy uploads. Will that be fine?

YousefED:
> @YousefED If we need uploadFile then we should create a similar uploadFileUppy for handing uppy uploads. Will that be fine?

I think we just need something like this?

```typescript

async function uploadFileViaUppy(file: File) {
uppy.addFile(file);
await uppy.upload();
}

...
uploadfile: uploadFileViaUppy
```

pseudo code, but I hope this clarifies the intended behavior for you

ezhil56x:
> > @YousefED If we need uploadFile then we should create a similar uploadFileUppy for handing uppy uploads. Will that be fine?
> 
> I think we just need something like this?
> 
> ```typescript
> 
> async function uploadFileViaUppy(file: File) {
> uppy.addFile(file);
> await uppy.upload();
> }
> 
> ...
> uploadfile: uploadFileViaUppy
> ```
> 
> pseudo code, but I hope this clarifies the intended behavior for you

I tried this though, uppy.addFile() needs the file as blob type rather than file type. Creating a new uploadFileUppy like  uploadFile will be a good option, but uppy packages should be added for that which increases the size. I am confused about what to do. What do you suggest?

YousefED:
> > > @YousefED If we need uploadFile then we should create a similar uploadFileUppy for handing uppy uploads. Will that be fine?
> > 
> > 
> > I think we just need something like this?
> > ```ts
> > async function uploadFileViaUppy(file: File) {
> > uppy.addFile(file);
> > await uppy.upload();
> > }
> > 
> > ...
> > uploadfile: uploadFileViaUppy
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > pseudo code, but I hope this clarifies the intended behavior for you
> 
> I tried this though, uppy.addFile() needs the file as blob type rather than file type. Creating a new uploadFileUppy like uploadFile will be a good option, but uppy packages should be added for that which increases the size. I am confused about what to do. What do you suggest?

see https://stackoverflow.com/a/33855825

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -113,7 +113,7 @@ export type BlockNoteEditorOptions<
    * @param file The file that should be uploaded.
    * @returns The URL of the uploaded file OR an object containing props that should be set on the file block (such as an id)
    */
-  uploadFile: (file: File) => Promise<string | Record<string, any>>;
+  uploadFile: (file: File) => Promise<string | Record<string, any> | void>;
```
YousefED, line 116: these changes are wrong I think? `uploadFile` shouldn't be able to be void

examples/01-basic/11-file-uploading-uppy/App.tsx
```diff
@@ -27,45 +40,9 @@ export default function App() {
...
-  return (
-    <BlockNoteView editor={editor} filePanel={false}>
-      <FilePanelController
-        filePanel={() => {
-          return <Dashboard uppy={uppy} width={400} height={500} />;
-        }}
-      />
-    </BlockNoteView>
-  );
+  return <BlockNoteView editor={editor} />;
```
YousefED, line 47: Why did we remove the Dashboard? This was quite nice and part of the original requirements

packages/core/src/editor/BlockNoteEditor.ts
ezhil56x: Resolved

examples/01-basic/11-file-uploading-uppy/App.tsx
ezhil56x: It is not possible to use dashboard when we need to use the existing uploadFile which is in [UploadTab.tsx](https://github.com/TypeCellOS/BlockNote/blob/main/packages/react/src/components/FilePanel/DefaultTabs/UploadTab.tsx)

Either we can implement in this way [https://github.com/TypeCellOS/BlockNote/pull/842/files/06053acfc998b70845b8bf689e8a8241d5770496..a3ed8fc41d24ec00614e6bd72a83da8fcd0541b6#diff-48f7b0a43a51a764debb003cbd5e52382948064e2af0b1a76f1c2968b8c1d714](https://github.com/TypeCellOS/BlockNote/pull/842/files/06053acfc998b70845b8bf689e8a8241d5770496..a3ed8fc41d24ec00614e6bd72a83da8fcd0541b6#diff-48f7b0a43a51a764debb003cbd5e52382948064e2af0b1a76f1c2968b8c1d714)

Or implement in this way
[https://github.com/TypeCellOS/BlockNote/pull/842/files/56efb1f33d262fcd3ca203b8973bdc84d05c01db#diff-10e76f1f7e0513fa475c80821b41e5a38ca78b76e840007cee48c2e30f3cdda9](https://github.com/TypeCellOS/BlockNote/pull/842/files/56efb1f33d262fcd3ca203b8973bdc84d05c01db#diff-10e76f1f7e0513fa475c80821b41e5a38ca78b76e840007cee48c2e30f3cdda9)

YousefED: We can use the second option, but still implement uploadFile. In that case for the panel, the Dashboard and direct uppy upload is used.

For other cases (such as file drop), the editor will call `uploadFile` and your implementation will make sure files are uploaded via uppy.

Hope this makes sense why I think we need both

ezhil56x: @YousefED I would suggest to go with 2nd only as it supports both file drop as well as selecting files for uploading. Implementing both on an example wont be possible. If this is fine, I can update my PR with the 2nd implementation

YousefED:
@ezhil56x I've updated the PR. As you can see I've been able to use:
- both `uploadFile` and `Dashboard`
- make sure the dashboard updates the correct block

I've also added some documentation, cleaned up the code a bit, and added some Uppy plugins so that it's a nice demo. Let me know what you think!

examples/01-basic/11-file-uploading-uppy/App.tsx
YousefED: should be fixed now

examples/01-basic/11-file-uploading-uppy/App.tsx
YousefED: fixed, as you can see in my changes it's certainly possible

examples/01-basic/11-file-uploading-uppy/App.tsx
YousefED: see updated example code!

ezhil56x:
@YousefED 
Great!
Tested, seems to be working good as expected

YousefED (approved):

Issue #655, opened 2024-03-19
Table doesn't save columns width after refresh


serhiiAntonyshynFulcrumRocks:
**Describe the bug**
Table doesn't save columns width after refresh

**To Reproduce**
I resize the columns width for editor table. 
After that I save the content to the api ( I save all the content that editor return in editor.topLevelBlocks from onEditorContentChange ), but after refresh table has the default width for column
It happens because the styles object is empty and there is no information for updated column width

How can I pass column width from editor table to styles object?

**Misc**
- Node version: 20.9.0
- Package manager: npm 
- Browser: Chrome
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

jehaad1:
any updates?

PR #846, opened 2024-06-16
feat: persist table resize

halimt92:
When resizing a table, prosemirror changes the colgroup's col widths without transmitting this data to the table nodes. This is related to [this issue](https://github.com/TypeCellOS/BlockNote/issues/655). The fix consists of listening to the mouse up event and checking for a change in any of the columns from prosemirror. If there is a change that should affect a cell, we update the cells of the table. This will re-render the table. We persist the width information in a property for the node. Thus we had to change the typing of cell contents.

### DIFF
--- a/packages/core/src/api/nodeConversions/nodeConversions.ts
+++ b/packages/core/src/api/nodeConversions/nodeConversions.ts
@@ -172,13 +172,14 @@ export function tableContentToNodes<
     const columnNodes: Node[] = [];
     for (const cell of row.cells) {
       let pNode: Node;
-      if (!cell) {
+      if (!cell || cell.length === 0) {
         pNode = schema.nodes["tableParagraph"].create({});
       } else if (typeof cell === "string") {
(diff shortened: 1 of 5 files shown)


halimt92:
@matthewlipski Greetings.

My team and I are testing the library, and we encountered a blocker (basically the issue that I linked to the PR). 

Basically our aim is to add images in tables, but without being able to resize, that would be pointless.

I was able to fix it using the method described above, however, I am still new to this library and new to prosemirror, therefore I am not sure if this is the approach that you like to take regarding this issue.

Cheers.

matthewlipski:
Thanks for your submission! The code is a great start point, and after checking it out I found that a lot of the code can be simplified without losing functionality. Unfortunately I don't have push rights to your repo, so I just branched off of this and made #856 to make things easier. Closing this PR since #856 contains the same commits and more.

halimt92:
@matthewlipski thanks for your feedback, it's good to know that we understood the concepts well! I will continue proposing new PRs whenever we find limitations that we encounter in the future.

Cheers

PR #847, opened 2024-06-17
fix: ShadCN Tailwind scoping

### LINKED ISSUE #845: Isolate the style with a parent class

import  @blocknote/shadcn/style.css to project will cause unexpected sideeffect. this is what I have done:

```makefile
gen:
	echo '.BlockNoteView { ' > ./styles.css
	curl -s 'https://cdn.jsdelivr.net/npm/@blocknote/react@0.14.1/dist/style.css' >> ./styles.css
	curl -s 'https://cdn.jsdelivr.net/npm/@blocknote/shadcn@0.14.1/dist/style.css' >> ./styles.css
	echo ' }' >> ./styles.css

```

1. add a parent class
2. use css module
3. or some css-in-js

**Additional context**

**Bonus**
Not yet

matthewlipski:
The `shadcn` package uses TailwindCSS, so when building a project that imports both `@blocknote/shadcn` and Tailwind, it creates duplicate CSS declarations. This isn't a huge problem on its own, but when building projects using TailwindCSS, only the rules that are actually used are included due to optimization. This can lead to cases where the Tailwind classes created by the `shadcn` package take priority over those created by the parent app, and cause styling to break as they don't include certain conditional classes such as `min-width`.

This PR makes the Tailwind classes used by `@blocknote/shadcn` have a `bn-` prefix, which basically scopes them to the editor only and stops them from conflicting with those used outside the editor.

This PR also scopes the styles of each UI library to the corresponding editor instead of to all editors. This fixes a very niche edge case where you have 2 editors using different UI libs, and also the examples, since those can have styles sheets from other examples loaded.

Closes #845

### DIFF
--- a/packages/ariakit/src/index.tsx
+++ b/packages/ariakit/src/index.tsx
@@ -105,7 +105,7 @@ export const BlockNoteView = <
   return (
     <ComponentsContext.Provider value={components}>
       <BlockNoteViewRaw
-        className={mergeCSSClasses("bn-aria-kit", className || "")}
+        className={mergeCSSClasses("bn-ariakit", className || "")}
         {...rest}
       />
     </ComponentsContext.Provider>
--- a/packages/ariakit/src/index.tsx
+++ b/packages/ariakit/src/index.tsx
@@ -1,4 +1,9 @@
-import { BlockSchema, InlineContentSchema, StyleSchema } from "@blocknote/core";
+import {
+  BlockSchema,
+  InlineContentSchema,
+  mergeCSSClasses,
+  StyleSchema,
+} from "@blocknote/core";
 import {
   BlockNoteViewRaw,
   Components,
@@ -95,9 +100,14 @@ export const BlockNoteView = <
 >(
   props: ComponentProps<typeof BlockNoteViewRaw<BSchema, ISchema, SSchema>>
 ) => {
+  const { className, ...rest } = props;
+
   return (
     <ComponentsContext.Provider value={components}>
-      <BlockNoteViewRaw {...props} />
+      <BlockNoteViewRaw
+        className={mergeCSSClasses("bn-aria-kit", className || "")}
+        {...rest}
+      />
     </ComponentsContext.Provider>
   );
 };
--- a/packages/shadcn/src/index.tsx
+++ b/packages/shadcn/src/index.tsx
@@ -121,7 +121,7 @@ export const BlockNoteView = <
     <ShadCNComponentsContext.Provider value={componentsValue}>
       <ComponentsContext.Provider value={components}>
         <BlockNoteViewRaw
-          className={mergeCSSClasses("bn-shad-cn", className || "")}
+          className={mergeCSSClasses("bn-shadcn", className || "")}
           {...rest}
         />
       </ComponentsContext.Provider>
--- a/packages/shadcn/src/index.tsx
+++ b/packages/shadcn/src/index.tsx
@@ -107,8 +107,8 @@ export const BlockNoteView = <
 
   const componentsValue = useMemo(() => {
     return {
-      ...shadCNComponents,
       ...ShadCNDefaultComponents,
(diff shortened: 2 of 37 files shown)


YousefED (approved):
Nice. The only downside I see is that when providing your own ShadCN components, you still need the blocknote shadcn stylesheet in this case, because we pass some `bn-` prefixed classes. Not sure if we can find an approach that doesn't have this though?

packages/shadcn/src/lib/utils.ts
```diff
@@ -1,5 +1,9 @@
 import { type ClassValue, clsx } from "clsx";
-import { twMerge } from "tailwind-merge";
+import { extendTailwindMerge } from "tailwind-merge";
+
+const twMerge = extendTailwindMerge({
```
YousefED, line 4: Why do we need this? Aren't all classnames we pass to cn already prefixed?

matthewlipski: Yes, this is just so that `twMerge` recognizes that the `bn-` prefixed Tailwind classes are in fact still Tailwind classes, otherwise it treats them like regular classes so it doesn't combine and order them properly.

e.g. if you do `twMerge("w-10", "w-16")` you'll get `"w-16"`, but `twMerge("bn-w-10", "bn-w-16")` you'll get `"bn-w-10 bn-w-16"` unless you extend it to recognize the `bn-` prefix.

YousefED: clear! maybe add this as a comment

packages/ariakit/src/index.tsx
```diff
@@ -95,9 +100,14 @@ export const BlockNoteView = <
 >(
   props: ComponentProps<typeof BlockNoteViewRaw<BSchema, ISchema, SSchema>>
 ) => {
+  const { className, ...rest } = props;
+
   return (
     <ComponentsContext.Provider value={components}>
-      <BlockNoteViewRaw {...props} />
+      <BlockNoteViewRaw
+        className={mergeCSSClasses("bn-aria-kit", className || "")}
```
YousefED, line 108: ariakit uses "ariakit" as one word, let's use that?

packages/shadcn/src/index.tsx
```diff
@@ -115,7 +120,10 @@ export const BlockNoteView = <
   return (
     <ShadCNComponentsContext.Provider value={componentsValue}>
       <ComponentsContext.Provider value={components}>
-        <BlockNoteViewRaw {...rest} />
+        <BlockNoteViewRaw
+          className={mergeCSSClasses("bn-shad-cn", className || "")}
```
YousefED, line 124: same comment as for ariakit, shadcn is also 1 word afaik

YousefED (commented):
I suppose the latest change to the container rules is because styles were conflicting when changing examples in the playground, right? Curious; which rules were conflicting? Because most of the rules already were prefixed (`bn-ak-*`)?

matthewlipski:
> I suppose the latest change to the container rules is because styles were conflicting when changing examples in the playground, right? Curious; which rules were conflicting? Because most of the rules already were prefixed (`bn-ak-*`)?

Yeah both playground and sometimes with docs though it was inconsistent, as NextJS doesn't seem to load stylesheets in a  predefined order in the built env so you would sometimes see conflicting styles there too.

Regarding the conflicting styles that's a good point, we really only need the `bn-xyz ...` prefix for more generic rules like this one in `mantine`'s styles:
```css
/* Toolbar styling */
.bn-toolbar {
  background-color: var(--bn-colors-menu-background);
  border: var(--bn-border);
  border-radius: var(--bn-border-radius-medium);
  box-shadow: var(--bn-shadow-medium);
  flex-wrap: nowrap;
  gap: 2px;
  padding: 2px;
  width: fit-content;
}
```

YousefED (approved):

PR #849, opened 2024-06-18
 Update: ShadowRoot support

dcadavez66:
1. Core package updated to ensure BlockNote can be used in ShadowRoots;
2. pasteExtension.ts updated to also support image pasting into the editor;
3. Updated version for the Arabic translation file;

TODO:The changes related to the first point primarily involve replacing the use of _document_ with the _root_ property, which can be either Document or ShadowRoot. This property is accessed from pmView. In files where pmView is not available, the _root_ property should be accessed from _editor._tiptapEditor.view.root_.
To ensure consistency and ease for future modifications, there should be a standardized method for accessing this value. This will enable everyone making future changes to use this standard method instead of directly accessing the document.


packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -283,22 +286,42 @@ export class SideMenuView<
       this.pmView.dom.firstChild! as HTMLElement
     ).getBoundingClientRect().x;
 
-    document.body.addEventListener("drop", this.onDrop, true);
-    document.body.addEventListener("dragover", this.onDragOver);
+    // document.body.addEventListener("drop", this.onDrop, true);
+    this.pmView.root.addEventListener(
+      "drop",
+      this.onDrop as EventListener,
```
YousefED, line 292: Can we make this work without the cast?

YousefED (changes requested):
Great! Thanks. Two questions:

- can you move Arabic translations update and the Paste handler to separate PRs?
- Can you remove the commented out old code? We don't want the old code as comments in the codebase, and we can just see the differences here on github

After this we'll double check that all references to `document` have indeed been updated to use `view.root` instead

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
dcadavez66: Without the cast no, there is another way where we explicitly type it with the type of the handler. Like so:

this.pmView.root.addEventListener(
      "drop",
      (event: Event) => this.onDrop(event as DragEvent),
      true
    );
    
Let me know the one you prefer to keep as standard and I can do the necessary updates!

packages/core/src/extensions/Placeholder/PlaceholderPlugin.ts
```diff
@@ -16,7 +16,12 @@ export const PlaceholderPlugin = (
       if (nonce) {
         styleEl.setAttribute("nonce", nonce);
       }
-      document.head.appendChild(styleEl);
+      if (editor._tiptapEditor.view.root instanceof ShadowRoot) {
+        editor._tiptapEditor.view.root.append(styleEl);
+      } else if (editor._tiptapEditor.view.root instanceof Document) {
```
matthewlipski, line 21: Don't need `else if` here, `else` should be fine given the typing. Same goes in a few other places where you check `instanceof`.

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -146,12 +146,12 @@ function setDragImage(view: EditorView, from: number, to = from) {
   dragImageElement.className =
     dragImageElement.className + " bn-drag-preview " + inheritedClasses;
 
-  document.body.appendChild(dragImageElement);
+  view.root.appendChild(dragImageElement);
```
matthewlipski, line 149: Causes an error on drag & drop, need to check if `view.root instanceof ShadowRoot` otherwise append to `view.root.body`

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
matthewlipski: Can't we also conditionally add the listeners only if `view.root instanceof Document`? Since shadow roots don't support listeners anyway

matthewlipski (changes requested):
Looks good! Just a few changes that need to be made and we can merge:) Also make sure to check the comment regarding event listener type casting

matthewlipski:
I added most of the changes mentioned in my review, only thing left is the type casting comment! @dcadavez66 @YousefED

YousefED (approved):

matthewlipski (approved):

PR #854, opened 2024-06-20
feat: Add Emoji Picker Option

### LINKED ISSUE #808: Integrate emoji picker
An emoji picker that works similar to the one in Notion:

emoji

Main requirements:
[ ] Easy to update with new emojis, possibly use a library like https://www.npmjs.com/package/emoji-mart, https://www.npmjs.com/package/react-input-emoji, or different
[ ] As seen in the gif above, the focus should remain on the editor, while the user filters. Perhaps it needs to be implemented as a [custom suggestion menu](https://www.blocknotejs.org/docs/ui-components/suggestion-menus), but that currently doesn't have support for keyboard nav to left / right, so that would need to be added. It might be necessary to create your own version of `SuggestionMenuController` for this (and share common code nicely)
[ ] Should be accessible (both keyboard navigation and screen reader support would be ideal)
[ ] Component should be selectable both via the slash menu, and when the user types a trigger character (`:`)

This will require to get up to speed with the codebase and understand how suggestions work internally, so I don't recommend this issue for beginners

hunxjunedo:
/claim #808 
fixes #808 
this pull request adds the emoji picker option which is accessible by both the trigger character (:) and the slash menu.

### Recordings
[Screencast from 19-06-2024 20:11:04.webm](https://github.com/TypeCellOS/BlockNote/assets/89797440/8c6db2cd-a106-43db-acc6-2dd9522aade5)

https://github.com/TypeCellOS/BlockNote/assets/89797440/aa1637e1-8bf1-47b6-a3a6-1dea329103c8

### Approach
I started with a custom suggestion menu as suggested, with custom component and `getItems` function. The function fetches all the emojis from the Emojimart at initial render, and then returns the emojis as `items` on every search. This approach may sound unfamiliar as using emoji-mart as the picker would have been preferable, however the main issue with almost all emoji-pickers is that either they don't allow filtering out the emojis shown, or even if indirectly allowed (like emoji-mart allows to hide some emoijs ), they don't update dynamically to reflect those changes. Therefore I was left with an option of creating a custom picker, which worked pretty fine. I then proceeded to add the functionality in the `useSuggestionMenuKeyboardNavigation.ts` file to handle all sorts of navigation **forward and backward arrows** if it's for the `emojiMenu`.

Secondly, the task was to make this menu accessible by the slash-menu, therefore I added a custom option to slash menu for emojis, which was similar to other options except that this option doesn't insert anything into the editor, instead it just simulates a `keydown` when clicked, which is listened by the `onKeyDown` function specified in the plugin `props` in the `suggestionPlugin.ts` file, which then carries the logic and opens the `emojiMenu` 

### Tests

- The code is vigorously tested on multiple network throttling levels to ensure that the emojis don't take too much time to load.
- All types of examples are used, including `showing document Json` and the one with `mentions menu`
- All types of keyboard navigation is tested, including right and left arrows.


packages/core/src/extensions/SuggestionMenu/SuggestionPlugin.ts
```diff
@@ -248,6 +248,27 @@ export class SuggestionMenuProseMirrorPlugin<
       },
 
       props: {
+        //STEP 6: the final thing is to listen to the simulated keydown event generated by the onClick of emoji option in slash menu
```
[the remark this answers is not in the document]
YousefED: Update after reviewing further: I suppose this is for the `onItemClick` slash menu handler below. It's an interesting hack (cool that you got it working), but I think it's a little bit too hacky. imo we should create an API that hides / shows specific suggestionmenus so that we can make this a lot more "idiomatic".

This would also solve some related requests on Discord (https://discord.com/channels/928190961455087667/1204207715929296958). @matthewlipski maybe you could help with this?

packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
```diff
@@ -66,12 +68,42 @@ export function SuggestionMenuController<
...
+        charCode: 0,       // character code value (unsigned long), optional
+        bubbles: false,    // bubbles flag (boolean), optional
+        cancelable: false, // cancelable flag (boolean), optional
+        composed: false,   // composed flag (boolean), optional
+        ctrlKey: false,    // control key flag (boolean), optional
+        altKey: false,     // alt key flag (boolean), optional
+        shiftKey: false,   // shift key flag (boolean), optional
+        metaKey: false     // meta key flag (boolean), optional
+      });
+      editor.domElement.dispatchEvent(keyboardEvent);
```
YousefED, line 89: see comment above; a bit hacky

packages/react/src/components/SuggestionMenu/emojisMenu.tsx
```diff
@@ -0,0 +1,20 @@
...
+  items: string[],
+  emojiInsert: (item : never) => void,
+  selectedIndex: number
+}
+  export default function EmojiMenu({items, emojiInsert, selectedIndex} : Props) : JSX.Element {
+  //this is the component which renders emoji picker with the emojis searched.
+
+
+    return (
+    <div style={{display: 'grid', justifyItems: 'center', background: '#1d1d1d', borderRadius: 20, gap: 7, padding: 20, maxHeight: '30vh', minWidth: '30vh', overflowY: 'scroll', gridTemplateColumns: 'repeat(10, 1fr)'}}>
```
YousefED, line 12: we should move these styles to CSS imo similar to how we do this for other components (cc @matthewlipski, what's the best approach for this?)

packages/react/src/components/SuggestionMenu/hooks/useSuggestionMenuKeyboardNavigation.ts
```diff
@@ -7,7 +7,8 @@ export function useSuggestionMenuKeyboardNavigation<Item>(
   editor: BlockNoteEditor<any, any, any>,
   query: string,
   items: Item[],
-  onItemClick?: (item: Item) => void
+  isEmojiMenu: any,
```
YousefED, line 10: Great job with the keyboard functionality. I'd like to propose to make cleanly separate this hook of "emoji" related concerns. We could do this by changing `isEmojiMenu` to a property `gridWidth`, and set that to 1 for the suggestionmenu or 10 for the emoji menu

packages/react/src/editor/BlockNoteDefaultUI.tsx
```diff
@@ -15,8 +17,47 @@ export type BlockNoteDefaultUIProps = {
...
+async function search(value: string) {
+  if(value == ''){
+    //Don't do unnecessary linear search if no search query
+    return Object.values(Data.emojis).map((emoji : any)=>(
+      emoji.skins[0].native
+    ))
+  }
+  const emojisToShow : any[] = []
+  //begin the linear search
+  Object.values(Data.emojis).forEach((emoji : any)=>{
```
YousefED, line 39: Maybe use this? https://github.com/missive/emoji-mart?tab=readme-ov-file#%EF%B8%8F%EF%B8%8F-headless-search

packages/react/src/editor/BlockNoteDefaultUI.tsx
```diff
@@ -29,7 +70,16 @@ export function BlockNoteDefaultUI(props: BlockNoteDefaultUIProps) {
       {props.formattingToolbar !== false && <FormattingToolbarController />}
       {props.linkToolbar !== false && <LinkToolbarController />}
       {props.slashMenu !== false && (
+        <>
         <SuggestionMenuController triggerCharacter="/" />
+        {/*STEP 1:  the suggestion menu for the emojis, with custom component and getItems */}
+        <SuggestionMenuController
```
YousefED, line 76: Couple of things:
- I think we should get rid of the `isEmoji` property and just use an `onItemClick` handler here instead of having the logic to insert emojis in the `SuggestionMenuController`
- let's put this behind a flag (`props.emojiMenu` instead of `props.slashMenu`)
- bit of cleanup, let's separate the Emoji related code to a `<EmojiMenuController>` (init, search, etc) so that we don't have so much emoji related code in `BlockNoteDefaultUI`?

packages/react/src/components/SuggestionMenu/emojisMenu.tsx
YousefED, line 12: Let's make sure this works well in both dark and light modes

YousefED (commented):
Great stuff @hunxjunedo! Nice that you got the keyboard navigation to work. I left a number of comments which I think will be helpful to make the code clean and a bit more generic.

YousefED:
@flipvh wdyt of this? You can see a preview here: https://blocknote-git-fork-hunxjunedo-main-typecell.vercel.app/basic/minimal

The UX is very much based on Notion. There are some things to be improved (i.e.: dark / light mode), but I think the basics look great. wdyt?

hunxjunedo:
Acknowledged the comments, updating on that soon 👍

packages/core/src/blocks/Emoji.tsx
```diff
@@ -0,0 +1,26 @@
+import { createInlineContentSpec } from "../schema";
+ 
+
+export const Emoji = createInlineContentSpec(
```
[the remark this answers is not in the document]
hunxjunedo: I need your opinion on that. Both are fine to me, not much of an issue.

packages/react/src/editor/BlockNoteDefaultUI.tsx
hunxjunedo: thanks for reminding, I intended to use it but just forgot.

packages/react/src/components/SuggestionMenu/emojisMenu.tsx
hunxjunedo: added the styles in editor/styles.css as it had direct access to the CSS variables, which also helped in implementing dark mode.

packages/react/src/editor/BlockNoteDefaultUI.tsx
hunxjunedo: the thing is, despite the simulation of colon key, the rest of the stuff is still working as a custom suggestion menu. The only reason I added this simulation was to make sure the emoji menu was accessible thru the slash menu, otherwise it was working perfect as a custom suggestion menu handled by the `suggestionMenuController`. Then when I moved towards the slash menu thing, I first tried rendering the colon `inline` to trigger the emoji menu, which didn't work, leaving me with the option of simulation. But still it couldn't trigger the `handleTextInput` as it only responds when the user directly inputs (see prosemirror docs). Therefore I came up with a hack to first trigger the `onKeyPress`, then call the rest of the logic which is responsible for showing the menu. Both things have to exist in order to make it work. It will require collaboration and some time to take a different approach, may even require a separate issue.

YousefED: @matthewlipski could you help with a programmatic way to open the menu?

flipvh:
I like it too! Maybe in a second round, the use of emoji option in the command menu can be made more intuitive. Now it just renders a `:` without any feedback on what to do. Perhaps it would be nice if it can open the emoij suggestions menu too? Great work!

hunxjunedo:
@flipvh if I don't get you wrong, do you mean that the emoji option in the slash menu doesn't open the Emoji Menu ? Because I've tested it and can be seen in the demo that it works fine, is that a browser issue ?

hunxjunedo:
@YousefED it's been almost a week, what's the progress of this PR ? Are we moving towards merging it ? Is this a low-priority ?

packages/core/src/blocks/Emoji.tsx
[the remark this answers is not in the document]
jueinin: @YousefED it maybe better for insert a emoji inline content.

it's easier to extend. for example, we can add a hover, click feature.

matthewlipski:
I'll look into this PR tomorrow so we can try to finish it up and merge👍

package.json
```diff
@@ -33,5 +33,10 @@
     "prepublishOnly": "npm run build && cp README.md packages/core/README.md && cp README.md packages/react/README.md",
     "postpublish": "rm -rf packages/core/README.md && rm -rf packages/react/README.md",
     "clean": "lerna run --stream clean"
+  },
```
matthewlipski, line 36: Should be in the `react` package's `package.json`, not here. Or ideally in `mantine`, `shadcn`, and `ariakit`, once you extract the emoji components to those.

packages/core/src/blocks/Emoji.tsx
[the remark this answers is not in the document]
matthewlipski: +1 I think inline content is the better solution for the reasons mentioned by @jueinin

packages/core/src/blocks/Emoji.tsx
```diff
@@ -0,0 +1,26 @@
...
+  //STEP 4: this component recieves an emoji, and insets it in the line
+  {
+    type: "emoji",
+    propSchema: {
+      emoji: {
+        default: "Unknown",
```
matthewlipski, line 10: Should change default to just `""`

packages/core/src/blocks/Emoji.tsx
```diff
@@ -0,0 +1,26 @@
+import { createInlineContentSpec } from "../schema";
```
matthewlipski, line 1: You should move this file to a new directory called `inlineContent`, and structure it the same way that `blocks` is structured.

packages/core/src/blocks/defaultBlocks.ts
```diff
@@ -30,6 +30,7 @@ import { FileBlock } from "./FileBlockContent/FileBlockContent";
 import { ImageBlock } from "./ImageBlockContent/ImageBlockContent";
```
matthewlipski, line 30: Looking at this, we will probably want to move this file out of the `blocks` directory and rename it to `defaultSchema`. It's fine for now though, just leaving this as a note to sort this out later.

packages/react/src/editor/BlockNoteDefaultUI.tsx
matthewlipski: Sure! If you look in the side menu, notice we can press the `+` button to open the slash menu without having to press the `/` key. This is done with the following code:
```tsx
this.pmView.dispatch(
  this.pmView.state.tr.scrollIntoView().setMeta(suggestionMenuPluginKey, {
    triggerCharacter: "/",
    fromUserInput: false,
  })
);
```
All you should need to do to replicate this for the emoji menu is change `triggerCharacter: "/"` to `triggerCharacter: ":"`. Then we can get rid of all the extra code for simulating the colon input.

packages/core/src/blocks/Emoji.tsx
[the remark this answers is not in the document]
matthewlipski: +1 I think inline content is the better solution for the reasons mentioned by @jueinin

packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
```diff
@@ -66,12 +68,42 @@ export function SuggestionMenuController<
...
+    icon: <MdEmojiEmotions />,
+  });
+
   const getItemsOrDefault = useMemo(() => {
     return (
       getItems ||
       ((async (query: string) =>
         filterSuggestionItems(
-          getDefaultReactSlashMenuItems(editor),
+          //STEP 6: SHOW THE EMOJI OPTION IN THE SLASH MENU
```
matthewlipski, line 105: You should move the emoji slash menu item to `getDefaultSlashMenuItems`/`getDefaultReactSlashMenuItems`. Should be able to copy the same structure as the existing items.

packages/core/src/blocks/Emoji.tsx
matthewlipski, line 10: Should change default to just `""`

matthewlipski (changes requested):
Nice! Overall works & looks great - in addition to the review comments, there are a couple of broader changes that I think we still need:
- Option to disable emoji menu in `BlockNoteDefaultUI`, like with the slash menu.
- Apply the fix to open the emoji menu programmatically, see [here](https://github.com/TypeCellOS/BlockNote/pull/854#discussion_r1659127238),  and remove code for the simulated input approach.
- General refactor to decouple the emoji menu from the suggestion menu.
  - Move the slash menu item to `defaultSlashMenuItems`/`reactDefaultSlashMenuItems`.
  - Like @YousefED suggested, we should separate the `useSuggestionMenuKeyboardNavigation` into 2 hooks, the second being `useGridSuggestionMenuKeyboardNavigation`. Likewise the `SuggestionMenuController` and `SuggestionMenuWrapper` should take a `grid: boolean` prop to swap between them.
  - We should also rename the `EmojiMenu` component to a more general `GridSuggestionMenu`, that way we can hook it up to the `componentsContext` later (don't worry about the components context thing for now, we'll sort it out later).

matthewlipski:
So I implemented the fix to opening the menu and started with the refactor, this should give you a good starting point to work off for the remaining changes👍

hunxjunedo:
@matthewlipski thanks for taking charge and the valuable reviews. It's night here in my time zone, I'll do what's necessary tomorrow morning.

hunxjunedo:
@matthewlipski I've addressed the requested changes, can you please finalize.

iandreileo:
Hi Folks, any news here? :D

matthewlipski (approved):

matthewlipski:
Ok, I cleaned up the code and made it so that the emoji picker can be changed for the `ariakit`/`shadcn` packages. There are still some areas that could be improved:

- `ariakit`/`shadcn` implementation is the same as `mantine` for the moment, so there is a fair bit of duplicate code. However, since the grid suggestion menu is hooked up to the `ComponentsContext`, this should be very easy to change if we want and it now follows the correct architecture.
- Emojis aren't grouped into labels, unlike Notion. I think this would take quite a lot of extra effort esp to change regarding keyboard navigation, so don't think it's worth doing now.
- Loader & empty item implementations are also missing atm. This is a more major UX issue that may need to be fixed before merging.
- The grid width in CSS is tied to the keyboard navigation logic, which is not great for obvious reasons.
- There's inline content code in the `blocks` directory in `core`. I guess this was kind of an issue already that has gotten worse now that we have a third inline content type.
- Docs need updating.

@YousefED which of these do you think we still need to sort out before merging?

packages/ariakit/src/suggestionMenu/GridSuggestionMenuItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    if (overflow === "top") {
+      itemRef.current.scrollIntoView(true);
+    } else if (overflow === "bottom") {
+      itemRef.current.scrollIntoView(false);
+    }
+  }, [isSelected]);
+
+  return (
+    <p
```
YousefED, line 30: I don't think `<p>` makes sense for this

packages/core/package.json
```diff
@@ -54,6 +54,8 @@
     "clean": "rimraf dist && rimraf types"
   },
   "dependencies": {
+    "@emoji-mart/data": "^1.2.1",
+    "@emoji-mart/react": "^1.1.1",
```
YousefED, line 58: emoji-mart/react can't be a core dependency I think?

packages/mantine/src/suggestionMenu/GridSuggestionMenuItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }, [isSelected]);
+
+  return (
+    <p
+      className={className}
+      ref={mergeRefs(ref, itemRef)}
+      id={id}
+      role="option"
+      onClick={onClick}
+      // className={index === selectedIndex ? "grid-item-selected" : ""}
```
YousefED, line 38: remove commented out code?

packages/react/src/components/SuggestionMenu/types.tsx
```diff
@@ -18,5 +24,5 @@ export type SuggestionMenuProps<T> = {
   loadingState: "loading-initial" | "loading" | "loaded";
   selectedIndex: number | undefined;
   onItemClick?: (item: T) => void;
-  emojiInsert?: (emoji: never) => void
+  emojiInsert?: (emoji: never) => void;
```
YousefED, line 27: does this need to be here?

YousefED (commented):
Great! And thanks for clear description + follow up items.

> - ariakit/shadcn implementation is the same as mantine for the moment, so there is a fair bit of duplicate code. However, since the grid suggestion menu is hooked up to the ComponentsContext, this should be very easy to change if we want and it now follows the correct architecture.

I think ok for now

> - Emojis aren't grouped into labels, unlike Notion. I think this would take quite a lot of extra effort esp to change regarding keyboard navigation, so don't think it's worth doing now.

Also ok for now

> - Loader & empty item implementations are also missing atm. This is a more major UX issue that may need to be fixed before merging.

Agree this needs to be fixed. @hunxjunedo can you add this? The example code is already there

> - The grid width in CSS is tied to the keyboard navigation logic, which is not great for obvious reasons.

Can we think of a simple solution? For example, using a CSS variable or data- attribute so that this information has a single source of truth? (maybe timebox to half an hour)

> - There's inline content code in the blocks directory in core. I guess this was kind of an issue already that has gotten worse now that we have a third inline content type.

Let's discuss IC or not @matthewlipski (on call)

> - Docs need updating.

Must: explain emoji options
Nice to have: explain GridSuggestionMenu

> - Need unit/e2e tests.

Yes, an e2e test for this would be great, can you point @hunxjunedo to a test that he can take as base?

abdullah1308:
When can I expect to use this feature in blocknote ?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1038,4 +1038,16 @@ export class BlockNoteEditor<
       this._tiptapEditor.off("selectionUpdate", cb);
     };
   }
+
+  public openSelectionMenu(triggerCharacter: string) {
```
YousefED, line 1042: `suggestionMenu`?

and should we document this method as well?

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -649,14 +648,8 @@ export class SideMenuView<
...
-    // Focuses and activates the suggestion menu.
-    this.pmView.focus();
-    this.pmView.dispatch(
-      this.pmView.state.tr.scrollIntoView().setMeta(suggestionMenuPluginKey, {
-        triggerCharacter: "/",
-        fromUserInput: false,
-      })
-    );
+    // Focuses and activates the slash menu.
+    this.editor.openSelectionMenu("/");
```
YousefED, line 652: nice

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1038,4 +1038,16 @@ export class BlockNoteEditor<
...
+    this.prosemirrorView.focus();
+    this.prosemirrorView.dispatch(
+      this.prosemirrorView.state.tr
+        .scrollIntoView()
+        .setMeta(this.suggestionMenus.plugin, {
```
YousefED, line 1047: doesn't need to be `suggestionMenuPluginKey`?

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -0,0 +1,37 @@
+import data, { Emoji, EmojiMartData } from "@emoji-mart/data";
+import { init, SearchIndex } from "emoji-mart";
+
+import { BlockSchema, InlineContentSchema, StyleSchema } from "../../schema";
+import { BlockNoteEditor } from "../../editor/BlockNoteEditor";
+import { checkDefaultInlineContentTypeInSchema } from "../../blocks/defaultBlockTypeGuards";
+import { DefaultGridSuggestionItem } from "./DefaultGridSuggestionItem";
+
+const emojiMartData = data as EmojiMartData;
```
YousefED, line 9: TBD: this might be a significant increase in package size

packages/core/src/i18n/locales/en.ts
```diff
@@ -104,6 +104,12 @@ export const en = {
       aliases: ["file", "upload", "embed", "media", "url"],
       group: "Media",
     },
+    emoji: {
+      title: "Emoji",
+      subtext: "Used for inserting an emoji",
```
YousefED, line 109: I think we should get rid of the "Used for" part in these sentences. Imo it doesn't add any value.

Compare to notion:

image

packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
```diff
@@ -14,33 +14,70 @@ import { useUIPluginState } from "../../hooks/useUIPluginState";
...
+  DefaultReactSuggestionItem,
+  SuggestionMenuProps,
+} from "./types";
 
 type ArrayElement<A> = A extends readonly (infer T)[] ? T : never;
 
 type ItemType<GetItemsType extends (query: string) => Promise<any[]>> =
   ArrayElement<Awaited<ReturnType<GetItemsType>>>;
 
+// There's a lot going on with the typing here but tl;dr:
```
YousefED, line 30: To review (Y)

docs/pages/docs/ui-components/suggestion-menus.mdx
```diff
@@ -31,20 +31,56 @@ Passing `slashMenu={false}` to `BlockNoteView` tells BlockNote not to show the d
 
 `getItems` should return the items that need to be shown in the Slash Menu, based on a `query` entered by the user (anything the user types after the `triggerCharacter`).
 
-### Replacing the Suggestion Menu Component
+### Replacing the Slash Menu Component
```
YousefED, line 34: I'm afraid extending this doc with grid + emoji will add to many conepts in one

YousefED (commented):
(review in progress)

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -0,0 +1,37 @@
...
+  query: string
+): Promise<DefaultGridSuggestionItem[]> {
+  if (!checkDefaultInlineContentTypeInSchema("text", editor)) {
+    return [];
+  }
+
+  const emojisToShow =
+    query === ""
+      ? Object.values(emojiMartData.emojis)
+      : ((await SearchIndex.search(query)) as Emoji[]);
```
YousefED, line 27: I got this error at least once:

image

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  return emojisToShow.map((emoji: Emoji) => ({
+    id: emoji.skins[0].native,
+    onItemClick: () => {
+      if (checkDefaultInlineContentTypeInSchema("text", editor)) {
```
YousefED, line 32: do we need this check?

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export async function getDefaultEmojiPickerItems<
+  BSchema extends BlockSchema,
+  I extends InlineContentSchema,
+  S extends StyleSchema
+>(
+  editor: BlockNoteEditor<BSchema, I, S>,
+  query: string
+): Promise<DefaultGridSuggestionItem[]> {
+  if (!checkDefaultInlineContentTypeInSchema("text", editor)) {
```
YousefED, line 20: do we need this check?

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
YousefED: package size grows from 330kb to 440kb. Not ideal, but also not a dealbreaker at this moment.

It could be improved by adding support for dynamic data loading (https://github.com/missive/emoji-mart?tab=readme-ov-file#-data)

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -0,0 +1,37 @@
+import data, { Emoji, EmojiMartData } from "@emoji-mart/data";
+import { init, SearchIndex } from "emoji-mart";
+
+import { BlockSchema, InlineContentSchema, StyleSchema } from "../../schema";
+import { BlockNoteEditor } from "../../editor/BlockNoteEditor";
+import { checkDefaultInlineContentTypeInSchema } from "../../blocks/defaultBlockTypeGuards";
+import { DefaultGridSuggestionItem } from "./DefaultGridSuggestionItem";
+
+const emojiMartData = data as EmojiMartData;
+init({ emojiMartData });
```
YousefED, line 10: quite a lot happens in this init function. I would only call this the first time `getDefaultEmojiPickerItems` is called

hunxjunedo:
@matthewlipski please let me know if there's anything I have to do.

packages/core/src/editor/BlockNoteEditor.ts
matthewlipski: It's functionally the same thing, `setMeta` takes either a plugin key or the plugin itself

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
matthewlipski: In theory yes as we can't insert an emoji unless `text` is in the inline content schema. In practice, I don't think pretty much anyone is likely to use the editor without the `text` IC. But eh imo it's kind of the same as with the paragraph block in the slash menu items - we also check if that's in the schema (despite 99% of people using the paragraph block) and there's not really any issues there.

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
matthewlipski: In theory yes as we can't insert an emoji unless text is in the inline content schema. In practice, I don't think pretty much anyone is likely to use the editor without the text IC. But eh imo it's kind of the same as with the paragraph block in the slash menu items - we also check if that's in the schema (despite 99% of people using the paragraph block) and there's not really any issues there.

docs/pages/docs/ui-components/suggestion-menus.mdx
matthewlipski: I think it's ok tbh, but maybe better to have a separate page for grid suggestion menus and leave the current suggestion menu page as-is?

YousefED: Imagine you're new to blocknote and open this page. Not too much information to grasp all concepts? Besides, I think the different UI pages now follow the same "flow" making it easier to grasp, I think we're also breaking that with this change?

matthewlipski: Hmm ok, so do you think the suggestion of splitting the grid suggestion menu w/ emoji picker and regular suggestion menu w/ slash menu makes sense? That would keep the structure consistent and have less information on a single page.

matthewlipski:
@hunxjunedo All good, we're just polishing off the remaining TODOs. Should be merged sometime this week, maybe tomorrow depending on how much will need to change based on @YousefED's review

docs/pages/docs/ui-components/suggestion-menus.mdx
YousefED: - Yes, my suggestion would be to move the emoji and grid suggestionmenu info to a single page under Advanced. I think customizing these components is pretty niche.
- The emojiPicker needs to be added to the props here: https://blocknote-website-1u5ylvcga-typecell.vercel.app/docs/editor-basics/setup#rendering-the-editor-with-blocknoteview

packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
YousefED: I'm concerned the complexity of the code is not worth the wins we get out of it. The typing was already pretty complex (maybe too much), but now by adding a different "type" it increases too much I think. I think we've got three options:

1. Remove some of the typing tricks added in this PR for columns / grids etc. 
2. Revise the original code. Maybe we don't need all the defaults but it's better to explicitly pass in things like `defaultsuggestionmenuitems` from `BlockNoteDefaultUI`, etc.
3. Creating a separate GridSuggestionMenuController, that would be pretty much the same as the existing `SuggestionMenuController` but defaulting to emojis and grids

I think option 3 is the least effort while at least not making the code less readable from what it currently is. wdyt?

matthewlipski: Yeah agreed tbh, there will be a bit of code duplication but I think the other 2 options have bigger drawbacks (either worse typing or worse experience creating custom suggestion menus)

docs/pages/docs/advanced/grid-suggestion-menus.mdx
```diff
@@ -0,0 +1,53 @@
...
+title: Grid Suggestion Menus
+description: In addition to displaying Suggestion Menus as stacks, BlockNote also supports displaying them as grids.
+imageTitle: Grid Suggestion Menus
+---
+
+import { Example } from "@/components/example";
+
+# Grid Suggestion Menus
+
+Grid Suggestion Menus appear when the user enters a trigger character, and text after the character is used to filter the menu items.
```
YousefED, line 11: Let's use this:

> Grid Suggestion Menus are similar to regular [Suggestion Menus](/docs/ui-components/suggestion-menus#additional-features), but results are organized in a grid, and users can use all arrow keys (including left, right) on their keyboard to navigate the results.

and remove the `Additional Features` part at the bottom

packages/react/src/components/SuggestionMenu/GridSuggestionMenu/GridSuggestionMenuWrapper.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { FC, useCallback, useEffect } from "react";
+
+import { useBlockNoteContext } from "../../../editor/BlockNoteContext";
+import { useBlockNoteEditor } from "../../../hooks/useBlockNoteEditor";
+import { useCloseSuggestionMenuNoItems } from "../hooks/useCloseSuggestionMenuNoItems";
+import { useGridSuggestionMenuKeyboardNavigation } from "./hooks/useGridSuggestionMenuKeyboardNavigation";
+import { useLoadSuggestionMenuItems } from "../hooks/useLoadSuggestionMenuItems";
+import { GridSuggestionMenuProps } from "./types";
+
+export function GridSuggestionMenuWrapper<Item>(props: {
```
YousefED, line 11: Hmm, I see you separated both the controller, wrapper and keyboard hook. Wouldn't it suffice to just separate the controller? That's where most of the difference in implementation is. Now we introduce a lot of duplicate code

matthewlipski: The reason I did it this way was mostly for understandability. Basically, splitting just the controller means that you have a hierarchy like this:

- Controller (split)
- Wrapper (shared)
- Keyboard nav hook (shared)
- Actual Menu (split)

I feel like if you never saw the codebase before, it would be somewhat confusing that the suggestion menu is split between regular and grid at the top level and bottom level, but not in between. Having it fully split, while introducing more duplication, makes the code and typing simpler + (imo) makes it easier to understand without having worked on the code ever or in a long time.

hunxjunedo:
Hey @matthewlipski any updates on this ? How is everything going ? Please let me know if I can be of any use.

hunxjunedo:
hey @YousefED ! since the PR has been merged, when can I except this to be completed ?

PR #856, opened 2024-06-20
feat: Table cell/column widths in API

### LINKED ISSUE #655: Table doesn't save columns width after refresh

matthewlipski:
You can already resize table cells in BlockNote, but these use premade TipTap nodes and so their widths are not really accessible to get/set via the BlockNote API. This PR modifies the typing of `TableContent` so that you can now read and update table cell widths. Additionally, it fixes a small bug that was preventing the cell widths from being parsed correctly.

The changes I made are based on #846, but I don't have push rights to that remote repo so this PR replaces that one.

Closes #655

### DIFF
--- a/docs/pages/docs/editor-basics/default-schema.mdx
+++ b/docs/pages/docs/editor-basics/default-schema.mdx
@@ -117,7 +117,7 @@ type TableBlock = {
   id: string;
   type: "table";
   props: DefaultProps;
-  content: TableContent[];
+  content: TableContent;
   children: Block[];
 };
 ```
--- a/docs/pages/docs/editor-basics/document-structure.mdx
+++ b/docs/pages/docs/editor-basics/document-structure.mdx
@@ -73,13 +73,16 @@ The `styles` property is explained below.
 
 While most blocks use an array of `InlineContent` objects to describe their content (e.g.: paragraphs, headings, list items). Some blocks, like [images](/docs/editor-basics/default-schema#image), don't contain any rich text content, so their `content` fields will be `undefined`.
 
-[Tables](/docs/editor-basics/default-schema#table) are also different, as they contain `TableContent`. Here, each table cell is represented as an array of `InlineContent` objects:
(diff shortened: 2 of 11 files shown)


YousefED (commented):
Hmm, downside is that this would break all existing documents stored right? Let's think a bit more about backwards compatibility before merging this

halimt92:
> Hmm, downside is that this would break all existing documents stored right? Let's think a bit more about backwards compatibility before merging this

If you guys use a minWidth of 100px and change the minWidth of the paragraphs instead of the width, it would not affect the existing tables right @YousefED ?

matthewlipski:
> > Hmm, downside is that this would break all existing documents stored right? Let's think a bit more about backwards compatibility before merging this
> 
> If you guys use a minWidth of 100px and change the minWidth of the paragraphs instead of the width, it would not affect the existing tables right @YousefED ?

It's not quite that simple - the width of the table cells was/is already stored within TipTap's `TableCell` node, this PR is more about exposing that to the BlockNote API so the widths are preserved when saving and loading the doc. But because we're adding a width to table cells in the API, the type definition of `CellContent` needs to be changed and so existing documents with the previous `CellContent` implementation will break.

matthewlipski:
Also just to add, it is a good point that cells should have a minimum width (should be able to extend the `TableCell` node with `addAttributes` to do that), and we should make sure to do that before merging because right now it makes the UX pretty janky.

osawa-koki (approved):

PR #858, opened 2024-06-21
add disableExtensions option to allow disabling built in extensions by name.

fridaystreet:
Without completely needing to refactor the way tiptap extensions and core blocknote extensions are added, this should be fairly simple approach to just allow a list of extension names to be passed in that will then be removed from the inbuiult extensions before being passed to tiptap.

This will allow you to the override these extensions if you wish without completely disbaling blocknote via 
```
    disableExtensions:['link'],
    _tiptapOptions: {
      extensions: [
        Link,
      ]
    }
  })
```

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -75,7 +75,7 @@ export type BlockNoteEditorOptions<
 > = {
   // TODO: Figure out if enableBlockNoteExtensions/disableHistoryExtension are needed and document them.
   enableBlockNoteExtensions: boolean;
-
+  disableExtensions: string[];
   /**
    * A dictionary object containing translations for the editor.
    */
@@ -280,6 +280,7 @@ export class BlockNoteEditor<
       inlineContentSpecs: this.schema.inlineContentSpecs,
       collaboration: newOptions.collaboration,
       trailingBlock: newOptions.trailingBlock,
+      disableExtensions: newOptions.disableExtensions
     });
 
     const blockNoteUIExtension = Extension.create({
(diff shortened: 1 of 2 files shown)


matthewlipski (changes requested):
Looks great! The only thing that might be a bit finicky is that `enableBlockNoteExtensions` also exists, and `disableExtensions` is basically the same thing but lets you disable specific extensions instead of all of them. Basically what I'm saying is you can probably just remove the `enableBlockNoteExtensions` option since your new one is the same but better:)

matthewlipski (approved):

YousefED (approved):

fridaystreet:
@matthewlipski ok great thanks. Yeah at first I actually just used enableBlockExtensions and made it 

string[] | boolean

but after looking at it a few times it didn't feel like it made sense at it would imply you need to pass an array of all of them not just the disabled ones. 

I wasn't sure if there was some bigger picture reason for being to just disable/enable the whole thing with enableBlockNoteExtensions so opted for the extra parameter.  I agree possibly it adds confusion with 2, but then again, if someone does want to disable all of them, they'd have to maintain an array full of all the extension names so maybe it is better to leave both.

I guess another option could be to just have explicit arrays for both, that way if you're looking to only enable some or only disable some it's easy, but probably overkill and confusing

enableExtensions string[] | boolean | undefined
disableExtensions string[] | boolean | undefined 

Thanks for reveiwing it though. let me know if you want me to submit another PR

matthewlipski:
All good, I think the use case of disabling all TipTap extensions is extremely rare since at that point you're probably better off just using a different editor, so I went ahead with removing `enableBlockExtensions`.

PR #870, opened 2024-06-26
fix: Copy selection starting in nested block

matthewlipski:
It's possible to select & copy nested blocks without content from the parent block being selected. ProseMirror still tries to preserve the nesting structure in this case, but because content from the parent block is not selected, it basically violates the schema when creating the `Fragment` for the selected content. I documented this pretty extensively in the code so recommend checking that out for more details.


packages/core/src/api/exporters/html/htmlConversion.test.ts
```diff
@@ -101,3 +102,204 @@ describe("Test HTML conversion", () => {
...
+
+  // When the selection starts in a nested block, the Fragment from it omits the
+  // `blockContent` node of the parent `blockContainer` if it's not also
+  // included in the selection. In the schema, `blockContainer` nodes should
+  // contain a single `blockContent` node, so this edge case needs to be tested.
+  describe("No block content", () => {
+    it("Selection within a block's children", () => {
+      // Fragment created when the selection starts and ends within a block's
+      // children.
+      const copiedFragment = Fragment.from(
```
YousefED, line 136: might be a bit cleaner to first create the blocknote json (more readable), convert that to prosemirror (we have a function for this), and then select the relevant part from it?

(I might be wrong here, wdyt?)

YousefED (commented):
Nice, and clear comments. Should we specifically also test the use-case from the bug report, which is about list structures? Or does the fact that it's a list not really matter?

matthewlipski:
> Nice, and clear comments. Should we specifically also test the use-case from the bug report, which is about list structures? Or does the fact that it's a list not really matter?

Nah the list items are totally irrelevant here, I think they got specifically reported just because nesting list items is a much more common use case than nesting other types of blocks.

YousefED (approved):

Issue #860, opened 2024-06-24
TypeError: use_prefers_color_scheme__WEBPACK_IMPORTED_MODULE_2__ is not a function


Ark07:
**Describe the bug**

I was trying to integrate rich text editor , and came across this amazing package Blocknote , reduces more than half of the work required with tip tap

But When i added webpack 5 to my application
It started giving below error 
`TypeError: use_prefers_color_scheme__WEBPACK_IMPORTED_MODULE_2__ is not a function `

<img width="1430" alt="Screenshot 2024-06-24 at 8 12 06 AM" src="https://github.com/TypeCellOS/BlockNote/assets/18495043/40f0f0bd-2ba7-452b-bdf7-50aa69ff5ca6">

Package.json
```
 "dependencies": {
    "webpack": "^5.92.1",
    "webpack-cli": "^5.1.4",
    "@blocknote/core": "^0.13.5",
    "@blocknote/mantine": "^0.13.5",
    "@blocknote/react": "^0.13.5",
    "@babel/core": "^7.24.7",
    "@babel/preset-env": "^7.24.7",
    "@babel/preset-react": "^7.24.7",
 }
```

webpack.config
```
entry: 'app/app.js',
 output: {
    path: path.resolve(process.cwd(), 'build'),
    publicPath: '/',
    filename: '[name].js',
    chunkFilename: '[name].chunk.js',
  },
```
Related issue maybe ? https://github.com/devforth/painterro/issues/123 , https://github.com/webpack/webpack/issues/12675

- Node version: 20
- Browser:  Chrome

Thank you in advance for your help!

matthewlipski:
Hmm this might be a pain to diagnose and fix, especially since it's more likely a bundling issue than something specific to BlockNote. That being said, could you try updating your BlockNote version to 0.14.1 and see if the problem is still there?

Ark07:
hi @matthewlipski @YousefED, thank you for taking time to check the issue
please find below PR for webpack 5 fix https://github.com/TypeCellOS/BlockNote/pull/872

PR #872, opened 2024-06-27
Fix: [#860] TypeError: use_prefers_color_scheme__ is not a function

Ark07:
usePrefersColorScheme is exported from use-prefers-color-scheme as named function and default

importing usePrefersColorScheme as named function will help in fixing the webpack 5 issue #860


matthewlipski:
Great, thanks!

PR #875, opened 2024-06-27
chore: add export and change tiptap options

jueinin:
fix https://github.com/TypeCellOS/BlockNote/issues/876


matthewlipski:
Could you add some details about how this fixes #876? Also see my comments under that issue

PR #922, opened 2024-07-14
fix: position of side menu controllers in `rtl`

zaaakher:
This PR simply makes the side menu buttons to be in the correct position when layout is RTL.

# Before
blocknote_rtl_before

# After
blocknote_rtl_after

### DIFF
--- a/packages/react/src/components/SideMenu/SideMenuController.tsx
+++ b/packages/react/src/components/SideMenu/SideMenuController.tsx
@@ -20,6 +20,7 @@ export const SideMenuController = <
   S extends StyleSchema = DefaultStyleSchema
 >(props: {
   sideMenu?: FC<SideMenuProps<BSchema, I, S>>;
+  dir?: "rtl" | "ltr";
 }) => {
   const editor = useBlockNoteEditor<BSchema, I, S>();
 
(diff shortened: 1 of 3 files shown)


matthewlipski (approved):

YousefED (changes requested):
(see comment above)

zaaakher:
@YousefED you're right, it should ideally adapt to RTL without the any prop-drilling.

After a little digging, it seems that the `floatingStyles` coming from the `useFloating` hook is the one preventing `-start` from being dynamic with the layout direction.

When I comment out this line https://github.com/TypeCellOS/BlockNote/blob/b20292a3bf7de955010900c7bc7e9e7244d6a30f/packages/react/src/hooks/useUIElementPositioning.ts#L51

The `-start` positioning works fine (but the fixed/absolute positioning breaks)

blocknote_rtl_floating

zaaakher:
Digging into the source-code of floating-ui, I think this is the root cause of the issue:
https://github.com/floating-ui/floating-ui/blob/cea9065811ff347cfe388720b3cc91165c8cebe7/packages/react-dom/src/useFloating.ts#L159

I think it should be `inset-inline-start`

zaaakher:
It seems that `floatingStyles.transform` is the one positioning it and not `left`/`insetInlineStart` as I thought.

Based on the response from floating-ui here https://github.com/floating-ui/floating-ui/pull/2982#issuecomment-2234669030 it seems the expected automatic direction flip based on dir layout isn't yet supported nativly in floating-ui.

Other than the minor prop-drilling suggested in this PR, I can't honestly think of an alternative way to do it. 😅

PR #926, opened 2024-07-15
docs: Updated custom UI example

### LINKED ISSUE #867: static toolbar example broken CSS?
https://www.blocknotejs.org/examples/ui-components/custom-ui

image

matthewlipski:
This PR reworks the custom UI example and adds a non-pro example for how to make a static formatting toolbar. The new custom UI example uses Material UI to implement replacements for the default Mantine components, and is extensively commented to show users how to customize elements and use other component libraries that aren't MUI.

Closes #867


YousefED (commented):
Nice. I think this looks good but we can make a couple of improvements. Mainly to explain to readers what the concepts are behind the code. What's the architecture?

Otherwise, we just provide them with some sample replacements that they can edit, but might not go beyond that or really understand it because the concepts aren't explained.

Specifically, I'd suggest:

- For every component, link / explain which built-in component is replaced. e.g. something like: "The MUIFormattingToolbar replaces the default FormattingToolbar (link) with a MUI-based one. It includes the same components as the default FormattingToolbar, but renders them using MUI elements instead. You can also adjust this code to add or remove specific buttons."
- Explain when a user would implement a new package (e.g.: "The approach in this example replaces individual components. Alternatively, you can .... This is how the default BlockNote UI layers for Mantine / ShadCN / Ariakit are created")

examples/03-ui-components/12-custom-ui/App.tsx
```diff
@@ -28,30 +33,55 @@ export default function App() {
...
+        sideMenu={false}
+        // Disables remaining UI elements.
+        linkToolbar={false}
+        filePanel={false}
+        tableHandles={false}
+        emojiPicker={false}>
+        {/* Adds the custom Formatting Toolbar. */}
+        {/* `FormattingToolbarController isn't used since the custom toolbar */}
+        {/* is static and always visible above the editor. */}
+        <MUIFormattingToolbar />
```
YousefED, line 66: Shall we also create a separate example with a static toolbar? (default mantine one?)

examples/03-ui-components/12-custom-ui/MUISideMenu.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { TextBlockSchema } from "./schema";
+
+// Custom component to replace the default Block Side Menu.
+export function MUISideMenu(props: SideMenuProps<TextBlockSchema>) {
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+
+  // Since the side menu is positioned by the top-left corner of a block, we\
+  // manually set its height based on the hovered block so that it's vertically
+  // centered.
+  const sideMenuHeight = useMemo(() => {
```
YousefED, line 22: maybe extract this into a reusable hook?

examples/03-ui-components/12-custom-ui/README.md
```diff
@@ -1,6 +1,6 @@
-# Custom UI
+# Using Material UI
```
YousefED, line 1: Maybe a mix? "Custom all UI elements (Material)"?

It feels like the purpose of the demo is to be a "cookbook" / showcase for how to customize all of the UI, the purpose is not necessarily "how to use Material UI". If it were the latter, we should just create a Material UI package imo. Let's try to make the title + description reflect this purpose

matthewlipski: So should it be a mix of MUI + the default Mantine UI elements? Or do you suggest implementing the remaining elements (link toolbar, table handles, file panel) in MUI? Since right now those are disabled.

YousefED: No sorry, I meant, maybe the title should be a mix? (you changed it to "Material UI", but I don't think that conveys the purpose of the example)

matthewlipski: Oh right, yeah I think "UI with Third-Party Components" is probably the most accurate for the use case that the demo is targeted at

examples/03-ui-components/12-custom-ui/MUISideMenu.tsx
matthewlipski: I don't think it makes sense here, since in the `react` package we return `data-*` which are used in CSS, whereas in the example we return numbers which are used in the `sx` prop. If MUI was styled using vanilla CSS it would make sense to use the same hook for both, but the convention is to style your components with the `sx` prop instead.

YousefED:
Looks great!

PR #939, opened 2024-07-16
chore: upgrade vite and vitest

YousefED:


matthewlipski (approved):
Looks good, tested built JS files on a NextJS app and didn't cause any issues in both built and dev envs.

PR #945, opened 2024-07-17
fix: Dropping blocks above editor

### LINKED ISSUE #934: Block drag to top of document doesn't work
**Describe the bug**
Block drag to top of document doesn't work

Example:
https://www.loom.com/share/8db9d97aaa6f470c988cc59e954017aa

**To Reproduce**
- Go to basic setup guide: https://www.blocknotejs.org/examples/basic/minimal
- Enter more than one block
- Try to drag any block to the top

**Misc**
- Node version:
- Package manager:
- Browser: 

matthewlipski:
Even though the drop cursor still appears when the mouse cursor is beyond the editor vertically, letting go of the mouse button doesn't actually trigger a drop in the editor. This PR fixes that by restricting the mouse cursor y-coordinate to the editor's bounding box on drop.

Closes #934


YousefED (approved):
looks good (if you remove the blur handling from 944)

PR #951, opened 2024-07-19
fix: make package exports explicit

### LINKED ISSUE #952: Unable to import from BlockNote packages when using `NodeNext` for `moduleResolution`.
**Describe the bug**
Using `NodeNext` as the `moduleResolution` value in a project's `tsconfig.json` breaks the ability to import from `@blocknote/core`, `@blocknote/react` and `@blocknote/server-util`. Any exports in those packages that use the `export * from` syntax are not importable and result in the error below.

```
Module '"@blocknote/react"' has no exported member 'Components'.ts(2305)
```

Mantine went through a similar issue [here](https://github.com/mantinedev/mantine/issues/4991). The resolution there was to explicitly append the `.js` extension to the exports. However, I'm not 100% sure this wouldn't be a breaking change.

**To Reproduce**
* Open a project using BlockNote and update the `tsconfig.json` to use `NodeNext` for `module` and `moduleResolution`.
* Restart the TS server if necessary and review errors

**Misc**
- Node version: v22.1.0
- Package manager: pnpm
- Browser: N/A

danielrearden:
This PR makes all the exports explicit for `@blocknote/core`, `@blocknote/react` and `@blocknote/server-util`. Without this change, the packages are not usable inside TypeScript applications that specify `NodeNext` as the `moduleResolution` value in their `tsconfig.json`.

Fixes #952

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,47 +1,255 @@
 import * as locales from "./i18n/locales";
-export * from "./api/exporters/html/externalHTMLExporter";
-export * from "./api/exporters/html/internalHTMLSerializer";
-export * from "./api/testUtil";
-export * from "./blocks/AudioBlockContent/AudioBlockContent";
-export * from "./blocks/FileBlockContent/FileBlockContent";
-export * from "./blocks/ImageBlockContent/ImageBlockContent";
-export * from "./blocks/VideoBlockContent/VideoBlockContent";
+export {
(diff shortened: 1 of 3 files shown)


gajus (approved):

danielrearden:
Closing this PR since it does not fully resolve the issue. While this change does surface the exports, they are not strongly typed.

PR #953, opened 2024-07-19
Feat/equation

### LINKED ISSUE #741: Custom Block for Katex

jkcs:
close #741
/claim #741

### DIFF
--- a/examples/03-ui-components/13-equation/README.md
+++ b/examples/03-ui-components/13-equation/README.md
@@ -1,8 +1,8 @@
-# Alert Block
+# Inline Equation
 
-In this example, we create a custom `Alert` block which is used to emphasize text. In addition, we create a Slash Menu item which inserts an `Alert` block.
+In this example, we create a custom `Inline Equation`
 
-**Try it out:** Press the "/" key to open the Slash Menu and insert an `Alert` block!
+**Try it out:** Press the "/" key to open the Slash Menu and insert an `Equation` block!
 
 **Relevant Docs:**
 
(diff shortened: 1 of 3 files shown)


matthewlipski:
Really nicely done! I have a few questions r.e. the implementation:

I see that the equations are implemented as TipTap nodes first and then converted into BlockNote inline content. Which functionality are we gaining from this vs implementing as a React custom block without content and storing the Latex as a prop instead? Keyboard navigation & storing the Latex in `content` instead of `props` are obvious ones, is there more?

I ask this because imo it would be preferable to keep things within the BlockNote API rather than diving into TipTap's. So if the difference in functionality is not too bad then I think that might make more sense + we could keep the current implementation as an advanced/pro example. I appreciate that the current UX is basically 1:1 with Notions (great job on that btw), it's just that we wanna minimize code complexity for these examples.

@YousefED wdyt? Should we just have an example using the TipTap API, have a simplifed one using the BlockNote API, or both?

matthewlipski:
I think the suggestion of having this as an example of using a TipTap node to create inline content is a good one👍

Do you want to give implementing it strictly using the BlockNote API a shot? You can simplify things quite a bit imo. Regarding keyboard navigation, I think it's best to just set `content: "none"` in the inline content config, which should make the arrow keys just ignore it. Also recommend saving the Latex to an `equation` prop instead of the block's content, since I think you'll otherwise be limited by the BlockNote API.

I don't know if we want to extend the API in this PR tbh, it should be possible to implement the simplified version in the current state and API design tends to take quite a while.

PR #956, opened 2024-07-19
chore: use import extensions in all packages (except docs)

YousefED:
Closes https://github.com/TypeCellOS/BlockNote/issues/952


YousefED:
Blocked by https://github.com/vercel/next.js/issues/59744

matthewlipski:
> Blocked by [vercel/next.js#59744](https://github.com/vercel/next.js/issues/59744)

Isn't my suggestion still relevant though? SInce this is a NextJS issue and we only really need this in packages that are on npm. Or are there other problems with that?

YousefED:
> > Blocked by [vercel/next.js#59744](https://github.com/vercel/next.js/issues/59744)
> 
> Isn't my suggestion still relevant though? SInce this is a NextJS issue and we only really need this in packages that are on npm. Or are there other problems with that?

Good point, the issue is that the examples span both the playground and the website. So I thought it wouldn't be possible. However, might be wrong, maybe I can use the "old style" in the examples as well. will investigate

PR #970, opened 2024-07-30
fix: windows gen file

### LINKED ISSUE #955: dev-scripts path issue on Windows
**Describe the bug**
Running `npm run gen` throws the following error on Windows:

image

**To Reproduce**
- Be on Windows 10
- Run `npm install`
- Run `npm run gen` from the root of the project or from examples dir
- See the same error as the screenshot above

**Misc**
- Node version: 18.20.3
- Package manager: pnpm
- Browser: Chrome

---

I was hitting my head for an hour trying to figure this one out. When I comment out `fs.writeFileSync` the error is gone, and this seems to be originating from the way we're getting the path/dir :

```
const dir = path.parse(import.meta.url.replace("file://", "")).dir;
```

This issue is in every instance of the above code, in `gen.ts`, `genDocs.ts`, and `util.ts`

---
I was only able to generate the example code in #954 using my MacOS. So this issue is exclusively on Windows.

jkcs:
/claim #955
close #955

### DIFF
--- a/packages/dev-scripts/examples/gen.ts
+++ b/packages/dev-scripts/examples/gen.ts
@@ -1,3 +1,4 @@
+import { fileURLToPath } from "node:url";
 import * as glob from "glob";
 import * as fs from "node:fs";
 import * as path from "node:path";
@@ -17,7 +18,7 @@ import { Project, getExampleProjects, groupProjects } from "./util";
  * (The downside of this is that we have some almost duplicate, generated files in the repo,
  * but the upside is anyone can run npm start in any of the examples (and that we can point a codesandbox / repl to the examples directory))
  */
-const dir = path.parse(import.meta.url.replace("file://", "")).dir;
+const dir = path.parse(fileURLToPath(import.meta.url)).dir;
 
 async function writeTemplate(project: Project, templateFile: string) {
   const template = await import(templateFile);
--- a/packages/dev-scripts/examples/gen.ts
+++ b/packages/dev-scripts/examples/gen.ts
@@ -5,7 +5,12 @@ import * as path from "node:path";
 import prettier from "prettier";
 import React from "react";
 import ReactDOM from "react-dom/server";
-import { Project, getExampleProjects, groupProjects } from "./util";
+import {
+  Project,
+  getExampleProjects,
+  groupProjects,
+  replacePathSepToSlash,
+} from "./util";
 
 /**
  * This script reads the examples in the /examples folder. These folders initially only need an App.tsx, .bnexample.json and README.md file.
@@ -58,7 +63,7 @@ async function writeTemplate(project: Project, templateFile: string) {
 
 async function generateCodeForExample(project: Project) {
   const templates = glob.sync(
-    path.resolve(dir, "./template-react/*.template.tsx").replace(/\\/g, "/")
+    replacePathSepToSlash(path.resolve(dir, "./template-react/*.template.tsx"))
   );
 
   for (const template of templates) {
--- a/packages/dev-scripts/examples/gen.ts
+++ b/packages/dev-scripts/examples/gen.ts
@@ -1,4 +1,4 @@
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import * as glob from "glob";
 import * as fs from "node:fs";
 import * as path from "node:path";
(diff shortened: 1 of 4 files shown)


zaaakher:
Good progress 👍 

I just tested it on both Windows 10 and MacOS and here are the current results


| OS      | Updated `examples.gen.tsx` | adds example files |
| ------- | ------------------------ | ------------------ |
| Windows | ✅                       | ❌                 |
| MacOS   | ✅                       | ✅                 |



By "adds example files" I mean the `package.json`, `tsconfig.json`, `index.html`, `main.tsx`, and `vite.config.ts` in each example that should be automatically generated by the script.

zaaakher:
@jkcs I think the only part missing is in the `writeTemplate` function in `gen.ts` (line 27)

zaaakher:
I just tested it again and everything seems to be working as expected in both Windows 10 and MacOS 👍 

| OS      | Updates `examples.gen.tsx` | Adds example files |
| ------- | ------------------------ | ------------------ |
| Windows | ✅                       | ✅                 |
| MacOS   | ✅                       | ✅                 |

@jkcs Thanks for your work 👍

packages/dev-scripts/examples/gen.ts
```diff
@@ -57,7 +58,7 @@ async function writeTemplate(project: Project, templateFile: string) {
 
 async function generateCodeForExample(project: Project) {
   const templates = glob.sync(
-    path.resolve(dir, "./template-react/*.template.tsx")
+    path.resolve(dir, "./template-react/*.template.tsx").replace(/\\/g, "/")
```
YousefED, line 61: instead of replace, using `path.join` is cleaner?

packages/dev-scripts/examples/util.ts
```diff
@@ -114,7 +116,7 @@ export type Files = Record<
 
 export function getProjectFiles(project: Project): Files {
   const dir = path.resolve("../../", project.pathFromRoot);
-  const files = glob.globSync(dir + "/**/*", {
+  const files = glob.globSync(glob.win32.convertPathToPattern(dir + "/**/*"), {
```
YousefED, line 119: Now we're also calling win32.convertPathToPattern on mac / unix right? Does that make sense? It seems a bit weird to me!

packages/dev-scripts/examples/util.ts
```diff
@@ -140,7 +142,11 @@ export function getProjectFiles(project: Project): Files {
  */
 export function getExampleProjects(): Project[] {
   const examples: Project[] = glob
-    .globSync(path.join(dir, "../../../examples/**/*/.bnexample.json"))
+    .globSync(
+      glob.win32.convertPathToPattern(
```
YousefED, line 146: same question as above

packages/dev-scripts/examples/util.ts
```diff
@@ -162,9 +168,8 @@ export function getExampleProjects(): Project[] {
         .split(path.sep);
 
       const group = {
-        pathFromRoot: path.relative(
-          path.resolve("../../"),
-          path.join(directory, "..")
+        pathFromRoot: slash(
```
YousefED, line 171: I'm not sure `slash` is needed (would prefer not to add this dependency). maybe we can just prevent passing / to path.resolve? e.g.: `path.resolve("..","..")`

YousefED (commented):
great! some questions

packages/dev-scripts/examples/gen.ts
YousefED: ah, and the is that `glob` always expects a fwd slash I suppose?

packages/dev-scripts/examples/util.ts
YousefED: Yes, I'd love to see this code a bit clearer / more uniform. When reviewing, it's not entirely clear to me why we have 3 different methods:
- `replace`
- `slash`
- `convertPathToPattern`

Can we either use 1 method (ideally with a separate, well named helper function). Or, if multiple methods are necessary, add a comment in code why we use each function?

packages/dev-scripts/examples/util.ts
```diff
@@ -201,3 +202,13 @@ export function getExampleProjects(): Project[] {
   // });
   return examples;
 }
+
+export function replacePathSepToSlash(path: string) {
+  const isExtendedLengthPath = path.startsWith("\\\\?\\");
```
YousefED, line 207: what's an example of this?

YousefED (approved):
Thanks @jkcs for your contribution and collaboration and @zaaakher for testing :)

PR #974, opened 2024-07-30
feat: add issue form for submitting blocks

zaaakher:
Adds `share_block.yml` to `ISSUE_TEMPLATE` to streamline the block submission process


zaaakher:
> Great! Do you think we need Block Code? Because if people submit a stackblitz or similar, all code should be in there? I think it might be a bit cumbersome for people to submit code in a text box

Indeed! will fix it

YousefED:
Cool! I'll merge, let's see if it gets used. If not we either need to point more people to it, or remove it again

PR #980, opened 2024-08-01
feat: AI block UI

matthewlipski:
ai

This PR adds all the UX/UI for AI integration in BlockNote according to the [Notion doc](https://www.notion.so/typecell/Spec-1dd474b463fa42098ae40013b244855b). The UI is not yet plugged into an API for querying AI models, so it uses a mock function which accepts a string prompt and returns a string AI response.

There are a few differences in the implementation vs the spec listed in the doc:

- The AI inline toolbar now instantly opens after selecting a prompt. This also means there are no AI related loaders outside of the AI inline toolbar. Currently the loader is just an `Updating...` string but can easily be changed.
- The "Accept", "Retry", and "Revert" buttons shown after an AI response is inserted are an entirely separate toolbar, instead of being part of the formatting toolbar.
    - Inserting content via the AI suggestion menu also shows this separate toolbar.

TODOs:

- [x] AI Block update button text isn't being set to `Updating...` on click.
- [x] Autofocus not working for prompt input in the AI button popover in the formatting toolbar.
- [ ] Error is thrown when calling `mockAIReplaceSelection` or `mockAIInsertAfterSelection` in `AIInlineToolbar` while React strict mode is enabled. Everything works fine for the user though.
- [x] The AI suggestion menu items are currently derived from dictionary entries, which we probably want to change.
- [ ] AI menu close on escape.
- [ ] AI menu close on backspace.
- [ ] AI menu fade in.
- [ ] AI menu smaller height.
- [ ] AI menu plugin should pass previous block pos when current block is empty. Currently this position change is done in the controller via CSS.


YousefED (commented):
In general looks good, I only looked at the overall architecture for now (not line-by-line). So far I have a few main comments:

**Extracting package**
Just realizing this now, I think we should try to ship this as a separate package. Two reasons for this: (a) it's likely this may be licensed differently (i.e.: GPL / Pro only) and (b) it would be a good "proof of the pudding" to see whether we can externalize larger plugins / extensions in a way that still makes sense


**Duplicated toolbar code?**
I think we're starting to see a lot of similar code with the AI Toolbar and plugin that we already have for FilePanel and LinkToolbar, right? If so, it might be a sign that we need to generalize this code and reduce the duplicate code

**AI specific**
- We need a way to pass in the `mockAIModelCall` function
- The AI Model call might need access to the rest of the document / blocks above / below the block
- I think the current mockup is some good first steps but the Notion AI Block is definitely more polished. Specifically, I like their "Generated by AI x minutes ago" context and don't really like our "show prompt" button.

wdyt?

packages/xl-ai/src/AIExtension.ts
```diff
@@ -0,0 +1,195 @@
...
+export function getAIExtension(editor: BlockNoteEditor<any, any, any>) {
+  return editor.extension(AIExtension);
+}
+
+// TODO: Revisit - this pattern might be a bit iffy
+// Make editor non-editable after calling the LLM, until the user accepts or
+// reverts the changes.
+// useEffect(() => {
+//   editor.isEditable = aiResponseStatus === "initial";
+// }, [aiResponseStatus, editor]);
```
nperez0111, line 195: Unrelated, but, this is why I prefer that LLM responses be injected as decorations instead of modifying document content.

Imagine a user refreshes their window, the LLM response should be dropped, but it actually modified the document. Just a thought

packages/xl-ai/src/AIExtension.ts
```diff
@@ -0,0 +1,195 @@
...
+   *
+   * - I decided to create a separate store for the options as they feel quite different from "internal state"
+   *   above. They're not so much "state", but more "configuration". I still think it's useful to have in a Zustand
+   *   store (instead of a just a plain object), because consumers might want to display the settings in the UI.
+   *   Case can also be made to make this a plain and leave the responsibility of displaying settings to the consumer
+   *   (and not have a store for this at all). TO DISCUSS
+   */
+  public readonly options: ReturnType<
+    ReturnType<typeof createStore<Required<GlobalLLMCallOptions>>>
+  >;
```
nperez0111, line 66: Yea, I think that configuration should just be a plain object, if the UI needs to be able to modify it, then the configur option should just be a default/initial value & the value is stored into the state store.

I'd imagine configuration like `codeBlock.supportedLanguages`, & I think it'd be complex to have that value be modifiable at runtime.

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -698,6 +713,26 @@ export class BlockNoteEditor<
...
+  public extension<T extends BlockNoteExtension>(
+    ext: { new (...args: any[]): T } & typeof BlockNoteExtension,
+    key = ext.name()
+  ): T {
+    const extension = this.extensions[key] as T;
+    if (!extension) {
+      throw new Error(`Extension ${key} not found`);
+    }
+    return extension;
+  }
```
nperez0111, line 734: I'm just gonna skip over any of the extension bit. I think there are better ways to do it, but not pertinent right now to discuss & should be separately discussed & written up

packages/xl-ai/src/AIExtension.ts
nperez0111: I'd even go as far as to say that configuration should be `T extends Readonly<Record<string, any>>`

packages/xl-ai/src/AIExtension.ts
```diff
@@ -0,0 +1,195 @@
...
+   * zustand design considerations
+   *
+   * - setters are not added in this store, but as class methods so we can have control over what to expose as API
+   * - con: if we'd expose the store externally, consumers could still call store.setState :/ (for this store it's not desired, for `options` below, it is)
+   *
+   */
+  public readonly store = createStore<AIPluginState>()((set) => ({
+    aiMenuBlockID: undefined,
+    aiMenuResponseStatus: "initial",
+  }));
```
nperez0111, line 53: Agree that setters being in the store would be weird, assuming that you have access to the class methods. If you didn't have access to the class methods, it would basically force all logic to be in zustand, which I think is the wrong way about things.

Being able to update state you don't own is pretty problematic 🙁  (if someone called store.setState), but I don't think we need to solve it or disable it.

packages/xl-ai/src/AIExtension.ts
```diff
@@ -0,0 +1,195 @@
...
+} from "@blocknote/core";
+import { LanguageModel } from "ai";
+import { createStore } from "zustand/vanilla";
+import { PromptOrMessages, llm } from "./api";
+import { LLMRequestOptions } from "./api/streamTool/callLLMWithStreamTools";
+
+type AIPluginState = {
+  aiMenuBlockID: string | undefined;
+  aiMenuResponseStatus: "initial" | "generating" | "error" | "done";
+};
```
nperez0111, line 14: What if this was a union, would setState be able to enforce this properly?

Like, what if it was:

```ts
type AIPluginState = {
   aiMenuBlockID: undefined;
   aiMenuResponseStatus: 'initial'
} | {
   aiMenuBlockID: string;
   aiMenuResponseStatus: "generating" | "error" | "done";
 };
```

YousefED: good suggestion, see this commit: https://github.com/TypeCellOS/BlockNote/pull/980/commits/5a529cb49f4760e6ea56ea2179d9abfe824c7b8f

packages/xl-ai/src/AIExtension.ts
YousefED: > Being able to update state you don't own is pretty problematic 🙁 (if someone called store.setState), but I don't think we need to solve it or disable it.

agree - as long as everywhere else we follow the pattern that we expose separate setters, I think we should be good

packages/xl-ai/src/AIExtension.ts
YousefED: not sure about this, let's discuss later (not zustand specific)

PR #984, opened 2024-08-02
fix: Emoji suggestion slash menu does not display

### LINKED ISSUE #975: Emoji Slash - lack of identification and bug
**Describe the bug**

After selecting the **Emoji** option from the /. it still showing that **use / command**. 
**:** operation perfectly for emoji, better if the : is added as default when the Emoji option is selected.

image

jkcs:
close #975 

Currently unable to use ":" to open emoji, because it will cause ":" not to be cleared. Can we add a new `pluginState` to mark it?

file: `packages/core/src/extensions/SuggestionMenu/SuggestionPlugin.ts`

```js
clearQuery = () => {
    if (this.pluginState === undefined) {
      return;
    }

    this.editor._tiptapEditor
      .chain()
      .focus()
      .deleteRange({
        from:
          this.pluginState.queryStartPos! -
          (this.pluginState.fromUserInput
            ? this.pluginState.triggerCharacter!.length
            : 0),
        to: this.editor._tiptapEditor.state.selection.from,
      })
      .run();
  };

```


matthewlipski:
Hi, thanks for this PR! It looks good, but I'm seeing a bug where if you open the emoji picker (with the : key or slash menu), start typing, and then keep hitting backspace, the whole picker briefly flashes after clearing the last character of the query.

YousefED (commented):
Thanks for fixing this!

Should we give some more thought about naming conventions? When reviewing, to me `payload` seems quite generic, and also `fromUserInput` is not very clear what it does. I know `fromUserInput` was already there, but maybe we should rename it to make more clear what it does?

YousefED:
> What do you think of `deleteTriggerCharacter` and `extraData`?

`deleteTriggerCharacter` first sounds great if that's exactly what it does, way more descriptive.

`extraData` is just as generic as `payload`. I guess it's not just the naming in this case, but more the question whether we need a generic "bag" of properties like this, or can we write it more strongly typed?

matthewlipski:
Tbh I think we can do with just `deleteTriggerCharacter` and `ignoreMinQueryLength`. While only the grid suggestion currently supports `minQueryLength` and this implies that the regular suggestion menu also uses it, I think `minQueryLength` is really rarely used anyway outside the default emoji picker so I don't think this is a big concern. Imo if we just change `fromUserInput` to `deleteTriggerCharacter` and `payload.ignoreQueryLength` to `ignoreQueryLength`, we'll be ready to merge👍

YousefED:
Great! Btw @jkcs , are you on discord?

PR #992, opened 2024-08-07
feat: added german translation

Muvels:
### Changes
- Added: German language file (de.ts) containing translations.
- Tested: Manually verified the translations by switching the app language to German and checking each screen for correctness and consistency.


YousefED:
Thanks! Will merge. What are you building @Muvels ?

PR #994, opened 2024-08-07
feat: Add collaborative initial content example

mfanselmo:
Hey! This pr adds an example on how to load an existing document when using the collaborative mode.

This approach passing the yjs update vector works for me and my use case, but if you think it can be reworked let me know. 

Stackblitz example: https://stackblitz.com/edit/github-a6kavu-x9g2cp?file=package.json


AlexBSoft:
Thanks a lot! This is really helped me.

PR #985, opened 2024-08-02
Add feature  - Spanish dictionary.

aespinoza96:
This PR is to add the spanish internationalization and make it available.

PR #999, opened 2024-08-12
Feature: Add missing line in index.ts for locales

aespinoza96:
In the previous PR #985 the index.ts is missing the line to export the new language format.


matthewlipski:
Fixed in #1002

PR #1000, opened 2024-08-12
fix: Add exports for vanilla block helpers

matthewlipski:
Right now, the only way to extend default blocks is to redefine them entirely, using most of the existing code. While we should figure out a proper API for extending blocks at some point, right now it's not possible to even redefine the default vanilla blocks as some exports are missing. This PR adds them.


YousefED (approved):
PR 1000!

PR #1022, opened 2024-08-22
feat: Custom inline equation using the BlockNote API

jkcs:
## API Extension Changes:

1. Pass the Tiptap node to props for inline content to access node attr, nodeSize, etc.
2. Allow custom `renderHTML` implementation for inline content (in this case, to address copying issues)"

## Link
 +  #953
 +  #741


YousefED:
Thanks! The UX of this looks great. I hope to dive into the code coming week. I'd like to research what API surface we'd need to add so this can be built using purely the BlockNote API (and not TipTap internals). Will keep you posted, but lmk in case you have suggestions for this!

YousefED (commented):
Great stuff. This example shows where the BN API is lacking and is a great starting point to improve it.

I've implemented an `isSelected` prop (see https://github.com/TypeCellOS/BlockNote/pull/1022/commits/519f34a00aaa767f3686a59ababbd2bcb53fec9d) and made inline content selectable to make the logic a lot easier. We now don't need to listen for "cursor edge keyboard navigations" anymore, but can let prosemirror handle this. 

Let me know what you think @jkcs! Do you see any new bugs popping up with the new approach?

There are a couple of todos left with this new approach:

- [ ] Make sure the formatting toolbar doesn't show
- [ ] Research if there are any other downsides / side effects (for example for other examples) of setting `selectable: true`. Should we make this configurable or not?
- [ ] Implement remaining APIs (marked as TODO in code)
- [ ] Review RenderHTML (@jkcs see my PR comment about copying)

PR #1024, opened 2024-08-22
fix: Show formatting toolbar when editor isn't editable

matthewlipski:
There seems to have been a regression as the formatting toolbar no longer shows when the editor is not editable. It should be the responsibility of toolbar buttons to determine whether to be shown instead, as the file download button should always be visible when a file block is selected.

This PR changes the behaviour to show the formatting toolbar when the DOM selection is within the editor, as opposed to when the editor is focused. This means that clicking off the editor still hides the toolbar, but it doesn't force it to stay hidden when the editor isn't editable.


YousefED (approved):
Looks ok. Conceptually we could wonder if non-editable features (like download button) should be in a "formatting toolbar" or not, or what the best API for this would be. For now this seems ok though

Issue #1008, opened 2024-08-14
Error [ERR_REQUIRE_ESM]: require() of ES Module 


odufuwa-segun:
**Describe the bug**
Error when starting nodejs application

<img width="1340" alt="Screenshot 2024-08-14 at 2 55 01 PM" src="https://github.com/user-attachments/assets/dfa7451f-9efc-4661-8a4d-8b759c67d9c8">

node_modules/ts-node/dist/index.js:851
            return old(m, filename);
                   ^
Error [ERR_REQUIRE_ESM]: require() of ES Module node_modules/@blocknote/core/node_modules/rehype-parse/index.js from node_modules/@blocknote/core/dist/blocknote.umd.cjs not supported.
Instead change the require of index.js in node_modules/@blocknote/core/dist/blocknote.umd.cjs to a dynamic import() which is available in all CommonJS modules.
    at require.extensions.<computed> [as .js] (node_modules/ts-node/dist/index.js:851:20)
    at node_modules/@blocknote/core/dist/blocknote.umd.cjs:1:99
    at Object.<anonymous> (node_modules/@blocknote/core/dist/blocknote.umd.cjs:1:2397)
    at require.extensions.<computed> [as .js] (node_modules/ts-node/dist/index.js:851:20)
    at node_modules/@blocknote/server-util/dist/blocknote-server-util.umd.cjs:1:70
    at Object.<anonymous> (node_modules/@blocknote/server-util/dist/blocknote-server-util.umd.cjs:1:487)
    at require.extensions.<computed> [as .js] (node_modules/ts-node/dist/index.js:851:20)
    at Object.<anonymous> (src/yjs.ts:8:23)
    at m._compile (node_modules/ts-node/dist/index.js:857:29)
    at require.extensions.<computed> [as .ts] (node_modules/ts-node/dist/index.js:859:16)
    at Object.<anonymous> (src/socket.ts:4:18)
    at m._compile (node_modules/ts-node/dist/index.js:857:29)
    at require.extensions.<computed> [as .ts] (node_modules/ts-node/dist/index.js:859:16)
    at Object.<anonymous> (src/index.ts:32:21)
    at m._compile (node_modules/ts-node/dist/index.js:857:29)
    at require.extensions.<computed> [as .ts] (node_modules/ts-node/dist/index.js:859:16)
    at phase4 (node_modules/ts-node/dist/bin.js:466:20)
    at bootstrap (node_modules/ts-node/dist/bin.js:54:12)
[...]

PR #1026, opened 2024-08-22
Dynamic import of unified dependencies

YousefED:
make unified imports dynamic to support loading from non esm projects

closes https://github.com/TypeCellOS/BlockNote/issues/1008


matthewlipski (approved):
Looks good👍 This is pretty much the same thing we do as with the emojis for the emoji picker right? i.e.

```ts
// load dynamically because emoji-mart doesn't specify type: module and breaks in nodejs
emojiMart = await import("emoji-mart");
```

But with multiple modules instead of just 1

PR #1045, opened 2024-09-01
Fix error in suggestion-menus.mdx

kfrp:
Fixed an error that refers to an `openSuggestionMenu` method on editor which can be used for opening a suggestion menu programmatically. However the method is actually called `openSelectionMenu`.


matthewlipski:
Huh, bizarre we never noticed it's called `openSelectionMenu` and not `openSuggestionMenu`. Thanks for bringing this up! Closing this issue because the correct name should actually be `openSuggestionMenu`, so opened #1047 will fix the naming in code.

PR #1052, opened 2024-09-06
fix: External file drag/drop and copy/paste

matthewlipski:
This PR fixes 2 issues with dropping and pasting external files into the editor.

First, the files are now always inserted as new blocks instead of updating existing ones, and also dropping has been fixed to use the block under the mouse cursor instead of the currently selected one.

Second, using file extensions in a file block's `fileBlockAcceptMimeTypes` was throwing errors before, and is now fixed. File extensions and MIME types are now checked independently, whereas before, file extensions were just treated as MIME types causing some logic to fail.

### DIFF
--- a/packages/core/src/api/parsers/handleFileInsertion.ts
+++ b/packages/core/src/api/parsers/handleFileInsertion.ts
@@ -82,7 +82,7 @@ export async function handleFileInsertion<
     // Gets file block corresponding to MIME type.
     let fileBlockType = "file";
     for (const fileBlockConfig of fileBlockConfigs) {
-      for (const mimeType of fileBlockConfig.fileBlockAcceptMimeTypes || []) {
+      for (const mimeType of fileBlockConfig.fileBlockAccept || []) {
         const isFileExtension = mimeType.startsWith(".");
         const file = items[i].getAsFile();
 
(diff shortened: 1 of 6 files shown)


packages/core/src/api/parsers/handleFileInsertion.ts
```diff
@@ -72,9 +83,23 @@ export async function handleFileInsertion<
     let fileBlockType = "file";
     for (const fileBlockConfig of fileBlockConfigs) {
       for (const mimeType of fileBlockConfig.fileBlockAcceptMimeTypes || []) {
-        if (checkMIMETypesMatch(items[i].type, mimeType)) {
-          fileBlockType = fileBlockConfig.type;
-          break;
+        const isFileExtension = mimeType.startsWith(".");
```
YousefED, line 86: Let's change `const mimeType` to `const mimeTypeOrExtension` (technically, mimetype is different from a file extension). Maybe we should also change `fileBlockAcceptMimeTypes` to `fileBlockAccept`

packages/core/src/api/parsers/handleFileInsertion.ts
```diff
@@ -72,9 +83,23 @@ export async function handleFileInsertion<
...
+        const file = items[i].getAsFile();
+
+        if (file) {
+          if (
+            (!isFileExtension &&
+              file.type &&
+              checkMIMETypesMatch(items[i].type, mimeType)) ||
+            (isFileExtension &&
+              checkFileExtensionsMatch(
+                "." + items[i].getAsFile()!.name.split(".").pop(),
```
YousefED, line 96: `file` instead of`items[i].getAsFile()!`

YousefED (commented):
nice, some minor suggestions

PR #1060, opened 2024-09-08
fix: xss in link toolbar and file download

zeyu2001:
There is an XSS vulnerability in the file upload and link toolbar features.

To reproduce, create a file embed or a link with `javascript:alert(origin)`. Click on the download / open button, and `alert(origin)` will be run in the user's browser.

image

This fixes the XSS vulnerability by providing a sensible default when the protocol is `javascript:`.


YousefED:
Thanks @zeyu2001 , great spot. What are you using BlockNote for?

zeyu2001:
We're building a collaborative note taking app that has publishing features similar to notion - so for us it's important that content rendered in another user's browser doesn't contain dangerous links

matthewlipski (approved):
Looks great!

Issue #1058, opened 2024-09-07
Opting out from animations


softmarshmallow:
Is there a way I can easily opt-out from default animations? without getting too dirty with custom css

PR #1062, opened 2024-09-09
add option to disable animations

YousefED:
closes https://github.com/TypeCellOS/BlockNote/issues/1058

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -67,6 +67,8 @@ import { en } from "../i18n/locales";
 
 import { Transaction } from "@tiptap/pm/state";
 import { createInternalHTMLSerializer } from "../api/exporters/html/internalHTMLSerializer";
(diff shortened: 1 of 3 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -353,6 +362,10 @@ export class BlockNoteEditor<
           ...(this.filePanel ? [this.filePanel.plugin] : []),
           ...(this.tableHandles ? [this.tableHandles.plugin] : []),
           PlaceholderPlugin(this, newOptions.placeholders),
+          ...(this.options.animations ?? true
+            ? [PreviousBlockTypePlugin()]
+            : []),
+          NonEditableBlockPlugin(),
```
matthewlipski, line 368: I'm not sure `NonEditableBlockPlugin` should be under the `BlockNoteUIExtension` since it handles keystrokes for blocks without inline content. Let's maybe move it to `getBlockNoteExtensions` or back to the `blockContainer` node.

matthewlipski (changes requested):
Nice, just 1 comment

matthewlipski (approved):

PR #1068, opened 2024-09-12
fix: `blocksToHTMLLossy` props as `data-` attributes

### LINKED ISSUE #1050: Feedback for “Converting Blocks to HTML”
Hey team, i think i've caught a small bug if I change the alignment of text.

## Steps to Reproduce:

1. Change the alignment of text or elements within the editor.
2. Preview or output the signature to HTML.
3. Note that the alignment changes made in the editor are not reflected in the HTML output.

## Acceptance Criteria:

- The Editor should accurately reflect the alignment changes made within the editor when outputting to HTML for email signatures.
- Alignment changes such as center, left, right, or justify should be properly translated to the HTML output.
- Verify that the HTML output displays the text or elements with the correct alignment as set in the editor.

Link to Loom

https://www.loom.com/share/26bd2272809e4ab79235a17ff41df3b9

https://www.loom.com/share/26bd2272809e4ab79235a17ff41df3b9?focus_title=1&muted=1&from_recorder=1

matthewlipski:
This PR adds the `data-` attributes rendered based on a block's props to content of the block after stripping the block structure in `blocksToHTMLLossy`. This is to reduce the amount of data lost when using this function.

Closes #1050


packages/core/src/api/exporters/html/__snapshots__/fontSize/basic/external.html
```diff
@@ -1 +1 @@
-<p class="bn-inline-content"><span style="font-size: 18px" data-style-type="fontSize" data-value="18px" data-editable="">This is text with a custom fontSize</span></p>
```
YousefED, line 1: class name removed intentionally?

packages/core/src/api/exporters/html/util/simplifyBlocksRehypePlugin.ts
```diff
@@ -142,15 +168,39 @@ export function simplifyBlocks(options: SimplifyBlocksOptions) {
         // Lifts all children out of the current block, as only list items should allow nesting.
         tree.children.splice(i + 1, 0, ...blockGroup.children);
         // Replaces the block with only the content inside it.
-        tree.children[i] = blockContent.children[0];
+        const content = blockContent.children[0] as HASTElement;
```
YousefED, line 171: small code style suggestion: let's make sure we reuse the duplicated code between this part and 191+?

YousefED (commented):
Looks like a great improvement!

I think at some point we could either supply a stylesheet which uses the data attributes, or maybe consider setting certain properties on `style` attributes, but for now this solves the main issue

packages/core/src/api/exporters/html/__snapshots__/fontSize/basic/external.html
matthewlipski: Yep I figured it really doesn't add anything to include `bn-` class names since they're only used for block styling and blocks are removed here.

PR #1069, opened 2024-09-12
fix: Loading indicator when dropping/pasting files

matthewlipski:
This PR adds a `loading` prop to the default file blocks, which causes them to render a `Loading...` indicator when `true`. Unfortunately,  the delay is happening in `editor.uploadFile`, which is not really related to the block itself. So that means we don't have an existing way of checking if a file is being uploaded while the block is rendered.


matthewlipski:
Do you have anything in mind? I was thinking we could expose an `editor.fileUploadInProgress` property which stores `undefined` or a block ID depending on if a file is currently being uploaded, and which file block it's being uploaded for. Then ig we can check that in the render function and avoid setting it in the state. Does this sound like a reasonable idea?

YousefED (approved):

PR #1072, opened 2024-09-13
feat: SuggestionMenuController.tsx support ignoreQueryLength

jkcs:
In #984, we added the ignoreQuery state, and I think it applies `SuggestionMenuController` as well


matthewlipski:
Good spot, thanks!

matthewlipski (approved):

PR #1080, opened 2024-09-16
feat: Added parameter for updating custom inline content in `render`

matthewlipski:


YousefED (commented):
imo this looks good. Have you been able to test:
- the new method
- whether vanilla (non-react) IC still works as expected?

Maybe discuss if there's an easy way to unit test this

matthewlipski:
> imo this looks good. Have you been able to test:
> 
> * the new method
> * whether vanilla (non-react) IC still works as expected?
> 
> Maybe discuss if there's an easy way to unit test this

Can confirm that everything works as expected for both vanilla and React blocks. Tested by adding an `click` listener to the existing mentions IC examples which calls the update function.

PR #1087, opened 2024-09-18
feat: Selectable non-editable text in blocks without inline content

matthewlipski:
This PR makes it possible to have blocks without inline content which contain selectable, non-editable text. Currently, you cannot select text while the block is selected, and selecting text while a different block is selected in the editor leads to a mismatch in editor & DOM selections. You can toggle this functionality using the `canSelectText` entry in a block config.


packages/core/src/extensions/FullySelectedNode/FullySelectedNodeExtension.ts
```diff
@@ -0,0 +1,72 @@
...
+export const FullySelectedNodeExtension = Extension.create({
+  name: "fullySelectedNode",
+  onCreate() {
+    document.addEventListener("selectionchange", () => {
+      onSelectionChange(this.editor);
+    });
+  },
+  onDestroy() {
+    document.removeEventListener("selectionchange", () => {
+      onSelectionChange(this.editor);
```
YousefED, line 69: this doesn't work (you're removing a newly created function)

packages/core/src/extensions/FullySelectedNode/FullySelectedNodeExtension.ts
```diff
@@ -0,0 +1,72 @@
...
+    return;
+  }
+
+  const blockInfo = getBlockInfoFromPos(
+    editor.state.doc,
+    editor.state.selection.from
+  );
+
+  // Selected block has no inline content.
+  const selectedNodeHasNoContent =
```
YousefED, line 37: a) remove double negation
b) should we check whether the node has the option enabled?

packages/core/src/extensions/FullySelectedNode/FullySelectedNodeExtension.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (
+    !editor.view.dom.contains(selection.anchorNode) ||
+    !editor.view.dom.contains(selection.focusNode)
+  ) {
+    return;
+  }
+
+  // Node selection is active.
+  const isNodeSelection = "node" in editor.state.selection;
+  if (!isNodeSelection) {
```
YousefED, line 26: this check seems easier than the code above. move to top? (even before document.getSelection?)

packages/core/src/schema/blocks/types.ts
```diff
@@ -60,6 +61,7 @@ export type BlockConfig =
       type: string;
       readonly propSchema: PropSchema;
       content: "inline" | "none" | "table";
+      canSelectText?: boolean;
```
YousefED, line 64: maybe `allowTextSelection`?

packages/core/src/editor/tiptap.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+@keyframes ProseMirror-cursor-blink {
+    to {
+        visibility: hidden;
+    }
+}
+
+/* Edited section to replace `.ProseMirror-hideselection` with `.ProseMirror-fullyselected */
+.ProseMirror-fullyselected *::selection {
```
YousefED, line 59: let's review this file together and maybe rename to `fullyselectednode` or sth

packages/core/src/editor/Block.css
```diff
@@ -268,11 +268,13 @@ NESTED BLOCKS
 }
 
 [data-file-block] .bn-file-block-content-wrapper {
-  cursor: pointer;
   display: flex;
   flex-direction: column;
   justify-content: stretch;
-  user-select: none;
+}
```
YousefED, line 274: why this change?

packages/core/src/api/exporters/copyExtension.ts
```diff
@@ -65,20 +27,35 @@ const copyToClipboard = <
   // the selection to the parent `blockContainer` node. This is
   // for the use-case in which only a block without content is
   // selected, e.g. an image block.
-  if (
+  const fragment =
```
YousefED, line 30: discuss file on call

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -61,6 +65,75 @@ export type CustomBlockImplementation<
   ) => PartialBlockFromConfig<T, I, S>["props"] | undefined;
 };
 
+export function fixNodeViewTextSelection(
+  props: NodeViewRendererProps,
+  nodeView: NodeView
+) {
+  // Necessary for DOM to handle selections.
+  nodeView.ignoreMutation = () => true;
```
YousefED, line 73: shouldn't this only be the case for selection events?

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -61,6 +65,75 @@ export type CustomBlockImplementation<
...
+      return true;
+    }
+
+    // Prevent all drag events.
+    if (event.type.startsWith("drag")) {
+      event.preventDefault();
+      return true;
+    }
+
+    // Keyboard events should be handled by the browser. This doesn't prevent
```
YousefED, line 94: sure? what about backspace or delete?

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -61,6 +65,75 @@ export type CustomBlockImplementation<
...
+          props.editor.view.state.tr.setSelection(
+            NodeSelection.create(props.editor.view.state.doc, nodeStartPos)
+          )
+        );
+      }
+
+      // If the target element contains only text, the browser should handle
+      // the event to update the selection. Otherwise, the event should be
+      // handled by BlockNote to select the block.
+      return (
```
YousefED, line 127: let's discuss

PR #1099, opened 2024-09-24
feat: example for code block

JaeungJayJang:
Hello, This PR adds an example of creating custom code block.
The example is based on Custom Alert Block from the official site and also from [blocknote-code](https://github.com/defensestation/blocknote-code) by defensestation.

I am new to BlockNote and it was hard to find example on creating custom code block. I saw many people faced the same issue and I hope this can help many people. :)

Stackblitz example: https://stackblitz.com/edit/vitejs-vite-fnzdjj?file=index.html

### DIFF
--- a/examples/06-custom-schema/05-code-block/.bnexample.json
+++ b/examples/06-custom-schema/05-code-block/.bnexample.json
@@ -1,0 +1,18 @@
+{
+  "playground": true,
+  "docs": true,
+  "author": "JaeungJayJang",
+  "tags": [
+    "Intermediate",
+    "Blocks",
+    "Custom Schemas",
+    "Suggestion Menus",
+    "Slash Menu"
+  ],
+  "dependencies": {
+    "@uiw/react-codemirror": "^4.23.3",
+    "@uiw/codemirror-extensions-langs": "^4.23.3",
+    "@uiw/codemirror-theme-material": "^4.23.3",
+    "lucide-react": "^0.407.0"
+  }
+}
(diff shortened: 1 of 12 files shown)


trancethehuman:
It'd be awesome to be able to choose from a drop down list of programming languages too. This looks awesome.

JaeungJayJang:
@trancethehuman Thanks for the suggestion! Just added language selection using NativeSelect from Mantine/core.
it's applied on Stackblitz and I will push commit here as well

trancethehuman:
> @trancethehuman Thanks for the suggestion! Just added language selection using NativeSelect from Mantine/core. it's applied on Stackblitz and I will push commit here as well

wow that's a lot of options 🔥 

Wondering if you can add bash and text (yep, just regular unhighlighted text) options. A search box would probably do well here as well.

JaeungJayJang:
@trancethehuman I pulled entire language supports from CodeMirror XD
Bash is not provided from the CodeMirror's default language set yet, but you could try 'shell' as an alternative for now.
It seems that regular text does not exist; maybe it's because not passing anything into extension result unhighlighted text. I will play around to extend that feature.

You can try something like this for the serach box option:
https://stackblitz.com/edit/vitejs-vite-7pzgrg?file=Code.tsx
The search box uses useState and useState cannot be called inside render function. So I needed to create component and pass prop into it, but I am not sure how to set the type for this props 😅
this example uses component from mantine/core, but the style is not from mantine; the official doc for mantine require the app to be wrapped with MantineProvider, but then i had to modify main.tsx as well and I am not sure if that's okay for example purpose, so I added few style changes on **styles.css** file.

trancethehuman:
this is super good.

@YousefED is this good for codeblock?

lamtuanamc:
@JaeungJayJang great work, can you add `Copy button` to copy and keep format for code?

PR #1103, opened 2024-09-25
fix: Block update removing children when setting content

### LINKED ISSUE #1101: Using `editor.updateBlock` while changing the content of the block deletes the children
**Describe the bug**

When using the editor API `editor.updateBlock` and changing the block's content, the children are removed. I think this is unexpected behavior but please let me know if there is a reason for this, I have not looked very into detail in the code

**To Reproduce**

https://github.com/user-attachments/assets/441e3b99-1541-455c-943f-1d8d371c3d2e

https://stackblitz.com/edit/github-7jn7g7?file=SideMenuItems.tsx

matthewlipski:
Closes #1101

### DIFF
--- a/packages/core/src/api/blockManipulation/__snapshots__/blockManipulation.test.ts.snap
+++ b/packages/core/src/api/blockManipulation/__snapshots__/blockManipulation.test.ts.snap
@@ -712,3 +712,210 @@ Line2",
   },
 ]
 `;
+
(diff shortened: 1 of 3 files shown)


YousefED (commented):
Nice! However, as it's pretty core functionality, I think we should add more unit tests for all the different scenarios before merging this

YousefED (approved):
nice

PR #1104, opened 2024-09-26
fix: `elementOverflow` code

matthewlipski:


packages/react/src/util/elementOverflow.ts
```diff
@@ -1,9 +1,47 @@
...
+export function elementOverflow(
+  element: HTMLElement,
+  getContainer?: () => HTMLElement | undefined
+) {
+  let container = getContainer?.();
+  if (!container) {
+    container = defaultGetContainer(element);
+  }
+  if (!container) {
+    return "none";
```
YousefED, line 33: can this happen or should it be an error?

packages/react/src/util/elementOverflow.ts
```diff
@@ -1,9 +1,47 @@
...
+  return container;
+};
+
+export function elementOverflow(
+  element: HTMLElement,
+  getContainer?: () => HTMLElement | undefined
+) {
+  let container = getContainer?.();
+  if (!container) {
+    container = defaultGetContainer(element);
```
YousefED, line 30: why not just say `getContainer = defaultGetContainer` in the function signature?

packages/react/src/util/elementOverflow.ts
```diff
@@ -1,9 +1,47 @@
...
+    container.parentElement
+  ) {
+    container = container.parentElement;
+  }
+
+  if (
+    !container.classList.contains("bn-suggestion-menu") &&
+    !container.classList.contains("bn-grid-suggestion-menu")
+  ) {
+    return;
```
YousefED, line 18: let's make explicit with return undefined (instead of empty return statement)

packages/react/src/util/elementOverflow.ts
```diff
@@ -1,9 +1,47 @@
-export function elementOverflow(element: HTMLElement, container: HTMLElement) {
+// Default for `getContainer`, meant for use in suggestion menus. Traverses
+// ancestors of the passed element until a suggestion menu root is found.
+const defaultGetContainer = (element: HTMLElement) => {
+  let container: HTMLElement = element;
+
+  while (
```
YousefED, line 6: I think you can just use https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

YousefED (changes requested):

PR #1107, opened 2024-09-29
feat: Added `resolveFileUrl` block ID field

matthewlipski:
This PR adds a block ID field to `resolveFileUrl` as there seems to be a use case for it as per [this Discord thread](https://discord.com/channels/928190961455087667/1286268979253481503/1286268979253481503).

Also fixes the upload file button not including the block ID when calling `editor.uploadFile`, and the video block not calling `editor.resolveUrl`.

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -137,7 +137,7 @@ export type BlockNoteEditorOptions<
    * implementing custom protocols / schemes
    * @returns The URL that's
    */
-  resolveFileUrl: (url: string) => Promise<string>;
+  resolveFileUrl: (url: string, blockId?: string) => Promise<string>;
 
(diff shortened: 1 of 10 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -282,7 +282,10 @@ export class BlockNoteEditor<
   private onUploadStartCallbacks: ((blockId?: string) => void)[] = [];
   private onUploadEndCallbacks: ((blockId?: string) => void)[] = [];
 
-  public readonly resolveFileUrl: (url: string) => Promise<string>;
+  public readonly resolveFileUrl: (
+    url: string,
+    blockId?: string
```
YousefED, line 287: shouldn't this parameter be required? Or are there cases when it's undefined? If so, I think we should document those

PR #1111, opened 2024-09-30
fix: Table handles hide when selection is active

matthewlipski:
I experimented with a few different methods of achieving this, but this was the solution that gave the best UX while also adding not much complexity.

### DIFF
--- a/packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
+++ b/packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
@@ -127,6 +127,7 @@ export class TableHandlesView<
     };
 
     pmView.dom.addEventListener("mousemove", this.mouseMoveHandler);
+    pmView.dom.addEventListener("mouseup", this.mouseUpHandler);
 
     pmView.root.addEventListener(
       "dragover",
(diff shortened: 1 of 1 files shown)


packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -145,6 +146,16 @@ export class TableHandlesView<
       return;
     }
 
+    // Prevents table handles from showing while a selection is active.
+    if (!this.pmView.state.selection.empty) {
```
YousefED, line 150: won't this also hide the draghandles when the mouse is not down? i.e. when the user just selected a word in a table cell?

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -231,6 +242,11 @@ export class TableHandlesView<
     return false;
   };
 
+  mouseUpHandler = (event: MouseEvent) => {
+    // Have to wait for changes from view to propagate to state.
+    setTimeout(() => this.mouseMoveHandler(event), 10);
```
YousefED, line 247: this seems a bit hacky tbh

YousefED (approved):
nice

PR #1126, opened 2024-10-04
feat: Keyboard shortcuts to move selected block up/down

matthewlipski:


packages/core/src/api/blockManipulation/moveBlock.ts
```diff
@@ -0,0 +1,177 @@
...
+  editor.removeBlocks([block]);
+  editor.insertBlocks([block], referenceBlock, placement);
+
+  updateBlockSelectionFromData(editor, selectionData);
+}
+
+export function moveBlockUp(editor: BlockNoteEditor<any, any, any>) {
+  const editorSelection = editor.getSelection();
+  if (editorSelection && editorSelection.blocks.length > 1) {
+    return;
```
YousefED, line 120: add comment that for now this is not supported

packages/core/src/api/blockManipulation/moveBlock.test.ts
```diff
@@ -0,0 +1,194 @@
...
+  );
+
+  if (selectionType === "cell") {
+    editor._tiptapEditor.view.dispatch(
+      editor._tiptapEditor.state.tr.setSelection(
+        CellSelection.create(
+          editor._tiptapEditor.state.doc,
+          editor._tiptapEditor.state.doc.resolve(startPos + 3).before(),
+          editor._tiptapEditor.state.doc
+            .resolve(startPos + contentNode.nodeSize - 3)
```
YousefED, line 79: fyi; at some point I want to get rid of all manual positions and keep these in helper functions; the reason is that they make implicit assumptions about our schema; so it's not future-proof when we start supporting more complex pm schemas

YousefED: Actually, can we do these positions more "semantically"? Like $from.start(), $from.end() for the text selectiontype

PR #1127, opened 2024-10-05
Update style.css

jehaad1:
Since ":empty" pseudo-class doesn't supported in some of the browsers yet (such as my browser Arc), I added another solution to that's more supported by other browsers (such as my browser Arc).

### DIFF
--- a/packages/mantine/src/style.css
+++ b/packages/mantine/src/style.css
@@ -146,7 +146,8 @@
   width: fit-content;
 }
 
-.bn-mantine .bn-toolbar:empty {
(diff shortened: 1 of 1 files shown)


jehaad1:
Sorry I didn't know if I deleted the repo, it will close the pr,

I changed the css in the "packages/mantine/src/style.css"
line 149

from
```css
.bn-mantine .bn-toolbar:empty {
```
to
```css
.bn-mantine .bn-toolbar:empty,
.bn-mantine .bn-toolbar:not(:has(*)) {
```

PR #1147, opened 2024-10-13
add support for blocknote transactions

YousefED:
This is a POC for blocknote-level transactions. It should make it possible to clean our existing commands (those in `blockManipulation` further down by migrating away from tiptap-style commands). It also unlocks the possibility for userland-transactions. 

For example, the following bug is fixed:
- select a number of paragraphs
- change type to "heading"
- "undo" will undo them one by one (after the fix, "undo" will reset all headings back to paragraphs in one step as expected)

closes https://github.com/TypeCellOS/BlockNote/issues?q=is%3Aissue+is%3Aopen+undo

TODO:
- add tests
- See if there are bugs with this approach. For example, what if you call other functions in a transact block that depend on `editor.state`? It's likely they should read the state from the `activeTransaction` instead (diagnose, create tests, and fix)
- if there are other BlockNote APIs called in a loop (similar to blocktype select), migrate them to use `transact`

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -2,7 +2,6 @@ import { EditorOptions, Extension, getSchema } from "@tiptap/core";
 import { Node, Schema } from "prosemirror-model";
 // import "./blocknote.css";
 import * as Y from "yjs";
-import { insertContentAt } from "../api/blockManipulation/insertContentAt.js";
 import { insertBlocks } from "../api/blockManipulation/commands/insertBlocks/insertBlocks.js";
 import {
   moveBlockDown,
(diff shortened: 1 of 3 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -484,8 +485,28 @@ export class BlockNoteEditor<
...
+   * we can dispatch them as a single operation (thus creating only a single undo step)
+   */
+  public transact(func: () => void) {
+    this.activeTransaction = this._tiptapEditor.state.tr;
+    try {
+      func();
+      this._tiptapEditor.dispatch(this.activeTransaction);
+    } finally {
+      this.activeTransaction = undefined;
+    }
```
nperez0111, line 509: Just for awareness, Tiptap already has a way to capture transactions without immediately dispatching them: https://github.com/ueberdosis/tiptap/blob/44cae5badc07e715bf6554e7042a9e4e926e57b7/packages/core/src/Editor.ts#L414-L429

Issue #1168, opened 2024-10-18
blocksToFullHtml renders null for numbered lists


entropi:
**Describe the bug**
When using editor.blocksToFullHtml, numbered lists list `null` instead of the actual number.

**To Reproduce**
generate some blocks (in my case from `tryParseMarkdownToBlocks`) which contain numbered list items and then attempt to render those blocks to html using the `blocksToFullHTML`. Null is there instead of the number.

**Misc**

- [x] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

(PoggioLabs is the sponsor)

YousefED:
Thanks! We'll look into this next week 👍

PR #1174, opened 2024-10-22
fix: set numbered list data-index on export to html

YousefED:
closes https://github.com/TypeCellOS/BlockNote/issues/1168


curiousyuvi:
Really awaiting this fix, one of our customers wants this fixed in our tool.

jueinin:
waiting for it

PR #1175, opened 2024-10-22
fix: Unexpected update block

### LINKED ISSUE #771: Typed or Copied Content Replaces Custom Alert Block Instead of Staying Within It
Description:
I have customized a block type "alert" followed by a custom block on the BlockNote homepage. However, every time I type '-', '*', '+', '1.' or copy a text segment within the same editor, it replaces the "alert" block instead of placing the content within it.

Steps to Reproduce:

Create a custom "alert" block as per the documentation.
Type '-', '*', '+', '1.' or copy and paste a text segment within the "alert" block.
Observe that the content replaces the "alert" block instead of staying within it.
Expected Behavior:
The typed or copied content should remain within the "alert" block instead of replacing it.

Actual Behavior:
The "alert" block is replaced by the typed or copied content.

Link to Documentation:
[BlockNote Custom Schemas - Custom Blocks](https://www.blocknotejs.org/docs/custom-schemas/custom-blocks)

lamtuanamc:
fix #771

### DIFF
--- a/packages/core/src/api/blockManipulation/commands/removeBlocks/__snapshots__/removeBlocks.test.ts.snap
+++ b/packages/core/src/api/blockManipulation/commands/removeBlocks/__snapshots__/removeBlocks.test.ts.snap
@@ -1,5 +1,445 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`Test removeBlocks > Remove all child blocks 1`] = `
+[
+  {
+    "children": [],
+    "content": [
+      {
+        "styles": {},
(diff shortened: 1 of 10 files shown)


lamtuanamc:
@YousefED
I'm encountering an issue where, while typing in a heading block, entering text like "1." automatically converts the block to a list block. I’m working on a logic check to detect if the block is of the default type (paragraph) so that auto-detection for lists is applied only in certain cases.

Could you please confirm if this approach is suitable, or suggest a recommended way to handle list detection that would only apply to specific block types (like the default paragraph)?

PR #1186, opened 2024-10-25
feat: Multi-column resizing

matthewlipski:
todo:

- [x] hide sidemenu when resizing?
- [x] side menu position (now too far to the left) for the 2nd column etc


packages/multi-column/src/pm-nodes/Column.ts
```diff
@@ -20,6 +22,45 @@ export const Column = createStronglyTypedTiptapNode({
   priority: 40,
   // defining: true, // TODO
 
+  addAttributes() {
+    return {
+      width: {
+        // Why does each column have a default width of 1, i.e. 100%? Because
```
YousefED, line 28: Clear! Let's think whether we want to use "1" or "100%" as attribute

could we also use `flex-grow`? or is `flex-basis` better because it takes px values?

packages/multi-column/src/pm-nodes/Column.ts
```diff
@@ -47,35 +88,33 @@ export const Column = createStronglyTypedTiptapNode({
     ];
   },
 
-  // TODO, needed? + type of attributes
```
YousefED, line 50: let's discuss!

packages/multi-column/src/extensions/ColumnResize/ColumnResizeExtension.ts
```diff
@@ -0,0 +1,376 @@
...
+
+  // In a `columnList` node, we expect that the average width of each column
+  // is 1. However, there are cases in which this stops being true. For
+  // example, making one really wide column and then removing it will cause
+  // the width average to go down. This isn't really an issue until the user
+  // tries to add a new column, which will, in this case, be wider than
+  // expected. Therefore, this function traverses the document and
+  // normalizes the average column width to 1 whenever and wherever
+  // necessary.
+  update = (view: EditorView) => {
```
YousefED, line 301: this is relatively heavy as it runs on every update and traverses all nodes. Could we also fix this when inserting a column for example?

packages/multi-column/src/extensions/ColumnResize/ColumnResizeExtension.ts
```diff
@@ -0,0 +1,376 @@
...
+  ): ColumnDefaultState | ColumnHoverState => {
+    let columnElement = event.target as HTMLElement;
+
+    // Do nothing if the event target is outside the editor.
+    if (!this.view.dom.contains(columnElement)) {
+      return { type: "default" };
+    }
+
+    // Traverse ancestors of event target until a column element is found.
+    while (
```
YousefED, line 65: you can use `.closest`

packages/multi-column/src/extensions/ColumnResize/ColumnResizeExtension.ts
```diff
@@ -0,0 +1,376 @@
...
+      // Prevent unnecessary state updates (when the state before and after
+      // is the same).
+      const bothDefaultStates =
+        pluginState.type === "default" && newState.type === "default";
+      const sameColumnIds =
+        pluginState.type !== "default" &&
+        newState.type !== "default" &&
+        pluginState.leftColumn.id === newState.leftColumn.id &&
+        pluginState.rightColumn.id === newState.rightColumn.id;
+      if (bothDefaultStates || sameColumnIds) {
```
YousefED, line 217: why ||?

YousefED (commented):
well documented and typed, that really helps! I think it looks good, but let's go over the ux / comments / code tomorrow :)

packages/multi-column/src/extensions/ColumnResize/ColumnResizeExtension.ts
matthewlipski: `bothDefaultStates` and `sameColumnIds` check 2 separate cases:

1. The column drag bar was not visible before and is still not visible.
2. The column drag bar was visible before and is still visible, but between the same pair of columns.

It wouldn't make sense to use `&&` here, as in one case we check if both states have type `"default"`, and in the other we check if neither of them have it.

packages/multi-column/src/extensions/ColumnResize/ColumnResizeExtension.ts
matthewlipski: Ideally we would do it whenever removing a column (since this is when the average width across columns may change), but doing it on insert should give the same UX. The only small difference is when the document is saved the widths won't not be normalized. But again this doesn't change anything for the user and it's much less complex to do it on insert instead.

YousefED: ok, can you move this logic to when we "drop-to-new-column"?

packages/multi-column/src/pm-nodes/Column.ts
```diff
@@ -47,35 +88,33 @@ export const Column = createStronglyTypedTiptapNode({
...
-    blockOuter.className = "bn-block-outer";
-    blockOuter.setAttribute("data-node-type", "blockOuter");
-    for (const [attribute, value] of Object.entries(HTMLAttributes)) {
-      if (attribute !== "class") {
-        blockOuter.setAttribute(attribute, value);
-      }
-    }
-
     const blockHTMLAttributes = {
+      // TODO: Should the block DOM attributes from the editor options be added
```
YousefED, line 93: remove

PR #1216, opened 2024-11-05
chore: externalize shiki languages

YousefED:
make sure languages are not output to core/dist


sindras (approved):
Thanks for the quick investigation @YousefED !

Issue #1213, opened 2024-11-04
Option to reduce bundle size by removing some languages from "Code Block" feature?


sindras:
**Is your feature request related to a problem? Please describe.**

Follow-up from #1177, is there a way to reduce bundle size by opting out from some languages?
Even though the languages have been filtered out as per discussion in the PR, they are still bundled in the production assets.
Screenshot attached below.

Also there does not seem to be any documentation about configuring the code block in the docs website.

**Describe the solution you'd like**

1. Provide the option to configure the Code Block, especially supported languages.
2. Tree-shake the unused languages.
3. Update the documentation to include an example of configuring the Code Block.

**Additional context**

This is generated by running `npm run build` with Remix framework. 

<img width="400" alt="Screenshot 2024-11-03 at 8 34 54 PM" src="https://github.com/user-attachments/assets/176f4fb5-e16f-40cc-97b1-dff4065fcc83">

**Bonus**

I am willing to help if someone can point me in the right direction regarding the tree-shaking.
Thank you so much for creating BlockNote! ❤️

YousefED:
Unfortunately #1216 doesn't solve this, but only makes sure the languages are not part of the @blocknote/core bundle. Your application would still have the language files in the output.

_Note that for the end-user, this should not be an issue as languages are loaded dynamically_

However, I understand you might still want to reduce your deployment output.

To do this, seems like there are two options:
a) we need to update the code to use https://shiki.style/guide/bundles#fine-grained-bundle and make languages configurable differently (requires work on BlockNote side)
b) possibly, remove languages manually using your rollup / webpack config

sindras:
> Note that for the end-user, this should not be an issue as languages are loaded dynamically

Yes I noticed that the lang is loaded dynamically.

> However, I understand you might still want to reduce your deployment output.

Exactly, these assets are significant in terms of size. Our deployment infrastructure charged us to a higher-tier usage plan based on upload size. We could have saved budget if we can reduce this dependencies. Thanks for taking a look at this, though 👍 . I will consider the options.

sindras:
That's super helpful, thank you @areknawo ! 

Just curious, is this the right way to customize the CodeBlock? I am having a TS error because the default language values are coming as `string[]` of lang ids but the option seem to expect a `SupportedLanguageConfig[]` (id, name, match) array.

```tsx
import {
  BlockNoteSchema,
  defaultBlockSpecs,
  customizeCodeBlock,
[...]

Issue #1212, opened 2024-11-03
The blue background color of the programming language options in the code block is displayed in full white text.


NGC2207:
**Describe the bug**
When I created a code block and clicked the language button, the options were in full white text and only displayed when I hovered over them.

**To Reproduce**
This bug can also be reproduced in the demo on the official Blocknote page (https://www.blocknotejs.org/).

![demo](https://github.com/user-attachments/assets/5a1f77dd-f29f-4f8a-9518-d00bce3c82e5)

**Misc**
- Node version:
- Package manager:
- Browser:

NGC2207:
The bug appears to be caused by the browser. I am using Mint 22 on my computer. When I open the official website with Chrome installed via google-chrome-stable_current_amd64.deb, the background color of the code language selection bar becomes white. However, this problem does not occur when I use Firefox.

nperez0111:
Unfortunately this is a default select menu, so we wouldn't have much control over it anyway

Issue #1210, opened 2024-11-02
Cannot copy and paste tsx code into code block


imaimai17468:
**Describe the bug**

I tried to copy and paste a tsx I wrote in VSCode into a code block and got the following error.

```
Uncaught (in promise) ShikiError: Language `typescriptreact` not found, you may need to load it first
    at Object.getLanguage (index.mjs:1773:13)
    at codeToTokensBase (index.mjs:840:29)
    at codeToTokens (index.mjs:1160:14)
    at Object.codeToTokens (index.mjs:1881:39)
    at parser (shiki.js:13:36)
    at parser (blocknote.js:6962:309)
    at eval (index.js:173:29)
    at Fragment.nodesBetween (index.js:105:31)
    at Node.nodesBetween (index.js:1238:22)
    at Fragment.nodesBetween (index.js:107:23)
    at Node.nodesBetween (index.js:1238:22)
    at Fragment.nodesBetween (index.js:107:23)
    at Node.nodesBetween (index.js:1238:22)
    at Node.descendants (index.js:1245:14)
    at calculateDecoration (index.js:162:7)
    at Plugin.apply (index.js:112:41)
    at EditorState.applyInner (index.js:851:45)
    at EditorState.applyTransaction (index.js:811:45)
    at EditorState.apply (index.js:787:21)
    at fe.dispatchTransaction (index.js:4575:34)
    at EditorView.dispatch (index.js:5698:33)
    at doPaste (index.js:3633:10)
    at EditorView.pasteHTML (index.js:5644:16)
    at fr (blocknote.js:10067:60)
    at Plugin.paste (blocknote.js:10094:24)
    at eval (index.js:3095:26)
    at EditorView.someProp (index.js:5498:50)
    at runCustomHandler (index.js:3093:17)
    at view.dom.addEventListener.view.input.eventHandlers.<computed> (index.js:3062:53)
```

**To Reproduce**

This bug can also be reproduced in the initial code of stackblitz, the template for this issue.
It can also be reproduced in the demo on the official Blocknote page (https://www.blocknotejs.org/).

https://github.com/user-attachments/assets/d3fb0ec3-290f-4c4f-bf9c-bf00a6a9fc0b

**Misc**
- Node version: -
- Package manager: - 
- Browser: 
  - Arc
  - version1.67.0 (55463)
  - Chromium Engine Version 130.0.6723.92
  - (The same problem occurs in Chrome.)

PR #1219, opened 2024-11-06
fix: CodeBlock fixes (select menu, bundled languages, paste from VS Code)

areknawo:
Fixes CodeBlock issues:
- #1212 (by setting text color for select options);
- #1210 (by improving language parsing from pasted HTML & matching `typescriptreact` alias for TSX);

Improves on the #1213, fixing duplicate languages in the bundle output (by importing from only a single Shiki bundle);

### DIFF
--- a/packages/core/src/blocks/CodeBlockContent/defaultSupportedLanguages.ts
+++ b/packages/core/src/blocks/CodeBlockContent/defaultSupportedLanguages.ts
@@ -43,7 +43,6 @@ export const defaultSupportedLanguages: SupportedLanguageConfig[] = [
         "shellscript",
         "sql",
         "svelte",
-        "tsx",
         "typescript",
         "vue",
         "vue-html",
@@ -58,6 +57,7 @@ export const defaultSupportedLanguages: SupportedLanguageConfig[] = [
       id: lang.id,
       name: lang.name,
     })),
+  { id: "tsx", name: "TSX", match: ["tsx", "typescriptreact"] },
   {
     id: "haskell",
     name: "Haskell",
--- a/packages/core/src/blocks/CodeBlockContent/defaultSupportedLanguages.ts
+++ b/packages/core/src/blocks/CodeBlockContent/defaultSupportedLanguages.ts
@@ -1,4 +1,4 @@
-import { bundledLanguagesInfo } from "shiki/bundle/web";
+import { bundledLanguagesInfo } from "shiki";
 
 export type SupportedLanguageConfig = {
   id: string;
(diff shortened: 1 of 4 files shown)


sindras (approved):
:shipit: thank you for improving this!

matthewlipski (approved):

packages/core/src/blocks/CodeBlockContent/defaultSupportedLanguages.ts
```diff
@@ -1,4 +1,4 @@
-import { bundledLanguagesInfo } from "shiki/bundle/web";
+import { bundledLanguagesInfo } from "shiki";
```
YousefED, line 1: @areknawo if we're selecting 1 bundle, doesn't it make more sense to only use `"shiki/bundle/web";` to bring down size at least a little?

YousefED (commented):
sorry for the delay, we'll get this merged this week!

@areknawo I pushed some minor suggestions from @matthewlipski, and only have 1 question re. the correct bundle to import (web or default)

packages/core/src/blocks/CodeBlockContent/defaultSupportedLanguages.ts
areknawo: @YousefED Not sure if it'd work. `bundledLanguagesInfo` is used to verify if specific language ID can be loaded. If it was limited to languages from the web bundle, languages outside of it wouldn't work.

Apart from the check, it might also affect bundling. Importing the full package, all languages are included in the bundle and you can customize which one you choose to support (and externalize others from the bundle). If it was limited to the web bundle, I'm not sure if other language would work at all.

YousefED: If we're fine with only supporting languages from the web bundle we can limit to `shiki/bundle/web` right? I think that's ok for now until we get requests for more esoteric languages.

Also, I'm thinking we might want to shift to forcing users to pass in languages explicitly; while that would not disable syntax highlighting by default, I think it's also important to have a cleaner packaging / bundling experience

areknawo: In that case - yes.

As for the latter - I agree that switching to explicit imports would make sense. The only drawback would be more difficult customization and a question of balance between built-in defaults and languages you have to import (in that case, maybe it makes sense to remove the defaults completely?)

YousefED: decided on call: let's keep as is (full bundle), because otherwise we'll miss several languages. In a next iteration, we'll separate syntax highlighting to a separate package + make it configurable for consumers

YousefED (approved):

PR #1223, opened 2024-11-07
fix: File resize UX

matthewlipski:
This PR fixes 2 bugs regarding file resize UX:

1. File blocks with captions have incorrect width when caption is longer than the editor is wide (see Default Schema Showcase demo with a narrow browser window).
2. Resizing file blocks with captions only resizes the caption once the mouse button is released, yet the file preview is resized while moving the mouse.
3. Blocks which have an initial width wider than the editor don't get resized on the first attempt.

TODO:
Currently unit tests fail as something is going wrong when calling `updateBlock` on the initial render. `getBlockFromPos` is throwing an error as `getPos()` returns 0.

An alternative approach I got working was to check whether the default `previewWidth` is higher than the editor width when a file is initially added. However, that approach doesn't fix the case where the editor initial contents have a file that is too wide. So I think checking the width on update is the only solution that covers all cases.

### DIFF
--- a/packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
+++ b/packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
@@ -57,6 +57,17 @@ export const imageRender = (
   block: BlockFromConfig<typeof imageBlockConfig, any, any>,
   editor: BlockNoteEditor<any, any, any>
 ) => {
+  // Immediately updates & re-renders the block if it's wider than the editor.
+  if (
+    block.props.previewWidth > editor.domElement.firstElementChild!.clientWidth
+  ) {
+    editor.updateBlock(block, {
+      props: {
+        previewWidth: editor.domElement.firstElementChild!.clientWidth,
+      },
+    });
+  }
+
   const icon = document.createElement("div");
   icon.innerHTML = FILE_IMAGE_ICON_SVG;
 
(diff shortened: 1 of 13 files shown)


packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
```diff
@@ -57,6 +57,17 @@ export const imageRender = (
   block: BlockFromConfig<typeof imageBlockConfig, any, any>,
   editor: BlockNoteEditor<any, any, any>
 ) => {
+  // Immediately updates & re-renders the block if it's wider than the editor.
+  if (
```
YousefED, line 61: My first feeling is I don't really like this approach:
- what if a user opens the document on a small screen, and later on a large screen?
- what if the user is in read-only mode?

packages/core/src/editor/Block.css
```diff
@@ -372,20 +373,19 @@ NESTED BLOCKS
   height: 24px;
 }
 
-[data-file-block] .bn-visual-media-wrapper {
+[data-file-block] .bn-resize-handles-wrapper {
```
YousefED, line 376: make sure DO etc isn't using the old classnames. I suppose you didn't like the earlier name / just a cleanup right?

YousefED (changes requested):
I'm a bit concerned about this approach, see comment!

matthewlipski:
Closing in favor of #1305

PR #1225, opened 2024-11-07
fix: Typing after link doesn't revert to regular text

### LINKED ISSUE #1199: Pressing space when creating a link should convert the content type back to regular text
**Describe the bug**
Pressing space when creating a link should convert the content type back to regular text. Currently the only way to "exit" the link formatting is to press right arrow instead.

**To Reproduce**
* Go to https://www.blocknotejs.org
* Type text, select it and turn it into a link using the link menu
* Type one or more spaces
* Note that the spaces inserted after the link are still part of the link

**Misc**
- Node version: 23.1.0
- Package manager: pnpm
- Browser: Chromium: 129.0.6668.100 

matthewlipski:
Closes #1199


YousefED (commented):
interesting, let's review / debug together

PR #1228, opened 2024-11-08
fix: ShadCN: inherit <a> styling for inline content

serpent213:
By default, links are not styled when using ShadCN (and therefore hard to recognise). This patch makes them inherit the browser default, matching the Mantine behaviour.

(Not sure this is the preferred way to fix it...)


YousefED (commented):
@matthewlipski should this be added in this file or separately? If this file is meant as "Do not modify, it's copied directly from xxx", maybe mark it as such and add the change somewhere else?

matthewlipski:
> @matthewlipski should this be added in this file or separately? If this file is meant as "Do not modify, it's copied directly from xxx", maybe mark it as such and add the change somewhere else?

Yeah you're right, I changed it so the link styling gets reverted in `style.css` instead of removed in `tailwindStyles.css` and added a comment.

matthewlipski (approved):

PR #1235, opened 2024-11-11
fix: export updateBlockCommand, to allow extending heading block

m-risto:
This pull request adds an export statement to `index.ts` to include the `updateBlockCommand` function from `./api/blockManipulation/commands/updateBlock/updateBlock.js`. This change is necessary to facilitate the extension of the heading block, which requires access to the updateBlockCommand function.


YousefED (approved):
thanks. At some point it would be nice to export internal methods under a separate namespace

YousefED:
We'll release this tomorrow or wednesday!

PR #1137, opened 2024-10-10
Refactor: clean blockcontainer

YousefED:
This PR refactors the internal BlockNote Prosemirror / Tiptap commands:
- all functions extracted to separate files
- testcases added
- cleaned the "blockinfo" concept to be more readable and typesafe

removed: insertBlocks `nested` option. Use `updateBlock` instead.

PR #1236, opened 2024-11-11
fix: List item shortcuts throwing errors

matthewlipski:
There was a regression in #1137 which broke keyboard shortcuts for bullet and check list items, which this PR addresses


YousefED:
Nice. would it be easy to add tests for this?

matthewlipski:
> Nice. would it be easy to add tests for this?

Would have to be e2e tests I think, as this bug is not from incorrect logic, but passing in the wrong editor due to lack of type safety.

YousefED (approved):

PR #1240, opened 2024-11-12
fix: Various multi-column bugs

matthewlipski:
This PR fixes 4 bugs to do with multi-column blocks:

1. Dragging a block from one column to another in the same column list now works properly. Previously, it would copy the dragged block instead of moving it.
2. Fixed bug where creating a new column list above an existing one would merge them. This was due to incorrect positions being used in `removeBlocksWithCallback`.
3. Fixed bug where dropping a block inside a new column in a column list would sometimes throw an error, as `nodeDOM` would return `null` due to the cursor position being outdated. I've managed to reproduce this consistently using the document setup below, and dragging the `"bbbbb"` block between the `"c"` and `"d"` blocks.
4. Fixed error being thrown after live reload due to event listeners not getting cleaned up in the column resize extension.

Document to reproduce issue 3:
```json
[
  {
    type: "columnList",
    children: [
      {
        type: "column",
        children: [
          {
            type: "paragraph",
            content: "a",
          },
        ],
      },
      {
        type: "column",
        children: [
          {
            type: "paragraph",
            content: "bbbbb",
          },
        ],
      },
    ],
  },
  {
    type: "columnList",
    children: [
      {
        type: "column",
        children: [
          {
            type: "paragraph",
            content: "c",
          },
        ],
      },
      {
        type: "column",
        children: [
          {
            type: "paragraph",
            content: "d",
          },
        ],
      },
    ],
  },
  {
    type: "paragraph",
  },
],
```

### DIFF
--- a/packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
@@ -87,6 +87,7 @@ export function removeAndInsertBlocks<
     // `blockContainer`.
     const $pos = tr.doc.resolve(pos - removedSize);
     if (
+      $pos.node().type.name === "blockGroup" &&
       $pos.node($pos.depth - 1).type.name !== "doc" &&
       $pos.node().childCount === 1
     ) {
--- a/packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
@@ -1,5 +1,3 @@
-import { Node } from "prosemirror-model";
-
 import { Block, PartialBlock } from "../../../../blocks/defaultBlocks.js";
 import type { BlockNoteEditor } from "../../../../editor/BlockNoteEditor";
 import {
@@ -8,11 +6,11 @@ import {
   InlineContentSchema,
   StyleSchema,
 } from "../../../../schema/index.js";
+import { Node } from "prosemirror-model";
 import { blockToNode } from "../../../nodeConversions/blockToNode.js";
 import { nodeToBlock } from "../../../nodeConversions/nodeToBlock.js";
-import { removeBlocksWithCallback } from "../removeBlocks/removeBlocks.js";
 
-export function replaceBlocks<
+export function removeAndInsertBlocks<
   BSchema extends BlockSchema,
   I extends InlineContentSchema,
   S extends StyleSchema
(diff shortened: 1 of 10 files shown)


packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -39,12 +39,12 @@ export function replaceBlocks<
     editor,
     blocksToRemove,
     (node, pos, tr, removedSize) => {
-      if (node.attrs.id === idOfFirstBlock) {
+      if (blocksToInsert.length > 0 && node.attrs.id === idOfFirstBlock) {
```
YousefED, line 42: is it easy to add a unittest for this one? Also, do you know / understand why we need the + 1?

packages/xl-multi-column/src/extensions/DropCursor/MultiColumnDropCursorPlugin.ts
```diff
@@ -130,17 +130,24 @@ export function multiColumnDropCursor(
             (b) => b.id === blockInfo.bnBlock.node.attrs.id
           );
 
-          const newChildren = columnList.children.toSpliced(
-            position === "left" ? index : index + 1,
-            0,
-            {
+          const newChildren = columnList.children
```
YousefED, line 133: clear!

packages/xl-multi-column/src/extensions/DropCursor/MultiColumnDropCursorPlugin.ts
```diff
@@ -267,20 +274,22 @@ class DropCursorView {
...
-              : blockRect.right;
-          rect = {
-            left: left - halfWidth,
-            right: left + halfWidth,
-            top: blockRect.top,
-            bottom: blockRect.bottom,
-            // left: blockRect.left,
-            // right: blockRect.right,
-          };
+          if (block !== null) {
```
YousefED, line 277: does this actually fix the issue or only prevents the error? Maybe discuss this together?

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
matthewlipski: So I dug deeper into this and turns out that coincidentally, `removeBlocks` and `replaceBlocks` were both working despite using positions that were too wide by 1 in `tr.delete`.

packages/xl-multi-column/src/extensions/DropCursor/MultiColumnDropCursorPlugin.ts
YousefED: fixed

tommerty:
Can I ask that if the resizing/dragging of columns be disabled if editable is false? 🙏🏻 https://github.com/TypeCellOS/BlockNote/issues/1244

matthewlipski:
> Can I ask that if the resizing/dragging of columns be disabled if editable is false? 🙏🏻 #1244

Will fix shortly! But in a separate PR

PR #1241, opened 2024-11-12
fix: `getNearestBlockContainerPos` unclear

### LINKED ISSUE #1237: double clicking block resizer crashes

matthewlipski:
Closes #1237 

Currently, `getNearestBlockContainerPos` gets the nearest block, which may be a `columnList`, `column`, or `blockContainer`. I think the functionality is fine, but this PR renames it to `getNearestBlockPos` for better clarity and removes the error in `getBlockInfoFromSelection` for when something other than a `blockContainer` is returned. This fixes a bug that caused said error to be thrown by selecting a `column` node using triple click (I think) or Cmd+click.

### DIFF
--- a/packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
+++ b/packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
@@ -2,7 +2,7 @@ import { InputRule } from "@tiptap/core";
 import { updateBlockCommand } from "../../../api/blockManipulation/commands/updateBlock/updateBlock.js";
 import {
   getBlockInfoFromSelection,
-  getNearestBlockContainerPos,
+  getNearestBlockPos,
 } from "../../../api/getBlockInfoFromPos.js";
 import {
   PropSchema,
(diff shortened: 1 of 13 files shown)


packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -232,7 +241,7 @@ const checkListItemBlockContent = createStronglyTypedTiptapNode({
 
         // TODO: test
         if (typeof getPos !== "boolean") {
-          const beforeBlockContainerPos = getNearestBlockContainerPos(
+          const beforeBlockContainerPos = getNearestBlockPos(
```
YousefED, line 244: should we rename the variable and / or add a check?

YousefED (commented):
nice fix! What's the part that actually fixes the clicking?

matthewlipski:
> nice fix! What's the part that actually fixes the clicking?

Removing the error throw from `getBlockInfoFromSelection` - it was getting triggered in double/triple clicking the resize bar as that would select the column, and then the formatting toolbar buttons would try to render. They pretty much all use `getBlockInfoFromSelection` somewhere, and so the error would be thrown. Basically all the changes are to accommodate for the fact that we no longer throw this error.

YousefED (approved):

PR #1247, opened 2024-11-14
fix: Made column/table resize bars get disabled when editor is non-editable

### LINKED ISSUE #1244: Editable controls the ability to resize columns and tables
When editable is false, it would be great to have an option/prop to decide should blocks that can be resized like tables and columns the ability to do so. I've not found a meaningful way to control this, and as someone who's using this as a "blog writer" the ability to modify the width is a little strange for end users.

When editable is false, users shouldn't be able to resize.

I've made attempts with some CSS that does hide the resize on columns, but if the user happens to click and drag in the right place they can still grab it.

Example CSS:
```css
.bn-container[data-theming-readonly="true"] [data-node-type="column"] {
    box-shadow: none !important; /* Remove resize indicator */
    cursor: default !important; /* Remove resize cursor */
}
```

**Additional context**
There is likely a way to do this, but it's not outlined in documentation, hasn't been shared in the GitHub/Discord, or I'm just terrible at searching. I'm still learning the innards of Blocknote and a solution is potentially obvious to most..

**Bonus**

matthewlipski:


packages/core/src/blocks/TableBlockContent/TableBlockContent.ts
```diff
@@ -42,6 +45,25 @@ export const TableBlockContent = createStronglyTypedTiptapNode({
     );
   },
 
+  // Since we're using ProseMirror's default table column resizing plugin, we
```
YousefED, line 48: I think this is a bit hacky, as it won't hide the resize options for non-editable users. I'd rather ship this PR without this fix (i.e.: only fix multi-column) than with a hacky fix.

For a proper solution, I'd investigate whether we can disable the relevant prosemirror plugins (i.e.: columnResizing) when the editor is not editable. I'm not sure it's trivial to do this though when switching from editable to non-editable and back, as I'm not sure we can easily disable / enable plugins at that point

YousefED (changes requested):

YousefED (approved):

PR #1253, opened 2024-11-17
feature: expose floatingOptions on react elements

YousefED:
This will allow consumers to pass options to floating-ui.

For example, fixes https://github.com/TypeCellOS/BlockNote/issues/1255 so that `strategy: "fixed"` can be selected


matthewlipski (approved):

Ambercssa:
image

If it is introduced using iframe, there will still be this problem

PR #1257, opened 2024-11-18
fix: Remove redundant zindex values

YousefED:
I think we went a little overboard on zIndex values.

I think this makes more sense, wdyt? can you check if the changes are "safe"?


matthewlipski (commented):
Hmm while we're on the subject of z-indices maybe we should also take a look at the UI element controllers? Those also have the z-index set but it's between 1000-10000. I think it would make sense if we had the same scale for both those and the examples.

matthewlipski:
Well turns out the z-indices were just redundant, so I ended up removing them

PR #1276, opened 2024-11-25
feat: Better selections and moving multiple blocks

### LINKED ISSUE #1239: Can't use ctrl + ⬆️ / ⬇️ if I select multiple blocks
**Describe the bug**
If I select multiple blocks and try to move them using the keyboard short cut ctrl + arrow keys it doesn't work.

**To Reproduce**
https://github.com/user-attachments/assets/4a0771dc-b790-4fc5-b1e4-ea0f9376b911

**Misc**
- Browser: Firefox

matthewlipski:
💖 This feature is sponsored by [DINUM](https://www.numerique.gouv.fr/dinum/) 🇫🇷 and [ZenDiS](https://zendis.de/) 🇩🇪

## `moveBlocks`

`moveBlockUp` and `moveBlockDown` have been renamed to `moveBlocksUp` and `moveBlocksDown` as they now work when multiple blocks are selected. As well as this, they now work with columns and the behaviour follows Notion aside from a few cases. Namely, `column` and `columnList` blocks in the selection are replaced with just the blocks inside them when moving up/down.

ezgif-5-0c6b9e1637
## `getSelection`

This PR changes the logic for retrieving `getSelection().blocks`.

Generally, we now return the blocks spanned by the selection at the depth of the shallowest block spanned, aka the shared depth. However, when the block in which the selection starts is at a higher depth than the shared depth, we omit the first block at the shared depth. Instead, we include descendant blocks of the first block at the shared depth, starting at the start block at its original depth, adding the siblings after it, and reducing the depth until we reach the shared depth. This sounds a bit confusing, but basically it's to mimic Notion's behaviour (which you can see when moving multiple selected blocks up/down using Cmd+Shift+ArrowUp/ArrowDown).

Previously, `getSelection().blocks` would return all blocks spanned by the selection regardless of nesting level.

The code should also be much more performant as it no longer calls `doc.descendants`.

## `setSelection`

A `setSelection` method has also been added to the editor. This takes 2 block identifiers for the start and end blocks of the selection and makes the selection span them, as well as any blocks between. The blocks passed must have content of some kind (either `"inline"` or `"table"`), as the selection created is always a `TextSelection` in ProseMirror terms. The block identifiers must also point to 2 different blocks.

## Other changes

`getPrevBlock`, `getNextBlock` and `getParentBlock` methods have been added to the editor. These are pretty self-explanatory and share a lot of logic with the existing `getBlock`. All of these also now work for getting `column` & `columnList` blocks, whereas `getBlock` could previously only find regular blocks.

Fixed an error sometimes being thrown by the table handles when removing blocks. This was specifically happening with `moveBlocksUp` and `moveBlocksDown`.

Closes #1239


packages/core/src/api/blockManipulation/commands/getBlock/getBlock.ts
```diff
@@ -0,0 +1,127 @@
+import { Node } from "prosemirror-model";
```
YousefED, line 1: should this be in the `commands` folder?

packages/core/src/api/blockManipulation/commands/getBlock/getBlock.ts
```diff
@@ -0,0 +1,127 @@
...
+import { getNodeById } from "../../../nodeUtil.js";
+
+// Get a block from a block identifier, or return undefined if the block cannot
+// be found in the editor. The ProseMirror node that gets converted into a block
+// is the one that has the same ID as the block identifier. However, if the
+// `getDifferentNodeForConversion` function is provided, the node returned from
+// it will be converted instead, e.g. a sibling or parent. If
+// `getDifferentNodeForConversion` returns undefined, this function will also
+// return undefined.
+function getBlockFromNode<
```
YousefED, line 21: I think it's cleaner to just remove this method and inline the code in `getPrevBlock` / `getParentBlock`, what do you think?

My thoughts are that the function does not really do a lot, and the other methods (`getNodeById` and `nodeToBlock` are pretty clear). While `getDifferentNodeForConversion` makes the overall code a bit shorter, using a callback to me adds unnecessary complexity that's not worth it

packages/core/src/api/blockManipulation/commands/getBlock/getBlock.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    return nodeToBlock(
+      nodeToConvert,
+      editor.schema.blockSchema,
+      editor.schema.inlineContentSchema,
+      editor.schema.styleSchema,
+      editor.blockCache
+    );
+  } catch (e) {
```
YousefED, line 54: why is there a try / catch? Seems suspicious, can we remove it?

packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
```diff
@@ -0,0 +1,309 @@
...
+      type: "text" as const,
+      anchorBlockId: anchorBlockPosInfo.node.attrs.id,
+      headBlockId: headBlockPosInfo.node.attrs.id,
+      anchorOffset: selection.anchor - anchorBlockPosInfo.posBeforeNode,
+      headOffset: selection.head - headBlockPosInfo.posBeforeNode,
+    };
+  }
+}
+
+// `getBlockSelectionData` and `updateBlockSelectionFromData` are used to save
```
YousefED, line 70: let's use jsdoc for longer method comments?

Maybe we can use https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/convert-to-jsdoc-comments.md to enforce this as a rule, and to convert the codebase? (can also be done separately)

packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
```diff
@@ -0,0 +1,309 @@
...
+  } else if (prevBlock.children.length > 0) {
+    referenceBlock = prevBlock.children[prevBlock.children.length - 1];
+    placement = "after";
+  } else {
+    referenceBlock = prevBlock;
+    placement = "before";
+  }
+
+  if (!referenceBlock || !placement) {
+    return undefined;
```
YousefED, line 199: can you add a comment when this can happen?

packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
```diff
@@ -0,0 +1,309 @@
...
+
+  editor.removeBlocks(blocks);
+  editor.insertBlocks(flattenColumns(blocks), referenceBlock, placement);
+
+  updateBlockSelectionFromData(editor, selectionData);
+}
+
+// Checks if a block is in a valid place after being moved. This check is
+// primitive at the moment and only returns false if the block's parent is a
+// `columnList` block. This is because when moving blocks, columns are flattened
```
YousefED, line 158: > This is because when moving blocks, columns are flattened
// into regular blocks and therefore can only be nested inside `column` or
// other regular blocks

I don't think this is very related to flattening. Imo the main reason (a block can't be a direct child of a columnlist) is enough

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -0,0 +1,212 @@
...
+  // Minimum depth at which the blocks share a common ancestor.
+  const sharedDepth = $startBlockBeforePos.sharedDepth($endBlockBeforePos.pos);
+  const startIndex = $startBlockBeforePos.index(sharedDepth);
+  const endIndex = $endBlockBeforePos.index(sharedDepth);
+
+  // In most cases, we want to return the blocks spanned by the selection at the
+  // shared depth. However, when the block in which the selection starts is at a
+  // higher depth than the shared depth, we omit the first block at the shared
+  // depth. Instead, we include nested blocks of the first block at the shared
+  // depth, from the start block onwards and up to its depth. This sounds a bit
```
YousefED, line 73: Let's change this to add an (ascii art) example? (and then you can remove the reference to moveblocks / notion)

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -0,0 +1,212 @@
...
+    }
+  } else {
+    // Adds the first block spanned by the selection at the shared depth.
+    blocks.push(indexToBlock(startIndex, sharedDepth));
+  }
+
+  // Adds all blocks spanned by the selection at the shared depth, excluding
+  // the first.
+  for (let i = startIndex + 1; i <= endIndex; i++) {
+    blocks.push(indexToBlock(i, sharedDepth));
```
YousefED, line 113: do we need to go via `indexToBlock` or is there just an easeier way to get the node with `$pos.node(...)` and convert that?

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -468,38 +468,55 @@ export class TableHandlesView<
...
+      if (this.state.colIndex >= colCount) {
+        this.state.colIndex = colCount - 1;
       }
+    }
 
-      const row = tableBody.children[this.state.rowIndex];
-      const cell = row.children[this.state.colIndex];
-      this.state.referencePosCell = cell.getBoundingClientRect();
+    // Update bounding boxes.
+    const tableBody = this.tableElement!.querySelector("tbody");
```
YousefED, line 507: when can this happen? (table body doesn't exist)

YousefED (commented):
Such a great improvement both in UX and affected code!! Really happy with the parts we decided to improve. Definitely a bit more work than expected, but worth it for sure.

Couple of things:

- [X]  Should we close toolbars when moving blocks?
  - Notion doesn't do this so probably not necessary
- [X]  re. `the selection created is always a TextSelection in ProseMirror terms` -> I guess this is intentional "for now", but let's add a comment that this should be improved? Or in general add a comment (or create an issue) with your thoughts of what a better selection API would look like?
- [X]  lets add a gif (moving through multi column / etc) it looks pretty cool! And "sponsored by dinum / zendis" (both PR and when doing release, check file exporter PR for exact text)

Breaking changes:
- Are there any breaking changes in this PR for consumers? If so:
  - release should be major
  - make clear what changed (in PR and release notes)
  - see if any docs or examples need textual updating

packages/core/src/api/blockManipulation/commands/getBlock/getBlock.ts
matthewlipski: `getNodeById` throws an error if a node with the given ID cannot be found, but for `getBlocks`, we just want to return undefined if the block can't be found. Maybe we want to also return undefined instead of throwing an error for `getNodeById`? Not sure

packages/core/src/api/blockManipulation/selections/selection.ts
matthewlipski: I'm pretty sure we need `indexToBlock` here because `$pos.node(...)` only lets you get ancestor nodes at the given position whereas here, we need to find siblings of the ancestor at `sharedDepth`.

I guess we could do something where we get the start position of the next block by adding the node size of the previous block or something like that, but I think the index based approach is cleaner.

matthewlipski:
> Such a great improvement both in UX and affected code!! Really happy with the parts we decided to improve. Definitely a bit more work than expected, but worth it for sure.
> 
> Couple of things:
> 
> * [x]   Should we close toolbars when moving blocks?
>   
>   * Notion doesn't do this so probably not necessary
> * [x]   re. `the selection created is always a TextSelection in ProseMirror terms` -> I guess this is intentional "for now", but let's add a comment that this should be improved? Or in general add a comment (or create an issue) with your thoughts of what a better selection API would look like?
> * [x]   lets add a gif (moving through multi column / etc) it looks pretty cool! And "sponsored by dinum / zendis" (both PR and when doing release, check file exporter PR for exact text)
> 
> Breaking changes:
> 
> * Are there any breaking changes in this PR for consumers? If so:
>   
>   * release should be major
>   * make clear what changed (in PR and release notes)
>   * see if any docs or examples need textual updating

Re docs - we previously didn't add them for `moveBlockUp`/`moveBlockDown`, should we add them now? As for selections, `getSelection` doesn't need updating but maybe we want to add docs for `setSelection` (not sure as it's only kind of "good enough" atm).

YousefED (approved):

PR #1284, opened 2024-11-28
feat: Mobile-optimized formatting toolbar

areknawo:
This PR optimizes the formatting toolbar for mobile, limiting its width and enabling position shifting and overflow when necessary.

Additionally, it includes an `ExperimentalMobileFormattingToolbarController` for future reference and possible development. This controller aims to use the [Visual Viewport API](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API) to position the toolbar right above the virtual keyboard on mobile devices. It's currently marked **experimental** due to the flickering issue with positioning (caused by delays of the Visual Viewport API)


packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
```diff
@@ -0,0 +1,85 @@
+import { BlockSchema, InlineContentSchema, StyleSchema } from "@blocknote/core";
+import { UseFloatingOptions } from "@floating-ui/react";
+import { FC, CSSProperties, useMemo, useRef, useState, useEffect } from "react";
+import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
+import { useUIPluginState } from "../../hooks/useUIPluginState.js";
+import { FormattingToolbar } from "./FormattingToolbar.js";
+import { FormattingToolbarProps } from "./FormattingToolbarProps.js";
+
+export const ExperimentalMobileFormattingToolbarController = (props: {
```
YousefED, line 9: can you document this code with comments?

YousefED (changes requested):
Nice! Turned out more simple than expected and great job researching everything re. visualviewport. No questions about the code atm.

- Can you add an "example" (/examples) that uses the experimental controller?
- Create a PR description

After that, please assign to @matthewlipski so he can do final review / merge

examples/03-ui-components/14-experimental-mobile-formatting-toolbar/README.md
```diff
@@ -0,0 +1,10 @@
+# Experimental Mobile Formatting Toolbar
+
+This example shows how to use the experimental mobile formatting toolbar, which uses [Visual Viewport API](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API) to position the toolbar right above the virtual keyboard on mobile devices.
+
+Controller is currently marked **experimental** due to the flickering issue with positioning (caused by delays of the Visual Viewport API)
```
matthewlipski, line 5: When is the flickering visible? I don't think I saw it when playing around in the preview

matthewlipski (commented):
The formatting toolbar fixes for mobile look good!

The `ExperimentalMobileFormattingToolbarController` looks like a good first step, however I noticed a few UX issues on mobile that are pretty significant. Not sure if those something that needs to be cleared before merging since it is an experimental feature @YousefED.

- The toolbar opens after highlighting some text, but then stays open when the selection is collapsed. I think the behaviour should be the same as in the regular formatting toolbar positioner.
- Tapping an element in the formatting toolbar looks like it causes the editor to lose focus, as the keyboard also disappears.
- The dropdown menus in the formatting toolbar aren't visible when open.

YousefED:
Good UX feedback @matthewlipski. The experiment is mainly about the positioning. If that, and the code looks good, I'm ok with merging it for now. We can decide to iterate on it later. Another (valid) option is to separate it into a draft PR

matthewlipski (approved):

examples/03-ui-components/14-experimental-mobile-formatting-toolbar/README.md
YousefED: Did you test on mobile? When you scroll / down a large document the updating of the position is a bit delayed

YousefED (approved):

PR #1285, opened 2024-11-28
fix: Tab handling (indentation and toolbar focus)

### LINKED ISSUE #1142: Cannot indent a selection of blocks or a block that starts with a link
**Describe the bug**
When I select few bullet points for example and hit tab I cannot indent them I have to go one by one.
Also I can't indent a bullet point with a link.

**To Reproduce**
[Capture vidéo du 11-10-2024 10:33:32.webm](https://github.com/user-attachments/assets/a6734bf7-d424-41f4-b0db-5cbcd62a9cd2)

matthewlipski:
This PR addresses the UX issue of using the Tab key to both indent blocks and to navigate UI elements. There are 2 things implemented here:

First, an editor option (`tabKeyboardNavigation`) has been added to disable Tab navigation for the formatting and link toolbars, as well as the file panel. Currently, these elements prevent indenting a selection of blocks while they're open and first require closing using Escape. Setting `tabKeyboardNavigation` to `false` will make Tab and Shift+Tab always indent/unindent blocks, regardless if those elements are shown.

Second, the formatting and link toolbars now stay closed after indenting/unindenting, instead of re-opening each time. The file panel doesn't need changes in this regard.

Closes #1142

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -219,6 +219,8 @@ export type BlockNoteEditorOptions<
   setIdAttribute?: boolean;
 
   dropCursor?: (opts: any) => Plugin;
+
+  tabKeyboardNavigation?: boolean;
 };
 
 const blockNoteTipTapOptions = {
@@ -395,6 +397,7 @@ export class BlockNoteEditor<
       tableHandles: checkDefaultBlockTypeInSchema("table", this),
       dropCursor: this.options.dropCursor ?? dropCursor,
       placeholders: newOptions.placeholders,
+      tabKeyboardNavigation: newOptions.tabKeyboardNavigation,
     });
 
     // add extensions from _tiptapOptions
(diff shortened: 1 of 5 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -219,6 +219,8 @@ export type BlockNoteEditorOptions<
   setIdAttribute?: boolean;
 
   dropCursor?: (opts: any) => Plugin;
+
+  tabKeyboardNavigation?: boolean;
```
YousefED, line 223: What do you think of this?

```typescript
/***
Select desired behavior when pressing Tab (or Shift-Tab). Specifically, what should happen when a user has selected multiple blocks, the Formatting Toolbar is open, and presses `Tab`.
- `"prefer-navigate-ui"`: when any toolbars are open, change focus to the toolbar. The user needs to first press `Escape` to close the toolbar, and can then indent multiple blocks. Better for keyboard accessibility.
- `"prefer-indent"`: regardless of whether toolbars are open, indent the selection of blocks. In this case, it's not possible to access the toolbar with the keyboard.

@default "prefer-navigate-ui" because it's
*/
tabBehavior: "prefer-navigate-ui" | "prefer-indent" = "prefer-navigate-ui";
```

YousefED (changes requested):
Great! one comment

YousefED (approved):

PR #1300, opened 2024-12-03
fix: Placeholder error handling for inserting CSS rules

### LINKED ISSUE #1286: Firefox 115, the toolbar doesn't display
**Describe the bug**
On Firefox 115, the toolbar doesn't display at text selection nor when typing the / command

**To Reproduce**
Select a text on Firefox 115 select a text see no BlockNote toolbar.
Capture d’écran du 2024-11-28 18-49-37
Capture d’écran du 2024-11-28 18-55-31

**Misc**
- Node version:
- Package manager:
- Browser: Firefox 115

matthewlipski:
The placeholder extension inserts CSS rules which use pseudo-selectors, specifically `:has`, `:only-child`, and `:before`. These are not supported in older browsers, for example the `:has` selector only works in versions of Firefox after 120.

This PR mitigates the issue by wrapping insertion of the CSS rules in a try/catch block, so that if the CSS selectors aren't supported, the placeholders won't be visible but the error thrown won't break other things in the editor.

Closes #1286

### DIFF
--- a/packages/core/src/extensions/Placeholder/PlaceholderPlugin.ts
+++ b/packages/core/src/extensions/Placeholder/PlaceholderPlugin.ts
@@ -48,22 +48,30 @@ export class PlaceholderPlugin {
         for (const [blockType, placeholder] of Object.entries(placeholders)) {
           const mustBeFocused = blockType === "default";
 
-          styleSheet.insertRule(
-            `${getSelector(
-              blockType,
-              mustBeFocused
-            )}{ content: ${JSON.stringify(placeholder)}; }`
-          );
-
-          // For some reason, the placeholders which show when the block is focused
-          // take priority over ones which show depending on block type, so we need
-          // to make sure the block specific ones are also used when the block is
-          // focused.
-          if (!mustBeFocused) {
+          try {
             styleSheet.insertRule(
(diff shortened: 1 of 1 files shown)


YousefED (commented):
Looks good. The editor functions fine I suppose, except that placeholders won't be shown?

matthewlipski:
> Looks good. The editor functions fine I suppose, except that placeholders won't be shown?

Yep that's right

Ambercssa:
const getBaseSelector = (additionalSelectors = "") => {
          if (supportsHasPseudo()) {
            return `.bn-block-content${additionalSelectors} .bn-inline-content:has(> .ProseMirror-trailingBreak:only-child):before`;
          } else {
            const blocks = document.querySelectorAll(`.bn-block-content${additionalSelectors} .bn-inline-content > .ProseMirror-trailingBreak`)
            if (blocks && blocks.length === 1) {
              return `.bn-block-content${additionalSelectors} .bn-inline-content:before`;
            } else {
              return null
            }
          }
        };
        const supportsHasPseudo = (): boolean => {
          try {
            document.querySelector(":has(*)");
            return true;
          } catch {
            return false;
          }
        };
        
     I tried to modify it, and it seems to work too

Ambercssa:
This method will not prevent placeholders from displaying

Issue #522, opened 2024-01-17
bug with adding a link


olha-s:
If add a link without https://, then the current domain will be added to the link. 
For example, if you add a link to www.blocknotejs.org/, then the current domain will appear at the beginning, and it will look like this https://www.blocknotejs.org/www.blocknotejs.org/. 
If you add from https://, that is, https://www.google.com/, then everything works correctly.

The bug can be reproduced here https://www.blocknotejs.org/

I assume that there should be a check for added link and if the link does not contain https://, then only https:// should be added to the link

nickGaooooooooooo:
I believe this is a serious issue. When using the BlockNote editor in this project, the editor typically handles a document, and users usually don't write "https://" while editing the document. For example, they would write "[www.google.com.hk/](http://www.google.com.hk/)" instead of "https://www.google.com.hk/". However, the current BlockNote only directs to the full link when "https://" is added, which goes against the long-standing habits of many users.

PR #1304, opened 2024-12-04
fix: Link UX

### LINKED ISSUE #552: Cannot update a component (`to2`)
**Describe the bug**
Warning: Cannot update a component (`to2`) while rendering a different component (`Editor`). To locate the bad setState() call inside `Editor`

**To Reproduce**
I don't know the name, but adding `[initialContent]` to `useBlockNote` hook trigger the error
```
const editor: BlockNoteEditor = useBlockNote({
        initialContent:
            initialContent
                ? JSON.parse(initialContent) //as PartialBlock[]
                : undefined,
        ...
    }, [initialContent]);
```

**Misc**
- Node version: v20.10.0
- Package manager: pnpm
- Browser: Firefox

### LINKED ISSUE #1141: UX feedback with links
**Describe the bug**
Here's a feedback from one of our user.
When I create a link, I'd like to be able to exit the link by typing tab or space, like on Google docs. 
At present, I can only “exit” the link by clicking on an arrow.
Also when I hit ctrl + K the modal to input my link doesn't display so I don't know how to use the shortcut.

**To Reproduce**
https://github.com/user-attachments/assets/a0822d1a-cad1-4dd4-a8e2-a9fdef818978

matthewlipski:
This PR addresses 2 UX issues with links:

1. The keyboard shortcut (`Mod+K`) toggles the link mark but doesn't open the link editing menu, so no `href` is set. This has been changed so the shortcut now just opens the link creation popover in the formatting toolbar. I think the only real flaw of this approach is that the formatting toolbar needs to be open for the shortcut to work, but otherwise imo the UX is good. This issue could also be fixed but I think it's a pretty unlikely scenario and fixing it would add fairly significant complexity.
2. Setting a link without a protocol prefix would make the href get treated as a relative URL. This lead to cases where setting a link like `www.google.com` would redirect users to `<protocol>://<current-domain>/www.google.com` instead of `https://www.google.com`. This has been changed so that link URLs without a protocol are prefixed with `http://` when the user creates/updates a link.

Closes #1141
Closes #552


Blapi:
@matthewlipski this is amazing ! That would also close https://github.com/TypeCellOS/BlockNote/issues/522 :)

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -159,13 +165,45 @@ const getTipTapExtensions = <
...
+              href: "",
+            }),
+            0,
+          ];
+        }
+
+        // Modified section
+        // If the href does not start with a protocol, prefix with default one.
+        let prefix = this.options.defaultProtocol + "://";
+        for (const protocol of this.options.protocols) {
```
YousefED, line 192: lets use `https` as defaultProtocol (I think valid choice in 2024). We can do this where we add the extension

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -159,13 +165,45 @@ const getTipTapExtensions = <
...
+              ...HTMLAttributes,
+              href: "",
+            }),
+            0,
+          ];
+        }
+
+        // Modified section
+        // If the href does not start with a protocol, prefix with default one.
+        let prefix = this.options.defaultProtocol + "://";
```
YousefED, line 191: I think this makes sense, but only fixes it at the prosemirror level "when rendering". This means in the blocknote schema, it will still show the URL without href.

I think we should also fix the input validation, so that we add the "https://" prefix there, if there was no scheme entered

YousefED (changes requested):

packages/core/src/editor/BlockNoteExtensions.ts
matthewlipski: R.e. the first part I think that's what we want. I originally implemented it the other way, so `https://` was prepended to the URL when it's set, and saving that to the URL prop. But I slightly preferred the UX with this approach, as if you add a link to e.g. `www.google.com` and then edit it, you'll see the same link you added instead of  `https://www.google.com`. Imo this is better for users since the overall link is shorter and more legible + more intuitive for the user if BlockNote doesn't modify the links that users set.

YousefED: Hmm, the downside I see is that when the consumer then processes blocknote json (for whatever reason), they need to apply the same logic to the links.

Notion (and I think google docs as well) automatically add the https to the ux. While yes, you then edit something differently then you inserted, it makes it very explicit what the actual link is

YousefED (approved):

PR #1305, opened 2024-12-05
fix: File HTML export and resizing UX

### LINKED ISSUE #1036: Image dissappearing (width=0) from image block if BlockNote is not visible (not in viewport)
**Describe the bug**
<what's going wrong!?>
I have embedded BlockNotes in the cards on the big whiteboard. Sometimes cards can be outside the viewport. 
While refreshing images are not showing up. Some investigations showed it's because all images which are outside the viewport have width=0. 

If go to docs: @blocknote/reacr/src/blocks/ImageBlockContent.tsx
```
const [width, setWidth] = useState<number>(
    Math.min(
      props.block.props.previewWidth!,
      props.editor.domElement.firstElementChild!.clientWidth
    )
  );
```
So if we check props.editor.domElement.firstElementChild - it will be div with className = "bn-block-group" and while refreshing the whole and card is outside viewport - this element will be returning clientWidth = 0. If card (embedded blocknote) is visible the clientWidth will be > 0.
If check size of BlockNote component it's fine - width > 0. 

It was possible to fix by manual change size of the image block to set attribute width of image. 
p.s. please don't be mad on my code - i hope it's ok
```
editor.document.forEach((block, _) => {
      if (block.type === "image" && block.props.showPreview == true) {
        const outerBlocks = editor?.domElement?.getElementsByClassName("bn-block-outer")
        if (outerBlocks != null) {
            const innerBlock = Array.from(outerBlocks).find((item) => item.getAttribute("data-id") === block.id);
            if (innerBlock != null) {
                const imageBlocks = innerBlock.getElementsByClassName("bn-visual-media");
                Array.from(imageBlocks).forEach((imageBlock) => {
                    if (imageBlock.getAttribute("width") === "0") {
                        imageBlock.setAttribute("width", block.props.previewWidth.toString());
                    }
                });                            
            }
        }
      }
  })
```

Please let me know if there are any ways to avoid manual change parameters in DOM - maybe override ImageBlockContent or pass some other parameter to blocknote to remove clientWidth checking. 
Or maybe other ways to fix it. 
Thanks

here is some example how it works: 

https://github.com/user-attachments/assets/6b52aed9-eca3-4d6d-9afd-8ec9547716fb

**To Reproduce**

We are using TLDraw with cards with BlockNote component inside card. 

Tried to make 2 blocknotes with one outside the screen, but seems working ok in this case. 

**Misc**
- Node version: node: '22.5.1',
- Package manager: npm: '10.7.0',
- Browser: Chrome Version 128.0.6613.85 (Official Build) (64-bit)

### LINKED ISSUE #1049: Errors when using blocksToFullHTML with Image block
**Describe the bug**
When using `editor.blocksToFullHTML` you get errors when your blocks include an Image block. 
The error is different depending on if you are using client side `useCreateBlockNote` or serverside `ServerBlockNoteEditor.create()`

Given the following editor blocks
```js
[
  {
    id: "e6f5b974-86e8-4d80-ba65-2ccd2e620478",
    type: "image",
    props: {
      backgroundColor: "default",
      textAlignment: "left",
      name: "Realfoto-Logo-Blue-Main-Header.png",
      url: "https://realfoto.ca/app/uploads/2023/08/Realfoto-Logo-Blue-Main-Header.png",
      caption: "",
      showPreview: true,
      previewWidth: 512,
    },
    children: [],
  },
]
```

On the server side attempting to use `blocksToFullHTML` if the blocks contains an Image block causes the server to crash with the following error.
```
TypeError: Cannot read properties of undefined (reading 'view')
    at get domElement [as domElement] (/project/workspace/node_modules/@blocknote/core/dist/blocknote.umd.cjs:7:59473)
    at Object.Et [as render] (/project/workspace/node_modules/@blocknote/core/dist/blocknote.umd.cjs:5:22136)
    at toInternalHTML (/project/workspace/node_modules/@blocknote/core/dist/blocknote.umd.cjs:5:10378)
    at Xe (/project/workspace/node_modules/@blocknote/core/dist/blocknote.umd.cjs:5:3036)
    at i.serializeNodeInner (/project/workspace/node_modules/@blocknote/core/dist/blocknote.umd.cjs:5:5776)
    at /project/workspace/node_modules/prosemirror-model/dist/index.cjs:2707:32
    at Fragment.forEach (/project/workspace/node_modules/prosemirror-model/dist/index.cjs:229:9)
    at DOMSerializer.serializeFragment (/project/workspace/node_modules/prosemirror-model/dist/index.cjs:2682:16)
    at Xe (/project/workspace/node_modules/@blocknote/core/dist/blocknote.umd.cjs:5:3379)
    at i.serializeNodeInner (/project/workspace/node_modules/@blocknote/core/dist/blocknote.umd.cjs:5:5776)
```

On the client side the html rendered is 

```html
<div class="bn-block-group" data-node-type="blockGroup">
  <div
    class="bn-block-outer"
    data-node-type="blockOuter"
    data-id="1bdf0f1e-eea1-4bde-a151-c178ac583288"
  >
    <div
      class="bn-block"
      data-node-type="blockContainer"
      data-id="1bdf0f1e-eea1-4bde-a151-c178ac583288"
    >
      <div
        class="bn-block-content"
        data-content-type="image"
        data-name="Realfoto-Logo-Blue-Main-Header.png"
        data-url="https://realfoto.ca/app/uploads/2023/08/Realfoto-Logo-Blue-Main-Header.png"
        data-file-block=""
      >
        <div class="bn-file-block-content-wrapper">
          <div class="bn-file-and-caption-wrapper">
            <div class="bn-visual-media-wrapper">
              <img
                class="bn-visual-media"
                alt="Realfoto-Logo-Blue-Main-Header.png"
                contenteditable="false"
                draggable="false"
                width="512"
              />
            </div>
            <p class="bn-file-caption"></p>
          </div>
        </div>
      </div>
    </div>
  </div>
</div>
```
This doesn't appropriately apply the src to the `img` element and so you end up with a broken image.

**To Reproduce**
A client side example can be found [here](https://stackblitz.com/edit/github-bkvhze?file=App.tsx).
A server side example can be found [here](https://stackblitz.com/edit/node-skkkum?file=index.js).

**Misc**
- Node version: 18 & 20
- Package manager: yarn and npm
- Browser: Chrome

### LINKED ISSUE #1309: The video file has not been processed by the `editor.resolveFileUrl` method
In this line of code there is no conversion like images and audio via the `editor.resolveFileUrl` method

https://github.com/TypeCellOS/BlockNote/blob/71792c2fbdc99785cd8ce5b540d1a468e6c66131/packages/core/src/blocks/VideoBlockContent/VideoBlockContent.ts#L66

Refer to the render code of the image

```js
  editor.resolveFileUrl(block.props.url).then((downloadUrl) => {
    image.src = downloadUrl;
  });
```

matthewlipski:
This PR fixes a few issues:

### Problem

The `src` attribute for file blocks was not added when exporting to external HTML, since the `src` would only be set after the download URL was resolved.

### Solution

The `src` attribute is now set to `block.props.url` if `editor.resolveFileUrl` isn't defined, which will at least fix the issue for blocks which are not uploaded via the editor. Since `editor.resolveFileUrl` is `async`, we don't have a way of getting the correct URL when converting the block, which is done synchronously.

### Problem

Exporting file blocks using the server editor would throw an error. This was because of a typing issue, as even though `_tiptapEditor` is not defined for the `ServerBlockNoteEditor`, you can still use all of its functions as if it was. The error was being thrown as the image block `render` function was trying to access `_tiptapEditor.view.dom`.

### Solution

Image/video block max widths are now constrained by CSS instead of the editor `DOMRect`, meaning they no longer need to access `_tiptapEditor`.

Also, the typing has been adjusted such that at least `_tiptapEditor.view` can be undefined, as this can even happen for the regular client editor. This is a small improvement, but we should look into a more robust typing fix in the future.

### Problem

The width of image/video blocks is constrained to the width of the editor. This was done by checking if the width of an image/video block was wider than the editor `DOMRect` when rendering or resizing, then updating it if necessary. These updates were causing a number of issues.

First, it was not great code-wise as clamping the width of image/video blocks could be accomplished using CSS instead.

Second, upon loading, images outside the viewport horizontally would become invisible, as the editor `DOMRect` width would be 0. So any images/videos would also also have their widths set to 0 to not be wider than the editor.

Finally, image/video caption widths would only update once the user stopped resizing the image/video itself, leading to bad UX.

### Solution

As mentioned in the previous issue, constraining the image/video block widths is now done with CSS. The `previewWidth` of file blocks is no longer limited to the width of the editor. Instead, the blocks' widths are now capped using a CSS `max-width` rule. Since `previewWidth` no longer necessarily represents the visual width of a file block, the block's `referenceRect` is used to get the actual width when resizing.

This also fixes the second issue as we're no longer checking any `DOMRects` on initial render - we just set the width to `previewWidth` and CSS figures out if it needs to be clamped.

Finally, the `FileAndCaptionWrapper` & `ResizeHandlesWrapper` elements/components have been removed & restructured into `FileBlockWrapper` & `ResizableFileBlockWrapper`. `DefaultFilePreview` has also been renamed to `FileNameWithIcon` for better clarity. The changes made both fix the caption resizing issue, and streamline the rendered output by getting rid of a few wrapper divs.

### TODO

- [x] Update existing test snapshots
- [x] Add/update tests for `ServerBlockNoteEditor` exporting file blocks to HTML

Closes #1036 
Closes #1049 
Closes #1309


packages/react/src/blocks/AudioBlockContent/AudioBlockContent.tsx
```diff
@@ -21,18 +22,20 @@ export const AudioPreview = (
...
-
   return (
-    <audio
-      className={"bn-audio"}
-      src={resolved.downloadUrl}
-      controls={true}
-      contentEditable={false}
-      draggable={false}
-    />
+    <FileAndCaptionWrapper {...props}>
```
YousefED, line 26: were these missing before? (explain on call)

packages/react/src/blocks/VideoBlockContent/VideoBlockContent.tsx
```diff
@@ -20,29 +22,29 @@ export const VideoPreview = (
...
-    <ResizeHandlesWrapper {...props} width={width} setWidth={setWidth}>
-      <video
-        className={"bn-visual-media"}
-        src={resolved.downloadUrl}
-        controls={true}
-        contentEditable={false}
-        draggable={false}
-        width={width}
-      />
+    <ResizeHandlesWrapper {...props}>
```
YousefED, line 30: I'm curious about this new structure and why we need the context

packages/core/src/blocks/FileBlockContent/fileBlockHelpers.ts
```diff
@@ -358,21 +353,21 @@ export const createResizeHandlesWrapper = (
 
     editor.updateBlock(block, {
       props: {
-        previewWidth: getWidth(),
+        previewWidth: parseInt(wrapper.style.width.replace("px", "")),
```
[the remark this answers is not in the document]
matthewlipski: Not at the moment, when the new width is calculated during resizing it goes straight into the style.  I did this intentionally because I felt it was redundant to also save to a variable, since it's the same information in 2 different places. But I can change this if you feel the parsing is too hacky.

Also in case you're wondering, `block.props.previewWidth` only gets updated when the mouse button is released, not while the user is dragging a resize handle, so it's not useful here.

packages/react/src/blocks/FileBlockContent/fileBlockHelpers.tsx
```diff
@@ -118,36 +59,83 @@ export const AddFileButton = (
...
+        // Show loader while a file is being uploaded.
+        <div className={"bn-file-loading-preview"}>Loading...</div>
+      ) : props.block.props.url === "" ? (
+        // Show the add file button if the file has not been uploaded yet.
+        <AddFileButton {...props} />
+      ) : (
+        // Show the file preview and caption if the file has been uploaded.
+        <>
+          {props.block.props.showPreview === false || !props.children ? (
+            // Use default preview.
```
YousefED, line 112: change comment (it's not a preview)

packages/react/src/blocks/ImageBlockContent/ImageBlockContent.tsx
```diff
@@ -20,30 +19,23 @@ export const ImagePreview = (
     "contentRef"
   >
 ) => {
-  const [width, setWidth] = useState<number>(
-    Math.min(
-      props.block.props.previewWidth!,
-      props.editor.domElement.firstElementChild!.clientWidth
-    )
-  );
+  const imgRef = useRef<HTMLImageElement>(null);
```
YousefED, line 22: is this used?

YousefED (commented):
awesome progress!

YousefED (commented):
Looks good if there are no major changes in the moved files (can't really easily review that as the diff doesn't show)

I liked the convenience of the default `resolveFileUrl`, but don't think it's worth implementing a different option (e.g.: `hasCustomResolveFileUrl: boolean`) as that doesn't seem great either (unless you have a better idea :) )

Issue #1299, opened 2024-12-03
list keyboard shortcut other than "1"


YousefED:
When you add an indented table in an ordered list, the list won’t continue ordering.

See video:

https://github.com/user-attachments/assets/5571a44e-d266-444f-8e1c-1bba03ec9f20

The same issue can be recreated without a table:
- create a list item
- hit enter to create a second list item, remove it (backspace)
- now, a new list item can only be created by typing “1. “, but the user probable expects it to be created by typing “2. ”

I think there are 3 ways to fix it.

First of all, we need to be able to create a list item when any number is entered, not just `1. `. This should already fix the issue above. But then, we have the question what should happen when you enter `2. ` when you're not in a list (for example, you type `2.` in an empty document). Couple of options:

a) trigger regular “list-creation” always on typing “<number>. “, regardless of cursor position. This will have a bit weird behaviour because when you’d start a new list with “2. “, it will change it into “1. ” (as it creates a new list that starts as 1)
b) option (a), but the new list would start at “2.” this would require a list-start attribute of some sorts to be added. I think this is the approach notion has.
c) Only make typing “2. ” work when we’re typing it at the correct position (i.e. after a single list item). This is more inline with MS Word

Imo, let's go for option (b) as it's most inline with Notion

Issue #1298, opened 2024-12-03
copy paste tables to excel / google sheets


YousefED:
A common workflow involves copying data between tables and Excel.

**Expectation**: When users copy a (part of a) table to Excel or google sheets, they should see it displayed exactly as it appeared in the notebook.
**Actual**: When users copy a  table to Excel, it pastes as a single column with irregular spacing.

https://github.com/user-attachments/assets/5a8a5f2e-a215-4fff-8b42-14734794a0d4

matthewlipski:
Closed in #1319

PR #1311, opened 2024-12-09
fix: Custom start index for ordered lists (#1299), improved support for copy/pasting tables into speadsheets (#1298)

areknawo:
This PR:
- adds support for ordered lists starting at any index (>=1), insertable with `\d+\.` shortcut; (#1299)
- adds support for exporting `OL` `start` attribute in HTML & MD output;
- fixes copy/pasting of tables into spreadsheet applications (like Excel/Google Sheets); (#1298)


packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -172,6 +172,14 @@ function serializeBlock<
   if (listType) {
     if (fragment.lastChild?.nodeName !== listType) {
       const list = doc.createElement(listType);
+
+      if (listType === "OL") {
+        const index = block.props?.index;
+
+        if (index !== undefined) {
```
YousefED, line 179: let's only add when it's not 1?

YousefED (commented):
seems like a good start, I see one major bug when creating a list

1. first item
2. second item

and then indenting the second item, the index shows as `1`.
---
I also don't think we always want to export the list index to BlockNote JSON format. When I look at `basic/block-objects` example, now the index is always included. Imo we should only include it if there's an explicit (different from default) "start".

Not sure what the best approach for these two issues is. Perhaps we should keep `index` as-is, and include a `start` attribute only when needed?

YousefED (commented):
Couple of improvement points:

_Discussed on slack:_
- add tests
- make sure `start` doesn't show up for non-explicitly set list items

_other:_
- when starting a new list with 1., we shouldn't set start=1, as it's the default behavior
- found one bug with inconsistent data / UX, see recording:


https://github.com/user-attachments/assets/d79a0aa6-ef53-4653-88f2-8b6884f5b6ad

Issue #1321, opened 2024-12-13
Copy pasting an Emoji insert a page break


virgile-dev:
**Describe the bug**
When I copy paste an emoji from Emojipedia, for [ex](https://emojipedia.org/sparkles)

**To Reproduce**
![paste](https://github.com/user-attachments/assets/b6b9758b-9420-4a8e-8abb-0dcab3a7006a)

**Misc**
- Node version:
- Package manager:
- Browser: Firefox 133 
- [x] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

YousefED:
@virgile-dev how do you copy the emoji? When I hit the "copy" button (e.g. [here](https://emojipedia.org/dizzy)) it works fine when pasting in blocknote

virgile-dev:
Yep same as you.  Doesn't seem to be firefox related, just tested on chromium. The line break appears to be created when I copy paste the emoji at the end of a line not when I copy paste at the beginning of a line.
[Capture vidéo du 16-12-2024 11:10:17.webm](https://github.com/user-attachments/assets/ba9438b2-71bd-4a2d-8613-ee24080b16d0)

Issue #1198, opened 2024-10-30
Pasting text pastes it inside a new block


danielrearden:
**Describe the bug**
Pasting text pastes it on a new block instead of inside the current block. If I am typing inside a paragraph block or another block that includes `content`, I would expect pasted text to be added to the block at the cursor position. Instead, it pastes as a new block below the current one. Note: copying text from the editor itself works as expected -- this seems to only happen when copying from somewhere else.

**To Reproduce**
* Go to https://www.blocknotejs.org
* Copy text from somewhere else like a text file
* Paste the text into the editor using Ctrl/⌘-V
* Observe that text is pasted on the next block instead of inside the current block

**Misc**
- Node version: 23.1.0
- Package manager: pnpm
- Browser: Chromium: 129.0.6668.100 
- [x] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

PR #1324, opened 2024-12-14
fix: better paste handling

YousefED:
- Fixes pasting from an external spreadsheet into an existing table
- closes https://github.com/TypeCellOS/BlockNote/issues/1198 (and closes https://github.com/TypeCellOS/BlockNote/pull/1321, probably same)
- Fixes pasting text in Custom Blocks. For example, pasting text in an empty (no-text) alert block would previously replace the entire block (fixed by setting the custom block as `isolating`). Described in this issue: https://github.com/TypeCellOS/BlockNote/issues/771 (but doesn't fully close it as the issue is also about keyboard shortcuts)

TODO:
- fix build
- add unit tests for (a) pasting tables (b) 1198 (c) pasting inside a custom block


matthewlipski (approved):

packages/core/src/editor/transformPasted.ts
```diff
@@ -49,31 +51,45 @@ export function wrapTableRows(f: Fragment, schema: Schema) {
...
  */
 export function transformPasted(slice: Slice, view: EditorView) {
   let f = Fragment.from(slice.content);
   f = wrapTableRows(f, view.state.schema);
 
+  const nodeHasSingleChild = f.childCount === 1;
+  const nodeHasInlineContent = f.firstChild?.type.spec.content === "inline*";
+  const nodeHasTableContent = f.firstChild?.type.spec.content === "tableRow+";
+
+  const blockInfo = getBlockInfoFromSelection(view.state);
```
YousefED, line 77: I'm a bit concerned about getting the selection here. I don't think `transformPasted` is only called when pasting something at the current selection. We might need to use the `clipboardParser` instead that gets a `$context` parameter, but I'm not sure whether that's the best option (we should investigate the PM code + docs to make sure).

I suppose you're positive that we need this?

packages/core/src/editor/transformPasted.ts
```diff
@@ -49,31 +51,45 @@ export function wrapTableRows(f: Fragment, schema: Schema) {
...
+ * an image or two consecutive paragraphs, where PM tries to nest the pasted
+ * block(s) when it shouldn't. However, this fix is not applied in a few cases.
+ *
+ * The first case is when we paste a single node with inline content, e.g. a
+ * paragraph or heading. This is because we want to insert the content in-line
+ * for better UX instead of a separate block below.
  *
- * The exception is when we encounter blockGroups with listitems, because those actually should be nested
+ * The second case is when we paste a single node with table content, i.e. a
+ * table, inside an existing table. This is because the fix would tell instead
```
YousefED, line 66: typo? `This is because the fix would tell instead want to add its content to the existing table, again for better UX.`

PR #1331, opened 2024-12-18
feature: odt exporter

YousefED:
This PR adds an ODT exporter, with support for a customizable header & footer.


examples/05-interoperability/07-converting-blocks-to-odt/README.md
```diff
@@ -0,0 +1 @@
+# Exporting documents to .odt (Open Document Text)
```
YousefED, line 1: todo, add readme (follow similar pattern as readmes)

package.json
```diff
@@ -34,5 +34,6 @@
     "prepublishOnly": "npm run build && cp README.md packages/core/README.md && cp README.md packages/react/README.md",
     "postpublish": "rm -rf packages/core/README.md && rm -rf packages/react/README.md",
     "clean": "lerna run --stream clean"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
```
YousefED, line 38: todo, remove

packages/xl-odt-exporter/src/odt/odtExporter.tsx
```diff
@@ -0,0 +1,235 @@
...
+      }
+    }
+
+    return ret;
+  }
+
+  public async toODTDocument(
+    blocks: Block<B, I, S>[],
+    options?: {
+      header?: string;
```
YousefED, line 146: I think we want to be able to use more than a string as header / footer. Can you compare with the other exporters?

YousefED (commented):
I've left some early comments. Besides that, I found the following with the example:

- [x] example exported document doesn't open in Word for Mac
- [x] on libreoffice, "Numbered List Item Nested funky center" doesn't have the blue font color as in the example
- [x] Tables always seem to be full width. Also (probably related), column width is not taken into account (see "wide cell" in testDocument.ts)
- [x] The images have an extra border around them, can we get rid of this?
- [x] the first two colored bg paragraphs have a white space between them. On docx / pdf this does not happen, and also not with the red numbered lists. Not sure if possible to fix, but would be nice
- [x] Could you also fix the CI build?

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
```diff
@@ -212,18 +224,60 @@ export const odtBlockMappingForDefaultSchema: BlockMapping<
...
+    const { Buffer } = await import("buffer");
+    const blob = await exporter.resolveFile(block.props.url);
+    const arrayBuffer = await blob.arrayBuffer();
+    const base64 = Buffer.from(arrayBuffer).toString("base64");
+    //const dimensions = await getImageDimensions(blob);
+    const styleName = createParagraphStyle(
+      exporter as ODTExporter<any, any, any>,
+      block.props
     );
+    const editorWidth = 623;
```
YousefED, line 246: what kind of unit is this / where does this number come from?

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
```diff
@@ -212,18 +224,60 @@ export const odtBlockMappingForDefaultSchema: BlockMapping<
...
+          style:rel-height="scale"
+          style:rel-width={block.props.caption ? "100%" : width}
+          {...(!block.props.caption && {
+            "text:anchor-type": "as-char",
+          })}>
+          <DrawImage
+            xlink:type="simple"
+            xlink:show="embed"
+            xlink:actuate="onLoad">
+            <OfficeBinaryData>{base64}</OfficeBinaryData>
```
YousefED, line 260: does word / openoffice also use this base64 embedding when you create images? Or do they create seperate files (resources) in the zip file. If the latter, that would be preferred (stay as close as possible to what other authoring software does)

packages/xl-odt-exporter/src/odt/imageUtil.ts
```diff
@@ -0,0 +1,21 @@
+import type Sharp from "sharp";
+
+export async function getImageDimensions(blob: Blob) {
```
YousefED, line 3: this does not seem to be used, right?

packages/xl-odt-exporter/src/odt/odtExporter.test.ts
```diff
@@ -0,0 +1,75 @@
+import { BlockNoteSchema, defaultBlockSpecs, PageBreak } from "@blocknote/core";
+import { testDocument } from "@shared/testDocument.js";
+import { BlobReader, TextWriter, ZipReader } from "@zip.js/zip.js";
```
YousefED, line 3: you mentioned admzip didn't work for you. Curious what was the exact case, but if zip.js works better, let's also move the other exporter(s) to zip.js so that the codebase remains consistent?

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
areknawo: Width of the BN editor in `px`.
I needed to somehow get a `%` value from `previewWidth`, which seems to be a `px` value.
Not sure how else to get it, other than by using `editorWidth` - let me know if you have any suggestions.

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
areknawo: I moved to separate files, which seems to have fixed some Word compatibility issues.

packages/xl-odt-exporter/src/odt/odtExporter.test.ts
areknawo: The only other place `adm-zip` is used is in DOCX exporter tests, which run server-side. Client-side, in the browser, `adm-zip` didn't work well. Tried to patch `process` to fix it (https://github.com/cthackers/adm-zip/issues/529) but other Node-dependency-related issues popped up.

areknawo:
> example exported document doesn't open in Word for Mac
The ODT file can now be opened in Word, but there seem to be some compatibility issues.
Not sure if they are expected in Word or caused by the exporter.

packages/xl-odt-exporter/src/odt/odtExporter.tsx
areknawo: Indeed, other exporters offer more customizability by using underlying libraries.
In this case, I wasn't sure about exposing the XML/JSX syntax to the end-user - it's a bit unusual and might be confusing for some.

We can export certain elements as JSX components and allow users to use them for more customization, but it'll likely still be just a subset of what's possible;

packages/xl-odt-exporter/src/odt/odtExporter.test.ts
areknawo: Now using `@zip.js/zip.js` in DOCX exporter tests

areknawo:
> Tables always seem to be full width. Also (probably related), column width is not taken into account (see "wide cell" in testDocument.ts)

Fixed setting column width. When no column width is available, `style:use-optimal-column-width` is used, which results in the default, full-width table (when testing in LibreOffice on macOS, might depend on the software used);

>  The images have an extra border around them, can we get rid of this?

Removed the additional frame being added to images;

>  the first two colored bg paragraphs have a white space between them. On docx / pdf this does not happen, and also not with the red numbered lists. Not sure if possible to fix, but would be nice

It seems to be about line/paragraph spacing - the default value (`>1`), results in this look. We can reduce the spacing, but then styled paragraphs will look different from others.

As for the numbered lists case - it seems paragraphs with same background/styling are "joined" together, filling the gap. When the styles are different, the spacing remains white (tested in LibreOffice on macOS)

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
```diff
@@ -50,51 +48,43 @@ const createParagraphStyle = (
...
 
   if (props.textColor && props.textColor !== "default") {
     const color =
       exporter.options.colors[
         props.textColor as keyof typeof exporter.options.colors
       ].text;
-    styles["fo:color"] = color;
-  }
-
-  if (
```
YousefED: was this removed intentionally? if there are no specific styles, we don't need to register a custom style right?

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
```diff
@@ -234,17 +224,21 @@ export const odtBlockMappingForDefaultSchema: BlockMapping<
   },
 
   image: async (block, exporter) => {
-    const Buffer = (await import("buffer")).Buffer;
+    const odtExporter = exporter as ODTExporter<any, any, any>;
     const blob = await exporter.resolveFile(block.props.url);
-    const arrayBuffer = await blob.arrayBuffer();
-    const base64 = Buffer.from(arrayBuffer).toString("base64");
-    //const dimensions = await getImageDimensions(blob);
+    const fileExtension = block.props.url.split(".").pop();
```
YousefED: what if there's no `.` or sth in the url?

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
YousefED: why do you need the editorWidth?

packages/xl-odt-exporter/src/odt/odtExporter.tsx
areknawo: Maybe we can accept not just strings, but stringified XML or even `XMLDocument`s?
You can check the lastest changes for this concept - let me know if it's worth refining.

I'd lean towards only accepting stringified XML, since it's the most flexible.

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
areknawo: Manage to remove the dependency on `editorWidth`

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
areknawo: `fo:color` was moved to `textStyles`, since it should be used with `style:text-properties`

When no custom properties are applied, we're now returning the `parentStyleName`

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
areknawo: I'll need to detect the file extension somehow, likely with some mimetype - file extension detector. Do we have a limited subset of images to support?

areknawo:
- Improved Word compatibility;
- Applied feedback;

packages/xl-odt-exporter/src/odt/odtExporter.tsx
```diff
@@ -182,27 +197,35 @@ export class ODTExporter<
...
             <style:master-page
               style:name="Standard"
               style:page-layout-name="Mpm1"
               draw:style-name="Mdp1">
-              {options.header && (
+              {header && (
                 <style:header>
-                  <text:p text:style-name="MP1">{options.header}</text:p>
+                  <text:p
+                    text:style-name="MP1"
```
YousefED, line 209: should this be in a `text:p`?

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
YousefED: `const blob = await this.resolveFile(url);`

does `blob.type` have the correct mimetype?

packages/xl-odt-exporter/src/odt/odtExporter.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+      if (typeof xmlDocument === "string") {
+        stringifiedDoc = xmlDocument;
+      } else {
+        const serializer = new XMLSerializer();
+
+        stringifiedDoc = serializer.serializeToString(xmlDocument);
+      }
+
+      return stringifiedDoc.replace(
```
YousefED, line 171: Can you:
- add a comment why this is here / what it does
- include where the regex comes from
- check whether it's necessary for both cases? (i.e.: both when a string vs xmldocument are passed). I don't have the context to judge this

YousefED:
- The Inter font doesn't load on libreoffice
- libreoffice still shows border around caption
- in Word, Links are not styled (blue / underlined)
- in Word, "Open Video" (and file / audio) have the wrong font
- Documentation is missing (take [docx](https://www.blocknotejs.org/docs/editor-api/export-to-docx) as reference)


Table widths:
- the default tables are full width on libreoffice
- Explicit table widths (when you resize one column to be larger) don't work on either libreoffice or word

packages/xl-odt-exporter/src/odt/util/components.ts
```diff
@@ -0,0 +1,149 @@
+// import { XMLElement, createElement } from "./xml.js";
+import { createElement } from "react";
```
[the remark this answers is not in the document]
areknawo: Removed `components.tsx`

areknawo:
- Fixed font loading in LibreOffice;
- Fixed link styling in Word;
- Added documentation;
- Fixed table width rendering;
- Added default column width, instead of auto;
- Removed `components.tsx`, using JSX elements directly;

> libreoffice still shows border around caption
It seems that border/frame is not customizable, even directly from the LibreOffice.
@YousefED

packages/xl-odt-exporter/src/odt/imageUtil.ts
nperez0111, line 1: I'd use `image-meta` which is a much more lightweight lib, and is in plain JS (sharp is natively compiled I think). This would also work in both browsers & node

https://github.com/unjs/image-meta

packages/xl-odt-exporter/src/odt/odtExporter.tsx
```diff
@@ -0,0 +1,353 @@
...
+  private automaticStyles: Map<string, React.ReactNode> = new Map();
+  private pictures: Map<
+    string,
+    {
+      file: Blob;
+      fileName: string;
+      height: number;
+      width: number;
+    }
+  > = new Map();
```
nperez0111, line 40: Not immediately obvious what this is key-ed by.

I'd also have different naming, but just a nit:
```suggestion
  private images = new Map<
    string,
    {
      content: Blob;
      name: string;
      height: number;
      width: number;
    }
  >();
```

nperez0111:
The export looks really good. Good job here @areknawo

packages/xl-odt-exporter/src/odt/imageUtil.ts
YousefED: Thanks. Nice to get rid of sharp. 
- I've extracted the util file to @shared (as it was also used for docx)
- changed sharp to image-meta
- only import image-meta on node / vitest, as browser side it's not necessary (we use `createImageBitmap`)

PR #1341, opened 2024-12-24
fix: Multi-editor drag & drop

### LINKED ISSUE #657: Multiple editor instances cause duplicates on drag&drop
**Describe the bug**
Dragging an item from one editor instance to another does not remove it from the source instance. This leads to duplicates.

**To Reproduce**
https://stackblitz.com/edit/github-cz76gj?file=App.tsx

**Related**
- https://github.com/TypeCellOS/BlockNote/issues/173
- https://github.com/TypeCellOS/BlockNote/pull/174

### LINKED ISSUE #1194: Cannot use two block node editors at the same time
**Describe the bug**
In my react application, I have two completely different components that are both using BlockNote. Both components are visible at the same time under some circumstances. All features but one are working perfectly fine: Dragging & dropping blocks to re-arrange them.

If I drag and drop a block in one editor, this block is duplicated in the other editor!

My wild guess is, that both editors are listening to the same "drag-and-drop" event, and are both reacting to it accordingly.

**To Reproduce**
1. render two BlockNote editors next to each other
2. type some text in both editors (e.g. create 3 normal text blocks in each editor)
3. drag and drop text in one editor, you will notice the dragged text also appears in the other editor

**Misc**
```
    "@blocknote/core": "^0.17.1",
    "@blocknote/mantine": "^0.17.1",
    "@blocknote/react": "^0.17.1",
```

### LINKED ISSUE #1246: TypeError: Cannot read properties of undefined (reading 'nodeType')
**Describe the bug**
I have a few instances of BlockNote on my page. One on initial render, and others that come from a side nav when clicked. When I render an image or custom block in the side nav instance and perform a drag, I get the error a type error.

`TypeError: Cannot read properties of undefined (reading 'nodeType')`

https://github.com/user-attachments/assets/ff4ce173-30e6-4587-905f-fb306cd3da30

**To Reproduce**
1 - Mount an instance of BockNote on a page
2 - Conditionally mount a second instance of BlockNote
3 - Create a custom block
4 - Render custom block in the conditionally instance of BlockNote
5 - Perform drag on custom block

**Misc**
- Node version: v19.0.1
- Package manager: webpack (nextjs 14.1.4)
- Browser: Chrome

matthewlipski:
This PR fixes dragging & dropping blocks between different editors, and the `sideMenuDetection` option has been added. `sideMenuDetection` can be set to `"viewport"` or `"editor"` (`"viewport"` by default), and sets the detection area for showing the side menu.

Setting it to `"viewport"` makes the side menu appear whenever the cursor is beside a block, anywhere on the page. Setting it to `"editor"` makes the side menu only appear when the cursor is over the editor, or slightly to the left to account for the side menu area.

The same goes for dragging and dropping blocks, where `"viewport"` will clamp the mouse x-coordinate to the side of the editor.

Closes #657 
Closes #1194 
Closes #1246


examples/01-basic/04-all-blocks/App.tsx
```diff
@@ -1,207 +1,203 @@
...
-  useCreateBlockNote,
-} from "@blocknote/react";
-import {
-  getMultiColumnSlashMenuItems,
   multiColumnDropCursor,
   locales as multiColumnLocales,
   withMultiColumn,
 } from "@blocknote/xl-multi-column";
-import { useMemo } from "react";
-export default function App() {
```
YousefED, line 22: move to separate example?

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -14,6 +14,9 @@ import { EventEmitter } from "../../util/EventEmitter.js";
 import { initializeESMDependencies } from "../../util/esmDependencies.js";
 import { getDraggableBlockFromElement } from "../getDraggableBlockFromElement.js";
 import { dragStart, unsetDragImage } from "./dragging.js";
+import { DOMParser } from "prosemirror-model";
```
YousefED, line 17: import both from same?

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -249,7 +273,16 @@ export class SideMenuView<
   onDrop = (event: DragEvent) => {
     this.editor._tiptapEditor.commands.blur();
 
+    // ProseMirror doesn't remove the dragged content if it's dropped outside
```
YousefED, line 276: hmm, this would also remove the content when there's a drop in an invalid element right? e.g.: what happens if you drop on the playground sidebar?

packages/core/src/extensions/SideMenu/dragging.ts
```diff
@@ -202,6 +202,5 @@ export function dragStart<
     e.dataTransfer.setData("text/plain", plainText);
     e.dataTransfer.effectAllowed = "move";
     e.dataTransfer.setDragImage(dragImageElement!, 0, 0);
-    view.dragging = { slice: selectedSlice, move: true };
```
YousefED, line 205: not needed anymore because we do this in the other dragstart?

packages/core/src/blocks/HeadingBlockContent/HeadingBlockContent.ts
```diff
@@ -130,7 +130,9 @@ const HeadingBlockContent = createStronglyTypedTiptapNode({
           }
 
           return {
-            level: element.getAttribute("data-level"),
+            level:
```
YousefED, line 133: do we have unit tests for other parse scenarios? if so, add the cases that we missed there?

YousefED (commented):
nice! left some quick comments

examples/01-basic/04-all-blocks/App.tsx
matthewlipski: Ah will actually just revert this one

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
matthewlipski: Nah, it's a `drop` event so it only fires when you attempt to drop on a valid target - this would be an issue if we were using `dragend` instead.

packages/core/src/extensions/SideMenu/dragging.ts
matthewlipski: Exactly

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
```diff
@@ -118,18 +118,7 @@ const checkListItemBlockContent = createStronglyTypedTiptapNode({
   parseHTML() {
     return [
       {
-        tag: "div[data-content-type=" + this.name + "]", // TODO: remove if we can't come up with test case that needs this
-        getAttrs: (element) => {
```
YousefED: why are these parts removed exactly? they're parsed somewhere else?

packages/core/src/api/parsers/html/parseHTML.test.ts
```diff
@@ -59,6 +59,202 @@ async function parseHTMLAndCompareSnapshots(
 }
 
 describe("Parse HTML", () => {
+  it("Parse BlockNote HTML", async () => {
```
YousefED, line 62: Can we use something like `blocksToFullHTML` instead of this hardcoded HTML? It makes it difficult to maintain or understand this testcase. Or is the HTML sth different?

Also, I suppose you're saying these cases weren't handled in other places like copy and paste tests? (I figured we had quite some extensive test cases there)

matthewlipski: True we can, I'll move the cases to copy/paste tests to make things a bit cleaner.

The main issue I was running into was heading levels being parsed as `null` from BlockNote HTML, and this indeed was not covered by existing copy/paste tests. So to prevent issues in the future the new tests cover parsing all basic blocks from BlockNote HTML (except audio, video, and file blocks, since those are pretty similar to images).

packages/core/src/blocks/ListItemBlockContent/CheckListItemBlockContent/CheckListItemBlockContent.ts
matthewlipski: Yeah idk when we added `getAttrs` here but it's not needed and causes bugs - we already parse props in `propsToAttributes`.

PR #1357, opened 2025-01-07
Fix for numbered list item when exporting to PDF, DOCX

mfanselmo:
This PR https://github.com/TypeCellOS/BlockNote/pull/1326 introduced the ability to start numbered lists from indices other than 1. I noticed on the demo to export to pdf/docx this was not taked into account.

I was not able to run playwright locally and update snapshots if that is needed, can you give me guidance here or do it?


YousefED (commented):
Nice, thanks. We'll fix the playwright tests, I think it's unrelated. Could you:

- also add support to the docx exporter?
- add it to the unit tests of the exporters? (add a case to the default test case and update the related snapshots)

mfanselmo:
Hey! I spent quite a bit of time on this but ran into some tricky edge cases I couldn’t get to work perfectly. Let me break them down:

For PDFs, numbering is straightforward since we can manually set the numbering. However, Word has a few quirks:
To start a numbered list at a value other than 1, you have to define a specific configuration for that list type. It’s not as simple as just setting a start value. I handled this by maintaining a set of starting indices and manually referencing them via blocknote-numbered-list-{index}.
Word automatically increments list numbers, so if you need to restart numbering manually, you have to use the instance parameter for adjacent lists.
I came across a weird edge case where adjacent lists, nested under different parent blocks, don’t behave as expected—even though the instance parameter should fix it. For now, my code assigns a unique numeric ID to each list instance, but I believe this might be overkill and could be simplified. I left it in just in case someone wants to dig into it.

Here’s an example of content that doesn’t render correctly in Word. Based on how instance works, I would expect this to work fine:
image


I’ve updated the snapshots for both PDF and DOCX to test these cases. The PDF output works perfectly, but DOCX still has this one issue. @YousefED I will leave this one to this point for now, but if I find time in the future I will give it another go

packages/xl-docx-exporter/src/docx/docxExporter.ts
```diff
@@ -111,8 +111,25 @@ export class DOCXExporter<
     nestingLevel = 0
   ): Promise<Array<Paragraph | Table>> {
     const ret: Array<Paragraph | Table> = [];
+    let currentNumberingStart = 1;
+    let numberingInstanceCount = 0;
+    let isFirstNumberedListItem = false;
 
     for (const b of blocks) {
+      if (b.type === "numberedListItem") {
```
mfanselmo, line 119: This code is a bit messy and could definitely be improved, but its the best I could come up with to ensure a unique instance id is given to each numbered list. In theory this id should guarantee no weird continuations of non adjacent lists happen, but for some reason they still are. I would not merge this as is

matthewlipski:
Hi @mfanselmo are you still planning to work on this? Since you mentioned it's not ready to merge

mfanselmo:
Hey, not planning on working on this for now. Still missing the DOCX fix.

PR #1361, opened 2025-01-08
fix: NextJS 15 `Objects are not valid as a React child`

### LINKED ISSUE #1347: React error when using multi-column blocknote in nextjs 15 / react 19
**Describe the bug**
Clicking on the slash menu will cause the following bug:
```
Error: Objects are not valid as a React child (found: object with keys {$$typeof, type, key, ref, props, _owner, _store}). If you meant to render a collection of children, use an array instead.
```

**To Reproduce**
- Use the example nextjs provided
- Add the multi-column plugin, with the guide provided
- Click on the slash menu
[Stackblitz reproduction](https://stackblitz.com/~/github.com/TheMythologist/nextjs-blocknote-bug)

**Misc**
- Node version: 20
- Package manager: npm (i also got this error when I was using pnpm)
- Browser: any (specifically, I'm using Chromium-based browsers)

matthewlipski:
For some reason, NextJS 15 throws an error when rendering props (that aren't `children`) that are JSX elements, in some cases. In particular, the `icon` prop for suggestion menu items causes this error to be thrown.

The error should be thrown when trying to render an array or object as a JSX element, so I'm not sure why it's also being thrown for JSX elements, which should work just fine.

What's worse is that copying the exact same code and icons to a local file vs importing them from BlockNote also stops the error from being thrown. Unfortunately, I couldn't find any information on why this is happening. Maybe it's worth submitting an issue in NextJS? Note that this doesn't seem to be an issue in base React 19, but specifically NextJS 15.

Either way, this PR changes props which aren't `children` that have type `JSX.Element` to the type `FC` to fix the issue.

Closes #1347

### DIFF
--- a/examples/01-basic/04-all-blocks/App.tsx
+++ b/examples/01-basic/04-all-blocks/App.tsx
@@ -195,9 +195,16 @@ export default function App() {
     );
   }, [editor]);
 
+  console.log(slashMenuItems);
+
   // Renders the editor instance using a React component.
   return (
-    <BlockNoteView editor={editor} slashMenu={false}>
+    <BlockNoteView
+      editor={editor}
+      slashMenu={false}
+      onSelectionChange={() => console.log("change")}
+      onFocus={() => console.log("focus")}
+      onBlur={() => console.log("blur")}>
       <SuggestionMenuController
         triggerCharacter={"/"}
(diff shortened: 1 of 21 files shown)


matthewlipski:
See #1365

Issue #1344, opened 2024-12-26
TypeError: Cannot read properties of undefined (reading 'toLowerCase') in parseHTML of codeblock


Karl-Wu:
**Describe the bug**
Got TypeError: Cannot read properties of undefined (reading 'toLowerCase')  while calling tryParseHTMLToBlocks. 
The HTML is not well-formed which was converted by readability.js. 

**To Reproduce**
1. Open the following link
https://stackblitz.com/edit/github-xpg9ndgd?file=package.json,html-with-codeblock.ts,App.tsx
2. Check the output of dev console

**Misc**
- Node version:
- Package manager:
- Browser: 
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

flycran:
I'm having the same issue

YousefED:
@areknawo could you check this and add a related test?

PR #1362, opened 2025-01-10
fix: Code block language parsing

areknawo:
- Fixes code block language parsing (#1344);
- Adds unit tests for verifying expected code block language parsing behavior;


jameszyao:
we are facing the same problem thank you

matthewlipski (approved):

Issue #1179, opened 2024-10-23
Add a page break block


virgile-dev:
**Is your feature request related to a problem? Please describe.**
As you are working with export users want to decide where the page break happens.

**Describe the solution you'd like**
I'd like to have a type of block that act as a page break. When the document is exported a new page is created right after the page break block.
 It should be visible on the editor so that its position can be changed and be easily deleted.

YousefED:
close by #1372

PR #1372, opened 2025-01-15
feat: Page break

areknawo:
This PR adds a page break block, dedicated for use with exporters to force page breaks in the documents.

- Page break block represented as a dashed line;
- Support for page break in PDF & DOCX exporters;


packages/xl-pdf-exporter/src/pdf/pdfExporter.test.tsx
```diff
@@ -160,8 +160,8 @@ describe("exporter", () => {
     );
 
     const transformed = await exporter.toReactPDFDocument(testDocument);
-    const view = render(transformed);
-    const str = prettyDOM(view.container, undefined, { highlight: false });
+    const str = reactElementToJSXString(transformed);
```
YousefED, line 163: what's the benefit of this library?

YousefED (changes requested):
Nice!

Two things;
- I'm not sure we want it as part of the default schema. It's quite specific and only relevant if you're printing / exporting
- maybe add this CSS so it works for printing as well? https://www.w3schools.com/cssref/pr_print_pageba.php

packages/xl-pdf-exporter/src/pdf/pdfExporter.test.tsx
areknawo: Compared to `render` & `prettyDOM` it doesn't expect DOM/HTML-compatible JSX. Thus, it results in fewer warnings and fixes the `break` edge case I was facing (the `break` prop being removed from the snapshot due to incompatibility with DOM).

areknawo:
- [x] Add CSS styles to support page break in printing;
- [ ] Remove page break from the default schema;

YousefED (commented):
Couple of things:

- Can we align the drag handle better?

- can we make the design a bit nicer? e.g. dotted line that's grey instead of black dashes? Maybe @areknawo can later also add "Page break" in the middle, but that's non blocking
- The "Default schema" showcase actually doesn't show the default schema any more because it also adds multi-column + page break. We should update the example to reflect this
- Maybe add the page break block to the pdf and docx examples? Because now there's no easy way to test this functionality easily

YousefED (approved):

YousefED:
closes #1179

PR #1374, opened 2025-01-15
feat: Improved collaboration cursor UX

### LINKED ISSUE #1163: The nametag over other editors cursor hides the text behind it
## Bug Report

**Problematic behavior**
The name tad showing where the cursor of another editor is hides the text.

**Expected behavior/code**
Give the cursor a color and show the list of editors connected and their associated color somewhere on the page

**Steps to Reproduce**
1. Have to editors login on the same doc
2. Go to the same line 
3. See that the name hides the text

**Environment**
- docs.numerique.gouv.fr

**Additional context/Screenshots**
Capture d’écran du 2024-10-17 17-34-45

**Ideally I'd want to it to look like this**
First letter of the user displays at the beginning of the block and I can see where he is editing the block if it's text. 
Capture d’écran du 2024-10-17 17-49-36

matthewlipski:
This PR improves the collaboration cursor UX so that the name tag no longer annoyingly blocks user text selection. The UX changes are summarized here: https://github.com/TypeCellOS/BlockNote/issues/1163#issuecomment-2504180222.

> User 1 and 2 collaborate in real-time, from user 1's perspective:
> - When user 2 writes, their name is displayed
> - After X seconds of user 2's inactivity, their name disappears
> - When user 1 hovers over user 2's cursor, user 2's name is displayed for X seconds
> - The cursor remains visible at all times. To improve its visibility, we can either add a dot as suggested by Virgile, or increase its thickness.

ezgif-5-bce5e744ef

Closes #1163

### DIFF
--- a/packages/core/src/editor/BlockNoteExtensions.ts
+++ b/packages/core/src/editor/BlockNoteExtensions.ts
@@ -1,4 +1,5 @@
 import { AnyExtension, Extension, extensions } from "@tiptap/core";
+import { Awareness } from "y-protocols/awareness";
 
 import type { BlockNoteEditor, BlockNoteExtension } from "./BlockNoteEditor.js";
 
@@ -251,7 +252,11 @@ const getTipTapExtensions = <
       })
     );
     if (opts.collaboration.provider?.awareness) {
-      const defaultRender = (user: { color: string; name: string }) => {
+      const defaultRender = (user: {
+        clientID: number;
+        color: string;
+        name: string;
+      }) => {
         const cursor = document.createElement("span");
 
         cursor.classList.add("collaboration-cursor__caret");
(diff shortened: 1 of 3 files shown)


packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -265,14 +270,72 @@ const getTipTapExtensions = <
...
+              return;
+            }
+
+            if (hideTimeout) {
+              clearTimeout(hideTimeout);
+            }
+
+            if (
+              JSON.stringify(opts.editor.document) !== JSON.stringify(oldDoc)
+            ) {
```
nperez0111, line 295: This can be expensive with large documents. Maybe a deep equal algo, Or if these are prosemirror nodes their .eq method

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -251,28 +252,112 @@ const getTipTapExtensions = <
...
+
+        cursors.set(clientID, {
+          element: cursorElement,
+          hideTimeout: undefined,
+        });
 
-        label.classList.add("collaboration-cursor__label");
-        label.setAttribute("style", `background-color: ${user.color}`);
-        label.insertBefore(document.createTextNode(user.name), null);
+        cursorElement.addEventListener("mouseenter", () => {
```
YousefED, line 313: is this needed or can we just use css `:hover`?

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -251,28 +252,112 @@ const getTipTapExtensions = <
...
+        const clientID = clientState[0];
+
+        return (
+          cursors.get(clientID) || createCursor(clientID, user.name, user.color)
+        ).element;
       };
       tiptapExtensions.push(
         CollaborationCursor.configure({
-          user: opts.collaboration.user,
+          user: {
```
YousefED, line 357: we can revert this

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -251,28 +252,112 @@ const getTipTapExtensions = <
...
     );
     if (opts.collaboration.provider?.awareness) {
-      const defaultRender = (user: { color: string; name: string }) => {
-        const cursor = document.createElement("span");
+      const cursors = new Map<
+        number,
+        { element: HTMLElement; hideTimeout: NodeJS.Timeout | undefined }
+      >();
+
+      const awareness = opts.collaboration!.provider.awareness as Awareness;
```
YousefED, line 260: this can be a bit cleaner by

doing `const awareness = opts.collaboration.provider?.awareness` before line 254 and then reusing it (nitpicking a bit)

packages/core/src/editor/BlockNoteExtensions.ts
matthewlipski: We ended up showing the labels on selection changes as well as content changes, so removed this altogether

packages/core/src/editor/BlockNoteExtensions.ts
matthewlipski: I think we still want it - GDocs behaviour still has a delay for hiding after moving the mouse off the cursor. We could have a transition delay but it's added complexity in both cases so this is ok imo.

Issue #1200, opened 2024-10-30
Placeholder for the entire content


danielrearden:
**Is your feature request related to a problem? Please describe.**
Blocknote supports placeholders for individual components, but does not have a way to render a separate placeholder for when the content is completely empty. We can override the paragraph placeholder, but this isn't desirable since we only want the placeholder to show when there is no content, not for every empty paragraph.

Example use case, writing a reply to a social post and having the placeholder be something like "Reply to Sarah..."

**Describe the solution you'd like**
A separate placeholder property on the editor component that would be rendered only when the content is empty.

**Describe alternatives you've considered**
We've implemented this on our own, but it requires some hacky workarounds. Something supported out-of-the-box by Blocknote would be much nicer.

**Additional context**

**Bonus**
- [X]  I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

PR #1259, opened 2024-11-19
Feature/liveblocks v2

YousefED:
This PR makes BlockNote compatible with the new, simplified Liveblocks developer experience ❤️

Setting up a Liveblocks enabled editor with collaboration, comments, and mentions now practically becomes a one-liner:

```typescript
export function Editor() {
  const editor = useCreateBlockNoteWithLiveblocks();

  return (
    <BlockNoteView
      editor={editor}>
      <Threads editor={editor} />
    </BlockNoteView>
  );
}
```

## Changes
- Changed the Liveblocks example to use `@liveblocks/react-blocknote` (to be released). See [Liveblocks PR](https://github.com/liveblocks/liveblocks/pull/2079)
- We automatically add an "Add comment" button to the FormattingToolbar (of course, this can be customized)
- We've improved the mounting system on BlockNote. This was necessary to make it compatible with the Liveblocks UI components so that editor.view is immediately available.

## TODO:

- [ ] update [docs](https://www.blocknotejs.org/docs/advanced/real-time-collaboration#liveblocks)
- [x] **Blocking**: the new mounting system requires React MarkViews to fully work. This is WIP and might take a few weeks to land. If we like, we can create a different workaround until that's available
- [x] Move new mounting system to separate PR?
- [x] Translate comments strings to other langs

### UI elements
In contrast to Lexical / Tiptap / Prosemirror, BlockNote is a UI-included library. Currently, we let Liveblocks add UI elements to the editor. This has the disadvantage that the UI elements are in the style of Liveblocks, and have a different look & feel + customization API than default BlockNote UI elements.

We'd like to make UI primitives for collaboration a "first class citizen" of BlockNote that Liveblocks plugs into. If possible, it would be great to collaborate with Liveblocks and re-use parts of their primitives so we can help improve & maintain those together.

### Higher level API

Similarly, BlockNote should provide a higher level API for change detection / notifications that Liveblocks now does at the Prosemirror level to add support for mention-notifications. Liveblocks could then plug in to this higher level API.

PR #1376, opened 2025-01-16
feat: Comments

### LINKED ISSUE #695: Google Doc like comments
Firstly, Kudos to the great work on the project!

I wanted mentions/comments functionality as available on Google Doc editor for my website. Upon some research I found https://tiptap.dev/product/comments and https://liveblocks.io/comments both of which are paid options.
Is Blocknote planning on any similar feature in the near future? if not, could anyone possible give some pointers as to how we can implement a similar functionality as those but with Blocknote and our stack:
- Next.js
- Supabase(including Supabase-realtime if relevant)

### LINKED ISSUE #703: New placeholders option doesn't work with multiple editors
**Describe the bug**
The new `placeholders` option has a bug where when using more than one editor, the placeholder will be overwritten by the last editor instance.

For additional context on my use-case, I have multiple editors to support a different editing experience between "Title" and "Content" sections of a page (similar to notion).

**To Reproduce**
Create two editors. Ex.
```
const titleEditor = useCreateBlockNote({
    placeholders: {
      default: "Untitled",
    },
  });

const contentEditor = useCreateBlockNote();
```
The placeholder for both will be the default placeholder text.

Expectation: The title editor should have `Untitled` for its placeholder

**Misc**
- Blocknote: 0.12.4

### LINKED ISSUE #1109: Is there a comment function available
Select a certain paragraph of text and add comments/notes to it

### LINKED ISSUE #1148: Will you consider the comment function?
image

Select a paragraph and add a comment to it

YousefED:
This PR adds comments to BlockNote!

- We implement several Comment / Thread related components that can be customized similar to other BlockNote components.
- The implementation is completely backend agnostic as the storage interface (ThreadStore) is pluggable. This means you can "bring your own provider" if you like.

[See docs](https://blocknote-website-git-feature-comments-typecell.vercel.app/docs/collaboration/comments)

# ThreadStores

## YjsThreadStore
The YjsThreadStore provides direct Yjs-based storage for comments, storing thread data directly in the Yjs document. This implementation is ideal for simple collaborative setups where all users have write access to the document.
The downside of this is that comments require a more advanced permission model than text documents, and Yjs doesn't provide a straightforward way to secure document updates.

## RESTYjsThreadStore
The RESTYjsThreadStore combines Yjs storage with a REST API backend, providing secure comment management while maintaining real-time collaboration. This implementation is ideal when you have strong authentication requirements, but is a little more work to set up.

In this implementation, data is written to the Yjs document via a REST API which can handle access control. Data is still retrieved from the Yjs document directly (after it's been updated by the REST API), this way all comment information automatically syncs between clients using the existing collaboration provider.

[We created a separate repository with an example server.](https://github.com/TypeCellOS/BlockNote-demo-nextjs-hocuspocus)

## TipTap
We integrate directly with the [TiptapCollabProvider](https://github.com/ueberdosis/hocuspocus/blob/795619935fc510bb844ea6607af40bd199e2d1a0/packages/provider/src/TiptapCollabProvider.ts#L60) with `TiptapThreadStore`.

## LiveBlocks (later)
The UI Components are roughly based on the LiveBlocks open source components. LiveBlocks users will have two different ways to use LiveBlocks:
- Use LiveBlocks UI components: this can be done with https://github.com/TypeCellOS/BlockNote/pull/1259. This would support all liveblocks functionality (read status, attachments, etc), but the style is a bit different from the other BlockNote components.
- Use BlockNote UI components (matching your existing mantine / shadcn / ... version of BlockNote). These are "first-class" supported in BlockNote but miss some features compared to liveblocks (i.e.: read status, attachments, etc). We want to implement a _LiveBlocksThreadStore_ for this, but this work is currently blocked (we can't store BlockNote documents describing comments in LiveBlocks)

## TODO
- [ ] implement StreamThreadView https://github.com/TypeCellOS/BlockNote/pull/1454
- [x] merge https://github.com/TypeCellOS/BlockNote/pull/1453
- [x] Fix design / bugs when example is rendered in docs (emoji)
- [ ] Fix scrolling bug in docs (@nperez0111 )
- [x] show selection when composing a comment
- [x] test all updates with shadcn / ariakit
- [x] code review
- [x] i18n
- [x] docs
- [x] reduce space between avatar and username
- [x] composer is not always visible (see nextjs demo)
- [x] dragging a block removes a comment
- [x] fix sidemenu

TBD:
- [ ] table cell, move block, and API operations
- [ ] accessibility

### Yjs next steps
- [ ] overlapping marks (Yjs)
- [ ] security of rest store
- [ ] security of yjs store

### To fix later

Important:
- [ ] https://github.com/TypeCellOS/BlockNote-demo-nextjs-hocuspocus/issues/1
- [ ] comments on images and other block nodes
- [ ] reposition on window resize
- [x] copy comment when cut+paste (optionally also with copy / paste?)
- [ ] resolving newer comments deletes older already resolved ones at the same place. Probably a larger issue (yjs doesn't deal well with overlapping marks?)
- [ ] show loading / error states in UI for interacting with comments
- [ ] TBD: should the thread be deleted when the main (first) comment is removed?
- [ ] TBD: add server side validation for YjsThreadStore (make sure users can't edit other users comments, etc). To make this work, we'd need to monitor updates in HocusPocus for example

Nice to haves:
- [ ] cmd+enter to save?
- [ ] Implement LiveBlocksThreadStore (and possibly move to separate package)
- [x] hide emojis when editing
- [x] badge ux (hover background / border)
- [ ] close emoji popup (action bar) when adding reaction
- [ ] better mouse over area for showing action toolbar
- [ ] disable children (indentation) for comments content
- [ ] implement anchored threads
- [ ] positions of floating elements after window resize
- [ ] expose an API for accessing comments (currently, comments don't appear in the BlockNote API)
- [ ] when another user adds a reaction, your picker gets closed
- [ ] when closing more actions popover by clicking the trigger button, the action toolbar stays visible when hovering something else

closes #695 
closes #1109 
closes #1148 
closes #703 
closes https://github.com/TypeCellOS/BlockNote/issues/1200


YousefED:
@dmonad I'm working on a POC for comments here. You can see:

- The ThreadStore is pluggable. i.e.: a user could integrate their own API (e.g.: REST-based)
- There is a basic yjs implementation (https://github.com/TypeCellOS/BlockNote/pull/1376/files#diff-a4c2b8d1cdfd357eaad40431946a8b8a593d07571f40ce6e067e27cec48c654e). Of course, this comes with security issues without implementing strong checks on the server.

Two questions about this:
- You mentioned the only way to validate an update would be to apply it to a clone and compare changes (+ see if there's no prelim? updates added)
- IF the user wants to store the Threads Y.Map as part of the regular Y.Doc that contains the entire document (instead of in a separate Y.Doc), is there an easy way to see if an incoming update affects that specific Y.Map? This way we could only run the validation logic (including cloning etc.) when there's been an update to the Threads datastructure.

dmonad:
>     * You mentioned the only way to validate an update would be to apply it to a clone and compare changes (+ see if there's no prelim? updates added)

Even that is edgy… There are several parts of Yjs that could be exploited to sneak-in changes, even if you check `pendingStructs`. It might be possible to add deleted structs that will show up when a user undos changes. I generally advise against validating updates like this. It's a security hazard; unmaintainable unless you employ somebody that understands Yjs internals. I don't want anyone to copy this approach.

>     * IF the user wants to store the Threads Y.Map as part of the regular Y.Doc that contains the entire document (instead of in a separate Y.Doc), is there an easy way to see if an incoming update affects that specific Y.Map? This way we could only run the validation logic (including cloning etc.) when there's been an update to the Threads datastructure.

That is actually a more realistic approach. You could check each struct / delete whether it applies changes to only the children of a certain other type. This approach is also more memory efficient (no need to fork a document). However, you need to understand the Yjs internals to pull this off.

Ideally, comments are stored in a separate Yjs document with different permissions. If that is not possible, I recommend that read-only users are only able to add comments through an HTTP request (no manipulations on the Yjs document).

YousefED:
> I recommend that read-only users are only able to add comments through an HTTP request (no manipulations on the Yjs document).

I agree, read-only users should probably always use an HTTP request. However, for other users (with write access), I think we should also be able to validate changes. For example, a user should only be able to edit comments they created themselves. So I'm interested if / how we can validate this. Can also discuss this over a call, might be easier :)

packages/core/src/extensions/Comments/CommentMark.ts
```diff
@@ -0,0 +1,45 @@
...
+          return {
+            "data-bn-thread-id": (attributes as { threadId: string }).threadId,
+          };
+        },
+        default: "",
+      },
+    };
+  },
+
+  renderHTML({ HTMLAttributes }: { HTMLAttributes: Record<string, any> }) {
```
nperez0111, line 37: Depending on if you want to support copy-paste of comments, maybe parseHTML?

YousefED: Good point!

I think copy paste will need to be tested extensively :) for example, you could paste a comment that's not in the store, etc

packages/mantine/src/toolbar/Toolbar.tsx
```diff
@@ -36,8 +36,7 @@ export const Toolbar = forwardRef<HTMLDivElement, ToolbarProps>(
         // TODO: aria-label
         onMouseEnter={onMouseEnter}
         onMouseLeave={onMouseLeave}
-        justify={variant === "action-toolbar" ? "flex-end" : undefined}
```
YousefED: I think you moved this to CSS. do we have a specific pattern / reason for this?

packages/mantine/src/style.css
```diff
@@ -564,3 +564,79 @@
   display: flex;
   justify-content: space-between;
 }
+
+/* TODO: Clean up */
```
YousefED, line 569: let's do this :)

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -1,20 +1,16 @@
...
 
-import {
-  CommentData,
-  ThreadData,
-  mergeCSSClasses,
-  getDefaultEmojiPickerItems,
-  DefaultGridSuggestionItem,
-} from "@blocknote/core";
+import { CommentData, ThreadData, mergeCSSClasses } from "@blocknote/core";
+import data from "@emoji-mart/data";
```
YousefED, line 4: iirc this is a large package that we might only want to load when necessary (like we do for the regular emoji picker)

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -443,73 +425,86 @@ export const Comment = ({
...
+                          "bn-badge",
+                          "bn-comment-reaction"
+                        )}
+                        text={reaction.usersIds.length.toString()}
+                        icon={reaction.emoji}
+                        isSelected={user && reaction.usersIds.includes(user.id)}
+                        onClick={() => onReactionSelect(reaction.emoji)}
+                        mainTooltip={"Reacted by"}
+                        secondaryTooltip={`${reaction.usersIds.map(
+                          (userId) => userId + "\n"
```
YousefED, line 454: you can use `.join("\n")` here

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -244,6 +251,7 @@ const getTipTapExtensions = <
     ...(opts.trailingBlock === undefined || opts.trailingBlock
       ? [TrailingNode]
       : []),
+    CommentMark,
```
YousefED, line 254: todo, only if comments are enabled?

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -0,0 +1,372 @@
...
+
+  const user = useUser(editor, comment.userId);
+
+  const blockNoteContext = useBlockNoteContext();
+
+  if (!showDeleted && !comment.body) {
+    return null;
+  }
+
+  let actions: ReactNode | undefined = undefined;
```
YousefED, line 182: @matthewlipski it would be better here to extract the emoji picker component, and only call this when the user expands the picker + show a loader in between

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -0,0 +1,263 @@
...
+import { User } from "../../models/User.js";
+import { EventEmitter } from "../../util/EventEmitter.js";
+import { ThreadStore } from "./threadstore/ThreadStore.js";
+import { CommentBody, ThreadData } from "./types.js";
+import { UserStore } from "./userstore/UserStore.js";
+const PLUGIN_KEY = new PluginKey(`blocknote-comments`);
+
+enum CommentsPluginActions {
+  SET_SELECTED_THREAD_ID = "SET_SELECTED_THREAD_ID",
+}
```
nperez0111, line 15: Would recommend against using enums: https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  /**
+   * when a thread is resolved or deleted, we need to update the marks to reflect the new state
+   */
+  private updateMarksFromThreads = (threads: Map<string, ThreadData>) => {
+    const doc = new Y.Doc();
+    const threadMap = doc.getMap("threads");
+    threads.forEach((thread) => {
+      threadMap.set(thread.id, thread);
+    });
```
nperez0111, line 97: Is this doing anything? I might be missing something here

packages/core/src/index.ts
```diff
@@ -68,3 +68,15 @@ export * from "./extensions/UniqueID/UniqueID.js";
 export * from "./api/exporters/markdown/markdownExporter.js";
 export * from "./api/parsers/html/parseHTML.js";
 export * from "./api/parsers/markdown/parseMarkdown.js";
+
+export * from "./extensions/Comments/types.js";
+
+// TODO: how to organize this? The number of exports is growing
+// - we could namespaces things (i.e.: comments.yjs.YjsThreadStore)
+// - split to different package (@blocknote/comments)
+// - sub imports ( @blocknote/core/comments)
```
nperez0111, line 77: Big fan of sub-exports so you can keep bundle size to a minimum. Will need some tooling to split up the bundling

packages/react/src/editor/EditorContent.tsx
```diff
@@ -1,67 +1,66 @@
...
-}) => {
-  return (
-    <>
-      {Object.entries(renderers).map(([key, renderer]) => {
-        return createPortal(renderer.reactElement, renderer.element, key);
-      })}
-    </>
-  );
-};
+export function getInstance() {
```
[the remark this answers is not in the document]
nperez0111: Well, I recognize this code 😂

packages/core/src/extensions/Comments/threadstore/YjsThreadStore.ts
```diff
@@ -0,0 +1,470 @@
...
+      }
+
+      const key = `${this.userId}-${options.emoji}`;
+
+      const reactionsByUser = yComment.get("reactionsByUser");
+
+      reactionsByUser.delete(key);
+    }
+  );
+  // TODO: async / reactive interface?
```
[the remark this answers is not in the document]
YousefED: wdyt of zustand vs valtio? (https://react-community-tools-practices-cheatsheet.netlify.app/state-management/poimandres/)

I personally think the "selector" API of zustand is a bit cumbersome , but curious about your thoughts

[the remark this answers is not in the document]
nperez0111: Yep, valtio is a good option too. I've always had an uneasyness when it comes to Proxy based things since they can feel a bit "magic", but I do appreciate the simpler mental model of mutable updates.

I mostly picked zustand for the easy React integration, and small bundle size. 

https://bundlephobia.com/package/valtio@2.1.3
https://bundlephobia.com/package/zustand@5.0.3

packages/mantine/src/toolbar/ToolbarButton.tsx
```diff
@@ -8,10 +8,10 @@ import {
 
 import { assertEmpty, isSafari } from "@blocknote/core";
 import { ComponentProps } from "@blocknote/react";
-import { forwardRef } from "react";
+import { forwardRef, useMemo } from "react";
 
 export const TooltipContent = (props: {
-  mainTooltip: string;
+  mainTooltip?: string;
```
YousefED, line 14: why optional? it's weird to have a TooltipContent without any tooltip?

packages/mantine/src/toolbar/ToolbarButton.tsx
```diff
@@ -52,17 +52,10 @@ export const ToolbarButton = forwardRef<HTMLButtonElement, ToolbarButtonProps>(
...
-        withinPortal={false}
-        label={
-          <TooltipContent
-            mainTooltip={mainTooltip}
-            secondaryTooltip={secondaryTooltip}
-          />
-        }>
-        {/*Creates an ActionIcon instead of a Button if only an icon is provided as content.*/}
-        {children ? (
+    const button = useMemo(
```
YousefED, line 55: do we need the useMemo here?

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -235,10 +258,14 @@ export const Comment = ({
             </Components.Generic.Toolbar.Button>
           ))}
         {(canDeleteComment || canEditComment) && (
-          <Components.Generic.Menu.Root>
+          <Components.Generic.Menu.Root
+            position={"bottom-start"}
+            onOpenChange={(open) =>
```
YousefED, line 263: we now have this logic in two places (to make sure tooltip hides when clicked). Is this the cleanest way or can we find a way to make this the responsibility of one of the components themselves? or extract it in some other way that's reusable?

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -201,18 +209,33 @@ export const Comment = ({
...
-                mainTooltip="Add reaction"
-                variant="compact">
-                <RiEmotionFill size={16} />
+                mainTooltip={showReactionsPicker ? undefined : "Add reaction"}
+                variant="compact"
+                onClick={(e) => {
+                  setShowReactionsPicker(
+                    showReactionsPicker === "toolbar" ? undefined : "toolbar"
+                  );
+                  // Needed as the Picker component's onClickOutside handler
```
YousefED, line 222: interesting, why wasn't this needed before?

packages/react/src/components/Comments/Comment.tsx
matthewlipski: Turns out the `Picker` component fetches the emojis itself if a `data` prop isn't passed, which works out nicely. The response with the emojis is <80kBs, so imo it's better not to show a loader here as in most cases it'll flash for a sec before showing the picker, which is kinda jarring.

packages/mantine/src/badge/Badge.tsx
```diff
@@ -28,6 +30,8 @@ export const Badge = forwardRef<
   // assertEmpty in this case is only used at typescript level, not runtime level
   assertEmpty(rest, false);
 
+  const { isOpened } = useContext(PopoverContext);
```
YousefED, line 33: @matthewlipski  we now introduce a knowledge from a "dumb" component (badge) about "what it should do when it's used in a specific way" (as part of a Popover).

Isn't this behavior (close a tooltip on click) more about the relation between the tooltip and the button (badge) instead of between the badge and the popover?

I think it can be cleaner if we have a tooltip that's more like the radix / shadcn one:

https://www.radix-ui.com/primitives/docs/components/tooltip : `https://www.radix-ui.com/primitives/docs/components/tooltip`

packages/core/src/extensions/Comments/CommentsPlugin.ts
YousefED: lol thx

YousefED:
Bugs:

- [ ] - action menu closes when emojis are open
- [ ] - removing reactions doesn't work
- [ ] - "reacted by 1"
- [ ] - emoji background color doesn't match user
- [ ] - comment-only user doesn't show formatting toolbar
- [ ] - new comments don't always work
- [ ] - comments on images
- [ ] - selection when composing a comment
- [ ] - composer dissappears when other user is typing
- [ ] - should the thread be deleted when the main (first) comment is removed?

Nice to have:

- [ ] - better mouse over area for showing action toolbar
- [ ] - hide emojis when editing
- [ ] - badge ux (hover background / border)
- [ ] - reduce space between avatar and username
- [ ] - reposition on window resize
- [ ] - copy comment when cut+paste (optionally also with copy / paste?)
- [ ] - cmd+enter to save?
- [ ] - disable children (indentation)
- [ ] - close emoji popup (action bar) when adding reaction
- [ ] -

packages/mantine/src/comments/Editor.tsx
```diff
@@ -0,0 +1,55 @@
...
+  HTMLDivElement,
+  ComponentProps["Comments"]["Editor"]
+>((props, ref) => {
+  const { className, onFocus, onBlur, editor, editable, ...rest } = props;
+
+  assertEmpty(rest, false);
+
+  // When we click the edit button on a comment, we also want to focus the
+  // comment editor
+  useEffect(() => {
```
YousefED, line 21: TODO: implement here or in blocknote/react?

packages/mantine/src/menu/Menu.tsx
```diff
@@ -147,9 +146,8 @@ export const Menu = (props: ComponentProps["Generic"]["Menu"]["Root"]) => {
   return (
     <MantineMenu
       withinPortal={false}
-      middlewares={{ flip: true, shift: true, inline: false, size: true }}
-      onClose={() => onOpenChange?.(false)}
-      onOpen={() => onOpenChange?.(true)}
+      // middlewares={{ flip: true, shift: true, inline: false, size: true }}
```
YousefED, line 149: @matthewlipski was this removed intentionally?

packages/mantine/src/style.css
```diff
@@ -564,3 +565,118 @@
...
+  gap: 4px;
+}
+
+.bn-mantine .bn-badge .mantine-Chip-label > span:not(.mantine-Chip-iconWrapper) > span {
+  align-items: center;
+  display: inline-flex;
+  justify-content: center;
+}
+
+/* We need to get rid of the checked icon - you can set the icon prop to an
```
YousefED, line 677: @matthewlipski are you sure? I don't notice this space on https://mantine.dev/core/chip/#states

packages/mantine/src/toolbar/ToolbarButton.tsx
```diff
@@ -40,72 +44,97 @@ export const ToolbarButton = forwardRef<HTMLButtonElement, ToolbarButtonProps>(
...
+        aria-label={label}
+        // Needed as Safari doesn't focus button elements on mouse down
+        // unlike other browsers.
+        onMouseDown={(e) => {
+          if (isSafari()) {
+            (e.currentTarget as HTMLButtonElement).focus();
+          }
+        }}
+        onClick={(event) => {
+          setHideTooltip(true);
```
YousefED, line 101: TODO: document

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -0,0 +1,347 @@
...
+                        </Components.Generic.Toolbar.Button>
+                      </Components.Generic.Toolbar.Root>
+                    )}
+                  </>
+                )
+              : undefined
+          }
+        />
+      ) : (
+        // Soft deletes
```
YousefED, line 339: TODO

packages/shadcn/src/toolbar/Toolbar.tsx
```diff
@@ -48,20 +56,29 @@ export const ToolbarButton = forwardRef<HTMLButtonElement, ToolbarButtonProps>(
       isDisabled,
       onClick,
       label,
+      variant,
       ...rest
     } = props;
 
     // false, because rest props can be added by shadcn when button is used as a trigger
     // assertEmpty in this case is only used at typescript level, not runtime level
+    // TODO
```
YousefED, line 65: TODO

playground/package.json
```diff
@@ -21,8 +21,14 @@
     "@blocknote/xl-multi-column": "^0.24.1",
     "@emotion/react": "^11.11.4",
     "@emotion/styled": "^11.11.5",
-    "@liveblocks/client": "^1.10.0",
-    "@liveblocks/yjs": "^1.10.0",
+    "@tiptap/suggestion": "^2.7.1",
+    "@tiptap/core": "^2.7.1",
+    "@tiptap/react": "^2.7.1",
+    "@liveblocks/client": "file:../../liveblocks/packages/liveblocks-client",
```
YousefED, line 27: TODO

tests/src/end-to-end/shadcn/shadcn.test.ts-snapshots/shadcn-drag-handle-menu-chromium-linux.png
YousefED, line 1: @matthewlipski this seems like a regression

packages/ariakit/src/toolbar/ToolbarButton.tsx
```diff
@@ -32,12 +32,14 @@ export const ToolbarButton = forwardRef<HTMLButtonElement, ToolbarButtonProps>(
 
     // false, because rest props can be added by ariakit when button is used as a trigger
     // assertEmpty in this case is only used at typescript level, not runtime level
+
+    // TODO
```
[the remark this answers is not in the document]
YousefED: @matthewlipski

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -180,6 +186,12 @@ export type BlockNoteEditorOptions<
    */
   resolveFileUrl: (url: string) => Promise<string>;
 
+  resolveUsers: (userIds: string[]) => Promise<User[]>;
+
+  // TODO: decide "comments" or "threads"?
```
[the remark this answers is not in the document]
nperez0111: I think `comments` makes the most sense, since it options related to the commenting feature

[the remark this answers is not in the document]
YousefED: Ok :) Would you still name the TheadStore as is, or prefer CommentStore?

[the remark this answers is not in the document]
nperez0111: Would keep it as is. comments: { threadStore: Store }

docs/pages/docs/collaboration/comments.mdx
```diff
@@ -0,0 +1,138 @@
...
+const threadStore = new TiptapThreadStore(
+  userId, // The active user's ID
+  provider, // Tiptap collaboration provider
+  new DefaultThreadStoreAuth(userId, "editor"), // Authorization rules (see below)
+);
+```
+
+### ThreadStoreAuth
+
+The `ThreadStoreAuth` class defines the authorization rules for interacting with comments. Every ThreadStore implementation requires a `ThreadStoreAuth` instance. BlockNote uses the `ThreadStoreAuth` instance to deterine which interactions are allowed for the current user (for example, whether they can create a new comment, edit or delete a comment, etc.).
```
nperez0111, line 108: to determine

docs/pages/docs/collaboration/comments.mdx
```diff
@@ -0,0 +1,138 @@
...
+async function myResolveUsers(userIds: string[]): Promise<User[]> {
+  // fetch user information from your database / backend
+  // and return an array of User objects
+
+  return await callYourBackend(userIds); //
+
+  // Return a list of users
+  return users;
+}
+```
```
nperez0111, line 138: Maybe this should be a `Map<User['id'], User>` to make it order-independent and automatically de-duped

packages/core/src/editor/BlockNoteTipTapEditor.ts
```diff
@@ -184,7 +184,10 @@ export class BlockNoteTipTapEditor extends TiptapEditor {
 
     // emit the created event, call here manually because we blocked the default call in the constructor
     // (https://github.com/ueberdosis/tiptap/blob/45bac803283446795ad1b03f43d3746fa54a68ff/packages/core/src/Editor.ts#L117)
-    this.commands.focus(this.options.autofocus);
+    this.commands.focus(
+      this.options.autofocus ||
+        this.options.element.getAttribute("data-bn-autofocus") === "true"
```
nperez0111, line 189: This seems strange, I've always found the autofocus behavior strange in tiptap shouldn't we just use the normal HTML version and read that from the element?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1260,4 +1357,12 @@ export class BlockNoteEditor<
...
     );
   }
+
+  public get ForceSelectionVisible() {
+    return this.showSelectionPlugin.ForceSelectionVisible;
+  }
+
+  public set ForceSelectionVisible(forceSelectionVisible: boolean) {
+    this.showSelectionPlugin.ForceSelectionVisible = forceSelectionVisible;
+  }
```
nperez0111, line 1367: Casing seems wrong here.

packages/core/src/extensions/ShowSelection/ShowSelectionPlugin.ts
```diff
@@ -0,0 +1,52 @@
...
+const PLUGIN_KEY = new PluginKey(`blocknote-show-selection`);
+
+/**
+ * Plugin that shows adds a decoration around the current selection
+ * This can be used to highlight the current selection in the UI even when the
+ * text editor is not focused.
+ */
+export class ShowSelectionPlugin {
+  public readonly plugin: Plugin;
+  private forceSelectionVisible = false;
```
nperez0111, line 14: I feel like this may be better described as `enabled`

packages/mantine/src/badge/Badge.tsx
```diff
@@ -0,0 +1,87 @@
...
+        setHideTooltip(true);
+        onClick?.(event);
+      }}
+      wrapperProps={{
+        onMouseEnter,
+        onMouseLeave: () => setHideTooltip(false),
+      }}
+      variant={"light"}
+      icon={<></>}
+      ref={ref}>
```
nperez0111, line 49: Can null

packages/mantine/src/style.css
nperez0111: Maybe if you null the icon instead, it could work? They are likely using truthiness to tell if an icon was provided (and therefore should use padding)

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -0,0 +1,341 @@
...
+}
+
+/**
+ * The Comment component displays a single comment with actions,
+ * a reaction list and an editor when editing.
+ *
+ * It's generally used in the `Thread` component for comments that have already been created.
+ *
+ */
+export const Comment = ({
```
nperez0111, line 69: This component is doing an awful lot. Consider breaking this up into either multiple components, or extract a hook to keep the business logic separate.

packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
```diff
@@ -43,8 +43,7 @@ export class FormattingToolbarView implements PluginView {
       return false;
     }
 
-    // check view.hasFocus so that the toolbar doesn't show up when the editor is not focused or when for example a code block is focused
-    return !(!view.hasFocus() || empty || isEmptyTextBlock);
+    return !(empty || isEmptyTextBlock);
```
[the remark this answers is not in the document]
YousefED: We needed to remove this to actually be able to show the formatting toolbar when the editor is non-editable. (i.e.: comment-only user that wants to leave a comment).

We also register a blur handler that should hide the toolbar when the user blurs the editor. @matthewlipski and I have tested the basics for this, but I do agree this is a "sensitive" change that we maybe should test further

(let's do a joint testing session after tomorrow's standup?)

packages/core/src/editor/BlockNoteTipTapEditor.ts
YousefED: I played around with this, but somehow `element.autofocus` returns false even when I set it to true on the div. Maybe sth with browser contenteditable compatibility.

The good thing however is that for the user our API does feel more HTML-like (they can do `<BlockNoteView autofocus={true}>`)

docs/pages/docs/collaboration/comments.mdx
YousefED: Thanks. Might be a good idea, agree it's a better API with a Map, but also a bit more complex to explain maybe.

I'll decide when finalizing unless you feel very strong about it :)

packages/ariakit/src/ariakitStyles.css
```diff
@@ -1003,4 +1003,187 @@ responsible for the majority of the styling. */
 
 .bn-ak-link:where(.dark, .dark *) {
     color: hsl(204 100% 64%);
+}
+
+/* https://ariakit.org/components/hovercard */
+/*.bn-ak-button {*/
```
nperez0111, line 1009: Needed?

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -0,0 +1,297 @@
...
+
+const PLUGIN_KEY = new PluginKey(`blocknote-comments`);
+const SET_SELECTED_THREAD_ID = "SET_SELECTED_THREAD_ID";
+
+type CommentsPluginState = {
+  /**
+   * Store the positions of all threads in the document.
+   * this can be used later to implement a floating sidebar
+   */
+  threadPositions: Map<string, { from: number; to: number }>;
```
nperez0111, line 20: I'm working on some bugs in the suggestion plugin that will be related to this. It isn't safe to store positions like this. Would likely also need to keep around a relative position as well. But, we can figure this out later

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -0,0 +1,297 @@
...
+            );
+            tr.removeMark(trimmedFrom, trimmedTo, markType);
+            tr.addMark(
+              trimmedFrom,
+              trimmedTo,
+              markType.create({
+                ...mark.attrs,
+                orphan: isOrphan,
+              })
+            );
```
nperez0111, line 119: Can't this just be that you update the mark attr instead of removing & re-applying it?

nperez0111: nvm, I see there isn't an easy API for that. I just remember a step specifically for updating markattrs but isn't exposed on transactions for some reason

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+    this.userStore = new UserStore<User>(editor.resolveUsers);
+
+    // Note: Plugins are currently not destroyed when the editor is destroyed.
+    // We should unsubscribe from the threadStore when the editor is destroyed.
+    this.threadStore.subscribe(this.updateMarksFromThreads);
+
+    editor.onCreate(() => {
+      this.updateMarksFromThreads(this.threadStore.getThreads());
+      editor.onSelectionChange(() => {
```
nperez0111, line 151: Feels weird to register this after the create hook, why not just register in the constructor separately?

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -0,0 +1,297 @@
...
+    super();
+
+    if (!editor.resolveUsers) {
+      throw new Error("resolveUsers is required for comments");
+    }
+    this.userStore = new UserStore<User>(editor.resolveUsers);
+
+    // Note: Plugins are currently not destroyed when the editor is destroyed.
+    // We should unsubscribe from the threadStore when the editor is destroyed.
+    this.threadStore.subscribe(this.updateMarksFromThreads);
```
nperez0111, line 147: Yea, we need an API for this

packages/core/package.json
```diff
@@ -105,6 +105,7 @@
     "@types/emoji-mart": "^3.0.14",
     "@types/hast": "^3.0.0",
     "@types/uuid": "^8.3.4",
+    "@hocuspocus/provider": "^2.15.2",
```
nperez0111, line 108: I think this is peerOptional no?
https://stackoverflow.com/questions/62047806/how-do-i-set-a-peer-dependency-optional#66228639

TS might care for someone who imports it

packages/react/src/components/Comments/FloatingComposerController.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const state = useUIPluginState(comments.onUpdate.bind(comments));
+
+  // the reference position is updated automatically when the selection moves,
+  // this can happen when the document is updated by a remote user
+  const referencePos = useEditorSelectionBoundingBox(state?.pendingComment);
+
+  useEffect(() => {
+    editor.ForceSelectionVisible = !!state?.pendingComment;
+  }, [editor, state?.pendingComment]);
```
nperez0111, line 44: This is weird. Can we describe why this is needed, it also feels like it should not be handled by React, because it is being derived from the editor state?

packages/react/src/components/Comments/FloatingThreadController.tsx
```diff
@@ -0,0 +1,108 @@
...
+      return;
+    }
+
+    const el = editor.domElement?.querySelector(
+      `[data-bn-thread-id="${state?.selectedThreadId}"]`
+    );
+    if (el) {
+      setReference(el);
+    }
+  }, [setReference, editor, state?.selectedThreadId]);
```
nperez0111, line 74: Definitely strange, I don't quite understand how this all works

packages/react/src/components/Comments/useUsers.ts
```diff
@@ -0,0 +1,52 @@
...
+import { useCallback, useRef, useSyncExternalStore } from "react";
+
+export function useUser(
+  editor: BlockNoteEditor<any, any, any>,
+  userId: string
+) {
+  return useUsers(editor, [userId]).get(userId);
+}
+
+export function useUsers(
```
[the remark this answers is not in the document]
nperez0111: Yea, let's work on this

packages/ariakit/src/ariakitStyles.css
matthewlipski: Since these styles are copy/pasted from the ariakit docs, I think better to leave it in and commented out - makes it clearer why the styles aren't present if we have to update them

packages/mantine/src/style.css
matthewlipski: The Mantine docs don't have an example where they outright remove the check icon, only change it for smth else. Using `null` makes Mantine just use the default too unfortunately. So I believe this is still necessary

packages/ariakit/src/style.css
```diff
@@ -11,10 +11,11 @@
   gap: 0.5rem;
 }
 
-.bn-toolbar.bn-ak-toolbar {
-  overflow-x: auto;
-  max-width: 100vw;
+.bn-ak-toolbar {
```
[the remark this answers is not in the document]
matthewlipski: Whoops pretty sure this is left over from when I was styling the action toolbar, will fix

packages/core/src/extensions/Comments/CommentsPlugin.ts
matthewlipski: It needs to access the TipTap state, and bc we init the comments plugin in the `BlockNoteEditor` constructor, the state is still undefined at that time. Will add a comment

packages/react/src/components/Comments/FloatingThreadController.tsx
matthewlipski: @YousefED can you explain what's happening here? Seems a bit hacky to use `querySelector`

packages/core/package.json
```diff
@@ -116,6 +115,14 @@
     "vite-plugin-eslint": "^1.8.1",
     "vitest": "^2.0.3"
   },
+  "peerDependencies": {
```
YousefED, line 118: @nperez0111 does this makes sense as a peerdependency even if we're only using a type from this package?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1431,11 +1368,11 @@ export class BlockNoteEditor<
     );
   }
 
-  public get ForceSelectionVisible() {
-    return this.showSelectionPlugin.ForceSelectionVisible;
+  public getForceSelectionVisible() {
```
YousefED, line 1371: why not use a getter / setter for this? (like we do for "isEditable")?

packages/react/src/components/Comments/FloatingComposerController.tsx
```diff
@@ -32,17 +32,18 @@ export const FloatingComposerController = <
   }
 
   const comments = editor.comments;
+  useEffect(() => {
```
YousefED, line 35: was this a code style preference to go from:

```
useEffect(() => {	
    editor.ForceSelectionVisible = !!state?.pendingComment;	
  }, [editor, state?.pendingComment]);
```
to 

```
 useEffect(() => {
comments.onUpdate((state) =>
      editor.setForceSelectionVisible(state.pendingComment)
    );
  }, [comments, editor]);
```

or does it solve a bug?

In any case, we're not unsubscribing from `onUpdate` so that would need to be fixed I think

tests/src/end-to-end/theming/theming.test.ts
```diff
@@ -48,7 +48,8 @@ test.describe("Check Dark Theme is Automatically Applied", () => {
     await page.keyboard.press("ArrowLeft");
 
     await page.waitForTimeout(500);
-    expect(await page.screenshot()).toMatchSnapshot("dark-link-toolbar.png");
+    // expect(await page.screenshot()).toMatchSnapshot("dark-link-toolbar.png");
```
YousefED, line 51: dev leftover?

packages/core/src/editor/BlockNoteEditor.ts
YousefED: @matthewlipski can you add a comment?

PR #1388, opened 2025-01-23
Fix Code Block syntax highlighting for imported text 

### LINKED ISSUE #1231: Syntax highlighting in code blocks not working properly when added in the initial content.
**Describe the bug**
The syntax highlighting in code blocks is not working properly when I add initial codeblock content with multiple lines. 

**To Reproduce**
add Initial codeblock content with multiple lines to the editor

Example component where the content comes from the initialContent prop
```
"use client";
import React from "react";
import { BlockNoteSchema, defaultBlockSpecs, PartialBlock } from "@blocknote/core";
import "@blocknote/core/fonts/inter.css";
import { useTheme } from "next-themes";

interface ViewerProps {
    initialContent?: string;
}

    blockSpecs: {
      // Adds all default blocks.
      ...defaultBlockSpecs
    },
  });

const Viewer: React.FC<ViewerProps> = ({ initialContent }) => {
    const { theme } = useTheme()
    
        schema,
        initialContent: initialContent
            ? (JSON.parse(initialContent) as PartialBlock[])
            : undefined,
    });

    return (
        <div className="-mx-[54px] my-4">
            <BlockNoteView
                editor={editor}
                editable={false}
                onChange={() => {}}
                theme={theme as "light" | "dark"}
            />
        </div>
    );
};

export default Viewer;

```

**Misc**
- Node version: 20.9.0
- Package manager: 10.2.4
- Browser: Chrome, Brave, Firefox

Hansenq:
Code blocks work great when the user is typing or pasting text into the code block. When this happens, a newline `\n` character is inserted into the prosemirror node's text content. 

When code is imported, there's an off-by-one syntax highlighting error for every newline in the code block. Note that this bug only occurs when code is _imported_ (not typed or pasted). 

Incorrect result:
Output of problematic code

When importing a code block from HTML/Markdown/initial placeholder content, the imported code block is converted to an array of `Block`s in `markdownToBlocks`, then inserted into the document with `insertBlocks`. `insertBlocks` converts these blocks to Prosemirror Nodes using the `blocksToNode` function. This function calls `styledTextToNodes`, which converts `\n` characters into a HardBreak node. This HardBreak node breaks Shiki's syntax highlighting because it inserts `<br>` tags inside the CodeBlock's `<pre><code>` tag. Shiki expects everything inside `<pre><code>` tags to be a string and results in an off-by-one syntax highlighting error if there's a `<br>` tag. 

Expected result:
Proper syntax highlighting

Fixes #1231. Involves some prop drilling but I couldn't figure out a better way to solve this.


matthewlipski:
Thanks for this PR! Looks great, just made some small improvements

PR #1405, opened 2025-01-30
fix: improve collab cursor code

YousefED:
fixes:

- Heard from a user they sometimes pass `provider: undefined` in which case we'd throw an error (fixed with `provider?.awareness`)
- "I would like to use showCursorLabels: 'activity', but it is not compatible with renderCursor"

TODO:
- [x] test
- [x] organize code better, this should not belong in BlockNoteExtensions.ts

### DIFF
--- a/packages/core/src/editor/BlockNoteExtensions.ts
+++ b/packages/core/src/editor/BlockNoteExtensions.ts
@@ -1,10 +1,4 @@
 import { AnyExtension, Extension, extensions } from "@tiptap/core";
-import { Awareness } from "y-protocols/awareness";
-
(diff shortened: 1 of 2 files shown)


matthewlipski:
For the tests, I suppose we need to change our testing setup? Since I guess it only makes sense to use e2e tests for this, and we need to be able to capture screenshots of 2 editors in the same frame.

YousefED:
@matthewlipski manual testing is ok for this now (I didn't test the changes, only reorganized the code)

matthewlipski (approved):

PR #1413, opened 2025-02-04
fix: Drag and drop event handling

matthewlipski:
The changes introduced in #1341 had some bugs which this PR fixes. The changes are thoroughly documented in the code, but broadly speaking there are 2 main issues that are fixed:

1. `view.dragging` not being cleared & set properly on editors where blocks were not dropped.
2. Overly simplistic handling selection handling on drop.

### DIFF
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -283,75 +283,67 @@ export class SideMenuView<
 
     this.editor._tiptapEditor.commands.blur();
 
-    // Right now, multiple editors on the same page are only supported when
-    // `sideMenuDetection === "editor"`, so we only need to handle cases where
-    // blocks are dragged & dropped between editors then.
-    if (this.sideMenuDetection === "editor") {
-      // When ProseMirror handles a drop event on the editor while
-      // `view.dragging` is set, it deletes the selected content. However, if
-      // a block from a different editor is being dropped, this causes some
-      // issues that the code below fixes:
-      //
+    // When ProseMirror handles a drop event on the editor while
+    // `view.dragging` is set, it deletes the selected content. However, if
+    // a block from a different editor is being dropped, this causes some
+    // issues that the code below fixes:
+    if (!this.isDragOrigin && this.pmView.dom.contains(event.target as Node)) {
       // 1. Because the editor selection is unrelated to the dragged content,
       // we don't want PM to delete its content. Therefore, we collapse the
       // selection.
+      this.pmView.dispatch(
+        this.pmView.state.tr.setSelection(
+          TextSelection.create(
+            this.pmView.state.tr.doc,
+            this.pmView.state.tr.selection.to
+          )
+        )
+      );
+    } else if (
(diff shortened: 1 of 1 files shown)


packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -268,30 +274,71 @@ export class SideMenuView<
...
   };
 
-  /**
-   * If the event is outside the editor contents,
-   * we dispatch a fake event, so that we can still drop the content
-   * when dragging / dropping to the side of the editor
-   */
   onDrop = (event: DragEvent) => {
+    if (this.pmView.dragging === null) {
+      return;
```
YousefED, line 279: when can this happen?

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -268,30 +274,71 @@ export class SideMenuView<
...
+        // editors have had the chance to handle the event.
+        setTimeout(
+          () => this.pmView.dispatch(this.pmView.state.tr.deleteSelection()),
+          0
+        );
+      }
+
+      // PM only clears `dragging` on the editor that the block was dropped, so
+      // we manually have to clear it on all the others.
+      setTimeout(() => (this.pmView.dragging = null), 0);
```
YousefED, line 328: why does this need to be in a settimeout?

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -268,30 +274,71 @@ export class SideMenuView<
...
+
     this.editor._tiptapEditor.commands.blur();
 
-    // ProseMirror doesn't remove the dragged content if it's dropped outside
-    // the editor (e.g. to other editors), so we need to do it manually. Since
-    // the dragged content is the same as the selected content, we can just
-    // delete the selection.
-    if (this.isDragOrigin && !this.pmView.dom.contains(event.target as Node)) {
-      this.pmView.dispatch(this.pmView.state.tr.deleteSelection());
+    if (this.sideMenuDetection === "editor") {
```
YousefED, line 284: why only run this on `this.sideMenuDetection === "editor"`?

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
matthewlipski: Since PM uses `view.dragging` when handling the drop event, we also need to delay clearing it until we're sure that's done - I'll also add a comment here

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
matthewlipski: The PM default behaviour works fine when dragging and dropping blocks within the same editor, and the code here handles cases when blocks are being dragged and dropped between editors.

This assumes that `sideMenuDetection === "viewport"` only supports one editor at a time, and you need to use `sideMenuDetection === "editor"` to support multiple editors on the same page. That is currently true, but maybe in the future we will revisit it.

Issue #1419, opened 2025-02-07
bug: firefox cursor end of line issue


YousefED:
On firefox, somehow the collaboration cursor wraps to the next line when the other user is typing spaces at the end of the line.

_Root cause_
I did not find the root cause yet, it doesn't seem to happen on tiptap / y-prosemirror demos.

_possible fix_
when I remove the following cursor styles, the issue doesn't occur:

```css
margin-left: -1px; 
margin-right: -1px;
```

Ah fix could be to use `outline` as then the caret would still be shown, but the element can have width 0 which would probably solve most layout issues. However, before we go for that solution, perhaps it would be better to find the root cause

Another way to fix it is to remove the non-breaking space  (`\u2060`), but that will of course cause other issues

_reproduce_
in firefox, open a collaborative demo like https://blocknote-main.vercel.app/collaboration/y-sweet, and have another user type whitespace at end of line

nperez0111:
Resolved with https://github.com/TypeCellOS/BlockNote/pull/1423

PR #1422, opened 2025-02-10
fix: cursors will not break to new-line on firefox #1419

nperez0111:
This resolves an issue where in Firefox the cursor's caret was breaking to a new line in specific scenarios. Resolves (#1419)

Original issue:

https://github.com/user-attachments/assets/680a8469-284e-4f1a-acde-863638a45f59

After fix is applied:

https://github.com/user-attachments/assets/04d4aa04-2c09-4fa7-95cf-79eb6cb61d99

You'll also notice I repositioned the label to actually line-up to the caret element. I felt like it just looked wrong otherwise. 

Before:

image

After:

image

If this should be changed in another PR, then happy to do so, just wanted this to look great. Also willing to make the cursor thicker than 1 px it is right now

### DIFF
--- a/packages/core/src/editor/editor.css
+++ b/packages/core/src/editor/editor.css
@@ -121,7 +121,7 @@ Tippy popups that are appended to document.body directly
   max-height: 1.1rem;
   max-width: 20rem;
   padding: 0.1rem 0.3rem;
-  top: -19px;
+  top: -18px;
   left: -1px;
   border-radius: 3px 3px 3px 0;
 
--- a/packages/core/src/editor/editor.css
+++ b/packages/core/src/editor/editor.css
@@ -91,12 +91,13 @@ Tippy popups that are appended to document.body directly
 
 /* Render the username above the caret */
 .collaboration-cursor__label {
-  border-radius: 3px 3px 3px 0;
+  border-radius: 3px;
   font-size: 12px;
   font-style: normal;
   font-weight: 600;
   line-height: normal;
-  left: -1px;
+  left: 0;
+  transform: translate(-2px, -2px);
   overflow: hidden;
   position: absolute;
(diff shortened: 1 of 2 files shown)


YousefED (changes requested):
🥳 yay, first PR!

I'm not sure about the design changes. The original thinking was to make the cursor look similar to how it is in Google Docs. Could you compare with that? The idea is imo that it's sort of a flag that expands

Also; ended up going for the outline version? Was that because it also solves the "zoom" issue?

nperez0111:
Got it, looked into google docs more & I think I nailed the look. Google docs on the left, blocknote on the right
image

YousefED:
@nperez0111 oh no :( something we didn't think of; of course hovering the cursor to show the username doesn't work anymore if the cursor doesn't take any space (with the outline trick). It's now only possible to hover the "dot", not the caret. Ideas for a fix?

nperez0111:
> @nperez0111 oh no :( something we didn't think of; of course hovering the cursor to show the username doesn't work anymore if the cursor doesn't take any space (with the outline trick). It's now only possible to hover the "dot", not the caret. Ideas for a fix?

Ah yea, didn't realize that was intended. zero-width element is definitely going to have that behaviour. So I'll go back to the display: inline-block

nperez0111:
Alright, tried another stab at it. With display: inline-block, it worked for firefox, but then zooming at specific sizes in chrome would be incorrect.

So, I went with the same outline approach, but now I'm overlaying the element with a pseudo element, specifically to capture the mouse events. This is a nice in-between because the pseudo element is absolutely positioned so it should have no effect on other elements (e.g. to break wrapping).

YousefED (approved):
🙌

YousefED (approved):

PR #1427, opened 2025-02-11
fix: Cut/copy handling in non-editable blocks and empty selections

matthewlipski:
This PR does 2 things:

When the selection is empty, we no longer use our own cut/copy handling and instead let the browser handle the event (so nothing happens). Currently an error is thrown as we attempt to create clipboard data from an empty selection.

When the selection is within a descendant of a `"contenteditable"="false"` element, we no longer use our own cut/copy handling and instead let the browser handle the event. This means the user is selecting some content in a non-editable block. Currently the editor selection is copied instead as this scenario still triggers PM event handlers.

~TODO: Unit tests?~

### DIFF
--- a/packages/core/src/api/clipboard/toClipboard/copyExtension.ts
+++ b/packages/core/src/api/clipboard/toClipboard/copyExtension.ts
@@ -147,6 +147,34 @@ export function selectedFragmentToHTML<
   return { clipboardHTML, externalHTML, markdown };
 }
 
+const checkIfSelectionInNonEditableBlock = () => {
+  // Let browser handle event if selection is empty (nothing
+  // happens).
+  const selection = window.getSelection();
+  if (!selection || selection.isCollapsed) {
+    return true;
+  }
+
+  // Let browser handle event if it's within a non-editable
+  // "island". This means it's in selectable content within a
+  // non-editable block. We only need to check one node as it's
+  // not possible for the browser selection to start in an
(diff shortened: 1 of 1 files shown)


YousefED (commented):
Nice. What was the reason for working on this? Curious how / where you ran into it!

Code-wise, some questions / feedback:
- what if you're within a contenteditable="true" nested in a contenteditable="false"?
- you might be able to use `closest()` instead of looping upo the tree

nperez0111 (approved):
lgtm

matthewlipski:
> Nice. What was the reason for working on this? Curious how / where you ran into it!
> 
> Code-wise, some questions / feedback:
> 
> * what if you're within a contenteditable="true" nested in a contenteditable="false"?
> * you might be able to use `closest()` instead of looping upo the tree

Good point r.e. the `contenteditable="true"`, I think it would make sense to return false if we hit an editable element before we hit a non-editable one, but I'll do some manual testing with that first.

matthewlipski:
Ok looks like putting your `contentDOM` element within a `"contenteditable"="false"` element is not really supported by ProseMirror anyway (it gets very confused regarding positions). Alternatively if you had say, just a text box within a non-editable block, that text is not part of the editor state and so the browser should handle cut/copy events there.

Long story short looks like the current behaviour is fine

YousefED (approved):

Issue #1425, opened 2025-02-11
Table updates RFC


nperez0111:
**Is your feature request related to a problem? Please describe.**

Tables are missing a number of features, but to implement them, we must modify the block schema for tables to allow for additional complexity to support features like:
 - merging / splitting cells
 - cell background color
 - cell alignment?
 - table headers

**Describe the solution you'd like**

Given that a table's contents is just a multi-dimensional array of `InlineContent`, there is not a good place to capture things like `colspan`, `rowspan` and `background-color` per cell.

So, the proposal is to change `TableContent`s representation from this:

```ts
export type TableContent<
  I extends InlineContentSchema,
  S extends StyleSchema = StyleSchema
> = {
  type: "tableContent";
  columnWidths: (number | undefined)[];
  rows: {
    cells: InlineContent<I, S>[][];
  }[];
};
```

into this:

```ts
export type TableItemProps = {
  colspan?: number;
  rowspan?: number;
  backgroundColor?: string;
  alignment?: "left" | "right" | "center" | "justify";
};

export type TableHeader<
  I extends InlineContentSchema,
  S extends StyleSchema = StyleSchema
> = {
  type: "tableHeader";
  props: TableItemProps;
  content: InlineContent<I, S>[];
};

export type TableCell<
  I extends InlineContentSchema,
  S extends StyleSchema = StyleSchema
> = {
  type: "tableCell";
  props: TableItemProps;
  content: InlineContent<I, S>[];
};

export type TableContent<
  I extends InlineContentSchema,
  S extends StyleSchema = StyleSchema
> = {
  type: "tableContent";
  columnWidths: (number | undefined)[];
  rows: {
    cells: (InlineContent<I, S>[] | (TableCell<I, S> | TableHeader<I, S>)[])[];
  }[];
};
```

This would be backwards-compatible by still allowing `InlineContent` as an input type, but the output would be as either `tableCell`s or `tableHeader`s. This would follow the principle of being liberal in what you accept, but strict in what you output. This would mean that a document's output between the two versions would be different and since existing versions cannot be forwards-compatible, it would be a 1 way door to a migration of a document.

**Describe alternatives you've considered**

In order to store these values, they need to be stored on a cell-by-cell basis, so there is not an opportunity to do something like what was done for `columnWidths` that could be added separately to the schema.

Another possibility would be to store metadata adjacent to the `rows` in a separate multidimensional array of the same size as the `rows.cells` like so:

```ts
[...]

PR #1429, opened 2025-02-12
feat(tables): overhaul table cells

nperez0111:
This PR is attempting to address the following for table cells:
 - Backgrounds
 - Text Color
 - Alignment
 - colspan & rowspan
 - header cells


packages/core/src/blocks/defaultBlocks.ts
```diff
@@ -42,6 +42,7 @@ export const defaultBlockSpecs = {
   numberedListItem: NumberedListItem,
   checkListItem: CheckListItem,
   table: Table,
+  tableCell: TableCell,
```
YousefED, line 45: had a quick look as I was looking for the pm-view changes. Are you sure tablecell should be a separate block?

nperez0111: You are right, shouldn't be. Was just fiddling with the types to see what it needed. Removed now

packages/core/src/api/blockManipulation/setupTestEnv.ts
```diff
@@ -117,7 +117,20 @@ const testDocument: PartialBlock[] = [
...
+              content: ["Cell 2"],
+            },
+            {
+              type: "tableCell",
+              content: ["Cell 3"],
+            },
+          ],
         },
         {
           cells: ["Cell 4", "Cell 5", "Cell 6"],
```
ViniCleFer, line 136: @nperez0111 also here

packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.test.ts
```diff
@@ -275,7 +275,34 @@ describe("Test updateBlock", () => {
...
+                  colspan: 1,
+                  rowspan: 1,
+                  textAlignment: "left",
+                  textColor: "default",
+                },
+              },
+            ],
           },
           {
             cells: ["Cell 4", "Cell 5", "Cell 6"],
```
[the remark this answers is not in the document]
nperez0111: Nope, it is backwards-compatible so it can be mixed between the two representations. The only rule is that an entire row must be of the same shape.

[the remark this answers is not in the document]
ViniCleFer: nice, thank you 🙏🏽

nperez0111:
## TODOs

- [ ] merging table cells with content is not working
- [ ] splitting a cell (either vertically or horizontally) even when the cell is only a colspan and rowspan of 1
- [ ] dragging rows and columns over a merged cell shows the drop point at the wrong place
- [ ] Allow dragging a row or column right next to a merged cell
- [ ] generic button component for table cell handle instead of div
- [ ] playwright tests for new UI interactions

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -0,0 +1,301 @@
...
+/**
+ * This will return a grid of the occupied cells in the table.
+ *
+ * @returns The grid of occupied cells.
+ */
+export function getTableCellOccupancyGrid(
+  block: BlockFromConfigNoChildren<DefaultBlockSchema["table"], any, any>
+): OccupancyGrid | null {
+  const { height, width } = getDimensionsOfTable(block);
+
```
nperez0111, line 100: I need to document absolute vs relative positioning better

packages/react/src/components/TableHandles/TableCellHandleMenu/DefaultButtons/ColorPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+        sub={true}
+        className={"bn-menu-dropdown bn-color-picker-dropdown"}>
+        <ColorPicker
+          iconSize={18}
+          text={{
+            color: isPartialTableCell(currentCell)
+              ? currentCell.props.textColor
+              : "default",
+            setColor: (color) =>
+              editor.updateBlock(props.block, {
```
YousefED, line 62: might be a bit nicer to extract these longer methods

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -0,0 +1,301 @@
...
+    );
+  });
+}
+
+/**
+ * This will get all the cells in a column of the table block.
+ *
+ * @returns The column of the table.
+ */
+export function getColumn(
```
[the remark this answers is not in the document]
YousefED: also, afaik there's no way to retrieve the bottom right cell in this example, correct? Let's document explicitly what this method is used for to prevent we'll wrongly use it in the future. I could see how someone expects that iterating over all cells should be possible with getColumn / getRow, but it actually isn't

packages/react/src/components/TableHandles/TableHandleMenu/DefaultButtons/AddButton.tsx
```diff
@@ -84,8 +86,14 @@ export const AddColumnButton = <
           type: "tableContent",
           columnWidths,
           rows: props.block.content.rows.map((row) => {
-            const cells = [...row.cells];
-            cells.splice(props.index + (props.side === "right" ? 1 : 0), 0, []);
+            const cells = [...row.cells] as
```
YousefED, line 89: the addcolumn button breaks when there's a cell with multiple rowspans. Let's see if and how prosemirror-tables handles this?

alt text

after:

alt text

YousefED (commented):
Starting to look solid! Nice job, definitely more to it than expected but we're getting there :):)

I've posted a bunch of comments inline. Tried to quite extensive but of course there was a lot of ground to cover. I have two more generic comments:

- I'm thinking we should put this functionality behind an editor option. The functionality is amazing but I feel it should be a conscious choice for apps to offer the extra complexity to their users. wdyt?
- Although it would be nice, I'm not sure if I'd invest too much time in the playwright tests. At this point, it seems like most of the complexity in this PR come from the abs/rel positions, and how to do operations like "color a column" or "add a row" with tables that have cells with colspan / rowspans. Imo it would be a better investment to invest in unit cases for these edge case than to cover the basic scenarios with Playwright

packages/core/src/api/blockManipulation/tables/tables.test.ts
```diff
@@ -0,0 +1,1065 @@
...
+  children: [],
+} satisfies Block<
+  {
+    table: DefaultBlockSchema["table"];
+  },
+  any,
+  any
+>;
+
+const tableWithRowspan = {
```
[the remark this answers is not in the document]
nperez0111: Unit tests already have a 2x2: https://github.com/TypeCellOS/BlockNote/blob/cb36ad331db4680ec692cf317677413277a8fcfa/packages/core/src/api/blockManipulation/tables/tables.test.ts#L453-L472

packages/core/package.json
```diff
@@ -58,32 +58,32 @@
...
-    "@tiptap/extension-link": "^2.7.1",
-    "@tiptap/extension-paragraph": "^2.7.1",
-    "@tiptap/extension-strike": "^2.7.1",
-    "@tiptap/extension-table-cell": "^2.7.1",
-    "@tiptap/extension-table-header": "^2.7.1",
-    "@tiptap/extension-table-row": "^2.7.1",
-    "@tiptap/extension-text": "^2.7.1",
-    "@tiptap/extension-underline": "^2.7.1",
-    "@tiptap/pm": "^2.7.1",
+    "@tiptap/core": "^2.11.5",
```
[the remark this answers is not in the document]
nperez0111: I did it to get the latest prosemirror-tables -> @tiptap/extension-table -> @tiptap/core

Any reason not to change it? It already is getting the latest v2 anyway

packages/core/src/api/testUtil/partialBlockTestUtil.ts
```diff
@@ -103,7 +128,13 @@ export function partialBlockToBlockForTesting<
       contentType === "inline"
         ? []
         : contentType === "table"
-        ? { type: "tableContent", columnWidths: [], rows: [] }
+        ? {
+            type: "tableContent",
+            columnWidths: [undefined, undefined, undefined],
```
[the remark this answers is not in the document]
nperez0111: Right, should just be undefined

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -34,33 +35,66 @@ export function contentNodeToTableContent<
...
     };
 
-    if (index === 0) {
+    if (rowIndex === 0) {
       rowNode.content.forEach((cellNode) => {
-        // The colwidth array should have multiple values when the colspan of a
-        // cell is greater than 1. However, this is not yet implemented so we
-        // can always assume a length of 1.
-        ret.columnWidths.push(cellNode.attrs.colwidth?.[0] || undefined);
+        let colWidth = cellNode.attrs.colwidth as null | undefined | number[];
```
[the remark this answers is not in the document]
nperez0111: Hm, not sure that I get the concern here?

This is collecting all of the widths of the columns into a single array, based on the first row. So, `ret.columnWidths` should always equal `getDimensionsOfTable().width`

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -34,33 +35,66 @@ export function contentNodeToTableContent<
...
+      } satisfies TableCell<I, S>;
     });
 
     ret.rows.push(row);
   });
 
+  for (let i = 0; i < headerMatrix.length; i++) {
+    if (headerMatrix[i].every((isHeader) => isHeader)) {
+      ret.headerRows = (ret.headerRows ?? 0) + 1;
+    }
```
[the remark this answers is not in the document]
nperez0111: Yea, can be. Doesn't hurt, just saves some looping

packages/core/src/schema/blocks/types.ts
```diff
@@ -289,3 +319,103 @@ export type PartialBlockFromConfig<
 };
 
 export type BlockIdentifier = { id: string } | string;
+
```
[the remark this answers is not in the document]
nperez0111: Wasn't sure where to put them. Do you have a suggestion?

packages/core/src/schema/blocks/types.ts
```diff
@@ -289,3 +319,103 @@ export type PartialBlockFromConfig<
...
+    | null
+): content is TableCell<T, S> {
+  return (
+    isPartialTableCell(content) &&
+    content.props !== undefined &&
+    content.content !== undefined
+  );
+}
+
+export function getColspan(
```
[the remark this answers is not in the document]
nperez0111: It makes it more convenient, regarding all of the partialTableCell stuff. The thing is that with the way that I implemented the backwards-compatiblity, a table's content can still be specified as:
https://github.com/TypeCellOS/BlockNote/blob/290b43cec54d02cd2619a76b58f544a3853abfa8/packages/core/src/schema/blocks/types.ts#L173
Which means that everywhere we are using a table cell, I either have to map it into a table cell or somehow leave it. So I wanted these helpers to all be maximally flexible in their input, instead of always have to check which kind of cell we are dealing with and branching logic on that.

You have a few other comments related to partial table cells, and it is for the same reason

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -443,33 +457,72 @@ export class TableHandlesView<
 
     const { draggingState, colIndex, rowIndex } = this.state;
 
-    const rows = this.state.block.content.rows;
+    const newTable = this.state.block.content.rows.map((row) => {
```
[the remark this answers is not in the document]
nperez0111: It may not be partial, but it can be specified as inline instead of as a table cell

packages/react/src/components/TableHandles/ExtendButton/ExtendButton.tsx
```diff
@@ -21,6 +22,16 @@ import { RiAddFill } from "react-icons/ri";
 import { useComponentsContext } from "../../../editor/ComponentsContext.js";
 import { ExtendButtonProps } from "./ExtendButtonProps.js";
 
+function isCellEmpty<I extends InlineContentSchema, S extends StyleSchema>(
```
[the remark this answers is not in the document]
nperez0111: I've not seen this. If you see it again, can you show me how you did that?

packages/react/src/components/FormattingToolbar/DefaultButtons/TextAlignButton.tsx
```diff
@@ -60,14 +63,99 @@ export const TextAlignButton = (props: { textAlignment: TextAlignment }) => {
           editor.updateBlock(block, {
             props: { textAlignment: textAlignment },
           });
+        } else if (block.type === "table") {
```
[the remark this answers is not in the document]
nperez0111: prosemirror & tiptap don't have anything that would help with this. Especially the text selection across cells.

It could maybe be extracted to some other part of the codebase to not make this button so aware of this

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -443,33 +457,72 @@ export class TableHandlesView<
...
+      };
+    });
+    const columnWidths = this.state.block.content.columnWidths;
 
     if (draggingState.draggedCellOrientation === "row") {
-      const rowToMove = rows[draggingState.originalIndex];
-      rows.splice(draggingState.originalIndex, 1);
-      rows.splice(rowIndex, 0, rowToMove);
+      const row = getRow(this.state.block, rowIndex);
+      // TODO need to work on this logic
```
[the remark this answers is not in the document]
nperez0111: Should be gtg now

packages/core/package.json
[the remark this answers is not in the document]
YousefED: true, let's keep

packages/core/src/api/nodeConversions/nodeToBlock.ts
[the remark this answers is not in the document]
YousefED: Sorry I was indeed not very clear and it might not be a major issue.

My point was (I think), that if `attrs.colwidth.length` somehow doesn't match `attrs.colspan`, there's probably a bug somewhere and it might be smart to either throw an error or fix up the array manually?

Because if the lengths don't match up, columnWidths will become invalid, right?

packages/core/src/schema/blocks/types.ts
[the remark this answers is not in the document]
YousefED: Got it. Would it make our lives (and codebase) easier if we go only for partial instead of full backwards compatibility? e.g.: we could choose to only allow the "old" syntax in `initialContent` for example (when people are restoring from a database), but not when calling functions like `insertBlock` or `updateBlock` etc.

packages/react/src/components/TableHandles/ExtendButton/ExtendButton.tsx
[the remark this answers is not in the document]
YousefED: will check when the preview is up again!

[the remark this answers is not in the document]
nperez0111: Should be rebuilding now 😉

packages/core/src/schema/blocks/types.ts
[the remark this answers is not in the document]
nperez0111: It would slightly simplify all of these checks, but I would also say that it doesn't sting so bad, you should know what you are doing with table cells.

I wouldn't know how easy or not it would be to maintain separate typings like that either.

This also keeps caller's code working, though I don't know how important or not that is

YousefED:
Testing:

- drag drop table breaks
- the bottom extend button doesn't work well when I want to make the table smaller (drag to remove rows)
- When I toggle header row (enable), table gets resized to only 2 rows

Minor: 

- somehow the right-most cell is more difficult to resize compared to [main](https://blocknote-main.vercel.app/). I think it's because of the cell-handle. Is there an easy way to improve this?

packages/core/src/api/blockManipulation/tables/tables.test.ts
```diff
@@ -2,12 +2,21 @@ import { describe, expect, it } from "vitest";
...
-  getRow,
-  resolveAbsoluteTableCellIndices,
-  resolveRelativeTableCellIndices,
+  getCellsAtColumnHandle,
+  getCellsAtRowHandle,
+  getRelativeTableCellIndices,
+  getAbsoluteTableCellIndices,
 } from "./tables.js";
 
+/**
```
YousefED, line 11: wdym with this? (looks the same to me?)

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -224,78 +388,344 @@ export function resolveAbsoluteTableCellIndices(
...
+      return (
+        cells.findIndex((c) => c.row === cell.row && c.col === cell.col) ===
         index
-    );
-  });
+      );
+    });
+  } catch (e) {
+    // In case of an invalid index, return an empty array
+    return [];
```
YousefED, line 466: can this happen? otherwise better to just let the error bubble

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -224,78 +388,344 @@ export function resolveAbsoluteTableCellIndices(
...
+ * This moves a column from one index to another.
+ *
+ * @note This is a destructive operation, it will modify the {@link OccupancyGrid} in place.
+ */
+export function moveColumn(
+  block: BlockFromConfigNoChildren<DefaultBlockSchema["table"], any, any>,
+  fromColIndex: RelativeCellIndices["col"],
+  toColIndex: RelativeCellIndices["col"],
+  occupancyGrid: OccupancyGrid = getTableCellOccupancyGrid(block)
+): TableContent<any, any>["rows"] {
```
YousefED, line 557: as discussed; let's add a comment that it would be nice to throw an error when one of the cells has a colspan or rowspan (no need to actually implement the error at this point imo)

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -224,78 +388,344 @@ export function resolveAbsoluteTableCellIndices(
...
-        index
+  return getTableRowsFromOccupancyGrid(occupancyGrid);
+}
+
+/**
+ * This moves a row from one index to another.
+ *
+ * @note This is a destructive operation, it will modify the {@link OccupancyGrid} in place.
+ */
+export function moveRow(
```
YousefED, line 591: same as above

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -224,78 +388,344 @@ export function resolveAbsoluteTableCellIndices(
...
+
+      // Skip the cells that the rowspan takes up
+      absoluteRow += cell.rowspan;
+    }
+
+    // Then for each column, get the cell at the absolute row index as a relative cell index
+    const cells = new Array(occupancyGrid[0].length)
+      .fill(false)
+      .map((_v, col) => {
+        return getRelativeTableCellIndices(
```
YousefED, line 450: nitpicking on naming; I found it a bit confusing that this function returns the cells (and not just the indices). i.e.: I was confused how `getCellsAtRowHandle` could return cells while the "return" function returns indices

packages/react/src/components/TableHandles/TableCellHandle.tsx
```diff
@@ -40,7 +40,9 @@ export const TableCellHandle = <
       <Components.Generic.Menu.Trigger>
         {/* TODO we should probably make a generic button (wait for comments PR to land) */}
         <div>
-          {props.children || <MdSettings size={14} data-test={"tableHandle"} />}
+          {props.children || (
```
YousefED, line 43: fyi, Still a TODO above

YousefED (commented):
no major questions about the code at this point!

packages/core/src/blocks/TableBlockContent/TableBlockContent.ts
```diff
@@ -160,10 +159,10 @@ export const Table = createBlockSpecFromStronglyTypedTiptapNode(
     TableExtension,
     TableParagraph,
     TableHeader.extend({
-      content: "tableContent",
+      content: "tableContent*",
```
matthewlipski, line 162: What's the reason for this change? Since table cells should only be able to have a single paragraph in them

packages/core/src/editor/editor.css
```diff
@@ -156,7 +156,6 @@ Tippy popups that are appended to document.body directly
 
 .bn-editor [data-content-type="table"] th {
   font-weight: bold;
-  text-align: left;
```
matthewlipski, line 159: Looks like this still needs to be set as table headers default to center alignment, so they're still centered even when the text alignment is set to left

packages/mantine/src/style.css
```diff
@@ -510,14 +511,23 @@
...
+.bn-mantine .bn-table-cell-handle {
+  padding: 0 4px;
+}
+
 .bn-mantine .bn-table-handle svg {
   margin-inline: -4px;
 }
 
+.bn-mantine .bn-table-handle-not-draggable {
+  cursor: default;
```
matthewlipski, line 523: Instead of using `default` and `pointer`, maybe we should use `pointer` and `grab` instead? Slightly nicer imo

packages/react/src/components/FormattingToolbar/FormattingToolbar.tsx
```diff
@@ -20,13 +20,15 @@ import {
...
 import { FilePreviewButton } from "./DefaultButtons/FilePreviewButton.js";
+import { TableCellMergeButton } from "./DefaultButtons/TableCellMergeButton.js";
 import { TextAlignButton } from "./DefaultButtons/TextAlignButton.js";
 import { FormattingToolbarProps } from "./FormattingToolbarProps.js";
 
 export const getFormattingToolbarItems = (
   blockTypeSelectItems?: BlockTypeSelectItem[]
 ): JSX.Element[] => [
   <BlockTypeSelect key={"blockTypeSelect"} items={blockTypeSelectItems} />,
+  <TableCellMergeButton key={"tableCellMergeButton"} />,
```
matthewlipski, line 31: I would put this at the end of the formatting toolbar instead of at the start

packages/xl-pdf-exporter/src/pdf/util/table/Table.tsx
```diff
@@ -67,7 +67,8 @@ export const Table = (props: {
                 : { flex: 1 },
             ]}
             key={index}>
-            {props.transformer.transformInlineContent(cell)}
+            {/* TODO: fix this */}
+            {props.transformer.transformInlineContent(cell as any)}
```
matthewlipski, line 71: Still TODO?

packages/core/src/schema/blocks/types.ts
```diff
@@ -220,14 +239,25 @@ export type SpecificBlock<
...
+  content?: PartialInlineContent<I, S>;
+};
+
 export type PartialTableContent<
   I extends InlineContentSchema,
   S extends StyleSchema = StyleSchema
 > = {
   type: "tableContent";
   columnWidths?: (number | undefined)[];
+  headerRows?: number;
```
matthewlipski, line 257: How come these are numbers? Isn't it only possible to have 1 of each?

packages/react/src/components/SideMenu/DragHandleMenu/DefaultItems/TableHeadersItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+  SpecificBlock,
+  StyleSchema,
+} from "@blocknote/core";
+import { ReactNode } from "react";
+
+import { useComponentsContext } from "../../../../editor/ComponentsContext.js";
+import { useBlockNoteEditor } from "../../../../hooks/useBlockNoteEditor.js";
+import { DragHandleMenuProps } from "../DragHandleMenuProps.js";
+
+export const TableRowHeaderItem = <
```
matthewlipski, line 16: 2 things:

1. Imo would be nice to also have this on the 1st row table handle (same for column headers)
2. Not specific to this component, but menu items in general look wonky with ShadCN

packages/react/src/components/TableHandles/TableCellHandle.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { MdArrowDropDown } from "react-icons/md";
+import { useComponentsContext } from "../../editor/ComponentsContext.js";
+import { TableCellHandleProps } from "./TableCellHandleProps.js";
+import { TableCellHandleMenu } from "./TableCellHandleMenu/TableCellHandleMenu.js";
+
+/**
+ * By default, the TableCellHandle component will render with the default icon.
+ * However, you can override the icon to render by passing children.
+ */
+export const TableCellHandle = <
```
matthewlipski, line 19: Not sure about the name `TableCellHandle`, I think handle implies that you can drag it around. Maybe smth like `TableCellOptions` would be better?

matthewlipski (commented):
Nice! Left some comments - mostly questions and some small suggestions

packages/core/src/schema/blocks/types.ts
nperez0111: We discussed this here: https://github.com/TypeCellOS/BlockNote/issues/1425

But, the idea is that we could in the future support multiple headerRows and headerColumns, but for now we only expose in the UI a single toggle for this. Just gives us more room in the future to implement something more complex if needed

packages/react/src/components/FormattingToolbar/FormattingToolbar.tsx
nperez0111: Hm, not sure here. I want to make the feature more known since it only ever shows for table cells. I feel like that follows the existing pattern like the file specific buttons.

packages/react/src/components/SideMenu/DragHandleMenu/DefaultItems/TableHeadersItem.tsx
nperez0111: 1. Good idea
2. What specifically is wonky? I'm not familiar with it enough to know what feels off about it.

packages/core/src/editor/editor.css
nperez0111: Maybe this is a style thing, but table headers are usually centered.

Also, I didn't want this CSS to conflict with manually setting the text alignment on the cell level given it is a low specificity selector: 

```css
[data-text-alignment="right"] {
  justify-content: flex-end;
  text-align: right;
}
```

matthewlipski: Makes sense, though I think it's still weird that currently changing text alignment on a header cell via the formatting toolbar has no effect between alignment left/center. I think if we want them to be centered by default, we should do it through the text alignment prop.

packages/react/src/components/SideMenu/DragHandleMenu/DefaultItems/TableHeadersItem.tsx
matthewlipski: For 2: sorry, I should specify - menu items with a checkmark show a space for it even when unchecked:
Screenshot 2025-02-28 at 10 51 55
Screenshot 2025-02-28 at 10 52 08

matthewlipski:
I just noticed that after setting text alignment in a cell, the selection closes and formatting toolbar disappears. I don't think it's necessary to fix before merging, but smth we may want to look into next week

nperez0111:
> I just noticed that after setting text alignment in a cell, the selection closes and formatting toolbar disappears. I don't think it's necessary to fix before merging, but smth we may want to look into next week

@matthewlipski I took a quick look into this to see what we can do here, and this is what is happening here: https://github.com/TypeCellOS/BlockNote/blob/add526cd67c5e0b457e533beece73fa675d66e34/packages/react/src/components/FormattingToolbar/DefaultButtons/TextAlignButton.tsx#L104-L115

So, the updateBlock call is completely replacing the table content, which causes the selection to be lost in the replaceRange step. What I'd like out of this is 1 of two things, either make the selection more low-level so that we can set the position before & after to allow it to be in the same spot. Or, for the update block function to be smarter & not replace the whole range, just the "affected" porition, making a fine-grained update of the diff of the changes so that it knows it doesn't really need to mess with the selection at all. Maybe even both.

Definitely think it is out of scope for the moment

YousefED (commented):
had a quick glance at final state. two questions:

- example links to https://www.blocknotejs.org/docs/editor-basics/tables which doesn't exist
- did you want to simplify the types (we discussed this last week) or keep as is for now?

Also; enjoy the fun of closing all related issues 🙌🎉

nperez0111:
>     * example links to [blocknotejs.org/docs/editor-basics/tables](https://www.blocknotejs.org/docs/editor-basics/tables) which doesn't exist

Whoops, meant to link to advanced tables, I renamed it somewhere

>     * did you want to simplify the types (we discussed this last week) or keep as is for now?

Sorry, meant to write something about why I decided not to do this atm. Basically it came down to:
 - To make this change, I would have needed to de-couple the assumptions of the types that can be used for the `initialContent` vs. the partial (for updates) vs. the defined types. I didn't want to spend the time to work on this because it was only a reorganizing of the code that we've already validated here.
 - I took a look at what this would actually simplify, and it ended up only being like 5 callsites, so I felt like it was not worth the effort to literally get less functionality for only marginally simpler code.

PR #1437, opened 2025-02-17
refactor: support markviews

YousefED:
This PR adds support for prosemirror MarkViews, and uses them for BlockNote "styles" when defined with React (i.e.: https://www.blocknotejs.org/docs/custom-schemas/custom-styles#style-implementation-reactcustomstyleimplementation)

Previously we had to create a separate React root to make React custom styles work (see ReactRenderUtil), thanks to markviews this can now be cleaned up


packages/react/src/schema/markviews/CoreMarkViewOptions.ts
nperez0111, line 1: Hm, this is subtly incompatible with the implementation that I wrote for Tiptap V3: https://github.com/ueberdosis/tiptap/blob/next/packages/react/src/ReactMarkViewRenderer.tsx

May be worth reviewing the differences and seeing if we should upstream some of this or if we should follow the one I wrote for V3.

nperez0111 (approved):
Other than the Markview impl question, this looks good

packages/react/src/schema/markviews/CoreMarkViewOptions.ts
YousefED: any specific parts that seem incompatible? We don't expose markviews directly (only indirectly via React custom Styles), so I wonder if it's an issue. Of course, I could see if I can swap out the current implementation for the one you linked to

nperez0111: I looked into it more, specifically some of the CoreMarkViewUserOptions (`as` and `contentAs`), but it does not look like it is even used here so it should be fine.

We can deal with this later when we either want to expose this integration or migrate to the Tiptap V3 implementation of it. So, let's put it off until later

nperez0111 (approved):
We can move forward with this, just can't expose it on a public API without double-checking things

PR #1438, opened 2025-02-17
refactor: new react mounting system

YousefED:
This refactors the old mounting system for React. The biggest improvement is that we don't need `queueMicrotask` in several areas anymore which was a bit of a code smell and possible source for race conditions.

Also cleans up an outdated workaround re onDrop / onPaste that shouldn't be necessary with recent TipTap versions anymore


packages/react/src/editor/EditorContent.tsx
nperez0111, line 1: Nice code 😂

nperez0111 (approved):
This looks good code-wise. Let's hope tiptap merges v3 to make this even simpler!

PR #1439, opened 2025-02-17
fix: placeholder improvements

YousefED:
Placeholder improvements:
- correctly scope placeholders per editor
- add support for an "empty document" placeholder

Example: https://blocknote-git-fix-improve-placeholders-typecell.vercel.app/basic/custom-placeholder

I'm not sure the current API is great (the way they're passed in via the dictionary), but I think we can address this separately

closes https://github.com/TypeCellOS/BlockNote/issues/703
closes https://github.com/TypeCellOS/BlockNote/issues/1200

### DIFF
--- a/packages/core/src/extensions/Placeholder/PlaceholderPlugin.ts
+++ b/packages/core/src/extensions/Placeholder/PlaceholderPlugin.ts
@@ -1,5 +1,6 @@
 import { Plugin, PluginKey } from "prosemirror-state";
 import { Decoration, DecorationSet } from "prosemirror-view";
+import { v4 } from "uuid";
 import type { BlockNoteEditor } from "../../editor/BlockNoteEditor.js";
 
 const PLUGIN_KEY = new PluginKey(`blocknote-placeholder`);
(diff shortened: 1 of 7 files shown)


packages/core/src/extensions/Placeholder/PlaceholderPlugin.ts
```diff
@@ -26,54 +34,50 @@ export class PlaceholderPlugin {
...
 
-        const getSelector = (
-          blockType: string | "default",
-          mustBeFocused = true
-        ) => {
-          const mustBeFocusedSelector = mustBeFocused
-            ? `[data-is-empty-and-focused]`
-            : ``;
+        try {
+          // FIXME: the names "default" and "emptyDocument" are hardcoded
```
nperez0111, line 41: Unsure how important this is

nperez0111 (approved):
lgtm

packages/core/src/extensions/Placeholder/PlaceholderPlugin.ts
YousefED: yeah this relates to;

> I'm not sure the current API is great (the way they're passed in via the dictionary), but I think we can address this separately

PR #1440, opened 2025-02-17
refactor: reuse toolbar types and extract mantine context

YousefED:
- split `components` and `BlockNoteView` file for mantine / ariakit / shadcn to cleanup a bit
- extracted generic toolbar types (see [packages/react/src/editor/ComponentsContext.tsx](https://github.com/TypeCellOS/BlockNote/pull/1440/files#diff-3e7c63cb4b19e5cf416db35bb4d6d16b6cbdaf07e5e1ee205bf600e9003a1f8f)). This makes it clear that for example the formattingtoolbar and linktoolbar expect the same types. In the implementations (e.g. `mantine/src/components.tsx`) we actually pass the same `Toolbar` (and `ToolbarButton`) as the different components.

These changes were extracted from the Comments PR, so that it's easy for Nick to continue on this work when working on tables

### To discuss

does it still make sense to have a separate `FormattingToolbar` and `LinkToolbar`?

Pro:
- makes it easy to pass different components for these different cases (probably pretty niche case)

Con:
- more (internal) code (defining the components)

We _could_ also always use the generic component. And optionally, also pass in a parameter with the "location / usage" (not sure about right naming). e.g., instead of:

```
<Components.FormattingToolbar.Root>
```

we do:

```
<Components.Toolbar.Root usage="formatting-toolbar"
```


nperez0111 (approved):
This looks good to me

PR #1442, opened 2025-02-17
feat: add support for tiptap / liveblocks comments

YousefED:
This PR:
- adds an "add comment" button to the Formatting Toolbar if a commenting plugin has been registered on TipTap
- adds a support to tag certain Marks that are not relevant for BlockNote, such as Marks. TipTap / Liveblocks comment plugins currently rely on marks to associate Comments as annotations to the document, even though they're not really document content. By tagging Marks as `blocknoteIgnore` it's possible to use these marks while not affecting the rest of BlockNote. In Tiptap, this can be done like this:

```typescript
Mark.create(
...

extendMarkSchema(extension) {
    if (extension.name === "comment") { // "comment" is the name of the mark
      return {
        blocknoteIgnore: true,
      };
    }
    return {};
  }
...
```

### DIFF
--- a/packages/core/src/api/nodeConversions/nodeToBlock.ts
+++ b/packages/core/src/api/nodeConversions/nodeToBlock.ts
@@ -131,9 +131,9 @@ export function contentNodeToInlineContent<
       } else {
         const config = styleSchema[mark.type.name];
         if (!config) {
-          if (mark.type.spec.group?.includes("blocknoteIgnore")) {
+          if (mark.type.spec.blocknoteIgnore) {
             // at this point, we don't want to show certain marks (such as comments)
-            // in the BlockNote JSON output. These marks should have "blocknoteIgnore" in their group
+            // in the BlockNote JSON output. These marks should be tagged with "blocknoteIgnore" in the spec
             continue;
           }
           throw new Error(`style ${mark.type.name} not found in styleSchema`);
--- a/packages/core/src/api/nodeConversions/nodeToBlock.ts
+++ b/packages/core/src/api/nodeConversions/nodeToBlock.ts
@@ -104,11 +104,12 @@ export function contentNodeToInlineContent<
       return;
     }
 
-    if (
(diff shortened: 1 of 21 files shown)


packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -130,6 +131,11 @@ export function contentNodeToInlineContent<
       } else {
         const config = styleSchema[mark.type.name];
         if (!config) {
+          if (mark.type.spec.group?.includes("blocknoteIgnore")) {
```
nperez0111, line 134: Could we store something else in the spec rather than using the group for this?

Or maybe by the mark's name?

This feels kind of wrong to use the group for this

YousefED: fair. Prefer just to add anything else on the spec? (i.e.: `(mark.type.spec as any).blocknoteIgnore = true`)?

nperez0111: Yea, looks like the type should be already be open to allowing that sort of modifcation:

```
    /**
    Mark specs can include additional properties that can be
    inspected through [`MarkType.spec`](https://prosemirror.net/docs/ref/#model.MarkType.spec) when
    working with the mark.
    */
    [key: string]: any;
}
```

matthewlipski (approved):

packages/core/src/api/nodeConversions/nodeToBlock.ts
YousefED: nice, and it can be added in tiptap with:

extendMarkSchema(extension) {
    if (extension.name === "comment") { // if "comment" is the name of the comment mark
      return {
        blocknoteIgnore: true,
      };
    }
    return {};
  },

nperez0111: Nice, much cleaner!

PR #1443, opened 2025-02-17
fix: Blocks being deleted when dropped outside editor container

### LINKED ISSUE #1396: After dnd of a custom block, the editor sometimes loses the block or changes its data unexpectedly.
**Describe the bug**
Sometimes, the editor loses a block after dnd. I added logs to the onEditorContentChange function, and after dnd, the block is missing from the editor document. It seems like the block just disappears.

**To Reproduce**
https://drive.google.com/file/d/1zVuWurChspTwsIGIiwXM3iNOt1Hb0gN8/view?usp=sharing

**Misc**
Node version: v16.14.2
Package manager: pnpm
Browser: Chrome  132

### LINKED ISSUE #1435: Blocks disappearing
**Describe the bug**

When I drag blocks a certain way, they disappear. In dev mode I get the image I provided, but on the website / in build mode I imagine the block just disappears. This is a bug that exists since 0.22.0, and is obviously frustrating since users don't want their blocks / content to disappear.

**To Reproduce**
Drag the drag handle a tiny bit to the right and let go. The block disappears (works even in the example in the website)

**Misc**
- Package manager: Node / NPM
- Browser: Chrome

Image

matthewlipski:
This PR fixes an issue with block drag & drop that was introduced in PR #1413. The code block
```ts
if (this.isDragOrigin && !this.pmView.dom.contains(event.target as Node)) {
  setTimeout(() => this.pmView.dispatch(this.pmView.state.tr.deleteSelection()), 0);
}
```
is supposed to check if the dragged blocks come from the same editor instance as the handler, but are being dropped in a different editor instance. This is wrong as `!this.pmView.dom.contains(event.target as Node)` can be any element outside the current editor instance, and not necessarily another editor.

And so, all that ends up happening when a drop event occurs outside the current editor instance is that the selection is deleted.

Closes #1396
Closes #1435

### DIFF
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -283,14 +283,15 @@ export class SideMenuView<
 
     this.editor._tiptapEditor.commands.blur();
 
-    let parentEditorElement =
-      event.target instanceof Node ? event.target : null;
-    while (
-      parentEditorElement instanceof HTMLElement &&
-      !parentEditorElement.classList.contains("bn-editor")
-    ) {
-      parentEditorElement = parentEditorElement.parentElement;
-    }
+    // Finds the BlockNote editor element that the drop event occurred in (if
+    // any).
+    const parentEditorElement =
+      event.target instanceof Node
+        ? (event.target instanceof HTMLElement
+            ? event.target
+            : event.target.parentElement
+          )?.closest(".bn-editor") || null
+        : null;
 
     // Drop event occurred within an editor.
     if (parentEditorElement) {
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -284,9 +284,8 @@ export class SideMenuView<
     this.editor._tiptapEditor.commands.blur();
 
     let parentEditorElement =
-      event.target instanceof Node ? event.target.parentElement : null;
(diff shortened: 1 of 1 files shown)


packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -283,47 +283,56 @@ export class SideMenuView<
...
-          TextSelection.create(
-            this.pmView.state.tr.doc,
-            this.pmView.state.tr.selection.to
-          )
-        )
-      );
-    } else if (
-      this.isDragOrigin &&
-      !this.pmView.dom.contains(event.target as Node)
+    let parentEditorElement =
```
YousefED, line 286: can we use `closest` here, or not?

If not, on line 289, I think `parentEditorElement !== null` is clearer than `parentEditorElement instanceof HTMLElement`. Reasoning; checking for `instanceof` doesn't match nicely with the if statement of 296. i.e.: if parentElement would return for example a string (which ofc. doesn't happen, but well..), the if statement would pass.

Zoomed out; I don't think we have a cleaner way of checking whether an element is related to a BlockNote editor? Maybe @nperez0111 has thoughts on this

nperez0111: I'm unsure of this, but the way that this was solved in Tiptap was to store the currently dragged editor instance in a global var, so it could be distinguished whether the drag came from another instance.

https://github.com/ueberdosis/tiptap/pull/5893/files

PR #1446, opened 2025-02-19
fix(table): table cell merging

nperez0111:
How crazy would it be to just allow table cells to have multiple table paragraphs to allow them to merge cleanly, then just store that back in the json cleanly?

I made this as a separate PR since the other one is getting rather large & this is more experimental

### DIFF
--- a/packages/core/src/api/nodeConversions/nodeToBlock.ts
+++ b/packages/core/src/api/nodeConversions/nodeToBlock.ts
@@ -67,10 +67,8 @@ export function contentNodeToTableContent<
 
       return {
         type: "tableCell",
-        content: contentNodeToInlineContent(
-          cellNode.firstChild!,
-          inlineContentSchema,
-          styleSchema
+        content: cellNode.content.content.flatMap((child) =>
+          contentNodeToInlineContent(child, inlineContentSchema, styleSchema)
         ),
         props: {
           colspan: cellNode.attrs.colspan,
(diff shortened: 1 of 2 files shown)


YousefED:
interesting! very minimal solution so could be nice. Downside is I suppose that it makes the reasoning between blocknote schema vs table schema more complex.

Where is the merge logic located in PM makes it work like this? i.e.; where are we "breaking prosemirror" by only allowing a single tableparagraph? I suppose here? https://github.com/ProseMirror/prosemirror-tables/blob/b0130d3bea42c368d38c47dcd5ff09f40dfdd33e/src/commands.ts#L435

nperez0111:
Yep so the content.append is able to add together the two tableParagraphs (which become an array of tableParagraphs at that point) which is not valid for a tablecell. So Prosemirror does it's whole "find the right place to fit the content" algo and messes up the table.

Eh, the only thing that needs to be aware of this is nodeToBlock, which should just always try to conform it into the right shape. We already make some lies about the data model, like how table headers are technically possible anywhere, but we lossily convert that to only be allowed for the headerRow & headerCol.

The alternative to this would be to do the merging logic in blocknote JSON, but that would be quite a lot of work

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -65,10 +65,8 @@ export function contentNodeToTableContent<
 
       return {
         type: "tableCell",
-        content: contentNodeToInlineContent(
-          cellNode.firstChild!,
-          inlineContentSchema,
-          styleSchema
+        content: cellNode.content.content.flatMap((child) =>
```
YousefED, line 68: are we sure this doesn't load any information when we load the blocknote json in the editor again? i.e.: should there be a `join("\n")` for example, or similar?

YousefED (commented):
no objections except the comment added. Nice solution!

packages/core/src/api/nodeConversions/nodeToBlock.ts
nperez0111: I'll take a look at this again, but, yea there definitely is some sort of coalescing that can happen here.

PR #1453, opened 2025-02-21
refactor: add renderEditor boolean to BlockNoteView

YousefED:
This PR cleans up the BlockNoteView code a bit and adds support for `renderEditor={false}`

The idea is that the consumer can then "position" (render) the editor themselves, but still access the Context values that BlockNoteView is setting up. 

Example usage on a different branch: https://github.com/TypeCellOS/BlockNote/blob/42cdbd1a3c5c580465bb9430a9b81eaaa065aa2f/examples/07-collaboration/04-comments/App.tsx#L150

### DIFF
--- a/packages/react/src/editor/BlockNoteView.tsx
+++ b/packages/react/src/editor/BlockNoteView.tsx
@@ -18,11 +18,19 @@ import React, {
 import { useEditorChange } from "../hooks/useEditorChange.js";
 import { useEditorSelectionChange } from "../hooks/useEditorSelectionChange.js";
 import { usePrefersColorScheme } from "../hooks/usePrefersColorScheme.js";
-import { BlockNoteContext, useBlockNoteContext } from "./BlockNoteContext.js";
+import {
+  BlockNoteContext,
+  BlockNoteContextValue,
+  useBlockNoteContext,
+} from "./BlockNoteContext.js";
 import {
   BlockNoteDefaultUI,
   BlockNoteDefaultUIProps,
 } from "./BlockNoteDefaultUI.js";
+import {
+  BlockNoteViewContext,
+  useBlockNoteViewContext,
+} from "./BlockNoteViewContext.js";
 import { Portals, getContentComponent } from "./EditorContent.js";
 import { ElementRenderer } from "./ElementRenderer.js";
 import "./styles.css";
@@ -41,7 +49,17 @@ export type BlockNoteViewProps<
   theme?: "light" | "dark";
 
   /**
+   * Whether to render the editor element itself.
+   * When `false`, you're responsible for rendering the editor yourself using the `BlockNoteViewEditor` component.
+   *
+   * @default true
+   */
+  renderEditor?: boolean;
+
+  /**
    * Locks the editor from being editable by the user if set to `false`.
+   *
+   * @default true
    */
   editable?: boolean;
   /**
@@ -86,9 +104,12 @@ function BlockNoteViewComponent<
     sideMenu,
     filePanel,
     tableHandles,
+    autoFocus,
+    renderEditor,
     ...rest
   } = props;
 
+  const doRenderEditor = renderEditor ?? true;
   // Used so other components (suggestion menu) can set
   // aria related props to the contenteditable div
   const [contentEditableProps, setContentEditableProps] =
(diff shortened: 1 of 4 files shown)


packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -161,31 +148,57 @@ function BlockNoteViewComponent<
...
+        }}>
+        <ElementRenderer ref={setElementRenderer} />
+        {!editor.headless && (
+          <>
+            <Portals contentComponent={portalManager} />
+            {doRenderEditor ? (
+              <BlockNoteViewEditor>{children}</BlockNoteViewEditor>
+            ) : (
+              children
             )}
```
nperez0111, line 198: This sort of brings to mind what the difference is between this `renderEditor={false}` and `headless` being `true` on the editor. Can this be combined?

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -86,9 +104,12 @@ function BlockNoteViewComponent<
     sideMenu,
     filePanel,
     tableHandles,
+    autoFocus,
+    renderEditor,
     ...rest
   } = props;
 
+  const doRenderEditor = renderEditor ?? true;
```
nperez0111, line 112:
```suggestion
    renderEditor = true,
    ...rest
  } = props;
```

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -161,31 +148,57 @@ function BlockNoteViewComponent<
...
+
+  const editorProps = {
+    autoFocus,
+    className,
+    editorColorScheme,
+    mount,
+    contentEditableProps,
+    ref,
+    ...rest,
+  };
```
nperez0111, line 181: This should be memoized, no? Every render will force all consumers of the context to re-render too

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -161,31 +148,57 @@ function BlockNoteViewComponent<
...
+    <BlockNoteContext.Provider value={blockNoteContext}>
+      <BlockNoteViewContext.Provider
+        value={{
+          editorProps,
+          defaultUIProps,
+        }}>
+        <ElementRenderer ref={setElementRenderer} />
+        {!editor.headless && (
+          <>
+            <Portals contentComponent={portalManager} />
```
nperez0111, line 193: Feels like the portals should just be part of `BlockNoteViewEditor` to clean this up

packages/react/src/editor/BlockNoteView.tsx
YousefED: awesome, thanks

packages/react/src/editor/BlockNoteView.tsx
YousefED: I've just tested this, it would require the following code:

```tsx
  const viewContext = useMemo(
    () => {
      const defaultUIProps = {
        formattingToolbar,
        linkToolbar,
        slashMenu,
        emojiPicker,
        sideMenu,
        filePanel,
        tableHandles,
      };

      const editorProps = {
        autoFocus,
        className,
        editorColorScheme,
        contentEditableProps,
        ref,
        ...rest,
      };

      return {
        editorProps,
        defaultUIProps,
      };
    },
    // eslint-disable-next-line
    [
      formattingToolbar,
      linkToolbar,
      slashMenu,
      emojiPicker,
      sideMenu,
      filePanel,
      tableHandles,
      autoFocus,
      className,
      editorColorScheme,
      contentEditableProps,
      ref,
      // eslint-disable-next-line
      ...Object.values(rest),
    ]
  );```

And we'd need to add `React.memo` to the `BlockNoteViewEditor`, otherwise it will still rerender.

This would only prevent rerenders for cases where props that are not part of the `viewContext` are changed. For example, toggling editable (`<BlockNoteView editable={...}>`). The majority of props are passed down to the context anyway.

Do you think this is worth it?

packages/react/src/editor/BlockNoteView.tsx
YousefED: see update

packages/react/src/editor/BlockNoteView.tsx
YousefED: thanks, see update

packages/react/src/editor/BlockNoteView.tsx
nperez0111: Yea, maybe not worth it. I personally try to avoid context for this very reason, but it is too late to think on how to change that now

nperez0111 (approved):
👍  looks good

packages/react/src/editor/BlockNoteView.tsx
YousefED: ok, agree :) let's gradually do this (get rid of context) once we settled on other state management for plugins / stores / etc

PR #1454, opened 2025-02-21
feat: Threads sidebar

matthewlipski:
This PR adds an optional sidebar for threads that can be used instead of having the threads float next to their respective marks. The `ThreadsSidebar` component can be imported from BlockNote to add the sidebar to any editor - see the [docs](https://www.blocknotejs.org/docs/collaboration/comments#sidebar-view) for more.

https://github.com/user-attachments/assets/06056b46-2fb2-4728-be55-6aad1b429647

## `BlockNoteView` changes

The `ThreadsSidebar` component sits outside the editor but needs to have access to the editor context, including Mantine and theme variables. To do this, I modified what `BlockNoteView` and `BlockNoteViewEditor` render when `renderEditor` is false - here's an explanation for what's changed:

Previously:
`BlockNoteView`: Only provides context, doesn't render any elements.
`BlockNoteViewEditor`: Renders both `bn-container` and `bn-editor` elements.

Now:
`BlockNoteView`: Renders `bn-container` element.
`BlockNoteViewEditor`: Renders `bn-editor` element.

Since Mantine and theme variables are set on the `bn-container` element, this ensures that anything within `BlockNoteView` is styled properly. It's also worth noting that theme variables are used by the `react` package, and so we can't really make a similar change to e.g. the Mantine-specific `BlockNoteView` instead.

This change should not change anything when `renderEditor` is true or undefined.


packages/react/src/components/Comments/ThreadStreamView.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { useUIPluginState } from "../../hooks/useUIPluginState.js";
+import { Thread } from "./Thread.js";
+import { useThreads } from "./useThreads.js";
+
+/**
+ * The ThreadStreamView component can be used to display a list of comments in a sidebar.
+ *
+ * This component is similar to Google Docs "Show All Comments" sidebar (cmd+option+shift+A)
+ */
+export function ThreadStreamView(props: {
```
nperez0111, line 12: Only comment here is on the naming. ThreadStream is a bit strange to me.

packages/mantine/src/comments/Comment.tsx
```diff
@@ -37,7 +40,7 @@ const AuthorInfo = forwardRef<
       <Text fz="sm" fw={"bold"}>
         {authorInfo.username}
         <Text fz="xs" c="dimmed" span ml={"xs"}>
-          {timeString}
+          {timeString} {edited && "(edited)"}
```
matthewlipski, line 43: TODO: i18n

examples/07-collaboration/04-comments/App.tsx
```diff
@@ -27,7 +35,7 @@ async function resolveUsers(userIds: string[]) {
 // (but of course, you also use other collaboration providers
 // see the docs for more information)
 export default function App() {
-  const docId = "my-blocknote-document-with-comments";
+  const docId = "my-blocknote-document-with-comments-1";
```
matthewlipski, line 38: TODO: revert before merging

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const filteredThreads = Array.from(threads.values()).filter((thread) =>
+    props.filter === "resolved" ? thread.resolved : !thread.resolved
+  );
+
+  const sortedAndFilteredThreads =
+    props.sort === "position"
+      ? sortByPosition(editor, filteredThreads)
+      : props.sort === "newest"
+      ? sortByNewest(filteredThreads)
+      : sortByReplies(filteredThreads);
```
nperez0111, line 79: Just a nit, but, typically to avoid this sort of thing, I would make a function like:

```ts
sortThreads(editor: BlockNoteEditor, threads: ThreadData[], sortBy: 'replies' | 'newest' | 'position' = 'replies') {
 // Implementation
}
```

packages/react/src/editor/BlockNoteView.tsx
nperez0111, line 1: I'll let Yousef comment here, I don't quite grok this change & why or why not it is needed

nperez0111 (changes requested):
## More general comments

I like that clicking on a comment will focus on the comment in the sidebar, I wonder if that is enough though, maybe there should be a selection style? Or the other comments hide or move out of the way? 

I see that clicking a comment in the sidebar will select the comment in the editor. Is there a way to maybe highlight the comment in the editor when being hovered over like in the Tiptap version: https://tiptap.dev/docs/comments/getting-started/overview

Maybe this is more general to BlockNote, but putting my mouse over the thread comments, and the blocknote sidemenu shows up on the other side of the screen, it is sort of distracting, making me think that hovering the comment is showing me where the comment is in the document when it is just doing nothing. You can't even reach the menu, so I don't know why it would show up at all

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { ThreadData } from "@blocknote/core/comments";
+import { BlockNoteEditor } from "@blocknote/core";
+
+function sortByPosition(
+  editor: BlockNoteEditor<any, any, any>,
+  threads: ThreadData[]
+) {
+  const sortedThreads: ThreadData[] = [];
+
+  editor.prosemirrorState.doc.descendants((node) => {
```
YousefED, line 14: The commentsplugin actually does something similar and sets this info in `threadPositions`. It would be cleaner (and more efficient) to expose and reuse that

YousefED: Also, in your current implementation I think you accidentally filter out threads which don't have content anymore in the document (i.e.: when the text has been deleted). Make sure to double check edge-cases and that the `sort` function really only sorts (and does not filter)

packages/react/src/editor/BlockNoteDefaultUI.tsx
```diff
@@ -48,10 +48,10 @@ export function BlockNoteDefaultUI(props: BlockNoteDefaultUIProps) {
       {editor.tableHandles && props.tableHandles !== false && (
         <TableHandlesController />
       )}
-      {editor.comments && props.comments !== false && (
+      {editor.comments && (
         <>
           <FloatingComposerController />
-          <FloatingThreadController />
+          {props.comments !== false && <FloatingThreadController />}
```
YousefED, line 54: I'm not sure about this. With this change, users cannot use their own composer should they want to

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -174,11 +174,21 @@ function BlockNoteViewComponent<
           defaultUIProps,
         }}>
         <ElementRenderer ref={setElementRenderer} />
-        {renderEditor ? (
-          <BlockNoteViewEditor ref={ref}>{children}</BlockNoteViewEditor>
-        ) : (
-          children
-        )}
+        <div
```
YousefED, line 177: Nice improvement, I think this could work. We need to check all references to `bn-container` in the codebase to check whether this is safe (both in CSS and TS)

Let's also create a separate component `BlockNoteContainer` for this div

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -198,82 +208,49 @@ export const BlockNoteViewRaw = React.forwardRef(BlockNoteViewComponent) as <
...
-          ref={mount}
-          {...contentEditableProps}
-        />
-        {/* The UI elements such as sidebar, formatting toolbar, etc. */}
-        {children}
-      </div>
-    );
-  }
-);
+const EditorElement = (props: {
```
YousefED, line 241: Maybe rename to `ContentEditableElement` for clarity? (and update comment)

YousefED (changes requested):
Nice job! Love the extra work you put into the example. See inline comments + feedback here from testing:

- [x] getting `Warning: Each child in a list should have a unique "key" prop` (might not be threads specific, but let's fix it nevertheless)
- [x] Agree with @nperez0111 that a "selection style" for threads in the sidebar would be a big improvement (so that we not only select the sidebar comment when it's selected in the main editor, but that there's also a clear visual difference)
- [x] I'm not able to select text in any of the sidebar comments
- [ ] emoji picker doesn't work well in the sidebar (cut off + emoji doesn't get added)
- [x] let's hide / collapse comments in large threads, similar to gdocs (`3 more replies`)
- [x] selecting a comment in the sidebar should also scroll the editor to make sure the mark is in view
- [x] there should be a way to not `filter` threads at all (both in the example and as an option).
- [ ] For the above, we also need to show `Marked as resolved` for resolved threads to clearly distinguish resolved threads (similar to gdocs)
- [ ] Comments for which the content has been deleted should still be shown in the sidebar, and we should show `Original content deleted` or sth (similar to gdocs)
- [ ] (nice to have) when clicking a resolved thread, it should still select the text in the document where the orphaned mark is

packages/react/src/components/Comments/Thread.tsx
```diff
@@ -1,53 +1,20 @@
...
-
-  /**
-   * Component element props.
-   */
-  className?: string;
-  onFocus?: () => void;
-  onBlur?: () => void;
-  tabIndex?: number;
+  thread: ThreadData;
+  view: "floating" | "sidebar";
```
YousefED, line 15: same for `index` earlier. Do you prefer this component to have a `view` prop, or the props we had before?

packages/react/src/editor/BlockNoteDefaultUI.tsx
```diff
@@ -48,11 +49,11 @@ export function BlockNoteDefaultUI(props: BlockNoteDefaultUIProps) {
       {editor.tableHandles && props.tableHandles !== false && (
         <TableHandlesController />
       )}
-      {editor.comments && (
-        <>
-          <FloatingComposerController />
-          {props.comments !== false && <FloatingThreadController />}
-        </>
+      {editor.comments && props.commentsComposer !== false && (
+        <FloatingComposerController />
```
YousefED, line 53: I think both elements could be part of the same `comments` flag, wdyt? (i.e.: if you opt out if this you need to render all comment-releated elements yourself)

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -4,40 +4,46 @@ import { Thread } from "./Thread.js";
 import { useThreads } from "./useThreads.js";
 import { ThreadData } from "@blocknote/core/comments";
 import { BlockNoteEditor } from "@blocknote/core";
+import { useEffect, useMemo } from "react";
```
YousefED, line 7: I see you still need to resolve my previous comments to this file - so for now I'll skip re-reviewing :)

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -49,47 +55,74 @@ export function ThreadsSidebar(props: {
...
   );
-
   const selectedThreadId = state?.selectedThreadId;
 
   const threads = useThreads(editor);
 
-  const filteredThreads = Array.from(threads.values()).filter((thread) =>
-    props.filter === "resolved" ? thread.resolved : !thread.resolved
-  );
+  useEffect(() => {
```
YousefED, line 73: I'm not sure if this should be the responsibility of the sidebar per-se. Let's leave it for now and I'll review this in depth when the rest is ready to go

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -174,26 +174,47 @@ function BlockNoteViewComponent<
...
-          data-color-scheme={editorColorScheme}
-          {...rest}
-          ref={ref}>
-          {renderEditor ? (
-            <BlockNoteViewEditor>{children}</BlockNoteViewEditor>
-          ) : (
-            children
-          )}
-        </div>
+        <BlockNoteViewContainer
```
YousefED, line 177: Looks good :)

Note to self: test re-renders

playground/src/examples.gen.tsx
```diff
@@ -1204,6 +1204,30 @@
...
         }
+      },
+      {
+        "projectSlug": "comments-with-sidebar",
+        "fullSlug": "collaboration/comments-with-sidebar",
+        "pathFromRoot": "examples/07-collaboration/05-comments-with-sidebar",
+        "config": {
+          "playground": true,
+          "docs": true,
+          "author": "yousefed",
```
YousefED, line 1215: give yourself some credits :)

YousefED (commented):
great progress!

Didn't test yet as there's no preview environment

packages/react/src/components/Comments/FloatingThreadController.tsx
```diff
@@ -102,7 +105,11 @@ export const FloatingThreadController = <
 
   return (
     <div ref={ref} style={style} {...getFloatingProps()}>
-      <Component threadId={state.selectedThreadId} />
+      <Component
```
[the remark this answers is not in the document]
matthewlipski: High-level the `Comment` and `Thread` components were "dumbed down", so a parent component manages their states rather than the components themselves. That's why the `threadId` was changed to `thread`, as it's now the responsibility of the `FloatingThreadController` or `ThreadsSidebar` to fetch the thread and pass it to the `Thread` component, rather than having the `Thread` component itself fetch it.

This is also why the `selected` prop was added, so that the `ThreadsSidebar` can easily manage which thread is selected rather than the `Thread` component having to figure it out. Floating threads are always selected however, as only one is visible at a time.

The `view` prop is just to add a `onFocus` handler, `onBlur` handler, and `tabIndex` to the thread so the user can select threads in the sidebar. These aren't needed when the floating controller is used instead.

packages/react/src/components/Comments/Thread.tsx
```diff
@@ -127,40 +74,108 @@ export const Thread = ({
     newCommentEditor.removeBlocks(newCommentEditor.document);
   }, [comments, newCommentEditor, thread.id]);
 
-  showComposer =
-    showComposer && comments.threadStore.auth.canAddComment(thread);
+  const [showAllComments, setShowAllComments] = useState(
```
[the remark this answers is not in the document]
matthewlipski: Good point, google docs seems to only use this in the sidebar view

packages/react/src/editor/BlockNoteDefaultUI.tsx
matthewlipski: I think since the `ThreadsSidebar` is part of the API, I think it's gonna be annoying if people have to re-add the `FloatingComposerController` when they use it. Also this is a bit of a different situation vs the other flags, but for those it's the case that 1 flag is for 1 UI element, whereas here the `comments` flag would control multiple elements.

examples/07-collaboration/04-comments/App.tsx
```diff
@@ -122,95 +91,35 @@ function Document() {
...
-                    onClick: () => setCommentSort("replies"),
-                    isSelected: commentSort === "replies",
-                  },
-                ]}
-              />
-            </div>
-            <ThreadsSidebar filter={commentFilter} sort={commentSort} />
-          </div>
-        )}
+      // In other examples, `BlockNoteView` renders both editor element itself,
```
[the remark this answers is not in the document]
matthewlipski: Actually I can remove `renderEditor={false}` and just use some flexbox stuff to flip the order of the user select and editor, so we can keep the blocknote components

packages/react/src/components/Comments/Thread.tsx
matthewlipski: Here I think using `view` is nicer - `onFocus`, `onBlur`, and `tabIndex` can all be consolidated into a single prop. Also it's much clearer what `view` is for, whereas the props it replaces are all generic don't immediately tell you why that their point is.

packages/ariakit/src/comments/Card.tsx
```diff
@@ -37,3 +54,25 @@ export const CardSection = forwardRef<
...
+
+  assertEmpty(rest, false);
+
+  return (
+    <AriakitButton
+      className={mergeCSSClasses(
+        className,
+        "bn-ak-button bn-ak-secondary bn-ak-expand-sections-prompt"
+      )}
+      onMouseDown={(event) => event.preventDefault()}
```
nperez0111, line 72: comment here on why would be nice

packages/core/src/comments/threadstore/yjs/YjsThreadStore.ts
```diff
@@ -237,6 +237,7 @@ export class YjsThreadStore extends YjsThreadStoreBase {
 
     yThread.set("resolved", true);
     yThread.set("resolvedUpdatedAt", new Date().getTime());
+    yThread.set("resolvedBy", this.userId);
```
nperez0111, line 240: We will need to add this new field to the other thread stores.

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -28,37 +22,11 @@ import { ReactionBadge } from "./ReactionBadge.js";
...
-   * Whether to show the comment if it was deleted. If set to `false`, it will render deleted comments as `null`.
-   */
-  showDeleted?: boolean;
-
-  /**
-   * Whether to show reactions.
-   */
-  showReactions?: boolean;
-}
+  index: number;
```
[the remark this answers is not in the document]
nperez0111: Hm, yea. Agree, this feels less flexible than it was.

nperez0111:
There is a bug when replying to threads 

https://github.com/user-attachments/assets/ebc4e077-9124-4b87-a280-639bb16d5b11

nperez0111:
When I do sort by newest, it seems to be sorting by the newest thread, but I'd actually expect to see the newest comment.

May need to go through a thread to see what the newest comment is within a thread and sort the timestamps by that instead.

packages/react/src/components/Comments/Thread.tsx
```diff
@@ -1,84 +1,48 @@
...
+  highlightedText?: string;
+  maxCommentsBeforeCollapse?: number;
+};
 
 /**
  * The Thread component displays a (main) comment with a list of replies (other comments).
  *
  * It also includes a composer to reply to the thread.
  */
 export const Thread = ({
```
nperez0111, line 26: This component is doing too many things, would recommend splitting it up.

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -0,0 +1,188 @@
...
+          from: number;
+          to: number;
+        }
+      >
+    | undefined
+  >(
+    () =>
+      editor.comments?.plugin.getState(editor.prosemirrorState)
+        ?.threadPositions,
+    [editor.comments?.plugin, editor.prosemirrorState]
```
YousefED, line 82: I don't think this would update after a state update, right? (because there won't be a rerender when the state updates. We could also follow the existing pattern by accessing this via  `useUIPluginState`?

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -0,0 +1,188 @@
...
+          .getState(editor.prosemirrorState)
+          .threadPositions.get(b.id)?.from || Number.MAX_VALUE;
+
+      return threadA - threadB;
+    });
+  }
+
+  if (sort === "newest") {
+    return threads.sort((a, b) => {
+      return b.createdAt.getTime() - a.createdAt.getTime();
```
nperez0111, line 32: https://github.com/TypeCellOS/BlockNote/pull/1454#issuecomment-2721015972
```suggestion
      const aLastComment = a.comments.reduce((acc, comment) => {
        return comment.createdAt > acc.createdAt ? comment : acc;
      }, a.comments[0]);
      const bLastComment = b.comments.reduce((acc, comment) => {
        return comment.createdAt > acc.createdAt ? comment : acc;
      }, b.comments[0]);
      return bLastComment.createdAt.getTime() - aLastComment.createdAt.getTime();
```

packages/react/src/components/Comments/Thread.tsx
```diff
@@ -118,38 +76,111 @@ export const Thread = ({
...
+          return;
+        }
+
+        if (
+          !(event.relatedTarget instanceof Node) ||
+          !(event.target instanceof Node) ||
+          !event.target.contains(event.relatedTarget)
+        ) {
+          setShowAllComments(
+            thread.comments.length <= (maxCommentsBeforeCollapse || 5)
```
YousefED, line 150: I don't think this should be based on blur, but based on the state of whether a Thread is selected.

i.e.: when selecting a thread in the editor, it should expand, and when unselecting, it should collapse (didn't test this yet, but judging from the code)

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const threadDataToElement = (thread: ThreadData) => {
+      const threadPosition = threadPositions?.get(thread.id);
+
+      return (
+        <Thread
+          key={thread.id}
+          thread={thread}
+          view={"sidebar"}
+          selected={thread.id === selectedThreadId}
+          highlightedText={
```
YousefED, line 157: wow this is super fancy :) didn't think we'd get this in

- we might need to truncate it? for long parts of text
- in google docs I think they store the original content (like it was when the first comment was added). I think there are definitely some good arguments for that, but maybe ok to keep out of scope for now

packages/react/src/components/Comments/Thread.tsx
```diff
@@ -127,40 +74,108 @@ export const Thread = ({
...
-        })}
+          comments.selectThread(undefined);
+        }
+      }}
+      selected={selected}
+      tabIndex={view === "sidebar" ? 0 : undefined}>
+      <Components.Comments.CardSection className="bn-thread-comments">
+        {commentElements}
+        {thread.resolved && (
+          <Components.Comments.Comment
```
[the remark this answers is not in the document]
YousefED: ah got it, makes sense! They actually also keep track of all reopens and resolves and display those as separate "comments". Might be smart to account for this in the datamodel, I'll check

(TODO @YousefED )

packages/mantine/src/comments/Card.tsx
```diff
@@ -32,3 +64,22 @@ export const CardSection = forwardRef<
...
+export const ExpandSectionsPrompt = forwardRef<
+  HTMLButtonElement,
+  ComponentProps["Comments"]["ExpandSectionsPrompt"]
+>((props, ref) => {
+  const { className, children, onClick, ...rest } = props;
+
+  assertEmpty(rest, false);
+
+  return (
+    <MantineButton
```
YousefED, line 77: nice to use https://mantine.dev/core/divider/#with-label?

packages/react/src/components/Comments/ThreadsSidebar.tsx
YousefED: very curious why you prefer that above sort?

nperez0111: Let's say that I got a notification that a review was done on my document that already had a bunch of existing comments on them. If I had sorted the list by recency, I would expect to just review those items one by one, until I hit the item that was no longer apart of this review anymore.

Whereas, to see that a new comment happened in this sort, I'd have to dig through each thread to see which comment happened and whether I need to action on it or not.

Like, this is sorted by "newest", but especially since we only show to the granularity of a day, it actually feels like a random sort:
image

examples/07-collaboration/04-comments/App.tsx
```diff
@@ -7,10 +7,13 @@ import {
 import { BlockNoteView } from "@blocknote/mantine";
 import "@blocknote/mantine/style.css";
 import { useCreateBlockNote } from "@blocknote/react";
-import { MantineProvider, Select } from "@mantine/core";
 import { YDocProvider, useYDoc, useYjsProvider } from "@y-sweet/react";
 import { useMemo, useState } from "react";
-import { HARDCODED_USERS, MyUserType, getRandomColor } from "./userdata.js";
+
+import { SettingsSelect } from "./SettingsSelect";
+import { HARDCODED_USERS, MyUserType, getRandomColor } from "./userdata";
```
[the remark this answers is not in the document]
nperez0111: Hm, okay I'll take a look at this for you. I thought I had fixed this at one point.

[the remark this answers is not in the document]
nperez0111: Got it, I will apply it to main and you can take it up from there. But change ended up being:

```diff
diff --git i/docs/next.config.mjs w/docs/next.config.mjs
index 523499686..189f95930 100644
--- i/docs/next.config.mjs
+++ w/docs/next.config.mjs
@@ -142,7 +142,7 @@ const nextConfig = withAnalyzer(
 
       // https://github.com/vercel/next.js/issues/59744#issuecomment-2442603644
       config.resolve.extensionAlias = {
-        ".js": [".js", ".ts"],
+        ".js": [".js", ".ts", ".jsx", ".tsx"],
         ".mjs": [".mjs", ".mts"],
         ".cjs": [".cjs", ".cts"],
         ".jsx": [".jsx", ".tsx"],
```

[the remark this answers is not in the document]
nperez0111: On `main` now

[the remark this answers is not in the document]
matthewlipski: Thanks!

packages/core/src/comments/threadstore/yjs/YjsThreadStore.ts
matthewlipski: Actually had a brief talk with @YousefED over Slack about this:

**Matthew Lipski**
Hey I have a question about this comment: 
https://github.com/TypeCellOS/BlockNote/pull/1454#discussion_r1990635124
In the `TipTapThreadStore`, it seems like when converting from `TCollabThread` used by hocuspocus to `ThreadData` used by BlockNote, some fields, like `deletedAt`, aren't available and so they're left undefined. Looks like this is also the case for `resolvedBy`, so shouldn't it also stay undefined?


**Yousef**
for tiptap yes, sounds like that makes sense :slightly_smiling_face:but not for the rest store then


**Matthew Lipski**
For the rest store ig just the demo needs updating?
https://github.com/TypeCellOS/BlockNote-demo-nextjs-hocuspocus/tree/main
Since the client just posts to a `/threadId/resolve`, the only other thing the server needs is the user ID, which it already gets from the authorization header


**Yousef**
ah you’re right, sorry I didn’t check in depth
that should work automatically as it reuses the yjsthreadstore
so we should be all good, my bad

Seems like the others don't need updating, or did we miss smth?

packages/react/src/components/Comments/ThreadsSidebar.tsx
matthewlipski: Can't we just do this instead?
```ts
threads.sort(
  (a, b) =>
    b.comments[b.comments.length - 1].createdAt.getTime() -
    a.comments[a.comments.length - 1].createdAt.getTime()
);
```
Since new comments are pushed to the end of the array, do we need to check all of them or just the last one in the array?

nperez0111: Yea, probably fine. I did it against all of them to capture edits, but that is likely good enough

packages/core/src/comments/threadstore/yjs/YjsThreadStore.ts
nperez0111: Alright, then we are probably good

examples/07-collaboration/05-comments-with-sidebar/style.css
```diff
@@ -53,20 +53,27 @@
   min-width: initial;
 }
 
+.sidebar-comments-main-container .bn-thread-expand-prompt .mantine-Text-root,
```
YousefED, line 56: should the styles in this file all be here, or part of the library?

packages/react/src/components/Comments/ThreadsSidebar.tsx
YousefED: fyi, I think gdocs simply sorts by thread creation time. In this case I think we could just keep it simple next time; probably both solutions are somewhat confusing to the end user. Best way to address nicks concerns is to implement notifications down the line ;)

for now let's keep what we have in the codebase

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -117,66 +180,47 @@ export function ThreadsSidebar(props: {
       state?.threadPositions
     );
 
-    const openThreads: ThreadData[] = [];
-    const resolvedThreads: ThreadData[] = [];
+    const ret: Array<{ thread: ThreadData; referenceText: string }> = [];
```
matthewlipski, line 183: How come we store the `referenceText` here instead of just calling `getReferenceText` when we render the `ThreadItem`s? Ig it's about preventing unnecessary rerenders?

packages/react/src/editor/BlockNoteDefaultUI.tsx
YousefED: then we should also name the other one `floatingThreads` or sth to be more descriptive?

imo a single `comments` flag is easiest atm

examples/07-collaboration/04-comments/README.md
```diff
@@ -1,13 +1,18 @@
-# Comments & Threads
+# Threads Sidebar View
```
YousefED, line 1: wrong file?

examples/07-collaboration/04-comments/README.md
```diff
@@ -1,13 +1,18 @@
...
 
-In this example, you can add comments to the document while collaborating with others. You can also pick user accounts with different permissions, as well as react to, reply to, and resolve existing comments.
+This is another example showcasing threads and comments. However, instead of
```
YousefED, line 3: this text ("This is another example") doesn't work because you can't assume the user has seen the other example

packages/react/src/components/Comments/Comments.tsx
```diff
@@ -3,45 +3,75 @@ import { ThreadData } from "@blocknote/core/comments";
...
+        showActions={false}>
+        <div className={"bn-resolved-text"}>
+          {dict.comments.sidebar.marked_as_resolved}
+        </div>
+      </Components.Comments.Comment>
+    );
+  }
+
+  // Collapses replies if needed.
+  if (
```
YousefED, line 61: could this collapse the "marked as resolved comment"? how does that work in gdocs?

packages/react/src/components/Comments/Comments.tsx
```diff
@@ -3,45 +3,75 @@ import { ThreadData } from "@blocknote/core/comments";
...
+    const resolvedCommentIndex =
+      thread.comments.findLastIndex(
+        (comment) =>
+          thread.resolvedUpdatedAt!.getTime() > comment.createdAt.getTime()
+      ) + 1;
+
+    comments.splice(
+      resolvedCommentIndex,
+      0,
+      <Components.Comments.Comment
```
YousefED, line 44: does this need a `key`?

packages/react/src/components/Comments/Comments.tsx
```diff
@@ -3,45 +3,75 @@ import { ThreadData } from "@blocknote/core/comments";
...
 
-export const Comments = ({ thread, collapse }: CommentsProps) => {
+export const Comments = ({
+  thread,
+  maxCommentsBeforeCollapse,
+}: CommentsProps) => {
   const Components = useComponentsContext()!;
   const dict = useDictionary();
+  const editor = useBlockNoteEditor();
+  const users = useUsers(editor, [thread.resolvedBy!]);
```
YousefED, line 21: can we make this work without the `!`?

packages/react/src/components/Comments/Thread.tsx
```diff
@@ -74,16 +73,6 @@ export const Thread = ({
...
 
-  const userIds = useMemo(() => {
-    return thread.comments.flatMap((c) => [
-      c.userId,
-      ...c.reactions.flatMap((r) => r.userIds),
-    ]);
-  }, [thread.comments]);
-
-  // load all user data
-  const users = useUsers(editor, userIds);
```
YousefED: ah, I suppose this was a bug because we didn't load the `resolvedBy` userid?

packages/react/src/components/Comments/ThreadsSidebar.tsx
```diff
@@ -116,6 +133,10 @@ export function getReferenceText(
   if (!threadPosition) {
     return "Original content deleted";
   }
+  
+  if (editor.prosemirrorState.doc.nodeSize < threadPosition.to) {
```
YousefED, line 137: add comment for this?

packages/react/src/index.ts
```diff
@@ -98,8 +98,15 @@ export * from "./components/TableHandles/TableHandleMenu/DefaultButtons/DeleteBu
 export * from "./components/TableHandles/TableHandleMenu/TableHandleMenu.js";
 export * from "./components/TableHandles/TableHandleMenu/TableHandleMenuProps.js";
 
+export * from "./components/Comments/Comment.js";
```
YousefED, line 101: can you add a todo for later? want to let @nperez0111 if we want to export these as @blocknote/react/comments

packages/react/src/components/Comments/Comments.tsx
matthewlipski: GDocs treats the "Marked as resolved" comments the same way as regular comments when it comes to collapsing threads. There is the caveat that GDocs doesn't let you reply to resolved threads (it will reopen them if you do), which we can quite easily change if needed. GitHub though, for example, does let you respond to resolved threads.

packages/react/src/components/Comments/Thread.tsx
matthewlipski: Good catch! I only tested resolving the thread on a user that also added a comment to the thread, so I didn't spot this but looks like you're right.

PR #1456, opened 2025-02-24
add LaTeX block for equation input

### LINKED ISSUE #741: Custom Block for Katex

Recxsmacx:
Fixes #741

/claim #741


matthewlipski:
Hi, just wondering why this was closed? Was about to take a look at it

PR #1458, opened 2025-02-24
fix: Drag & drop

matthewlipski:
This PR fixes 2 additional issues with drag & drop that we didn't previously catch:

1. When using the multi-column extension, it's possible to drag a block and drop it next to itself. BlockNote will attempt to create 2 columns containing the same block, but will throw an error it tries to remove the same block twice. This edge case has been fixed and now, nothing happens.
2. It's possible that a dragged block is dropped in an element that is not a valid drop target. In this case, only a `dragend` event fires, and not a `drop`. Since we only clear `view.dragging` on drop events, BlockNote will throw an error the next time the user attempts to drag a block, as it will think that a block is already being dragged. Therefore, this has been fixed by making `view.dragging` get cleared on `dragend` events instead.

### DIFF
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -178,6 +178,11 @@ export class SideMenuView<
       this.onDrop as EventListener,
       true
     );
+    this.pmView.root.addEventListener(
+      "dragend",
+      this.onDragEnd as EventListener,
+      true
+    );
     initializeESMDependencies();
 
     // Shows or updates menu position whenever the cursor moves, if the menu isn't frozen.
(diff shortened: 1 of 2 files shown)


YousefED (approved):
great fix that actually gets rid of the hacky settimeout!

packages/core/src/extensions/UniqueID/UniqueID.ts
```diff
@@ -309,8 +314,7 @@ const UniqueID = Extension.create({
               });
               return Fragment.from(list);
             };
-            // reset check
-            transformPasted = false;
```
YousefED, line 313: are you sure this can be removed? If I see it correctly, `parseFromClipboard` calls `transformPasted`.

This means the behavior of `parseFromClipboard` will now change depending on the value of `transformPasted` (which is determined by what was dropped previously)

matthewlipski: Yeah I think the only place where it might cause issues is in unit tests, since that's the only time we call `parseFromClipboard` ourselves. Otherwise, it'll get called only after any paste/drop event, where `transformPasted` is always set.

Unit tests pass though so I wouldn't worry about it.

YousefED: Does readd `transformPasted = false;` actually break anything?

If not, I think it's safer to readd it in case we would at some point in the future start calling anything that calls `transformPasted` (like `parseFromClipboard`) without taking UniqueID into account

PR #1459, opened 2025-02-24
feat: File insertion QoL

### LINKED ISSUE #1386: Drag and dropping an image creates an additional block
**Describe the bug**
When I drag and drop an image It creates a new block even if I try to drop it on a specific line.

**To Reproduce**
https://github.com/user-attachments/assets/4271149f-fcb1-493d-99f7-8d9952f10bb8

**Misc**
- Node version:
- Package manager:
- Browser: Firefox 133.0.3 on blocknotejs.org

matthewlipski:
Closes #1386

### DIFF
--- a/packages/core/src/api/clipboard/fromClipboard/handleFileInsertion.ts
+++ b/packages/core/src/api/clipboard/fromClipboard/handleFileInsertion.ts
@@ -1,4 +1,4 @@
-import { PartialBlock } from "../../../blocks/defaultBlocks.js";
+import { Block, PartialBlock } from "../../../blocks/defaultBlocks.js";
 import type { BlockNoteEditor } from "../../../editor/BlockNoteEditor";
 import {
   BlockSchema,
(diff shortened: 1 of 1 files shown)


YousefED (approved):

packages/core/src/api/clipboard/fromClipboard/handleFileInsertion.ts
```diff
@@ -143,11 +151,22 @@ export async function handleFileInsertion<
 
         const blockInfo = getBlockInfo(posInfo);
 
-        insertedBlockId = editor.insertBlocks(
-          [fileBlock],
-          blockInfo.bnBlock.node.attrs.id,
-          "after"
-        )[0].id;
+        if (
+          blockInfo.isBlockContainer &&
```
YousefED, line 155: might be nice to extract this somewhere as I think we use it in more than 1 place

PR #1463, opened 2025-02-25
fix: Mantine CSS variables element

matthewlipski:
Mantine theme CSS variables aren't being set properly at the moment, which means that an editor using the Mantine `BlockNoteView` may not have the CSS variables it needs when not within an existing `MantineProvider`. This is because `getRootElement` returns no element, but we need it to set `data-mantine-color-scheme` on the `.bn-mantine` element. Since it's a child of the `MantineProvider` and can't be found on render, we have to set `data-mantine-color-scheme` manually through `BlockNoteViewRaw`.

We pretty much never use Mantine theme variables as we use the BlockNote theme variables instead, but the `Badge` components from the comments PR does use them. Maybe we also want to change this to use strictly BlockNote theme variables? (ofc in the comments PR, not this one)

### DIFF
--- a/packages/mantine/src/BlockNoteView.tsx
+++ b/packages/mantine/src/BlockNoteView.tsx
@@ -12,7 +12,7 @@ import {
   usePrefersColorScheme,
 } from "@blocknote/react";
 import { MantineProvider } from "@mantine/core";
-import { useCallback } from "react";
+import { useCallback, useRef } from "react";
 import {
   applyBlockNoteCSSVariablesFromTheme,
   removeBlockNoteCSSVariables,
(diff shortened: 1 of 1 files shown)


packages/mantine/src/BlockNoteView.tsx
```diff
@@ -74,24 +74,28 @@ export const BlockNoteView = <
...
-        />
-      </MantineProvider>
+      <div className={"bn-mantine-css-variables"} ref={mantineCssVariablesRef}>
+        <MantineProvider
+          theme={mantineTheme}
+          cssVariablesSelector=".bn-mantine-css-variables"
+          // This gets the element to set `data-mantine-color-scheme` on. Since we
+          // don't need this attribute (we use our own theming API), we return
+          // undefined here.
+          getRootElement={() => mantineCssVariablesRef.current || undefined}>
```
nperez0111, line 90: Little unclear, the comment says it isn't needed, but we still are giving it the ref?

packages/mantine/src/BlockNoteView.tsx
```diff
@@ -74,24 +74,28 @@ export const BlockNoteView = <
...
-          className={mergeCSSClasses("bn-mantine", className || "")}
-          theme={typeof theme === "object" ? undefined : theme}
-          {...rest}
-          ref={ref}
-        />
-      </MantineProvider>
+      <div className={"bn-mantine-css-variables"} ref={mantineCssVariablesRef}>
+        <MantineProvider
+          theme={mantineTheme}
+          cssVariablesSelector=".bn-mantine-css-variables"
```
nperez0111, line 86: From what I understand of their docs, this is just about selector their stylesheet would reference the styles for. So, I don't think that this actually needs the element to exist, as it is only being applied through CSS.

Am I getting this wrong?

packages/mantine/src/BlockNoteView.tsx
nperez0111: Ah I see now, this is the change you wanna make, well the comment should be updated too.

But, am I crazy that all this is doing is setting that `data-mantine-color-scheme` can't we just set that attribute ourselves? Or maybe we don't know the color scheme?

Reason being that React normally doesn't like elements being "touched" by plain JS and on a re-render will just remove the extra attribute as though it were never there. Maybe that is the root cause of this issue?

But, IDK. I've never used Mantine nor understand the internals here

matthewlipski: Ah sorry, that's my bad I missed changing the comment.

I think when I wrote this comment originally I misunderstood what `cssVariablesSelector` and `getRootElement` were for. To clarify, `cssVariablesSelector` is for additional variables from the Mantine theme, and `getRootElement` is the base set of variables to make Mantine components work.

Either way, seems like it's more than just `data-mantine-color-scheme` being set - also CSS variables. Sorry again for the misleading comment.

matthewlipski:
Ok so I was also getting confused and not exactly understanding what's going on - @nperez0111 you're right in your comments, I added `data-mantine-color-scheme` manually to `BlockNoteViewRaw` and that worked just fine, so now the theme variables get added to `.bn-mantine` correctly and no extra div is needed👍

nperez0111 (approved):
👍 this looks good now!

PR #1472, opened 2025-02-28
build: parallelize across browsers

nperez0111:


YousefED (commented):
Cool! Looks good to me :) What's the part that runs the test in docker?

Note that I don't have any opinion on nx vs. x. vs. y (except that lerna is deprecated and we should get rid of it). I'd not invest too much further in nx unless you're confident it's a solid choice going forward compared to other solutions

nperez0111:
> Cool! Looks good to me :) What's the part that runs the test in docker?

We already merged that onto main last week: https://github.com/TypeCellOS/BlockNote/blob/main/.github/workflows/build.yml#L57

> Note that I don't have any opinion on nx vs. x. vs. y (except that lerna is deprecated and we should get rid of it). I'd not invest too much further in nx unless you're confident it's a solid choice going forward compared to other solutions

I've used NX and turborepo before, I think NX is better but, I agree that we should leave it alone for now and make a decision on it later. For now, was a very easy integration to just enable it and keep everything run with lerna

PR #1475, opened 2025-03-03
feat: support new table properties in PDF export

nperez0111:
This adds support for table cell text & background color, alignment, and header cells to the PDF export of tables.

One thing this does not support at the moment is rowspan & colspan tables, with the current approach it looks like it would take a good amount of work to get this to export properly

### DIFF
--- a/packages/xl-pdf-exporter/src/pdf/util/table/Table.tsx
+++ b/packages/xl-pdf-exporter/src/pdf/util/table/Table.tsx
@@ -30,6 +30,9 @@ const styles = StyleSheet.create({
     wordWrap: "break-word",
     whiteSpace: "pre-wrap",
   },
+  headerCell: {
+    fontWeight: "bold",
+  },
   bottomCell: {
     borderBottom: "1px solid #ddd",
   },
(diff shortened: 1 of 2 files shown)


packages/xl-docx-exporter/src/docx/util/Table.tsx
nperez0111, line 1: table headers are not supported atm because they require modifying the run of the paragraph to make the cell's content bold.

Table paragraph color seems to not be working at the moment either, maybe to do with the inheritance of paragraph within a paragraph's color?

packages/xl-pdf-exporter/src/pdf/util/table/Table.tsx
nperez0111, line 1: rowspan & colspan change up how this layouting would work, I think we'd need to rewrite this completely to allow them to be supported

YousefED: pdfkit just merged this last week: https://github.com/foliojs/pdfkit/pull/1577 which hopefully should solve a lot of edge cases related to tables.

I'm checking with the author of react-pdf: https://github.com/diegomura/react-pdf/issues/3110

packages/xl-docx-exporter/src/docx/util/Table.tsx
```diff
@@ -12,33 +19,79 @@ export const Table = (
...
+                    : cell.props.textAlignment === "right"
+                    ? "right"
+                    : cell.props.textAlignment === "justify"
+                    ? "distribute"
+                    : (() => {
+                        throw new UnreachableCaseError(
+                          cell.props.textAlignment
+                        );
+                      })(),
+                run:
```
YousefED, line 80: - you could add "bold" here as well for headers

My guess why it's not working might be because the children of the paragraph overwrite all the styles. You could break / log here to confirm: https://github.com/TypeCellOS/BlockNote/blob/11c492146fbb7819586e994d046a7f70a4330179/packages/xl-docx-exporter/src/docx/docxExporter.ts#L98

If that's not the case, then maybe the styles don't inherit the nested runs and we'd need to find another way.

lmk if you need help with digging in further

PR #1481, opened 2025-03-04
fix: Comments feedback & flaky test

matthewlipski:
This PR addresses the remaining feedback from #1376

Also fixes an e2e test that was temporarily bypassed because it was flaky


tests/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "build": "tsc",
     "lint": "eslint src --max-warnings 0",
-    "playwright": "npx playwright test",
+    "playwright": "npx playwright test --ui",
```
nperez0111, line 8: Probably not intentional

nperez0111 (approved):
Otherwise gtg

tests/package.json
matthewlipski: It is actually! Since we really only use this command to check why an individual test is failing, I figured we may as well add the `--ui` flag to make this easier. I would previously set `--headed` instead, but `--ui` is much nicer.

nperez0111: Alright, well then the playwright:ui task is redudant then

PR #1484, opened 2025-03-04
fix: Removed `@emoji-mart/react` dependency

matthewlipski:
`@emoji-mart/react` is not compatible with React 19 and is only one file, so this PR changes it from a dependency to just a copy of the file

### DIFF
--- a/packages/react/src/components/Comments/EmojiMartPicker.tsx
+++ b/packages/react/src/components/Comments/EmojiMartPicker.tsx
@@ -1,0 +1,23 @@
+// From https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart-react/react.tsx
+import React, { useEffect, useRef } from "react";
+import { Picker } from "emoji-mart";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export default function EmojiPicker(props: any) {
+  const ref = useRef(null);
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Comments/EmojiMartPicker.tsx
```diff
@@ -0,0 +1,23 @@
+// From https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart-react/react.tsx
+import React, { useEffect, useRef } from "react";
+import { Picker } from "emoji-mart";
```
nperez0111, line 3: Does this include the whole bundle of emoji mart data? Or is that separate?

Do we need this as a dynamic import?

matthewlipski: Don't think so, iirc from doing manual testing the emojis are fetched on initial render. Diving into the source code seems to support this:

https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart/src/components/Picker/Picker.tsx#L112

The `_init` gets called in the `Picker` component's `componentWillReceiveProps` function, so either way it's not until the first render.

nperez0111: Alright, cool, wasn't aware of how it works

PR #1485, opened 2025-03-04
fix: forward refs to the final component

nperez0111:
There was a console error being logged because we were passing a ref without using forward ref

image


matthewlipski (approved):

YousefED:
Nice catch. React 19 actually gives warnings with `forwardRef` btw (because it's deprecated). I'm not sure what the best practice is to make it work well on both 18+19

PR #1486, opened 2025-03-04
fix: Link toolbar update error

### LINKED ISSUE #1479: Move up multiple blocks - content disappear
**Describe the bug**
When we move up multiple blocks, sometime the content disappears. 
This bug seems to happen when we have multiple links in the content.

**To Reproduce**

https://github.com/user-attachments/assets/031b047c-27d7-4ac6-b6d6-fa952663fe69

Seems to happen around here:
https://github.com/TypeCellOS/BlockNote/blob/add526cd67c5e0b457e533beece73fa675d66e34/packages/core/src/extensions/SideMenu/MultipleNodeSelection.ts#L27-L33

### Logs
```
28886.03723a0b571b25df.js:4 Uncaught TypeError: Cannot read properties of undefined (reading 'nodeSize')
    at i.nodesBetween (28886.03723a0b571b25df.js:4:82183)
    at i.textBetween (28886.03723a0b571b25df.js:4:82418)
    at C.textBetween (28886.03723a0b571b25df.js:4:94573)
    at oJ.update (94018.ecfc4dfe86ae539e.js:1:201052)
    at tK.updatePluginViews (849baa33.6a24fe58beb16197.js:1:84466)
    at tK.updateStateInner (849baa33.6a24fe58beb16197.js:1:83505)
    at tK.updateState (849baa33.6a24fe58beb16197.js:1:80653)
    at iA.dispatchTransaction (456144eb.04539d1d8e6e2cc8.js:75:5453)
    at tK.dispatch (849baa33.6a24fe58beb16197.js:1:92273)
    at iA.dispatch (94018.ecfc4dfe86ae539e.js:1:243760)
nodesBetween	@	28886.03723a0b571b25df.js:4
textBetween	@	28886.03723a0b571b25df.js:4
textBetween	@	28886.03723a0b571b25df.js:4
update	@	94018.ecfc4dfe86ae539e.js:1
updatePluginViews	@	849baa33.6a24fe58beb16197.js:1
updateStateInner	@	849baa33.6a24fe58beb16197.js:1
updateState	@	849baa33.6a24fe58beb16197.js:1
dispatchTransaction	@	456144eb.04539d1d8e6e2cc8.js:75
dispatch	@	849baa33.6a24fe58beb16197.js:1
dispatch	@	94018.ecfc4dfe86ae539e.js:1
(anonymous)	@	94018.ecfc4dfe86ae539e.js:1
ou	@	94018.ecfc4dfe86ae539e.js:1
removeBlocks	@	94018.ecfc4dfe86ae539e.js:1
os	@	94018.ecfc4dfe86ae539e.js:1
(anonymous)	@	94018.ecfc4dfe86ae539e.js:1
moveBlocksDown	@	94018.ecfc4dfe86ae539e.js:1
Shift-Mod-ArrowDown	@	94018.ecfc4dfe86ae539e.js:1
(anonymous)	@	456144eb.04539d1d8e6e2cc8.js:1
(anonymous)	@	28886.03723a0b571b25df.js:4
(anonymous)	@	849baa33.6a24fe58beb16197.js:1
someProp	@	849baa33.6a24fe58beb16197.js:1
e0.keydown	@	849baa33.6a24fe58beb16197.js:1
e.dom.addEventListener.e.input.eventHandlers.<computed>	@	849baa33.6a24fe58beb16197.js:1
```

**Misc**
- Browser: Chrome Version 133.0.6943.98 (Official Build) (64-bit)

matthewlipski:
This PR fixes an bug with the link toolbar, where `mouseHoveredLinkMark` may be defined even when the mouse cursor isn't hovering any links. This is because the plugin would check if a link is currently hovered by the keyboard or mouse on every state update. State updates may cause the link under the mouse cursor move or be removed outright (such as moving blocks up & down, where it's briefly removed when replacing blocks). However, as no `mouseover` event is fired, `mouseHoveredLinkMark` doesn't get cleared and so an error is thrown if the link doesn't exist.

Therefore, this PR changes the behaviour so that only `keyboardHoveredLinkMark` is considered when updating the link toolbar, unless the update was explicitly called by the `mouseover` handler.

Closes #1479

### DIFF
--- a/packages/core/src/extensions/LinkToolbar/LinkToolbarPlugin.ts
+++ b/packages/core/src/extensions/LinkToolbar/LinkToolbarPlugin.ts
@@ -52,7 +52,7 @@ class LinkToolbarView implements PluginView {
 
     this.startMenuUpdateTimer = () => {
       this.menuUpdateTimer = setTimeout(() => {
-        this.update(this.pmView);
+        this.update(this.pmView, undefined, true);
       }, 250);
     };
 
@@ -190,7 +190,7 @@ class LinkToolbarView implements PluginView {
     }
   }
 
-  update(view: EditorView, oldState?: EditorState) {
+  update(view: EditorView, oldState?: EditorState, fromMouseOver = false) {
     const { state } = view;
 
     const isSame =
@@ -235,7 +235,7 @@ class LinkToolbarView implements PluginView {
       }
     }
 
-    if (this.mouseHoveredLinkMark) {
+    if (this.mouseHoveredLinkMark && fromMouseOver) {
       this.linkMark = this.mouseHoveredLinkMark;
       this.linkMarkRange = this.mouseHoveredLinkMarkRange;
     }


packages/core/src/extensions/LinkToolbar/LinkToolbarPlugin.ts
```diff
@@ -190,7 +190,7 @@ class LinkToolbarView implements PluginView {
     }
   }
 
-  update(view: EditorView, oldState?: EditorState) {
+  update(view: EditorView, oldState?: EditorState, fromMouseOver = false) {
```
nperez0111, line 193: Am I crazy to think that no one else is calling this update method?

matthewlipski: Ah this is part of the `PluginView` spec from ProseMirror, it automatically gets called on state updates (https://prosemirror.net/docs/ref/#state.PluginView). It's just harder to see that since in our case the view does a lot of stuff.

matthewlipski: Also I guess normally you'd pass in a plain object, so the fact that we're using a class also obfuscates things

nperez0111: Ohh ok. Eh, this definitely is a hack. But, I think it is okay for now

PR #1490, opened 2025-03-04
feat: markdown pasting & custom paste handlers

nperez0111:
This implements markdown pasting, and custom paste handlers.

Adds

 - `editor.pasteText` which will paste text content into the editor
 - `editor.pasteHTML` which will convert HTML into BlockNote HTML and then paste that into the editor
 - `editor.pasteMarkdown` which will convert the markdown into BlockNote HTML and then paste that into the editor
 - `editorOptions.pasteHandler` for allowing the developer the ability to customize how content is pasted into the editor while still allowing the default paste behavior.


packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -61,15 +64,37 @@ export const createPasteFromClipboardExtension = <
...
+                  return true;
+                }
+
                 if (format === "text/html") {
+                  if (editor.settings.pasteBehavior === "prefer-markdown") {
+                    // Use plain text instead of HTML if it looks like Markdown
+                    const plainText =
+                      event.clipboardData!.getData("text/plain");
+
+                    if (is(plainText)) {
```
YousefED, line 80: better name for `is` 😅?

packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -61,15 +64,37 @@ export const createPasteFromClipboardExtension = <
...
                   const htmlNode = nestedListsToBlockNoteStructure(data.trim());
                   data = htmlNode.innerHTML;
                   view.pasteHTML(data);
                   return true;
                 }
 
-                view.pasteText(data);
-
+                // Convert Markdown to HTML first, then paste as HTML
+                markdownToHTML(data).then((html) => {
```
YousefED, line 95: should this be behind a flag as well? (paste as text vs markdown)

packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -61,15 +64,37 @@ export const createPasteFromClipboardExtension = <
...
+
+                    if (is(plainText)) {
+                      // Convert Markdown to HTML first, then paste as HTML
+                      markdownToHTML(plainText).then((html) => {
+                        view.pasteHTML(html);
+                      });
+                      return true;
+                    }
+                  }
                   const htmlNode = nestedListsToBlockNoteStructure(data.trim());
```
YousefED, line 88: Might be worth checking whether we need this for the other cases as well (`nestedListsToBlockNoteStructure`)

packages/core/src/api/parsers/markdown/is.ts
```diff
@@ -0,0 +1,60 @@
+// Headings H1-H6.
```
YousefED, line 1: curious how many false positives we get. I didn't review the regexpes obviously.

Figured this might be useful: https://chatgpt.com/share/67d12910-3d78-8009-8bc0-ddd23dd7cba9

YousefED (commented):
Great stuff!

What do you think about making this pluggable? i.e.: you could register a paste handler / preprocessor, and our "default preprocessor" changes the mimetype to `text/markdown` if we think it's a better fit.


(just realized; we actually also have a client requesting `Event listeners for "paste" events so we can do our own manipulation before inserting into the document`)

docs/pages/docs/editor-basics/setup.mdx
```diff
@@ -66,6 +76,8 @@ The hook takes two optional parameters:
 
 `initialContent:` The content that should be in the editor when it's created, represented as an array of [Partial Blocks](/docs/manipulating-blocks#partial-blocks).
 
+`pasteHandler`: A function that can be used to override the default paste behavior. See [Custom Paste Behavior](/docs/advanced/custom-paste-behavior) for more.
```
nperez0111, line 79: Want to add a demo

docs/pages/docs/editor-basics/setup.mdx
```diff
@@ -33,6 +33,16 @@ type BlockNoteEditorOptions = {
...
   }) => Plugin;
   initialContent?: PartialBlock[];
+  pasteHandler?: (context: {
+    view: EditorView;
+    event: ClipboardEvent;
+    editor: BlockNoteEditor;
+    defaultPasteHandler: (context: {
+      pasteBehavior?: "prefer-markdown" | "prefer-html";
+    }) => boolean | undefined;
+    convertHtmlToBlockNoteHtml: (html: string) => string;
```
YousefED, line 43: API feedback:

- what about exposing `pasteHTML` `pasteMarkdown` and `pasteText` on the editor instead of the conversion functions? this way consumers also don't need to call into prosemirror `view.paste...` functions.
- I'm not sure I like passing the `defaultPasteHandler` vs exporting that function directly (still in doubt)

packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -8,9 +8,101 @@ import {
...
+
+  if (format === "text/markdown") {
+    convertMarkdownToBlockNoteHtml(data).then((html) => {
+      view.pasteHTML(html);
+    });
+    return true;
+  }
+
+  if (format === "text/html") {
+    if (pasteBehavior === "prefer-markdown") {
```
YousefED, line 78: code style feedback:

instead of having these checks spread here and below (L95), I feel like it would be cleaner to check the flag earlier (or even in a separate "middleware" function) that would just set the format to "text/markdown" and then let the default behavior paste that. makes sense / wdyt? (I didn't completely think it out)

packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -8,9 +8,101 @@ import {
...
+        });
+        return true;
+      }
+    }
+
+    view.pasteHTML(convertHtmlToBlockNoteHtml(data));
+    return true;
+  }
+
+  if (pasteBehavior === "prefer-markdown" && isMarkdown(data)) {
```
YousefED, line 95: I'm not sure if this single flag is what we want. We now always turn on "detection" and for both markdown and HTML. is that desirable? 

Other options I can think of what the user might want:
- always paste plain text as markdown (i.e.: paste markdown instead of plain text at L103), regardless of detection
- Only run the detection for plain text, not for HTML (i.e.: always prefer regular HTML paste)?
- (Other combinations like only run the detection for HTML, not for plain text. Not sure if that makes sense)

Happy to think this through together. I don't think we need all options to be possible via different flags, especially since users can now add their own paste handlers. But I do think there should at least be a way to not rely on the auto-detection yet still paste text as markdown

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -212,6 +212,51 @@ export type BlockNoteEditorOptions<
...
+   * @returns The function should return `true` if the paste event was handled, otherwise it should return `false` if it should be canceled or `undefined` if it should be handled by another handler.
+   *
+   * @example
+   * ```ts
+   * pasteHandler: ({ defaultPasteHandler }) => {
+   *   return defaultPasteHandler({ pasteBehavior: "prefer-html" });
+   * }
+   * ```
+   */
+  pasteHandler?: (context: {
```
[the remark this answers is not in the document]
YousefED: I think current approach is ok. You could do something like `return true / false` to enable chaining but at this point I'm not sure it's worthwhile (don't see a lot of scenarios where that would be useful)

docs/pages/docs/editor-basics/setup.mdx
nperez0111: - I like exposing the methods, to keep it simple for users, I exposed pasteHTML, pasteText & pasteMarkdown methods
 - I think having too many things exposed on a top-level API make it hard for discoverability, this function would only be useful for pasting content, so it makes sense to expose on the handler you'd use it at.

YousefED:
@nperez0111 didn't review yet; but I just noticed (working on sth unrelated) that actually inline content markdown gets parsed. try pasting `xx \*\*hello world\*\*`. Did you know this / know where this is done? Tiptap feature?

nperez0111:
> @nperez0111 didn't review yet; but I just noticed (working on sth unrelated) that actually inline content markdown gets parsed. try pasting `xx \*\*hello world\*\*`. Did you know this / know where this is done? Tiptap feature?

https://github.com/ueberdosis/tiptap/blob/dd0a25f5b48867992dd68d3effd5c98ddafcab69/packages/extension-bold/src/bold.tsx#L134-L145

Bold already has a paste rule set up. As well as a number of other Tiptap extensions. What does not work (in blocknote) though would be things across paragraphs or code blocks for example

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1437,4 +1472,41 @@ export class BlockNoteEditor<
...
+  }
+
+  /**
+   * Paste HTML into the editor. Defaults to converting HTML to BlockNote HTML.
+   * @param html The HTML to paste.
+   * @param raw Whether to paste the HTML as is, or to convert it to BlockNote HTML.
+   */
+  public pasteHTML(html: string, raw = false) {
+    let htmlToPaste = html;
+    if (!raw) {
```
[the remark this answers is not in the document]
nperez0111: We use it here: https://github.com/TypeCellOS/BlockNote/blob/af0c5ffaf90d979b72b4697806b2402c1eb117ee/packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts#L55

docs/pages/docs/advanced/paste-handling.mdx
```diff
@@ -0,0 +1,59 @@
+---
```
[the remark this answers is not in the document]
nperez0111: I think it is in the sidebar?

examples/01-basic/13-custom-paste-handler/.bnexample.json
matthewlipski, line 1: I would move this example to the "interoperability" section tbh, especially since the docs for paste handling are in the "advanced" section, it's a bit weird to have the example in the "basic" section

examples/01-basic/13-custom-paste-handler/README.md
```diff
@@ -0,0 +1,9 @@
+# Custom Paste Handler
+
+In this example, we change the default paste handler to append some text to the pasted content when the content is plain text.
+
+**Try it out:** Use the buttons to copy some content to the clipboard and paste it in the editor to trigger our custom paste handler
```
matthewlipski, line 5: Missing period at end of sentence

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -212,6 +216,39 @@ export type BlockNoteEditorOptions<
...
+      /**
+       * Whether to auto-detect markdown in the clipboard and paste that as rich text.
+       * @default true
+       */
+      autoDetectMarkdown?: boolean;
+      /**
+       * Whether to interpret plain text as markdown and paste that as rich text.
+       * @default true
+       */
+      plainTextAsMarkdown?: boolean;
```
matthewlipski, line 248: Imo the difference between `autoDetectMarkdown` and `plainTextAsMarkdown` is not super clear. Don't they kind of do the same thing, except `autoDetectMarkdown` will prioritize `text/plain` as Markdown over `text/html`?

packages/core/src/api/clipboard/fromClipboard/acceptedMIMETypes.ts
```diff
@@ -1,6 +1,7 @@
 export const acceptedMIMETypes = [
   "vscode-editor-data",
   "blocknote/html",
+  "text/markdown",
```
matthewlipski, line 4: Out of curiosity, do you know any apps that use `text/markdown` for the clipboard? Since AFAIK it's not an official MIME type. I know Notion for example just puts Markdown in `text/plain`

nperez0111: It's got an RFC: https://www.rfc-editor.org/rfc/rfc7763

It is trivial for us to implement, and makes it explicit rather than having to parse plain text.

I don't have example apps that use this. And, it'd be kind of annoying to check haha

examples/01-basic/13-custom-paste-handler/README.md
nperez0111, line 5:
```suggestion
**Try it out:** Use the buttons to copy some content to the clipboard and paste it in the editor to trigger our custom paste handler.
```

packages/core/src/editor/BlockNoteEditor.ts
matthewlipski: I think I would change the options slightly to this:

`plainTextAsMarkdown`: Whether to parse `text/plain` content from the clipboard as Markdown content.

`prioritizeMarkdownOverHTML`: Whether to prioritize Markdown content in `text/plain` over `text/html` when pasting from the clipboard. `plainTextAsMarkdown` must be set to `true` for this to have any effect.

Thoughts on this?

PR #1491, opened 2025-03-05
fix: import emoji-mart dynamically like (#950)

nperez0111:
This resolves the emoji-mart import dynamically just like what was done in #950 to resolve the same issue.

### DIFF
--- a/packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
+++ b/packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
@@ -10,11 +10,19 @@ import {
 import { DefaultGridSuggestionItem } from "./DefaultGridSuggestionItem.js";
 
 // Temporary fix for https://github.com/missive/emoji-mart/pull/929
-let emojiMart: typeof import("emoji-mart") | undefined;
-let emojiData: EmojiMartData | undefined;
+let emojiLoadingPromise:
+  | Promise<{
+      emojiMart: typeof import("emoji-mart");
+      emojiData: EmojiMartData;
+    }>
+  | undefined;
 
 async function loadEmojiMart() {
-  if (!emojiMart || !emojiData) {
+  if (emojiLoadingPromise) {
+    return emojiLoadingPromise;
+  }
+
+  emojiLoadingPromise = (async () => {
     // load dynamically because emoji-mart doesn't specify type: module and breaks in nodejs
     const [emojiMartModule, emojiDataModule] = await Promise.all([
       import("emoji-mart"),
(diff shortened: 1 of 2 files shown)


YousefED (commented):
nice. I think we might want to also copy or reuse the pattern for loading the data. My guess is that it'll now it from a cdn, better to keep it isolated so we limit external dependencies

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -9,13 +9,32 @@ import {
...
-  | Promise<{
-      default: EmojiMartData;
-    }>
-  | undefined;
-
+// Temporary fix for https://github.com/missive/emoji-mart/pull/929
 let emojiMart: typeof import("emoji-mart") | undefined;
+let emojiData: EmojiMartData | undefined;
+
+async function loadEmojiMart() {
```
nperez0111, line 16: Decided to make both of these the same, and also made them Node.js safe

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -9,13 +9,32 @@ import {
...
+    ]);
+
+    emojiMart =
+      "default" in emojiMartModule ? emojiMartModule.default : emojiMartModule;
+    emojiData =
+      "default" in emojiDataModule
+        ? (emojiDataModule.default as EmojiMartData)
+        : (emojiDataModule as EmojiMartData);
+
+    emojiMart.init({ data: emojiData });
```
YousefED, line 33: I think this needs an `await`

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
```diff
@@ -9,13 +9,32 @@ import {
 } from "../../schema/index.js";
 import { DefaultGridSuggestionItem } from "./DefaultGridSuggestionItem.js";
 
-let data:
-  | Promise<{
-      default: EmojiMartData;
-    }>
-  | undefined;
-
+// Temporary fix for https://github.com/missive/emoji-mart/pull/929
```
YousefED, line 12: fyi, the reason data was global and could be a promise would be to prevent duplicate loads when `loadEmojiMart` is called twice in a row.

Didn't help readability and it might be a premature optimization so I'm ok with the new version as long as we don't see this causing any issues

nperez0111: Ah, I can de-dupe easily. I'll pull that in.

packages/core/src/extensions/SuggestionMenu/getDefaultEmojiPickerItems.ts
nperez0111: Hm, wasn't clear that it was async in the docs https://github.com/missive/emoji-mart?tab=readme-ov-file#-emoji-component , but won't hurt to await

PR #1492, opened 2025-03-05
fix: scope cursor caret CSS to be more specific

nperez0111:
To help in our liveblocks integration, we are scoping the styles of our cursors to be more specific (based on the `.collaboration-cursor__base` classname) so that we do not conflict with the styles they have setup to style the default tiptap cursors

So, for their CSS to work properly, they will need to provide their own cursor rendering logic to match the default Tiptap one


nperez0111:
To only prefix the classname? Or prefixing in addition to this?

Issue #1468, opened 2025-02-28
Exporting doc in official examples cannot run properly


qsm771563597:
https://github.com/TypeCellOS/BlockNote/tree/main/examples/05-interoperability/06-converting-blocks-to-docx

1.npm install  
2.npm run dev
3.click   doc export button 
4. open the console interface ( An error occurred)

![Image](https://github.com/user-attachments/assets/dd4bbe7c-0a6e-4246-a56b-1063b6d0480f)

qsm771563597:
![Image](https://github.com/user-attachments/assets/31a1d464-838b-43e4-a5ea-96e2395bca8a)
add "const Buffer = (await import("buffer")).Buffer;  globalThis.Buffer = Buffer;"  before  use   blob

YousefED:
I think this has been fixed last week

PR #1493, opened 2025-03-05
fix: pull from the default export which is available in both ESM and CJS (#1468)

nperez0111:
This resolves a bug in CJS module resolution where the buffer package was incorrectly accessing a named export instead of the default export which is available both in ESM & CJS builds


YousefED (commented):
nice

PR #1496, opened 2025-03-05
feat: add sentry to website for error detection

nperez0111:
This adds Sentry to our website to help with tracking errors that may occur on our docs website.

We have a few decisions to make for this:
 - [ ] Should we self-host this? Sentry does allow self-hosting their error aggregator.
 - [ ] If not, we will need to make an official Sentry SaaS acc

 We can tweak the error capturing over time, to reasonable collection rates. But, for now it is good to get the full firehose.

 I double-checked and Sentry does not track users via the use of cookies, so no need for a cookie-banner, and should be GDPR compliant.


YousefED:
invite sent

docs/pages/about.mdx
```diff
@@ -3,6 +3,10 @@ import CTAButton from "../components/pages/landing/shared/CTAButton";
 
 # About BlockNote
 
+<button onClick={() => { throw new Error("test") }}>
```
YousefED: don't forget ;)

YousefED (commented):
do / cab we add version / release metadata? i.e.: https://docs.sentry.io/platforms/javascript/configuration/releases/

PR #1499, opened 2025-03-05
fix: upgrade y-prosemirror and remove clientID hack

nperez0111:
Applies fix #1497 as a hotfix onto v0.23.2


nperez0111:
Released 0.23.2-hotfix.0 to resolve this

PR #1503, opened 2025-03-06
fix: resolve unneeded error being thrown

nperez0111:
resolves: https://blocknote-js.sentry.io/issues/32147104


YousefED (approved):
nice, do you also know why this could happen? (i.e.: could it indicate there's a different bug?)

nperez0111:
> nice, do you also know why this could happen? (i.e.: could it indicate there's a different bug or are there expected scenarios?)

The drag over handler is listening to the root element, so you are not guaranteed to be hovering over a table: https://github.com/TypeCellOS/BlockNote/blob/aed9fb5044a0da73276c322b685dd838193247d0/packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts#L186-L189 in which case, the previous state should still be maintained for the drop to actually switch the rows

PR #1505, opened 2025-03-06
fix: Backspace in empty block deletes previous block

### LINKED ISSUE #605: Remove blank line under a 'none' content block will remove 'none' block together
**Describe the bug**
When I need to remove a blank line under a 'none' type content, it never works, remove the 'none' content together.
I think remove a blank line is reasonable action.

```
const blockConfig = {
  type: "image" as const,
  propSchema: imagePropSchema,
  content: "none",
} satisfies CustomBlockConfig;
```
I tried, all `content: "none",` will cause this bug.

**To Reproduce**
1. Insert a 'none' content block, like 'image'
2. Insert a blank line under the 'none' content block
3. Press Backspace in the blank line

**Misc**
- Node version:
- Package manager:
- Browser: 

### LINKED ISSUE #1478: Improve delete key behavior near image blocks

I experience unexpected behavior when working with images in BlockNote. When editing a paragraph adjacent to an image block and pressing the Delete key at the start of the subsequent line, the image gets deleted rather than the cursor moving into the image block.

Examples of current behavior vs expected behavior:

Expected (Google Docs behavior):
https://github.com/user-attachments/assets/1e1f97f8-2438-4d63-8b23-36a991ea8711
Cursor moves into image block when deleting at line start

Current (BlockNote behavior):
https://github.com/user-attachments/assets/f0b24e02-7ff8-479e-a75d-e05328059d20
Image gets deleted when using Delete key at line start

The cursor should move into the image block when pressing Delete at the beginning of the subsequent line, rather than deleting the image. This should mirror the behavior seen in Google Docs and other mainstream text editors.

Martinrsts:
This suggestion is for resolving some weird behaviour with the backspace like: Fixes #605 Fixes #1478 .
Also, when a block is empty and backspace is clicked, it makes more sense to remove the block as seen in here:

Old behaviour:
https://github.com/user-attachments/assets/85669900-267a-4f1f-8642-57585d2a77dd

New behaviour:
https://github.com/user-attachments/assets/9f7a8eff-83e0-4e06-af58-cdc275c162d6

### DIFF
--- a/packages/core/src/extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
+++ b/packages/core/src/extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
@@ -292,6 +292,24 @@ export const KeyboardShortcutsExtension = Extension.create<{
                 throw new Error(`todo`);
               }
 
+              const currentBlockNotTableAndNoContent =
+                blockInfo.blockContent.node.type.spec.content === "" ||
+                (blockInfo.blockContent.node.type.spec.content === "inline*" &&
+                  blockInfo.blockContent.node.childCount === 0);
+
+              if (currentBlockNotTableAndNoContent) {
+                const pos = state.tr.doc.resolve(
+                  blockInfo.bnBlock.beforePos - 2
+                );
+                state.tr.setSelection(new TextSelection(pos));
+                return chain()
+                  .deleteRange({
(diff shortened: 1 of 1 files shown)


matthewlipski:
Thanks for this! The change is nice, though I ended up re-writing it to make it clearer what's happening and sort out some edge cases.

matthewlipski (approved):

PR #1515, opened 2025-03-10
docs: suggestions architecture WIP

nperez0111:
This is an Architecture Decision Record (ADR) that describes how & why we are choosing to implement suggestions and versioning for BlockNote.


adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,172 @@
...
+    - The content that was changed (i.e. additions or deletions)
+    - Updates to content (i.e. changing an attribute)
+    - Which user identifiers modified the content
+    - Ideally, a timestamp of when the content was modified
+    - Changes should only show content that was present in both documents (i.e. if some content was added and subsequently deleted in some version between the two snapshots, because it is not present in the final document of either snapshot, it should not show)
+  - These changes can be visualized in a number of ways: as a unified diff, split diff, or even stacked diff. _See [Appendix](#displaying-changes) for more information_
+
+**Prior Art:**
+
+- Google Docs [Video here]
```
YousefED, line 27: maybe also link to ink and switch projects to indicate how big the problem / solution space can be:
- https://www.inkandswitch.com/upwelling/
- https://www.inkandswitch.com/patchwork/notebook/
-

adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Summary
+
+This document was purposefully written from the perspective that Y.js is one of many document storage and synchronization mechanisms. And, we would like to keep our editor de-coupled from these specific implementation details as much as possible to encourage code re-use between different implementations.
+
+## Appendix
+
+This is a collection of scattered thoughts on the _diff_-erent aspects of diffing  (sorry, had to), and their complications.
+
+### Diffing HTML
```
YousefED, line 99: This section implies there's something related to "HTML" that makes it difficult, I don't think this is the case. I would rename this section "Keeping user intent / operations intact" and describe it from a user instead of a technical perspective if that makes sense

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
```diff
@@ -0,0 +1,32 @@
...
+## Changes as _attributes_
+
+Kevin has already implemented a preliminary version of the `y-diff` between two documents. Specifically, by using `ychange` attributes to add metadata onto elements to indicate whether they were modified between the two versions. This approach works for simple documents at the moment, with some limitations around:
+
+- Certain features are missing (due to this being preliminary work) like
+  - XYZ
+- Controlling the display of the diff:
+  - y-prosemirror is controlling how the diffs are rendered so doing something like a stacked diff would not be possible
+  - Trying to simplify changes to a more human understandable version as described in the HTML diffing section would not be possible, since y-prosemirror directly writes the changes it finds to elements
+- Inability to edit content while displaying diffs
```
YousefED, line 15: maybe add this to requirements :)

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
```diff
@@ -0,0 +1,32 @@
+
+# Diff two Y.js snapshots
+
+Y.js keeps track of metadata around what changes have been made to a document in order to efficiently synchronize changes. So, to compare a document with it's previous version, you can leverage this metadata by taking a snapshot of the full history of a document, and compare the insertions and deletions that have been made between the two snapshots. Because they share history, you can figure out what has been modified from _this point_ in history compared to _that point_ in history. For clarity, let's call this a `y-diff` for now to make sure it is different than a full content diff.
+
```
YousefED, line 5: basically, everything below here is about how to display the changes right (i.e.: how to bring them to the editor-level)? that wasn't immediately clear to me

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
```diff
@@ -0,0 +1,32 @@
...
+## Changes as _marks_
+
+Another approach, would be to model what is currently being represented as attributes, but instead use marks similar to [`@handlewithcare/prosemirror-suggest-changes`'s schema](https://github.com/handlewithcarecollective/prosemirror-suggest-changes?tab=readme-ov-file#schema) where the content changes are represented as marks. This would allow flexibility in display (because of support for [markviews](https://prosemirror.net/docs/ref/#view.MarkView) which can decouple content from presentation).
+
+Though, I question whether it _should_ be the responsibility of `y-prosemirror` to be both finding the difference and modifying the content of the document to display those differences. It feels like `y-prosemirror` should be scoped to the integration of yjs & prosemirror, and this is somewhat out of scope for it. I would be okay with `y-prosemirror` providing utilities to find the `y-diff` between documents in the y-prosemirror format, but the responsibility of how that integrates with prosemirror should be the responsibility of the caller of that function, rather than, built into the `y-sync` plugin.
```
YousefED, line 22: I think this paragraph is key

adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+### How should Y.js snapshots be compared
+
+Y.js already has versioning in the form of snapshots of the Y.js document, at a point in time. In Y.js, the snapshots have the nice property of having some form of a history associated to those edits. To figure out what changed, you could calculate a diff between the two histories of the document. Doing a comparison based on the document content alone, would not meet our requirement of being able to display who made the change and when. Luckily, in Y.js, there is additional metadata that we can extract from Y.js which can tell us which `clientID` made the change (`clientID` being an ID a user assumes when making edits to a document). More on how to calculate that diff here: [Diff two Y.js Snapshots](./diff-yjs-snapshots.md)
+
+This would be the most straightforward way to capture changes across versions of a document. The question then becomes how to display those changes within a prosemirror-based editor, which sort of depends on the approach taken for suggestions.
+
+### How should suggestions be stored
+
+There are a few different mental models we can use to represent or derive these changes:
```
YousefED, line 71: can you explain a little bit here what "these changes" are? what makes suggestions different, etc?

adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+```html
+<p>Some text</p>
+<p>More text</p>
+```
+
+Let's say that the user deletes the range of text from "text" and collapses the two paragraphs into 1 paragraph, resulting in this document:
+
+```html
+<p>Some More text<p>
```
dmonad, line 117: You probably meant `<p>Some text</p>`

adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Summary
+
+This document was purposefully written from the perspective that Y.js is one of many document storage and synchronization mechanisms. And, we would like to keep our editor de-coupled from these specific implementation details as much as possible to encourage code re-use between different implementations.
+
+## Appendix
+
+This is a collection of scattered thoughts on the _diff_-erent aspects of diffing  (sorry, had to), and their complications.
+
+### Keeping user intent, while diffing
```
YousefED, line 103: was thinking about another edge-case earlier today - writing it down here to share and so I don't forget :)

- when a user is in "suggestion" mode, deleting a word, and adding the same word again should probably not result in a deletion and insertion

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
```diff
@@ -0,0 +1,40 @@
...
+I won't go into detail about how to implement the `y-diff` between two documents, given that it is partially implemented in y-prosemirror, but will instead focus on how those changes end up being represented within the editor.
+
+We have a few different options:
+
+### Changes as _attributes_
+
+Kevin has already implemented a preliminary version of the `y-diff` between two documents. Specifically, by using `ychange` attributes to add metadata onto elements to indicate whether they were modified between the two versions. This approach works for simple documents at the moment, with some limitations around:
+
+- Certain features are missing (due to this being preliminary work) like
+  - XYZ
```
dmonad, line 17: I would be super interested in feedback here. I remember that the approach - albeit buggy in certain scenarios - is solid.

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
```diff
@@ -0,0 +1,40 @@
...
+- Controlling the display of the diff:
+  - y-prosemirror is controlling how the diffs are rendered so doing something like a stacked diff would not be possible
+  - Trying to simplify changes to a more human understandable version as described in the HTML diffing section would not be possible, since y-prosemirror directly writes the changes it finds to elements
+- Inability to edit content while displaying diffs
```
dmonad, line 21: This is not an inherent limitation

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
dmonad: How would this diff be represented?

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Representing these changes in the editor
+
+I won't go into detail about how to implement the `y-diff` between two documents, given that it is partially implemented in y-prosemirror, but will instead focus on how those changes end up being represented within the editor.
+
+We have a few different options:
+
+### Changes as _attributes_
+
+Kevin has already implemented a preliminary version of the `y-diff` between two documents. Specifically, by using `ychange` attributes to add metadata onto elements to indicate whether they were modified between the two versions. This approach works for simple documents at the moment, with some limitations around:
```
dmonad, line 14: I feel that this is basically what you are looking for. You could compute a y-diff based on the ychange attributes. That might make sense, so you can compute a more optimized diff (e.g. filtering out non-edits like `delete x • insert x`) and to get more control over how the diff is rendered.

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
nperez0111: Good question. I think the Delta format is a good starting point, but would need additional metadata.

Or, we can go with something more specific to Prosemirror since this would be implemented for `y-prosemirror`.

We can talk about it 😄

adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+- A user can take _snapshots_ of the current document state, and store that in a separate data store
+- These snapshots can be viewed independently at any point in time later
+- These snapshots can be _compared_ with each other
+  - This allows the user to see what changed between the two snapshots
+  - Changes should show:
+    - The content that was changed (i.e. additions or deletions)
+    - Updates to content (i.e. changing an attribute)
+    - Which user identifiers modified the content
+    - Ideally, a timestamp of when the content was modified
```
nperez0111, line 21: Flesh out timestamp _means_ and at what granularity (typed first keystroke, or when snapshot was taken, or when user assumed a clientID)

nperez0111: Can be derived from snapshotted version?

adr/2025_03_10-suggestions/suggestions-as-patches.md
```diff
@@ -0,0 +1,23 @@
...
+
+- This stores only the minimal representation of a change
+- Since it is stored separately
+  - it is not possible to be "lost" due to changes in the editor content.
+  - it can be permission-ed separately
+
+### Cons
+
+- This stores the change in a separate data structure, which would need to be either maintained by mapping new positions through, or using relative positions.
+- Being in a different data structure presents risk for being invalidated by changes to the document (which may not have been intentional).  Imagine moving a paragraph that had a suggestion on it, it could result in the suggestion pointing into an element which no longer exists.
```
nperez0111, line 23: significantly different from versioning

adr/2025_03_10-suggestions/suggestions-as-patches.md
```diff
@@ -0,0 +1,23 @@
...
+## Implementation Details
+
+One easy implementation of this is to have a suggestion be a serialization of a [Prosemirror Transform](https://prosemirror.net/docs/ref/#transform). This transform can be [rebased](https://prosemirror.net/docs/guide/#transform.rebasing) through document transformations, to ensure that it's positions are not invalidated. Unfortunately, using this with the current y-prosemirror implementation, makes this approach untenable, as the positions will almost immediately be invalidated. A workaround may be to explore using [relative positions](https://docs.yjs.dev/api/relative-positions) as the anchors to reference for these positions instead of the absolute positions that prosemirror uses. But, this would require some research & testing.
+
+### Pros
+
+- This stores only the minimal representation of a change
+- Since it is stored separately
+  - it is not possible to be "lost" due to changes in the editor content.
+  - it can be permission-ed separately
```
nperez0111, line 18: how to edit these changes? Store them completely separately in a separate data store

adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+```html
+<p>Some <del>text</del><ins>More text</ins></p>
+<del><p>More text</p></del>
+```
+
+They ultimately result in the same in the final document but, the difference is in their semantics and user expectations. There are probably ways to simplify these changes in such a way that it is closer to what the user actually intended with the change (e.g. Prosemirror's [simplifyChanges](https://github.com/ProseMirror/prosemirror-changeset)), but it gets complicated rather quickly and we have to face that it may not always be easy to represent/display.
+
+Whatever the approach, we need to make sure that the changes are displayed in a way that is easy to understand and aligns with the user's intent.
+
```
nperez0111, line 143: Make clear: this is a tradeoff of a nested element approach, but we are purposefully not choosing to implement a flat YText.

nperez0111: Consider other limitations of the diff-based approach.

 - Possibly, undo-redo may change the result of the diff
 - moving content shows large deletes & insertions

adr/2025_03_10-suggestions/diff-yjs-snapshots.md
nperez0111: Check whether tracked changes within a document also is limited to this or not

adr/2025_03_10-suggestions/suggestions-as-patches.md
nperez0111: And from existing mark-based approaches. relative positions have some edges cases to be figured out.

adr/2025_03_10-suggestions/suggestions-as-history.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Cons
+
+- Requires diffing an entire YDoc to derive what the change actually was
+  - Performance might be an issue especially since it scales by number of users who suggest
+- Unclear whether to diff the content at Y.js level or to diff the JSON document
+  - If diffing at the Y.js level, not everyone uses Y.js and we may not want to pay that bundle/complexity cost
+  - If diffing on the JSON, are we just recreating the version snapshot comparison and duplicating the work involved?
+- Complicated to setup
+  - Storage of the branched versions need to be kept up to date (i.e. must apply new updates since the branch) and need to store that document separately
```
nperez0111, line 31: Perhaps, we could have a hybrid approach, where suggestions are stored into a different document (as a different history), but when in Prosemirror, we lie to Prosemirror & represent them as marks, this would allow us to re-use the approaches around tracked changes at the Prosemirror level, keep all of the benefits of marks just naturallly being transformed, and defer two sources of truth to a separate layer like y-prosemirror to sync the suggestions document and the root document.

adr/2025_03_10-suggestions/suggestions-as-history.md
```diff
@@ -0,0 +1,31 @@
...
+
+This would work very similarly to how the version diffing would work, and is intended to mostly re-use that logic.
+
+## Implementation
+
+To implement this, each user who suggests changes is given a separate document which they would have access to modify the contents of. This allows the user to own that document, while guaranteeing the original document stays unmodified.
+
+![Branched document histories merged](./images/doc+suggestions_merged.png)
+
+When viewing what suggestions are available on a document, the frontend would have to "collect" the suggestions from each suggestion document, apply an updates those documents may not have seen yet, diff the documents to see what changes are in the suggestion document compared to the original document and finally display those changes to the user in the original document somehow.
```
nperez0111, line 15: Kevin suggested that we instead have a server-maintained "suggestions doc" where all suggestions would be recorded into. Rather than having to have something which collects these suggestions into one document.

adr/2025_03_10-suggestions/suggestions-as-history.md
nperez0111: Reason being that it is quite unclear how to have 2 sources of truth for editing a single Prosemirror document, who is responsible for synchronizing the change? The problem with the editor having to be aware of this is that it potentially complicates all of the editor content to have to become "suggestion-aware", like if you are editing in a paragraph it has to choose where to save that change into.

adr/2025_03_10-suggestions/suggestions-as-content.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Pros
+
+- [There is prior art](https://github.com/handlewithcarecollective/prosemirror-suggest-changes) that we can leverage (and [this one](https://github.com/Atypon-OpenSource/manuscripts-track-changes-plugin) and [this one](https://github.com/davefowler/prosemirror-suggestion-mode/) and [this one](https://gitlab.coko.foundation/wax/wax-prosemirror/-/blob/master/wax-prosemirror-core/src/utilities/track-changes/trackedTransaction.js?ref_type=heads))
+- Importing from something like DOCX, would be a 1-1 mapping as you can just map the DOCX insertions and deletions to our equivalents
+
+## Cons
+
+- Adds complexity to displaying content in the editor, may need to rely on things like decorations to hide content we do not want to display
+- Since it is not a separate document, the permissions have to be the same as having write permissions to the document
```
nperez0111, line 22: The main trade-off of this approach, is around having a suggestions-only user (without having to resort to some server undoing logic, inspecting the changes).

Theoretically, to get around this, you could use cryptography to sign the content of these changes to ensure that they have not been tampered with across editors, but a suggestion user would still have enough permissions to write into the document when they should not be able to.

Perhaps you can offload this to the server so that it is the only entity able to write these marks into the document.

adr/2025_03_10-suggestions/README.md
```diff
@@ -0,0 +1,178 @@
...
+### How should suggestions be stored
+
+Suggestions are different from versioning in that, they are not a history of the document, but a set of pending changes that can be applied to the document. They may be rejected, or may be applied to the document. So, in a sense, they are a patch to the document. But, there are a few different mental models we can use to represent or store these changes to the document:
+
+- [Seeing a suggestion as a _branch of the document history_, and the suggestion as being the diff between the two branches](./suggestions-as-history.md)
+- [Storing document _patches_ and later applying that patch to a document](./suggestions-as-patches.md)
+- [Suggestions as _document content_ which stores the changes within the document](./suggestions-as-content.md)
+
+### AttributionManager
```
nperez0111, line 81: Kevin will write more about the attribution manager, but to give a summary of how it will work:

 - The AttributionManager will sit on the server and be able to attribute updates & elements in a trusted store (the server), to specific users. Those users will be atrtributed to specific clientIDs & clocks.
 - Technically, a user can re-use another's, clientID, so the attribution manager must take care to actually known at what clock the change was made, to correctly identify the user.
 - The attribution manager will likely be implemented in a network-agnostic way similar to the awareness protocol, so that clients can receive the attribution of users from the server.

juliomuhlbauer:
Related to versioning, I think Lix is doing a really great job at this: https://lix.opral.com/

PR #1519, opened 2025-03-11
feat: use slim shiki build and pre-compiled languages

### LINKED ISSUE #1487: BlockNote codeBlock's embedded Shiki significantly increases bundle size
**Describe the bug**
BlockNote with Shiki is causing a significant bundle size issue. Unnecessary language files (like emacs-lisp.mjs) are being included in the client bundle, increasing the bundle size by approximately 788KB (192KB gzipped).

This is occurring in a Turborepo monorepo environment with pnpm as the package manager, where our web workspace uses Next.js 14.

**To Reproduce**
1. Create a Next.js 14 project (in our case, within a Turborepo monorepo)
2. Install BlockNote: `npm install @blocknote/core @blocknote/react @blocknote/mantine`
3. Create a basic BlockNote editor with code block support
4. Run `next build` and analyze the bundle
5. Notice large chunks containing Shiki language files that aren't being used

**BlockNote version:**
```
"@blocknote/core": "^0.23.6",
"@blocknote/mantine": "^0.23.6",
"@blocknote/react": "^0.23.6",
```

**Code setup:**
```typescript
// BlockNote.tsx
import { BlockNoteSchema, customizeCodeBlock, defaultBlockSpecs } from '@blocknote/core';
import { BlockNoteView, useCreateBlockNote } from '@blocknote/react';

const customCodeBlock = customizeCodeBlock({
  defaultLanguage: 'typescript',
  supportedLanguages: [
    { id: 'javascript', match: ['javascript', 'js'], name: 'JavaScript' },
    { id: 'typescript', match: ['typescript', 'ts'], name: 'TypeScript' },
    { id: 'html', match: ['html'], name: 'HTML' },
    { id: 'css', match: ['css'], name: 'CSS' },
    { id: 'json', match: ['json'], name: 'JSON' },
    { id: 'markdown', match: ['markdown', 'md'], name: 'Markdown' },
  ],
});

  blockSpecs: {
    ...defaultBlockSpecs,
    codeBlock: customCodeBlock,
  },
});

export const BlockNote = ({ initialContent }) => {
    initialContent,
    schema,
  });

  return (
    <BlockNoteView
      editor={editor}
      editable={false}
      formattingToolbar={false}
    />
  );
};
```

I've tried using webpack externals to exclude unnecessary language files:

```javascript
// next.config.js
config.externals = [
  ...(config.externals || []),
  function(context, request, callback) {
    if (/node_modules[\/\\]@shikijs[\/\\]langs[\/\\]dist[\/\\](?!javascript|typescript|html|css|json|markdown).+\.mjs$/.test(request)) {
      console.log('Excluded Shiki language file:', request);
      return callback(null, 'commonjs {}');
    }
    callback();
  }
];
```

But the bundle still includes all Shiki language files:
Image

parsed: static/chunks/b9c69e33.fdaaa1746205dfd1.js **(785.91 KB)**
gziped: static/chunks/b9c69e33.fdaaa1746205dfd1.js **(192.3 KB)**

**Misc**
- Node version: 22.11
- Package manager: pnpm
- Browser: Chrome
- Environment: Turborepo monorepo with Next.js 14 in web workspace
- [x] I'm a sponsor and would appreciate if you could look into this sooner than later 💖

### Questions

1. Is there a way to completely remove or replace Shiki in BlockNote's codeBlock?
2. Is it possible to isolate just the code block rendering to a server component? This would keep Shiki on the server side and prevent it from being included in the client bundle.

nperez0111:
This significantly cuts down on the bundle size of the package by leveraging shiki's new slim builds and precompiled languages

Fixes #1487 #1213

This could probably be reduced further by completely separating the default languages, but for this would require a breaking change to the API for code blocks to properly work.

The ideal would be for the user to specifically pick which languages and themes for the code block feature, rather than us providing any defaults for this.

Perhaps we can have defaults built into a separate package? To make it easier for those already using us?

Either way, this would require some thought, while the current change would still be compatible with the current API


YousefED (commented):
This is great!

If I understand it correctly; this mostly makes sure unused languages that are not in our "defaultSupportedLanguages" don't appear in the app output. However, languages that are, of course still appear in the app output. correct?

> The ideal would be for the user to specifically pick which languages and themes for the code block feature, rather than us providing any defaults for this.
> 
> Perhaps we can have defaults built into a separate package? To make it easier for those already using us?

I think it's worthwhile to tackle the issue completely and invest a bit more to also address this. 

- I think it's ok to make this a breaking changes as long as it's clearly documented
- if it's not too much overhead (in code / docs / maintenance) I think having some sensible defaults easily available makes sense indeed
- we should decide whether by default we don't want any syntax highlighting at all, or only support 1-3 langs
- while we're at it, maybe we should also make it easy for consumers to load other themes

flipvh:
Great news. My biased opinion would be to not add any and let the consumers decide to include one or multiple somehow. I would see the code block as a more 'advanced functionality' that will be used by consumers that should be fine with doing some config.  If you throw in an example of that on your site, the first round of beers in de Biergarten are on me 🍻  😸.

nperez0111:
Alright, I've exposed the codeBlock configuration on the editor instance, and so this would let you configure the languages & themes that are supported by your app. By default, *no languages or themes are bundled*, to configure syntax highlighting, you can install the new package `@blocknote/code-block` which provides a decent set of languages and light & dark themes to use.

If you want more control than this, I'll write a guide on how easy it will be to integrate your list of languages, using the shiki-codegen tool.

This isn't completely ready yet, as I'll need to:
 - [ ] add an example
 - [ ] write a guide on how to configure with shiki-codegen
 - [ ] make sure the new package is being bundled appropriately & TS works within the repo

packages/core/src/api/exporters/html/__snapshots__/codeBlock/defaultLanguage/internal.html
```diff
@@ -1 +1 @@
-<div class="bn-block-group" data-node-type="blockGroup"><div class="bn-block-outer" data-node-type="blockOuter" data-id="1"><div class="bn-block" data-node-type="blockContainer" data-id="1"><div class="bn-block-content" data-content-type="codeBlock"><pre><code class="bn-inline-content language-javascript">console.log('Hello, world!');</code></pre></div></div></div></div>
\ No newline at end of file
+<div class="bn-block-group" data-node-type="blockGroup"><div class="bn-block-outer" data-node-type="blockOuter" data-id="1"><div class="bn-block" data-node-type="blockContainer" data-id="1"><div class="bn-block-content" data-content-type="codeBlock"><pre><code class="bn-inline-content">console.log('Hello, world!');</code></pre></div></div></div></div>
```
YousefED, line 1: is it easy / needed to add tests whether the exporting / importing to / from md / html still works for custom languages?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -156,6 +157,11 @@ export type BlockNoteEditorOptions<
     showCursorLabels?: "always" | "activity";
   };
 
+  /**
+   * Options for code blocks.
+   */
+  codeBlock?: CodeBlockOptions;
```
YousefED, line 163: minor thought; is it confusing that the term `codeBlock` is used for all of:
- the blocknote block (defined in core)
- the option for the theme / highlighting (defined here)
- the package for advanced theme / highlighting

examples/04-theming/06-code-block/README.md
```diff
@@ -0,0 +1,33 @@
+# Code Block Highlighting
```
YousefED, line 1: The pattern we currently use is that we have:
- max 1 or 2 paragraphs in the example as description
- an optional "try it out" line to explain how the user can see the functionality in the demo
- links to relevant docs

The idea is that the examples can really be a larger set of examples to click through and play with without too much reading - and that all documentation should be in /docs.

Could we change this? or lmk if you like a different setup

YousefED (commented):
awesome!! only some comments / qs about naming and docs

packages/core/src/api/exporters/html/__snapshots__/codeBlock/defaultLanguage/internal.html
nperez0111: I think we have some existing tests for this: https://github.com/TypeCellOS/BlockNote/blob/013818d40a2fedaf192ddc7021a8457b2eac37bc/packages/core/src/api/exporters/html/__snapshots__/codeBlock/python/internal.html#L0-L1

Just didn't need to modify it because it was only the default language that changed (it went from being `javascript` by default to `text`)

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: `codeBlock` (this option) makes sense as it is the bundle of options for configuring the code-block block type.
`@blocknote/code-block`(the package) I think could maybe be renamed to `@blocknote/defaults` with an export of `codeBlock` which could make sense if we decide to add more things like this. Or, similar.

examples/04-theming/06-code-block/README.md
nperez0111: Totally makes sense. Updated

examples/04-theming/06-code-block/App.tsx
```diff
@@ -0,0 +1,73 @@
...
+import "@blocknote/mantine/style.css";
+import { useCreateBlockNote } from "@blocknote/react";
+// This packages some of the most used languages in on-demand bundle
+import { codeBlock } from "@blocknote/code-block";
+
+export default function App() {
+  // Creates a new editor instance.
+  const editor = useCreateBlockNote({
+    codeBlock,
+    initialContent: [
```
matthewlipski, line 12: Just a minor thing - normally in the demos we pass `PartialBlock[]` to `initialContent` rather than `Block[]`. This is to make it a bit easier to read, since you don't need to specify the ID or children, and can set the content to a string.

examples/04-theming/07-custom-code-block/App.tsx
```diff
@@ -0,0 +1,95 @@
...
+      // This creates a highlighter, it can be asynchronous to load it afterwards
+      createHighlighter: () =>
+        createHighlighter({
+          themes: ["dark-plus", "light-plus"],
+          langs: [],
+        }),
+    },
+    initialContent: [
+      {
+        id: "fc832df4-bd15-49d2-8d64-140c27f29692",
```
matthewlipski, line 36: Same thing as previous comment r.e. using `PartialBlock[]` instead of `Block[]`

matthewlipski, line 36: Same thing as previous comment r.e. using `PartialBlock[]` instead of `Block[]`

packages/core/src/api/exporters/markdown/__snapshots__/codeBlock/defaultLanguage/markdown.md
```diff
@@ -1,3 +1 @@
-```javascript
-console.log('Hello, world!');
-```
+    console.log('Hello, world!');
```
matthewlipski, line 1: Not sure we can really do much about it either (since the HTML -> MD conversion is not handled by us), but this seems to be missing the code block formatting.

matthewlipski (approved):
Looks good! Only a few suggestions

packages/core/src/api/exporters/markdown/__snapshots__/codeBlock/defaultLanguage/markdown.md
nperez0111: Yea, I spent like 30 minutes looking into it. It does not want to properly convert it if it does not have a language specified

nperez0111: I will just output language-text, I don't see the harm in it

```text
This example proves it should work fine
```

PR #1520, opened 2025-03-12
fix: small issues related to comments

YousefED:
- removes dependency + check for collaboration, so theoretically you should be able to use comments without yjs
- removes `blocknoteIgnore` (wasn't used, we directly use a property on the mark spec)


nperez0111 (approved):
👍  lgtm

PR #1524, opened 2025-03-13
fix: resolve several sentry errors

nperez0111:
This is just a series of bug fixes for Sentry errors we are seeing

### DIFF
--- a/packages/core/src/schema/inlineContent/types.ts
+++ b/packages/core/src/schema/inlineContent/types.ts
@@ -126,27 +126,19 @@ export type PartialInlineContent<
 > = PartialInlineContentElement<I, T>[] | string;
 
 export function isLinkInlineContent<T extends StyleSchema>(
-  content: InlineContent<any, T> | undefined
+  content: InlineContent<any, T>
 ): content is Link<T> {
-  return content !== undefined && content.type === "link";
+  return content.type === "link";
 }
 
 export function isPartialLinkInlineContent<T extends StyleSchema>(
-  content: PartialInlineContentElement<any, T> | undefined
+  content: PartialInlineContentElement<any, T>
 ): content is PartialLink<T> {
-  return (
-    content !== undefined &&
-    typeof content !== "string" &&
-    content.type === "link"
-  );
+  return typeof content !== "string" && content.type === "link";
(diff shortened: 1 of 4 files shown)


packages/core/src/schema/inlineContent/types.ts
```diff
@@ -126,19 +126,27 @@ export type PartialInlineContent<
 > = PartialInlineContentElement<I, T>[] | string;
 
 export function isLinkInlineContent<T extends StyleSchema>(
-  content: InlineContent<any, T>
+  content: InlineContent<any, T> | undefined
```
YousefED, line 129: I want to be careful updating these types, what do these changes fix exactly / where was the error? shouldn't the typescript compiler break anyway if we had been passing undefined to the original functions?

nperez0111: This was the sentry issue: https://blocknote-js.sentry.io/issues/32289262/?environment=vercel-production&query=is%3Aregressed&referrer=issue-stream&stream_index=2

I think you cannot assume that the index access into the array works, we should probably turn on `noUncheckedIndexAccess` on in our tsconfig: https://www.typescriptlang.org/tsconfig/#noUncheckedIndexedAccess

But, I stand by my previous statement that these functions are meant to assert that something is the correct type and therefore should also assert that something is there first before trying to assert that (i.e. check for undefined or null prior)

YousefED: Ok. I usually take the opposite approach; https://martinfowler.com/ieeeSoftware/failFast.pdf - let's discuss when you're back :)

nperez0111: I'd agree with that in general, but I see a type-guard as a parser: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/ 😉

YousefED: Thanks for that and for the issue! Let's discuss when you're back!

YousefED: fixed the issue at different level so we can discuss when you're back, just good to get the fixes in now

PR #1529, opened 2025-03-14
feat: position storage WIP

nperez0111:
This is still in progress, the idea here is to make a class that can be used for keeping track of positions in the editor across collaborative or single user transactions.

This would be useful not only for the suggestions plugin, but for anything that would need to hold onto a position across transactions.

I still am working through setting up the tests to prove that it is able to keep track of those positions, but I had it preliminarily working for the suggestion plugin at one point, but then decided to solve this in a more generic way.

### DIFF
--- a/packages/core/src/api/positionMapping.ts
+++ b/packages/core/src/api/positionMapping.ts
@@ -1,0 +1,175 @@
+import { Transaction } from "prosemirror-state";
+import { Mapping } from "prosemirror-transform";
+import {
+  absolutePositionToRelativePosition,
+  relativePositionToAbsolutePosition,
+  ySyncPluginKey,
+} from "y-prosemirror";
+import { BlockNoteEditor } from "../editor/BlockNoteEditor.js";
+import * as Y from "yjs";
+import { ProsemirrorBinding } from "y-prosemirror";
+
+export function isRemoteTransaction(tr: Transaction) {
+  return tr.getMeta(ySyncPluginKey) !== undefined;
+}
+type YSyncPluginState = {
+  doc: Y.Doc;
+  binding: Pick<ProsemirrorBinding, "type" | "mapping">;
+};
+type RelativePosition = symbol;
+
+/**
+ * This class is used to keep track of positions of elements in the editor.
+ * It is needed because y-prosemirror's sync plugin can disrupt normal prosemirror position mapping.
+ *
+ * It is specifically made to be able to be used whether the editor is being used in a collaboratively, or single user, providing the same API.
+ */
+export class PositionStorage {
+  private readonly editor: BlockNoteEditor;
+  /**
+   * Whether the editor has had a remote transaction.
+   */
+  private hadRemoteTransaction = false;
+  /**
+   * A map of an ID to the position mapping.
+   */
+  private readonly positionMapping = new Map<
+    string,
+    {
+      position: number;
+      relativePosition: RelativePosition | undefined;
+      mapping: Mapping;
+      side: "left" | "right";
+    }
+  >();
+
+  constructor(editor: BlockNoteEditor) {
+    this.editor = editor;
+    this.onTransactionHandler = this.onTransactionHandler.bind(this);
+  }
+
+  /**
+   * Mounts the position storage.
+   */
+  public mount() {
+    this.editor._tiptapEditor.on("transaction", this.onTransactionHandler);
+
+    return this.unmount.bind(this);
+  }
+
+  /**
+   * Unmounts the position storage.
+   */
+  public unmount() {
+    this.positionMapping.clear();
+    this.editor._tiptapEditor.off("transaction", this.onTransactionHandler);
+  }
+
+  /**
+   * This will be called whenever a transaction is applied to the editor.
+   *
+   * It's used to update the position mapping or tell if there was a remote transaction.
+   */
+  private onTransactionHandler({ transaction }: { transaction: Transaction }) {
+    console.log("onTransactionHandler", transaction);
+    if (this.hadRemoteTransaction) {
+      // If we have already had a remote transaction, we rely only on relative positions
+      return;
+    }
+
+    if (isRemoteTransaction(transaction)) {
+      this.hadRemoteTransaction = true;
+    } else {
+      this.positionMapping.forEach(({ mapping }) => {
+        mapping.appendMapping(transaction.mapping);
+      });
+    }
+  }
+
+  /**
+   * Stores a position for a given ID. To consistently track the position of an element.
+   *
+   * @param id An ID to store the position of.
+   * @param position The position to store.
+   * @param side The side of the position to store.
+   */
+  public set(id: string, position: number, side?: "left" | "right") {
+    const ySyncPluginState = ySyncPluginKey.getState(
+      this.editor._tiptapEditor.state
+    ) as YSyncPluginState;
+
+    if (!ySyncPluginState) {
+      // TODO unsure if this works
+      this.positionMapping.set(id, {
+        position,
+        relativePosition: undefined,
+        mapping: new Mapping(),
+        side: side ?? "left",
+      });
+      return this;
+    }
+
+    const relativePosition = absolutePositionToRelativePosition(
+      // Track the position before the position
+      position + (side === "left" ? -1 : 0),
+      ySyncPluginState.binding.type,
+      ySyncPluginState.binding.mapping
+    );
+
+    this.positionMapping.set(id, {
+      position,
+      relativePosition,
+      mapping: new Mapping(),
+      side: side ?? "left",
+    });
+
+    return this;
+  }
+
+  public get(id: string): number {
+    const storedPos = this.positionMapping.get(id);
+
+    console.log(storedPos);
+
+    if (!storedPos) {
+      throw new Error("No mapping found for id: " + id);
+    }
+
+    if (this.hadRemoteTransaction) {
+      // If we have had a remote transaction, we need to rely on the relative position
+      if (!storedPos.relativePosition) {
+        throw new Error("No relative position found for id: " + id);
+      }
+
+      const ystate = ySyncPluginKey.getState(
+        this.editor._tiptapEditor.state
+      ) as YSyncPluginState;
+      const rel = relativePositionToAbsolutePosition(
+        ystate.doc,
+        ystate.binding.type,
+        storedPos.relativePosition,
+        ystate.binding.mapping
+      );
+
+      if (rel === null) {
+        // TODO when does this happen?
+        return -1;
+      }
+
+      return rel + (storedPos.side === "left" ? 1 : -1);
+    }
+
+    return (
+      storedPos.mapping.map(
+        storedPos.position - (storedPos.side === "left" ? 1 : 0),
+        storedPos.side === "left" ? -1 : 1
+      ) + (storedPos.side === "left" ? 1 : 0)
+    );
+  }
+
+  public remove(id: string) {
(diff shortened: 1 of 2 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -367,6 +368,17 @@ export class BlockNoteEditor<
...
+   * Internal properties that are not part of the public API and may change in the future.
+   *
+   * @internal
+   */
+  public readonly '~internal': {
+    /**
+     * Stores positions of elements in the editor.
+     */
+    positionStorage: PositionStorage<BSchema, ISchema, SSchema>;
+  };
```
nperez0111, line 381: Let's have a discussion on this.

In general, here is what I'm thinking:

  - The editor instance is the main thing passed around and is essentially [a god object](https://en.wikipedia.org/wiki/God_object)
  - configuration is primarily done through the editor instance, making it the most convenient place to store and retrieve references to the current config
  
I think what we need to be aiming for here instead is a de-coupled approach like [prosekit](https://prosekit.dev), where the editor doesn't "have" these sorts of values, it can execute commands to move between states, but the values of these are stored elsewhere (like context).

Given that we don't have a good solution for state management at the moment, I propose here to just sweep things under the rug for a little while, and put them under an `~internal` property. Using `~` makes it both cumbersome (read intentional) and shows the value last in TS autocomplete [idea from standard-schema](https://x.com/colinhacks/status/1816860780459073933).

For configuration, I think that if the schema was defined separately, it would be much easier to configure things like tables and codeBlocks. For example:

```ts
import { defaultSchema, tableBlock } from '@blocknote/core/schema';|
import { BlockNoteEditor } from '@blocknote/core';

const editor = BlockNoteEditor.create({
  schema: defaultSchema.replace('table', tableBlock.configure({ option: true })
})

// editor.schema.table.config.option === true
```
 Obviously, there are a lot of different ways to do it, but the point is that the editor doesn't really have to know about this.
 
 What about all of the prosemirror plugins? It could maybe be something like:
 
 ```ts
 import { BlockNoteEditor, BlockNoteExtension, getDefaultExtensions } from '@blocknote/core';
 
 const editor = BlockNoteEditor.create({
   extensions: getDefaultExtensions.remove(['tables']).add({ myExt: BlockNoteExtension.create({ name: 'myExt', plugin: new Plugin() }) })
 })
 
// editor.extensions.myExt.plugin
 ```
 Again, different approaches here too

packages/core/src/api/positionMapping.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    if (!shouldMount) {
+      return;
+    }
+
+    if (!this.editor._tiptapEditor) {
+      throw new Error("Editor not mounted");
+    }
+    this.editor._tiptapEditor.on("transaction", this.onTransactionHandler);
+  }
```
nperez0111, line 66: An alternative to this, would be to have a prosemirror plugin that gets these transactions between state updates. Perhaps that is preferable?

packages/core/src/api/positionMapping.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  private readonly mapping = new Mapping();
+  private mappingLength = 0;
+
+  constructor(
+    editor: BlockNoteEditor<BSchema, ISchema, SSchema>,
+    { shouldMount = true }: { shouldMount?: boolean } = {}
+  ) {
+    this.editor = editor;
+    this.onTransactionHandler = this.onTransactionHandler.bind(this);
```
YousefED, line 56: this overrides a local function? maybe just taste / code style, but imo it's cleaner to do a closure in the definition (`private onTransactionHandler = () => ...`)

packages/core/src/api/positionMapping.ts
YousefED: this is fine imo, unless there's an advantage to doing this in a plugin, but I don't think so.

Another alternative would be to emit our own event in BlockNoteEditor.dispatch - but there's a change that misses things when people use _tiptapEditor.dispatch

packages/core/src/api/positionMapping.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (this.hadRemoteTransaction) {
+      // If we have already had a remote transaction, we rely only on relative positions, so no need to update the mapping.
+      return;
+    }
+
+    if (isRemoteTransaction(transaction)) {
+      this.hadRemoteTransaction = true;
+    } else {
+      this.mapping.appendMapping(transaction.mapping);
+      this.mappingLength += transaction.mapping.maps.length;
```
YousefED, line 83: can't get this from `this.mapping`?

YousefED: ah; after reading further; I suppose the answer is that you can use a single `PositionStorage` object to track multiple positions. That makes sense and is pretty nifty. On the other hand, having a single `PositionStorage` object per tracked position would make it a little bit more efficient I suppose

packages/core/src/api/positionMapping.ts
```diff
@@ -0,0 +1,145 @@
...
+      return () => {
+        const pos = this.mapping
+          .slice(trackedMapLength)
+          .map(position, side === "left" ? -1 : 1);
+
+        return pos;
+      };
+    }
+
+    const relativePosition = absolutePositionToRelativePosition(
```
YousefED, line 119: apparently yjs also supports `assoc` for left / right: https://github.com/yjs/yjs/blob/0e3195797fd6fce39b987f7aebd2529681f29e15/src/utils/RelativePosition.js#L163 - but it's not available in the prosemiror "absToRel" functions.

Could this cause any bugs (the fact that we now just do `rel-1` at the end?)

packages/core/src/api/positionMapping.ts
```diff
@@ -0,0 +1,145 @@
...
+        this.editor._tiptapEditor.state
+      ) as YSyncPluginState;
+      const rel = relativePositionToAbsolutePosition(
+        ystate.doc,
+        ystate.binding.type,
+        relativePosition,
+        ystate.binding.mapping
+      );
+
+      // This can happen if the element is deleted
```
YousefED, line 137: can this cause things to break in a weird way for the user?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -723,6 +735,13 @@ export class BlockNoteEditor<
       // but we still need the schema
       this.pmSchema = getSchema(tiptapOptions.extensions!);
     }
+
+    this["~internal"] = {
+      positionStorage: new PositionStorage<BSchema, ISchema, SSchema>(this, {
```
YousefED, line 740: hmm, instantiating this on the editor means we store the mapping of all transactions since the document starts. I'm starting to think that we don't need a "global" instance of this and the alternative approach (in my previous comment) is a bit cleaner. In that case, we also don't need to worry about the "internal" thingg)

packages/core/src/extensions/SuggestionMenu/SuggestionPlugin.ts
```diff
@@ -127,7 +127,7 @@ class SuggestionMenuView<
       .focus()
       .deleteRange({
         from:
-          this.pluginState.queryStartPos! -
+          this.pluginState.queryStartPos!() -
```
YousefED, line 130: can the `!` here (and in usages below) cause errors if a remote user deletes a piece of text?

packages/core/src/extensions/SuggestionMenu/SuggestionPlugin.ts
```diff
@@ -218,13 +218,21 @@ export class SuggestionMenuProseMirrorPlugin<
...
+                suggestionPluginTransactionMeta.triggerCharacter.length
+            );
             return {
               triggerCharacter:
                 suggestionPluginTransactionMeta.triggerCharacter,
               deleteTriggerCharacter:
                 suggestionPluginTransactionMeta.deleteTriggerCharacter !==
                 false,
-              queryStartPos: newState.selection.from,
+              // When reading the queryStartPos, we offset the result by the length of the trigger character, to make it easy on the caller
```
YousefED, line 232: thanks for this comment :)

YousefED (commented):
really like this approach to abstract it away and make it reusable! even better that it's unit tested etc.

I have one main question about the architecture and some minor comments, see inline

packages/core/src/api/positionMapping.ts
nperez0111: The yjs implementation is just subtracting on the index (because yjs normally does right association since that is how most writing is done & compresses well for runs). So, this should be exactly equivalent.

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: The tradeoff here is potential for memory leaks & ease-of-use.

What you are proposing would require having each position storage listen for transactions, but, when retrieving the value, there is nothing that would clean up that listener, so it'd just keep listening and leak memory.
Having 1 listener, and one map to keep track of the transactions avoids this by storing just a singleton, meaning that when the position is no longer referenced, it can be garbage collected.

What if we just add a cleanup function? Then storing the value would be a bit more of a hassle, since you would need to unsubscribe somewhere. And in something like the prosemirror plugin it is being used in now, you typically just stop referencing the value instead of needing to destructure it so there isn't a good place for it.
What if, on read of the value, we automatically unsubscribe? It makes it inconvenient to reference the value multiple times.


My first attempt at this API did it like you described, but I found the tradeoff with the listener to be problematic later.

YousefED: Can't you just unsubscribe / dispose the PositionStorage instance when the slash menu is closed, and we no longer need to track the position?

nperez0111: You could, but it's a less convenient API. I think I have something else that should work.

nperez0111: I was able to separate it from the editor instance by using a map of editor instance to the mapping. This will still use 1 mapping for the duration of an editor, but I think this is a fine trade-off for making it just as easy to store & use as a raw position value

matthewlipski (approved):
Nice! Love how clear the comments are and the unit tests also help a lot with understanding how to use `trackPosition`. I don't really have any comments on the code, but IIRC this change is supposed to fix a bug with the slash menu with collaboration enabled, so would be great if you could update the description with that.

YousefED (approved):
Sorry for the late review!

Nice improvements :) Keeping the map externally works and I suppose the `hadRemoteTransaction` only made things more complex / didn't really add anything.

I still think conceptually it smells a bit that something that registeres an event listener doesn't get cleaned up (a `.on` without a `.off` so to say) - but not a deal breaker (memory impact should be minor)

I'm sure the position storage is a useful abstraction to have 🙌

packages/core/src/api/positionMapping.ts
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * It is specifically made to be able to be used whether the editor is being used in a collaboratively, or single user, providing the same API.
+ *
+ * @param editor The editor to track the position of.
+ * @param position The position to track.
+ * @param side The side of the position to track. "left" is the default. "right" would move with the change if the change is in the right direction.
+ * @returns A function that returns the position of the element.
+ */
+export function trackPosition<
+  BSchema extends BlockSchema = DefaultBlockSchema,
```
YousefED, line 61: generic can be removed, editor can just be <any,any,any> as the `trackPosition` code doesn't care about the schema

PR #1533, opened 2025-03-16
feat: split out localization files for optimized bundle

nperez0111:
Is a breaking change, since I had to change the export, but an easy one to make

Before:
```txt
vite v5.3.4 building for production...
✓ 168 modules transformed.
dist/webpack-stats.json   14.00 kB │ gzip:  2.39 kB
dist/style.css            11.63 kB │ gzip:  2.82 kB
dist/comments.js          17.39 kB │ gzip:  3.51 kB │ map:    49.43 kB
dist/blocknote.js        416.35 kB │ gzip: 93.27 kB │ map: 1,103.85 kB
dist/webpack-stats.json   14.01 kB │ gzip:  2.41 kB
dist/style.css            11.63 kB │ gzip:  2.82 kB
dist/comments.cjs         11.72 kB │ gzip:  2.85 kB │ map:    47.30 kB
dist/blocknote.cjs       273.46 kB │ gzip: 72.57 kB │ map: 1,044.26 kB
✓ built in 506ms
```

After:

```txt
vite v5.3.4 building for production...
✓ 170 modules transformed.
dist/webpack-stats.json   14.50 kB │ gzip:  2.52 kB
dist/style.css            11.63 kB │ gzip:  2.82 kB
dist/en-BEb_5vQO.js        7.78 kB │ gzip:  1.91 kB │ map:  14.03 kB
dist/comments.js          17.39 kB │ gzip:  3.51 kB │ map:  49.43 kB
dist/locales.js          135.64 kB │ gzip: 27.15 kB │ map: 238.90 kB
dist/blocknote.js        274.20 kB │ gzip: 65.23 kB │ map: 853.96 kB
dist/webpack-stats.json   14.51 kB │ gzip:  2.52 kB
dist/style.css            11.63 kB │ gzip:  2.82 kB
dist/en-DoDAHwFo.cjs       5.34 kB │ gzip:  1.76 kB │ map:  12.34 kB
dist/comments.cjs         11.72 kB │ gzip:  2.85 kB │ map:  47.30 kB
dist/locales.cjs          93.50 kB │ gzip: 24.76 kB │ map: 210.56 kB
dist/blocknote.cjs       175.52 kB │ gzip: 47.01 kB │ map: 823.91 kB
✓ built in 588ms
```

Shaved ~100kb from cjs

Now, to import locales you can either do it individually like:

```js
import { en, nl, de } from '@blocknote/core/locales';
```

Or, all together like:

```js
import * as locales from '@blocknote/core/locales';
```

### DIFF
--- a/packages/core/src/i18n/index.ts
+++ b/packages/core/src/i18n/index.ts
@@ -1,2 +1,3 @@
 export * as locales from "./locales/index.js";
+export * from "./locales/index.js";
 export * from "./dictionary.js";
--- a/packages/core/src/i18n/index.ts
+++ b/packages/core/src/i18n/index.ts
@@ -1,3 +1,2 @@
-export * as locales from "./locales/index.js";
 export * from "./locales/index.js";
 export * from "./dictionary.js";
--- a/packages/core/src/i18n/index.ts
+++ b/packages/core/src/i18n/index.ts
@@ -1,0 +1,2 @@
+export * as locales from "./locales/index.js";
+export * from "./dictionary.js";
--- a/docs/components/pages/landing/hero/DemoEditor.tsx
+++ b/docs/components/pages/landing/hero/DemoEditor.tsx
@@ -2,9 +2,9 @@ import {
   BlockNoteSchema,
   combineByGroup,
   filterSuggestionItems,
-  locales,
   uploadToTmpFilesDotOrg_DEV_ONLY,
 } from "@blocknote/core";
+import { locales } from "@blocknote/core/locales";
 import "@blocknote/core/fonts/inter.css";
 import {
   getDefaultReactSlashMenuItems,
--- a/docs/components/pages/landing/hero/DemoEditor.tsx
+++ b/docs/components/pages/landing/hero/DemoEditor.tsx
@@ -4,7 +4,7 @@ import {
   filterSuggestionItems,
   uploadToTmpFilesDotOrg_DEV_ONLY,
 } from "@blocknote/core";
-import { locales } from "@blocknote/core/locales";
+import * as locales from "@blocknote/core/locales";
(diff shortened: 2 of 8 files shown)


YousefED (commented):
Nice! Q: doesn't it make more sense to split out the locales independently? I.e.: if I load Dutch, I don't need to pull in all other languages?

nperez0111:
> Nice! Q: doesn't it make more sense to split out the locales independently? I.e.: if I load Dutch, I don't need to pull in all other languages?

I'd have to test, but I think they should be tree-shaked based on access. So not sure if we need to be that granular. But, I can quickly test this

YousefED:
> I'd have to test, but I think they should be tree-shaked based on access. So not sure if we need to be that granular. But, I can quickly test this

good point! if it's tree-shaken then it's fine 👍

nperez0111:
After some testing, I found that `locales` cannot be tree-shaken, but the individual exports can be.

So, `import { locales } from '@blocknote/core/locales'` will be much larger than, `import { nl } from '@blocknote/core/locales'`

packages/core/src/i18n/index.ts
```diff
@@ -0,0 +1,3 @@
+export * as locales from "./locales/index.js";
```
YousefED, line 1: do we need both these first 2 lines? (if so, why exactly)?

nperez0111: If anything the first one is the redundant one.

Allows you to import locales either as:
`import { locales } from '@blocknote/core/locales';` or `import { en, nl, de } from '@blocknote/core/locales';`

The `locales` one is just convenient but, can be done with:
`import * as locales from '@blocknote/core/locales';`

So, I can remove it.

PR #1536, opened 2025-03-17
fix: table error in collaboration mode

YousefED:
fixes `Table block does not contain a 'tbody' HTML element. This should never happen.` when in collab mode.

It's a quick fix to fix just this issue. The TableHandlesPlugin definitely needs a refactor to:
a) make the code clearer
b) work better with collaboration. e.g.: we currently also store a nodepos in the state which is unsafe for collaboration scenarios (because the position can change when there are incoming changes)

I'll create a separate ticket for this


nperez0111 (approved):

nperez0111:
Agreed on all points

Issue #1543, opened 2025-03-18
Support custom schema for (React) comment editors


milanpoliak:
**Is your feature request related to a problem? Please describe.**
We want to implement mentions in comments. For that, we need to set up inline content specs, which is not possible since the comment schema is hardcoded in the React components.

**Describe the solution you'd like**
Being able to pass a custom schema to the comments plugin or components, while using the current comment schema as a fallback.

**Describe alternatives you've considered**
I'm working on creating our own set of comment components, which will use our custom schema, but that will make it harder for us to stay up to date with further BlockNote comment components development.

CodeXtinction:
+1 looking to implement mentions too

PR #1545, opened 2025-03-18
Commets/custom schema

milanpoliak:
Making it possible to pass custom schemas to comments.

Related issue: https://github.com/TypeCellOS/BlockNote/issues/1543

### DIFF
--- a/packages/react/src/components/Comments/Comment.tsx
+++ b/packages/react/src/components/Comments/Comment.tsx
@@ -25,8 +25,9 @@ import { useDictionary } from "../../i18n/dictionary.js";
 import { CommentEditor } from "./CommentEditor.js";
 import { EmojiPicker } from "./EmojiPicker.js";
 import { ReactionBadge } from "./ReactionBadge.js";
-import { schema } from "./schema.js";
+import { defaultCommentSchema } from "./schema.js";
 import { useUser } from "./useUsers.js";
+import { BlockNoteSchema } from '@blocknote/core';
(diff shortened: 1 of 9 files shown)


YousefED:
Awesome, thanks for creating this PR. Does this solve all your requirements? I figured it might be also be useful to allow all options of the editors to be easily customized? e.g. allow consumers to implement a function `getBlockNoteEditorOptions("comment-composer")` that returns the options for (in this case) the composer

milanpoliak:
> Awesome, thanks for creating this PR. Does this solve all your requirements? I figured it might be also be useful to allow all options of the editors to be easily customized? e.g. allow consumers to implement a function `getBlockNoteEditorOptions("comment-composer")` that returns the options for (in this case) the composer

Yeah, it solves all I need right now, but the idea of a function to provide all other options sounds really good. I won't have time to implement it today, but I can take a stab at it in the next few days.

> PS: what are you building?

We are migrating our news site from WordPress to a custom CMS

Harishri2002:
Is this issue solved? can i use this to implement mentioning in comment section?

ArihantBapna:
what's this pr waiting on atm? would help our use-case as well

nperez0111:
@ArihantBapna I think what we are waiting on here is for what was outlined here: https://github.com/TypeCellOS/BlockNote/pull/1545#issuecomment-2735528998

Also, there are a number of conflicts with main, now.

PR #731, opened 2024-05-08
feat: File blocks

matthewlipski:
This PR adds file blocks which can be extended to show previews for different file types. Currently, this includes an image preview by default. The file block replaces the existing image block.

PR #1553, opened 2025-03-21
fix: Selection when clicking past end of inline content

### LINKED ISSUE #1523: Trying to insert the cursor at the end of the element inserts the cursor at the start
The Alert component does work perfectly because of its flex-grow: 1.
I managed to fix it with this css style:
```css
.bn-block-content > * {
    width: 100%;
  }
```

https://github.com/user-attachments/assets/0fca76f0-d264-4083-9404-22f8dd9c1a82

matthewlipski:
Closes #1523

### DIFF
--- a/packages/core/src/editor/Block.css
+++ b/packages/core/src/editor/Block.css
@@ -37,6 +37,10 @@ BASIC STYLES
   outline: 4px solid rgb(100, 160, 255);
 }
 
+.bn-inline-content {
+  width: 100%;
+}
+
 /*
 NESTED BLOCKS
 */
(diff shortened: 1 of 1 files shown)


YousefED:
Was this a recent regression / do you know what caused this to occur now?

nperez0111 (approved):
Looks like it'd be easy to overlook this

matthewlipski:
> Was this a recent regression / do you know what caused this to occur now?

We made some changes to CSS in #731, and removed `flex-grow: 1` from `.bn-block-content`. I don't think there's a relevant difference between `flex-grow: 1` and the changes in this PR.

PR #1554, opened 2025-03-21
fix: Hard break keyboard shortcut not working in custom blocks

matthewlipski:
closes https://github.com/TypeCellOS/BlockNote/pull/1554

### DIFF
--- a/packages/core/src/editor/BlockNoteExtensions.ts
+++ b/packages/core/src/editor/BlockNoteExtensions.ts
@@ -1,6 +1,5 @@
 import { AnyExtension, Extension, extensions } from "@tiptap/core";
 import { Gapcursor } from "@tiptap/extension-gapcursor";
-import { HardBreak } from "@tiptap/extension-hard-break";
 import { History } from "@tiptap/extension-history";
 import { Link } from "@tiptap/extension-link";
 import { Text } from "@tiptap/extension-text";
@@ -17,6 +16,7 @@ import { CommentsPlugin } from "../extensions/Comments/CommentsPlugin.js";
 import type { ThreadStore } from "../comments/index.js";
 import { FilePanelProsemirrorPlugin } from "../extensions/FilePanel/FilePanelPlugin.js";
 import { FormattingToolbarProsemirrorPlugin } from "../extensions/FormattingToolbar/FormattingToolbarPlugin.js";
+import { HardBreak } from "../extensions/HardBreak/HardBreak.js";
 import { KeyboardShortcutsExtension } from "../extensions/KeyboardShortcuts/KeyboardShortcutsExtension.js";
 import { LinkToolbarProsemirrorPlugin } from "../extensions/LinkToolbar/LinkToolbarPlugin.js";
 import {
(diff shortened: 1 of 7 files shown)


YousefED (commented):
Looks nice - if possible; can we do this without the `commands` interface of Tiptap? (would rather use more vanilla PM code.

Also, why do we need the priority exactly?

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -179,7 +179,7 @@ const getTipTapExtensions = <
       types: ["blockContainer", "columnList", "column"],
       setIdAttribute: opts.setIdAttribute,
     }),
-    HardBreak.extend({ priority: 10 }),
```
nperez0111, line 182: Priority makes sense, since it needs to be higher priority

packages/core/src/extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -543,6 +543,7 @@ export const KeyboardShortcutsExtension = Extension.create<{
       Backspace: handleBackspace,
       Delete: handleDelete,
       Enter: handleEnter,
+      "Shift-Enter": () => this.editor.commands.setHardBreak(),
```
nperez0111, line 546: Unless something is using the `editor.commands.setHardBreak()`, I'd agree with @YousefED that this can probably even just be inlined into here & remove the `setHardBreak` command.

packages/core/src/editor/BlockNoteExtensions.ts
```diff
@@ -1,6 +1,5 @@
 import { AnyExtension, Extension, extensions } from "@tiptap/core";
 import { Gapcursor } from "@tiptap/extension-gapcursor";
-import { HardBreak } from "@tiptap/extension-hard-break";
```
nperez0111, line 3: We'll also want to remove this extension from package.json

YousefED:
@nperez0111 @matthewlipski let's maybe also think through whether this should be configurable or not.

- Do we want to make it easy to make custom blocks where `shift+enter` doesn't insert a break? (do we see use cases for this)?
- Do we want to make it easy to make custom blocks where just `enter` inserts a break? (do we see use cases for this)?

packages/core/src/extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -467,6 +467,36 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
             }
 
+            return false;
+          }),
+        // Creates a hard break if block is configured to do so.
+        () =>
+          commands.command(({ state }) => {
+            const blockInfo = getBlockInfoFromSelection(state);
+
+            const blockHardBreakShortcut =
```
YousefED, line 477: cleaner to set the default here imo with `?? "shift+enter"`

YousefED (commented):
looks nice imo!

YousefED:
@matthewlipski can you add docs for this + comment the BlockConfig in the code? (let's make this standard practice)

PR #1557, opened 2025-03-24
docs: Update Liveblocks

matthewlipski:


docs/pages/docs/collaboration/real-time-collaboration.mdx
```diff
@@ -95,88 +99,19 @@ const provider = new YPartyKitProvider(
...
+[//]  #[//]  #[//]  #[//]  #[//]  #
+[//]  #[//]  #[//]  #[//]  #
+[//]  #[//]  #
+[//]  #[//]  #
+[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #
+[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #
+[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #[//]  #
+[//]  #[//]  #[//]  #[//]  #[//]  #
+[//]  #[//]  #[//]  #[//]  #
+[//]  # ":::"
```
nperez0111, line 117: What is this about?

nperez0111 (approved):
Looks good to me otherwise

nperez0111:
> * The example is showing a `Duplicate extension names found: ['liveblocksMention']` warning, also with strict mode disabled. Is this smth we need to address? Seems to work well otherwise.

I'll take another look at this. Though Tiptap will de-dupe these and only respect one of them (whichever appeared first).

> * @nperez0111 when installing `@liveblocks/react-blocknote`, there was an issue with peer dependencies. Specifically, it needs `@blocknote/core: ^0.25.1` but `0.26.0` was provided. This seems like it should be fine as it's not a major version difference so I'm not sure why it's throwing an error. Any ideas why this might be happening? Right now we fixed it with `npm install --legacy-peer-deps`, but this seems hacky.

image
So, technically because this is a `0.x` package, that is the correct behavior since in semver `0.x` range is treated specially. I think what liveblocks will need to specify as their peer dep is: `0.25.1 - 1.0.0`
image

matthewlipski:
We should probably wait to merge this, since the Liveblocks guys are working on resolving the 2 bullet points in the PR description

PR #1559, opened 2025-03-24
chore: upgrade unified dependencies

YousefED:
was running into an issue (see test case) that has been addressed here: https://github.com/remarkjs/remark/issues/908


nperez0111 (approved):

packages/core/src/api/exporters/markdown/__snapshots__/complex/misc/markdown.md
```diff
@@ -1,4 +1,4 @@
-## **Heading ***~~2~~*
+## **Heading&#x20;***~~2~~*
```
matthewlipski, line 1: Is this correct? I know that it's a unicode space character but it doesn't seem quite right to have in the Markdown

matthewlipski (approved):

packages/core/src/api/exporters/markdown/__snapshots__/complex/misc/markdown.md
YousefED: I think this might be done to prevent markdown parsers from trimming text before the space

PR #1563, opened 2025-03-25
feat: Block quote

matthewlipski:
This PR adds a block quote to the default blocks. It also adds an editor flag, `showNestingIndicator` (defaults to true), which determines whether to show the line on the left of nested blocks, as this looks very similar to the line that's part of the block quote.


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -233,6 +233,11 @@ export type BlockNoteEditorOptions<
    */
   setIdAttribute?: boolean;
 
+  /**
+   * A flag indicating whether to show the grey left border on nested blocks.
+   */
+  showNestingIndicator: boolean;
```
nperez0111, line 239: Hm, I'm not sure I like this.
Should it be on the editor instance? It is purely for styling no?
Why do we need this? Can't this be done with CSS?

packages/core/src/blocks/QuoteBlockContent/QuoteBlockContent.ts
```diff
@@ -0,0 +1,65 @@
...
+import { createDefaultBlockDOMOutputSpec } from "../defaultBlockHelpers.js";
+import { defaultProps } from "../defaultProps.js";
+import { getBlockInfoFromSelection } from "../../api/getBlockInfoFromPos.js";
+import { updateBlockCommand } from "../../api/blockManipulation/commands/updateBlock/updateBlock.js";
+
+export const quotePropSchema = {
+  ...defaultProps,
+};
+
+export const QuoteBlockContent = createStronglyTypedTiptapNode({
```
nperez0111, line 14: Do we also want the `> ` input rule? To make it easy to make a blockquote?

packages/core/src/editor/BlockNoteEditor.ts
matthewlipski: I believe this was a suggestion from @YousefED, but I may have misunderstood. You're right though, you can just do the same thing with CSS:

```css
.bn-block-group
.bn-block-group
.bn-block-outer:not([data-prev-depth-changed])::before {
  border-left: none !important;
}
```

packages/core/src/blocks/QuoteBlockContent/QuoteBlockContent.ts
matthewlipski: Good call! Forgot about that

packages/core/src/editor/BlockNoteEditor.ts
YousefED: My idea was:
- to make the quoteblock properly work, the lines need to be disabled
- maybe it makes sense to disable the lines anyway (as per our "follow notion design" guideline)
- if we disable the lines by default; I think it makes sense to offer (existing) users an easy way to re-add them

Don't feel to strong about this I'll let @nperez0111 decide. We could also defer the border-left change to a later moment (it might make sense to also indent the draghandle when we do this)

matthewlipski: I'll revert the `showNestingIndicator` changes for now then - if people want to disable the nesting lines we can just paste the CSS styles above as a solution. I think it's a good idea to revisit this alongside the upcoming theming overhaul, as the "follow notion design" guideline will be highly relevant there too.

PR #1565, opened 2025-03-26
fix: Multi-block links

matthewlipski:
Currently, creating a new link which spans multiple blocks merges them together. This PR fixes that, and also makes the formatting toolbar button for creating links only show the URL field (previously URL + text), as this is more in line with Notion/GDocs UX.


nperez0111 (approved):
This looks good to me. I guess someone can re-enable the text input if they want?

matthewlipski:
Yeah they can, though I doubt this is going to be smth many people do anyway

PR #1567, opened 2025-03-26
chore: use pnpm instead of npm

nperez0111:
This PR introduces using pnpm instead of npm, pnpm is much better suited for:
 - monorepos
 - speed
 - running commands across packages (taking over part of lerna's role)

It also completely replaces lerna in favor of NX (which lerna was already based on). This gives us one tool to:

 - run commands and cache their results
 - release packages

This PR also renames several PNPM scripts, to be more standard. The goal should be for PNPM scripts to already have the most commonly used tasks that you might need within the repo. It should be possible to do everything you need from the root of the repo.


.github/workflows/build.yml
```diff
@@ -10,18 +10,26 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 100
+      - run: jq '.packageManager' package.json | tr -d '"pnpm@'
+        id: package-manager-version
```
nperez0111, line 19: I didn't want to have to keep the pnpm version in sync between the definition in `package.json` so I just pulled the value I wanted from the file, and parse out the version number.

So, version number for PNPM is defined by the `packageManager` field in the `package.json` and `.nvmrc` defines the node version. So, if you have nvm with corepack enabled, it should just always use pnpm instead of npm

.github/workflows/build.yml
```diff
@@ -33,32 +41,36 @@ jobs:
             nx-${{ env.NX_BRANCH }}-
             nx-
 
-      - name: Install Dependencies
-        run: npm ci --prefer-offline --no-audit
+      - run: sudo apt-get update && sudo apt-get install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev pkg-config
```
nperez0111, line 44: canvas dep wasn't building properly (hopefully we can remove it with Tiptap v3 soon!)

.github/workflows/build.yml
```diff
@@ -33,32 +41,36 @@ jobs:
...
 
       - name: Upload webpack stats artifact (editor)
         uses: relative-ci/agent-upload-artifact-action@v2
         with:
           webpackStatsFile: ./playground/dist/webpack-stats.json
           artifactName: relative-ci-artifacts-editor
 
+      - name: Soft release
+        id: soft-release
+        run: pnpx pkg-pr-new publish './packages/*' # --compact enable compact after release
```
nperez0111, line 66: This adds support for [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new) which allows us to publish preliminary versions of packages to their registry, which should decrease the burden of releasing things when people need them as they can always use the preliminary version. Also great for testing things out in development

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,61 @@
+# ./.github/workflows/publish.yml
+name: Publish
```
nperez0111, line 2: We now have automatic publishing, need to work out the kinks and then will write a guide on how to release stuff

.github/workflows/build.yml
nperez0111: Can see an example on this PR: https://github.com/TypeCellOS/BlockNote/pull/1567#issuecomment-2761268580

docs/package.json
```diff
@@ -4,18 +4,18 @@
...
-    "@blocknote/react": "^0.26.0",
-    "@blocknote/shadcn": "^0.26.0",
-    "@blocknote/xl-multi-column": "^0.26.0",
+    "@blocknote/ariakit": "workspace:*",
+    "@blocknote/core": "workspace:*",
+    "@blocknote/code-block": "workspace:*",
+    "@blocknote/mantine": "workspace:*",
+    "@blocknote/react": "workspace:*",
+    "@blocknote/shadcn": "workspace:*",
+    "@blocknote/xl-multi-column": "workspace:*",
```
nperez0111, line 18: `workspace:*` is PNPM specific to ensure that the docs & playground only ever reference the locally installed workspace version. Otherwise, it can technically be satisfied by anything matching the range. So, this syntax makes it explicit that we always want pnpm to symlink to the local package instead.

packages/react/src/hooks/useUIElementPositioning.ts
```diff
@@ -12,7 +12,14 @@ export function useUIElementPositioning(
   referencePos: DOMRect | null,
   zIndex: number,
   options?: Partial<UseFloatingOptions>
-) {
+): {
```
matthewlipski, line 15: Can we put this in a separate type declaration? There's just a lot going on here

tests/playwright.config.ts
```diff
@@ -95,10 +96,11 @@ const config: PlaywrightTestConfig = {
   // outputDir: 'test-results/',
 
   /* Run your local dev server before starting the tests */
-  // webServer: {
-  //   command: process.env.DEVSERVER ? "npm run start" : "npm run start:built",
-  //   port: 3000,
-  // },
+  webServer: {
```
matthewlipski, line 99: Unless this is commented out, I'm getting errors trying to run `test:updateSnaps` - seems like pnpm is not recognized in the Docker container? @YousefED can you see if this is the case for you? Went over it on call with Nick and we couldn't find a solution, so would be at least good to know if it's just a problem on my side.

package.json
```diff
@@ -1,39 +1,53 @@
...
-    "deploy": "lerna publish -- --access public",
-    "prepublishOnly": "npm run build && cp README.md packages/core/README.md && cp README.md packages/react/README.md",
+    "dev": "nx run @blocknote/example-editor:dev",
+    "dev:docs": "nx run docs:dev",
+    "build": "nx run-many --target=build",
+    "build:clean": "pnpm run clean && pnpm run gen && pnpm run clean && pnpm run build",
+    "build:site": "pnpm run gen && nx run docs:build:site",
+    "clean": "nx run-many --target=clean",
+    "deploy": "nx release --skip-publish",
+    "gen": "nx run @blocknote/dev-scripts:gen",
```
matthewlipski, line 41: Minor thing, but for whatever reason, the logs when generating are only dumped into the console after `gen:examples`/`gen:docs` run. This isn't a big deal, it's just `gen:docs` especially takes a minute to run and looks like it's stuck during that time.

package.json
```diff
@@ -1,39 +1,53 @@
...
+    "install-playwright": "cd tests && pnpx playwright install --with-deps",
+    "e2e": "cd tests && pnpm exec playwright test",
+    "e2e:updateSnaps": "cd tests && pnpm run test:updateSnaps",
+    "lint": "nx run-many --target=lint",
     "postpublish": "rm -rf packages/core/README.md && rm -rf packages/react/README.md",
-    "clean": "lerna run --stream clean"
+    "prepublishOnly": "pnpm run build && cp README.md packages/core/README.md && cp README.md packages/react/README.md",
+    "prestart": "pnpm run build",
+    "start": "serve playground/dist -c ../serve.json",
+    "test": "nx run-many --target=test",
```
matthewlipski, line 50: Getting failures in `xl-multi-column`:

```
      ⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯
      Error: The module '/Users/matthewlipski/Documents/Work/BlockNote/BlockNote/node_modules/.pnpm/canvas@2.11.2_encoding@0.1.13/node_modules/canvas/build/Release/canvas.node'
      was compiled against a different Node.js version using
      NODE_MODULE_VERSION 115. This version of Node.js requires
      NODE_MODULE_VERSION 127. Please try re-compiling or re-installing
      the module (for instance, using `npm rebuild` or `npm install`).
       ❯ Object..node node:internal/modules/cjs/loader:1732:18
       ❯ Module.load node:internal/modules/cjs/loader:1289:32
       ❯ Function._load node:internal/modules/cjs/loader:1108:12
       ❯ TracingChannel.traceSync node:diagnostics_channel:322:14
       ❯ wrapModuleLoad node:internal/modules/cjs/loader:220:24
       ❯ Module.require node:internal/modules/cjs/loader:1311:12
       ❯ require node:internal/modules/helpers:136:16
       ❯ Object.<anonymous> ../../node_modules/.pnpm/canvas@2.11.2_encoding@0.1.13/node_modules/canvas/lib/bindings.js:3:18
       ❯ Module._compile node:internal/modules/cjs/loader:1554:14
       ❯ Object..js node:internal/modules/cjs/loader:1706:10
      
      ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
      Serialized Error: { code: 'ERR_DLOPEN_FAILED' }
```

@nperez0111 any ideas?

matthewlipski (commented):
Great changes, overall experience with nx + pnpm is much nicer than lerna + npm. Just left a few comments, mainly commands that I've encountered some issues with.

package.json
nperez0111: Yep, this is because we switch pnpm versions. Just need to run `pnpm -r rebuild` locally & shouldn't be an issue anymore

nperez0111:
This is ready for review now @YousefED

docs/package.json
```diff
@@ -41,14 +41,15 @@
...
     "autoprefixer": "^10.0.1",
     "eslint": "^8",
     "eslint-config-next": "14.1.0",
     "postcss": "^8",
     "prettier-plugin-tailwindcss": "^0.5.11",
     "tailwindcss": "^3.3.0",
     "typescript": "^5"
-  }
+  },
+  "includedScripts": ["build:site", "lint"]
```
YousefED, line 54: I see this is an nx option. added for "safety" ?

examples/01-basic/01-minimal/package.json
```diff
@@ -23,15 +22,6 @@
     "@types/react": "^18.0.25",
     "@types/react-dom": "^18.0.9",
     "@vitejs/plugin-react": "^4.3.1",
-    "eslint": "^8.10.0",
     "vite": "^5.3.4"
-  },
-  "eslintConfig": {
```
YousefED, line 29: intentional to remove eslint from examples, or is it solved in a different way?

nx.json
```diff
@@ -1,9 +1,23 @@
 {
+  "$schema": "./node_modules/nx/schemas/nx-schema.json",
+  "parallel": 6,
+  "release": {
```
YousefED, line 4: curious how the eventual flow works (as both github actions and nx are involved). didn't really check yet (will do when you have this ready)

packages/ariakit/package.json
```diff
@@ -5,6 +5,11 @@
   "sideEffects": [
     "*.css"
   ],
+  "repository": {
```
YousefED, line 8: needed for pkg.new?

package.json
```diff
@@ -1,39 +1,53 @@
...
-    "packages/*",
-    "playground",
-    "tests",
-    "docs"
-  ],
   "devDependencies": {
+    "@nx/js": "20.6.4",
     "@typescript-eslint/eslint-plugin": "^5.5.0",
     "@typescript-eslint/parser": "^5.5.0",
+    "concurrently": "9.1.2",
```
YousefED, line 6: used?

package.json
```diff
@@ -1,39 +1,53 @@
...
-    "bootstrap": "lerna bootstrap --ci -- --force && patch-package",
-    "install-new-packages": "lerna bootstrap -- --force && patch-package",
-    "test": "lerna run --stream test",
-    "install-playwright": "cd tests && npx playwright install --with-deps",
-    "deploy": "lerna publish -- --access public",
-    "prepublishOnly": "npm run build && cp README.md packages/core/README.md && cp README.md packages/react/README.md",
+    "dev": "nx run @blocknote/example-editor:dev",
+    "dev:docs": "nx run docs:dev",
+    "build": "nx run-many --target=build",
+    "build:clean": "pnpm run clean && pnpm run gen && pnpm run clean && pnpm run build",
```
YousefED, line 37: do we still need to have these manual clean commands or does nx have some magic for it?

package.json
```diff
@@ -1,39 +1,53 @@
...
+    "e2e": "cd tests && pnpm exec playwright test",
+    "e2e:updateSnaps": "cd tests && pnpm run test:updateSnaps",
+    "lint": "nx run-many --target=lint",
     "postpublish": "rm -rf packages/core/README.md && rm -rf packages/react/README.md",
-    "clean": "lerna run --stream clean"
+    "prepublishOnly": "pnpm run build && cp README.md packages/core/README.md && cp README.md packages/react/README.md",
+    "prestart": "pnpm run build",
+    "start": "serve playground/dist -c ../serve.json",
+    "test": "nx run-many --target=test",
+    "bootstrap": "echo 'do nothing'"
```
YousefED, line 51: delete?

packages/dev-scripts/examples/template-react/package.json.template.tsx
```diff
@@ -8,9 +8,8 @@ const template = (project: Project) => ({
   scripts: {
     start: "vite",
     dev: "vite",
-    build: "tsc && vite build",
-    preview: "vite preview",
-    lint: "eslint . --max-warnings 0",
+    "build:prod": "tsc && vite build",
```
YousefED, line 11: you named this build:prod because you don't want it included when running build in root, right?

packages/dev-scripts/examples/util.ts
```diff
@@ -1,4 +1,4 @@
-import glob from "fast-glob";
+import { globSync } from "tinyglobby";
```
YousefED, line 1: what was the reason for this? curious

packages/react/src/hooks/useUIElementPositioning.ts
```diff
@@ -7,12 +7,21 @@ import {
 } from "@floating-ui/react";
 import { useEffect, useMemo } from "react";
 
+type UIElementPosition = {
```
YousefED, line 10: I'm gessing a ts change or sth that made this necessary?

packages/xl-docx-exporter/src/docx/__snapshots__/withCustomOptions/document.xml.rels
```diff
@@ -14,4 +14,5 @@
     <Relationship Id="FAKE-ID" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="" TargetMode="External"/>
     <Relationship Id="FAKE-ID" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/hyperlink" Target="https://www.blocknotejs.org" TargetMode="External"/>
     <Relationship Id="FAKE-ID" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/image" Target="media/e1037269513fbdc121e06bb1bc7cf37b9045c8fc.gif"/>
+    <Relationship Id="FAKE-ID" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/fontTable" Target="fontTable.xml"/>
```
YousefED, line 17: triggered by an upgrade?

packages/xl-multi-column/package.json
```diff
@@ -65,7 +73,9 @@
     "typescript": "^5.3.3",
     "vite": "^5.3.4",
     "vite-plugin-eslint": "^1.8.1",
-    "vitest": "^2.0.3"
+    "vitest": "^2.0.3",
+    "react": "^19",
```
YousefED, line 77: should this be ^18?

packages/xl-pdf-exporter/src/pdf/defaultSchema/index.ts
```diff
@@ -2,7 +2,11 @@ import { pdfBlockMappingForDefaultSchema } from "./blocks.js";
 import { pdfInlineContentMappingForDefaultSchema } from "./inlinecontent.js";
 import { pdfStyleMappingForDefaultSchema } from "./styles.js";
 
-export const pdfDefaultSchemaMappings = {
+export const pdfDefaultSchemaMappings: {
```
YousefED, line 5: weird that this was needed, typescript issue?

YousefED (approved):
awesome!! looking forward to using this

.github/workflows/build.yml
```diff
@@ -33,32 +37,36 @@ jobs:
...
             nx-
 
-      - name: Install Dependencies
-        run: npm ci --prefer-offline --no-audit
+      - run: sudo apt-get update && sudo apt-get install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev pkg-config
 
-      - name: Bootstrap packages
-        run: npm run bootstrap
+      - name: Install Dependencies
+        run: pnpm install && pnpm -r rebuild
```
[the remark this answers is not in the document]
nperez0111: rebuild tells pnpm to rebuild any deps from scratch. But, seems like it is not needed, so I removed it. Was just trying to get the canvas dep working

.github/workflows/build.yml
nperez0111: It allows you to install the "final" package into another repository so it is very useful for testing in a real project.
Like, for that bug fix for docs, we could have made a PR with the change and they could install this version instead of having to actually release something

docs/package.json
nperez0111: Oh, I was just testing this, it allows hiding scripts from NX. Totally not needed

nx.json
nperez0111: Here is an nx doc on how it will work: https://nx.dev/recipes/nx-release/publish-in-ci-cd

Essentially:
`nx release` locally, which changes the `package.json` versions.
It will commit & tag with the new version number
The new version number tag is what triggers the workflow to run, which will build and publish the package.

package.json
nperez0111: Fine to keep them, but NX has it's own cache.

package.json
nperez0111: vercel is using it for now, will remove when merged

packages/dev-scripts/examples/template-react/package.json.template.tsx
nperez0111: Exactly right, pnpm run build should probably not build each example, just the packages

packages/react/src/hooks/useUIElementPositioning.ts
nperez0111: Yep, otherwise:

```
 boolean,
YousefED1 day ago
I'm gessing a ts change or sth that made this necessary?

The inferred type of 'useUIElementPositioning' cannot be named without a reference to '.pnpm/@floating-ui+react-dom@2.1.2_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/@floating-ui/react-dom'. This is likely not portable. A type annotation is necessary.
```

Triggered because I regenerated the node_modules so it upgraded TS

packages/xl-docx-exporter/src/docx/__snapshots__/withCustomOptions/document.xml.rels
nperez0111: Yep, upgrade of a patch version

packages/xl-multi-column/package.json
nperez0111: I put 19 to test against that, but 19 is stable now so it shouldn't matter which

examples/01-basic/01-minimal/package.json
nperez0111: eslint is global, so still works locally

the referenced config file never worked for the sandboxes, so better to just remove altogether

pnpm-workspace.yaml
```diff
@@ -0,0 +1,12 @@
+packages:
+  - "packages/*"
+  - "examples/*/*"
+  - "playground"
+  - "docs"
+  - "shared"
```
YousefED, line 6: @nperez0111 I'm not sure `shared` should be part of the workspace.

The way it is now, shared contains some code (utility functions) that's shared by different packages (mainly, the exporters). However, it's not bundled in it's own package. Rather, the functions are just included in the consuming packages

a) should we remove it here?
b) I tried to remove it, but then got an error in one of the exporters (TS errors in `toMatchBinaryFileSnapshot`)

nperez0111: The workspace defines where to install packages from. By default, pnpm installs packages "scoped" i.e. a project can only resolve the packages that it declares.

So, `shared` has dependency on `image-meta` (imageUtil) & `@types/node` (binarySnapshotUtil). So when you remove it from the workspace, it can no longer find those packages since they exist in the project they are used in, not in it's source path.

There are two ways to resolve this. Make `shared` into a workspace package (leveraging it only as an project, not a package), or install it's deps like I did.

nperez0111: Or, remove the shared project altogether & just keep the files into their respective projects, which I'm most partial to since it keeps things simple.

YousefED: > Or, remove the shared project altogether & just keep the files into their respective projects, which I'm most partial to since it keeps things simple.

I think for code duplication that's not ideal. I think specifically for test utilities we should actually move a lot more to the `shared` package (test cases and utilities that are shared across packages) - unless you have a better solution for this.

> There are two ways to resolve this. Make shared into a workspace package (leveraging it only as an project, not a package), or install it's deps like I did.

I'm not sure I understand the difference between the two. Currently it's a workspace package right?

nperez0111: > I think for code duplication that's not ideal. I think specifically for test utilities we should actually move a lot more to the shared package (test cases and utilities that are shared across packages) - unless you have a better solution for this.

I think test utilities should be in/reusing the existing test package: https://github.com/TypeCellOS/BlockNote/blob/7df528c2fe4333393067ff74bf3115c929a56333/tests/package.json#L2 it's a package, so it is perfectly valid to import/export things from it.

So, the only one that I think is valid to live here is the image util, and even then, I don't see why this can't be either duplicated or exported from `@blocknote/core` (ideally their would actually be a `@blocknote/core/exporter` sub package for this reason)

> I'm not sure I understand the difference between the two. Currently it's a workspace package right?

Sorry, I could've explained that better. The difference is:

 1. workspace package: `@blocknote/shared` as a devDep to whatever uses it (so linking via pnpm workspaces)
 2. tricking vite to resolve it like we are doing now: https://github.com/TypeCellOS/BlockNote/blob/27494f6ddcd5600e139eb3743e3e2e1682c43002/packages/xl-docx-exporter/vite.config.ts#L23-L25
 
In either of these cases, you still need anything `shared` depends to be installed to this directory (whether you consider it a package or not).

YousefED: Thanks.

For (1), wouldn't it require us to publish that package to npm? (if we use functions from it in non-testing code)?

nperez0111: Nope, you can set it up where `@blocknote/shared` is declared as not "external" (i.e. vite/rollup should bundle it into the package that you are trying to build like docx-exporter). So, at build time the import would be resolved correctly within the repo & the consumer would not need that package.

YousefED: Ah makes sense. I suppose that's cleaner than what we currently have, because dependencies are better managed right?

Should we go for that approach? Also happy to discuss whether the package is needed at all - I think easier to go over that on a call

PR #1580, opened 2025-04-01
chore: update german translation

joh-klein:


joh-klein:
I'm from [dytab](https://github.com/dytab/ngx-blocknote) as well :)

Issue #829, opened 2024-06-10
Calling `editor.updateBlock()` in quick succession breaks the editor's undo/redo functionality


balu-deciphr:
**Describe the bug**
I'm using an (old) [example](https://github.com/TypeCellOS/BlockNote/blob/v0.11.2/packages/website/docs/docs/cursor-selections.md#demo-highlighting-blocks-spanned-by-selection) from the docs to highlight blocks spanned by the current selection. The highlighting works just fine, but this seems to break the editor's undo/redo functionality.

**To Reproduce**
The demo is [here](https://stackblitz.com/edit/github-vyxdkh?file=data.ts,App.tsx).

To reproduce the issue, please do the following:
1. Type something in the initialized content
2. Press `ctrl + z` or `cmd + z` to undo what you typed. It will take multiple attempts to undo just one character.
3. Remove the `onSelectionChange` prop from `<BlockNoteView />`
4. Type something again in the editor
5. Now the undo should work just fine.

It looks like the `editor.updateBlock()` calls within the `editor.forEachBlock()` causes the editor's history to break. Is there an alternate way to update the style of blocks?

**Misc**
- Node version: 20.11.1
- Package manager: npm
- Browser: Firefox 126.0.1, Google Chrome 125.0.6422.142

PR #1584, opened 2025-04-04
feat(core): add blocknote transactions

nperez0111:
This implements BlockNote transactions, which batch multiple dispatched transactions into 1 transaction

supercedes: https://github.com/TypeCellOS/BlockNote/pull/1147

Resolves: https://github.com/TypeCellOS/BlockNote/issues/829

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -800,41 +800,27 @@ export class BlockNoteEditor<
       return callback();
     }
 
-    let result: T = undefined as T;
-
     try {
       // Enter transaction mode, by setting a start state
       this.transactionState = this.prosemirrorState;
 
-      // Capture all tiptap transactions (tiptapEditor.dispatch'd transactions)
-      const tiptapTr = this._tiptapEditor.captureTransaction(() => {
-        // This is more of a safety mechanism, as we catch blocknote transactions separately
-        result = callback();
-      });
+      // Capture all dispatch'd transactions
+      const result = callback();
 
       // Any transactions captured by the `dispatch` call will be stored in `this.activeTransaction`
-      let activeTr = this.activeTransaction;
+      const activeTr = this.activeTransaction;
 
-      if (tiptapTr && activeTr) {
-        // If we have both tiptap & blocknote transactions, there is not a clean way to merge them as you'd need to know the order of operations
-        throw new Error(
-          "Cannot mix tiptap transactions with BlockNote transactions"
-        );
-      } else if (tiptapTr) {
-        // If we only have tiptap caught transactions, we can just use that
-        activeTr = tiptapTr;
-      }
-
+      this.transactionState = null;
       if (activeTr) {
+        this.activeTransaction = null;
         // Dispatch the transaction if it was modified
-        this._tiptapEditor.dispatch(activeTr);
+        this.dispatch(activeTr);
       }
+      return result;
     } finally {
       this.activeTransaction = null;
       this.transactionState = null;
     }
-
-    return result;
   }
 
   /**
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -94,18 +94,24 @@ import {
 import { Dictionary } from "../i18n/dictionary.js";
 import { en } from "../i18n/locales/index.js";
 
-import { Plugin, TextSelection, Transaction } from "@tiptap/pm/state";
+import {
+  EditorState,
+  Plugin,
+  Selection as ProsemirrorSelection,
+  TextSelection,
+  Transaction,
+} from "@tiptap/pm/state";
 import { dropCursor } from "prosemirror-dropcursor";
 import { EditorView } from "prosemirror-view";
 import { ySyncPluginKey } from "y-prosemirror";
 import { createInternalHTMLSerializer } from "../api/exporters/html/internalHTMLSerializer.js";
 import { inlineContentToNodes } from "../api/nodeConversions/blockToNode.js";
 import { nodeToBlock } from "../api/nodeConversions/nodeToBlock.js";
+import { nestedListsToBlockNoteStructure } from "../api/parsers/html/util/nestedLists.js";
+import { CodeBlockOptions } from "../blocks/CodeBlockContent/CodeBlockContent.js";
 import type { ThreadStore, User } from "../comments/index.js";
 import "../style.css";
 import { EventEmitter } from "../util/EventEmitter.js";
-import { CodeBlockOptions } from "../blocks/CodeBlockContent/CodeBlockContent.js";
-import { nestedListsToBlockNoteStructure } from "../api/parsers/html/util/nestedLists.js";
 
 export type BlockNoteExtensionFactory = (
   editor: BlockNoteEditor<any, any, any>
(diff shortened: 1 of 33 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -749,10 +863,18 @@ export class BlockNoteEditor<
     return this._tiptapEditor.view;
   }
 
+  /**
+   * The state of the editor when a transaction is captured, this can be continuously updated during the {@link transact} call
+   */
+  private transactionState: EditorState | null = null;
```
YousefED, line 869: are we sure we should store an `EditorState`, or a `transaction`? any trade-offs?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -726,9 +732,117 @@ export class BlockNoteEditor<
...
+   *   const otherTr = editor.transaction;
+   *   otherTr.insertText("Hello, world!");
+   *   editor.dispatch(otherTr);
+   * });
+   * ```
+   */
+  public transact<T>(callback: () => T): T {
+    if (this.transactionState) {
+      // Already in a transaction, so we can just callback immediately
+      return callback();
```
YousefED, line 800: I have the feeling we're not compatible with multiple nested `transact` calls. If I see it correctly, while "entering" should work, on "exit", we'd immediately dispatch when we exit the nested transaction.

If this is correct, I think we either need to fix it, or just throw an error here whene entering a nested transaction

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -726,9 +732,117 @@ export class BlockNoteEditor<
...
+   * ```ts
+   * const tr = editor.transaction;
+   * tr.insertText("Hello, world!");
+   * editor.dispatch(tr);
+   * ```
+   */
+  public dispatch(tr: Transaction) {
+    if (this.transactionState) {
+      const { state, transactions } =
+        this.transactionState.applyTransaction(tr);
```
YousefED, line 748: this would apply plugins, right? 
- do we want this?
- Should we check whether all plugins are compatible with this? i.e.: the state that they get will be different from the "actual editor state" (as the actual editor - or at least, the view - still has an old state)

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -749,10 +863,18 @@ export class BlockNoteEditor<
...
+   * The state of the editor when a transaction is captured, this can be continuously updated during the {@link transact} call
+   */
+  private transactionState: EditorState | null = null;
+
   /**
    * Get the underlying prosemirror state
    */
   public get prosemirrorState() {
+    if (this.transactionState) {
+      return this.transactionState;
```
YousefED, line 876: this can now be not in sync with the view. can this cause downstream issues you think?

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: This would apply the `filterTransaction` and `appendTransaction` events.
Both of these have function signatures provide the current editorState as an argument [filterTransaction](https://prosemirror.net/docs/ref/#state.PluginSpec.filterTransaction) and [appendTransaction](https://prosemirror.net/docs/ref/#state.PluginSpec.appendTransaction). So operating on the editor state they were given is what they should be doing anyway (since they are run during state transitions), it would be incorrect for them to be pulling anything from `editor.prosemirrorState`, `editor._tiptapEditor.state` or `editor._tiptapEditor.view.state`.

I've checked our plugins and this should not be an issue, I did not have to change any of the plugins for this.

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: Pretty sure that the logic is good here. The goal here is to only actually dispatch the active transaction until anything within the callback has been run. Which, I think is achieved here, even with nested calls.

Take this example for reference:

```txt
❯ node                                                                            
Welcome to Node.js v22.14.0.
Type ".help" for more information.
> let inTr = false
undefined
> function transact(cb) { if (inTr) { return cb() }
... inTr = true;
... const result = cb();
... console.log('dispatch');
... inTr = false;
... }
undefined
> (() => {
...   console.log("entering tr");
...   transact(() => {
...     console.log("in tr", inTr);
...     console.log("entering nested");
...     transact(() => {
...       console.log("in nested", inTr);
...     });
...     console.log("finished nested", inTr);
...   });
...   console.log("done with tr", inTr);
... })();
entering tr
in tr true
entering nested
in nested true
finished nested true
dispatch
done with tr false
undefined
```

Nested transactions should always be hitting the `if (this.transactionState) {}` branch, since they are within an outer transaction which would set this to a truthy value.
They'd have no "exit", since their callback is immediately executed, so they would never hit the automatic dispatch of the active transaction.

So, I'm pretty sure the logic is fine

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: The only way to implement a transaction mechanism is for the editor state to be temporarily de-coupled from the view state, so that a longer transaction can be "chained"/"built up". An alternative to this sort of API is Tiptap's chained fluent API where the transaction isn't applied until a specific call is made.

Personally, I like the chained API approach, but it forces your code to be rewritten to it, rather than your code leveraging it. It's like the difference between React & jQuery. React is not portable, but jQuery is

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: We also need the plugins to apply because after an `editor.dispatch` the plugins would normally be assumed to have already run, and we don't want to break this assumption just because the parent called the code within a transaction.

packages/core/src/editor/BlockNoteEditor.ts
YousefED: Another thing that could go wrong is calling `dispatch` and then trying to find the node in the view (dom), right? For example, getting an element by id. Previously this would have worked, so I suppose this is also something to check / be mindful of?

packages/core/src/editor/BlockNoteEditor.ts
YousefED: > The only way to implement a transaction mechanism is for the editor state to be temporarily de-coupled from the 
view state

An alternative is to not "de-couple" the state at all, but explicitly expose a transaction like I did in #1147, right? afaik you don't _need_ any magic / chaining at all, as long as your methods can operate on prosemirror transactions. I think this is the most prosemirror-native approach?

Trying to establish the pros / cons:

### This PR
- you can "magically" use `prosemirrorState`, and in most cases "everything should just work".
- pro: you can always read prosemirrorState after a dispatch
- con: `prosemirrorState` is a bit magic / might not reflect the actual state of the editor (view). This _could_ cause errors down the line

### #1147
- requires methods to be more aware of a transactions; i.e.: you should continue with an existing transaction if that's available on the editor, and read from `tr.doc`. This might be more cumbersome, but normally would all be abstracted away for consumers
- con: a `dispatch` call won't result in an updated `prosemirrorState` if you're in a transaction
- pro: because it's more explicit, maybe less "risky" / error-prone?

The approaches also differ in plugin handling but if I see it correctly we _could_ use any plugin approach with both of the solutions.

I'm not sure I have a clear preference. I like the ease of being able to access `prosemirrorState` in one approach, but it also feels more on the safe side (and easier to reason about the code paths) to do as little "magic patching" as possible especially in the core internals.

I might also be missing some pros/cons, so curious to your thoughts

packages/core/src/editor/BlockNoteEditor.ts
YousefED: ah ofc, you're right. might be worth a test case but definitely misinterpreted this

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: > An alternative is to not "de-couple" the state at all, but explicitly expose a transaction like I did in https://github.com/TypeCellOS/BlockNote/pull/1147, right?
Sorry, yes that is right.

I tried that approach when implementing this PR. I was able to implement everything at one point to use the current transaction as the source of truth for the state.

While this did work, it didn't feel "right" to start a transaction when just trying to read the current state. I would be more comfortable with this approach if, we did not expose the `editor.prosemirrorState` at all, and chose to only expose the `editor.transaction`. This would prevent confusion between doing the "right thing" (editor.transaction) and the "wrong thing"(editor.prosemirrorState), by only exposing the right thing.

Additionallly, I think that this would work if we had full control of all the callsites, but what is slightly problematic about this approach is external code.

For example, take this random command from `prosemirror-table` [which takes an editor state to execute the command](https://github.com/ProseMirror/prosemirror-tables/blob/5cf89e5fdfeba0b80c6eb908e4521ad8035b49ab/src/commands.ts#L340-L343), if called by us in a transaction, we have to generate an editor state for it to be called appropriately (with whatever is actually the most recent state). I think this is something that _we should be solving for our callers_, our job is to shield complexity from developers by making it a non-issue.

matthewlipski (commented):
Seems like @YousefED already covered the design decisions so just added a few questions about a few bits of code.

If I understand correctly, the main difference/point of contention between the BlockNote transactions and the underlying ProseMirror ones is:

- In PM you create a new set of transactions from `state.tr`, in which you should only refer to `tr.state`, before applying them with `state.applyTransaction`.
- In this PR, you call `editor.transact` to create/add to a set of transactions, after which you can refer to `editor.prosemirrorState` like usual before applying them with `editor.dispatch`.

So the advantage of this is that as a developer, you don't have to worry about transaction state vs editor state, and just have the editor state exposed to you. But the downside is you need to make sure that after calling `editor.transact`, you need to call `editor.dispatch` ASAP or the editor state gets desynced.

Am I getting that right?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -726,9 +732,117 @@ export class BlockNoteEditor<
...
+        this.transactionState.applyTransaction(tr);
+      // Set a default value if needed
+      const accTr = this.activeTransaction ?? this.transactionState.tr;
+
+      // Copy over the newly applied transactions into our "active transaction" which accumulates all transaction steps during a `transact` call
+      transactions.forEach((tr) => {
+        tr.steps.forEach((step) => {
+          accTr.step(step);
+        });
+        if (tr.selectionSet) {
```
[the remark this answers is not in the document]
nperez0111: Selections aren't stored as a `Step`, they are stored in a separate field. So, to preserve selections between, we have to copy it over separately.

You'll also notice that I serialize it to JSON, because technically the document in the dispatch'd transaction is a different reference than the accumulated `activeTransaction`

Issue #1541, opened 2025-03-18
Putting a watcher on blocks add / edition / removal


ClementEXWiki:
**Is your feature request related to a problem? Please describe.**
For a current project, I need to create some custom blocks that automatically change depending on the editor's content. One such block is a table of contents, which should be updated dynamically in real time as soon as headings are added / removed / modified anywhere in the editor.

**Describe the solution you'd like**
It would be extremely useful to have a way to _observe_ some changes in the editor, specifically related to some block types. For instance, an API on the editor that runs a function as soon as a block of type `heading` is added / removed / modified.

**Describe alternatives you've considered**
There are alternatives but none are ideal:
* Re-implementing the heading types (or other blocks if needed) and trigger an event manually when they are updated
* Putting a listener on the editor to watch for *any* change, which would require to traverse the entire AST to find headings and make a diff against the previous state
* Showing a button to manually refresh the ToC block's content, but this requires manual user interaction

**Additional context**
I gave an example with headings but it could really be anything else.

I have absolutely no idea how hard such a feature would be to implement in BlockNote

YousefED:
We'd love to work on this as I think it would unlock all sorts of interesting scenarios.

I think we should implement it with some additional requirements in mind:
- metadata about the source of the change (e.g. remote update, paste, user edit (typing), if we have this info)
- in the case of a remote update: who made the change (not sure if we have this)
- besides add / delete / update, we also need to think about what happens when a child block of a parent block has changed

Top of head; I see two ways to implement this:
- derived from prosemirror changes (listen to transactions and detect changed ranges -> transform to the API we need)
- alternatively, we could iterate through all blocks in the document with `nodeToBlock` and see which blocks have changed. Because we cache the return value of `nodeToBlock`, we should be able to compare this by reference, and it might not be as "expensive" as it sounds

ClementEXWiki:
That seems like a very good idea, having as much informations as we can will ensure the watcher won't be limited.

nperez0111:
I implemented something similar for Tiptap V3: https://github.com/ueberdosis/tiptap/pull/6022

I went with the prosemirror change approach (since that was the most obvious for Tiptap).

Issue #1566, opened 2025-03-26
onDelete, onRemove or onRemoveBlocks event handler


coxato:
**Is your feature request related to a problem? Please describe.**
I would like to have a specific event handler when a block or blocks are deleted, because I have some linked blocks that are being added as pairs, e.g (startBlock and endBlock) so if I delete a startBlock the endBlock is automatically deleted, because they are "pairs" in my implementation. 
So I don't want to do that validation on the `onChange` method every time.

**Describe the solution you'd like**
Having a `onDelete` or `onRemove` or `onContentDeleted` or `onRemoveBlocks` event, so it receives a callback with the deleted blocks[] as argument. 
something like this:
```js
  const editor = useCreateBlockNote();

  useEffect(() => {
    if(editor){
      // a function to know when blocks are deleted so we don't use onChange every time  
      editor.onRemoveBlocks((deletedBlocksArray) => {
        myCustomHandler(deletedBlocksArray);
      })
    }
  }, [editor]);
```

**Describe alternatives you've considered**
Right now I do a validation for ALL the blocks on every `onChange`, but I think isn't the best approach, I would like to do those validations only when a block is deleted, if we are editing or creating/adding blocks I don't need to do extra validations. 

is there any other possible solution?

nperez0111:
This is related to an issue we are tracking: https://github.com/TypeCellOS/BlockNote/issues/1566

Tim-Pet:
Hi there, this is an awesome feat! Are there any plans to expose block specific changes on a block level as well (so the change logic could get handled inside custom blocks)?

e.g.: 
`block.onChange(e => e.preventDefault(); window.alert("All links to this block have been deleted successfully"))`

nperez0111:
@Tim-Pet I don't completely understand what you are trying to achieve with this.

But, it does not make sense to expose this sort of API within a custom block, because the change is derived from what _already happened_, so the custom block's instance will just have been destroyed (or at least about to), so it wouldn't really be able to do anything about it.

Tim-Pet:
Ah fair point, thanks for pointing me in the right direction. 

> [@Tim-Pet](https://github.com/Tim-Pet) I don't completely understand what you are trying to achieve with this.
In the end I want to prevent certain blocks from deletion or apply custom behavior for an improved UX for complex custom blocks


[...]

Issue #482, opened 2023-12-16
Feature Request: more secure way to get image URL and deleteFile callback


ryendu:
I love the image block feature and the `uploadFile` method in the editor that allows you to define a custom function for uploading images to your server. To improve on this, I wish there would be an equivalent `deleteFile` callback invoked when the image block is deleted so we can delete the image from our database to save resources. I also think that defining the image as just a public URL is not the most secure, and we could also define a `getFile` method that will allow developers to define a custom function to retrieve a signed URL that expires after a certain duration.

i-am-chitti:
I recently used the library and felt the same blocker. Requirement of `deleteFile` is ➕ ➕ . 
@matthewlipski I'm interested in contributing to it. Let me know if I can take this.

YousefED:
Thanks. Both `deleteFile` and `getFile` are good suggestions.

What's tricky with `deleteFile` this is that a user could delete a block and then re-add it using cmd/ctrl+z. How would you go about this? Happy to accept contributions ofc!

ahmedbna:
Hey, any updates on `deleteFile` functionality?

matthewlipski:
Not at the moment unfortunately:/ the updates to file blocks were done as part of a client request so most likely `deleteFile` will be added also from client feedback or when we have time to implement it after some higher prio issues.

HenrikBechmann:
I think the most direct approach to the undo problem is to provide a message to the user if a file is not successfully loaded indicating that the file is not available, and that the block should be removed and a new one added.

A more sophisticated approach would be to provide a reload option, which would replace an existing file or supply a missing file. This would require a fileDelete trigger for any replaced file.

How do you handle duplicate file names?

HenrikBechmann:
In the meantime, is there a way to extend the editor specs for useCreateBlockNote to include a `onDeleteFile` property?

HenrikBechmann:
Here's what I'm doing in the meantime (using firebase):
My editor is modal ('display', 'edit').

When uploading a file, I record the filename in an array, saved to the same firestore document as the editor content
```
    async function uploadFile(file:File) {

        if (editorUploadedFiles.includes(file.name)) { // avoid invalidation of the first
            alert(file.name + ' has already been uploaded') // placeholder
            return null
        }

        const fileRef = ref(storage, workboxHandler.editRecord.profile.workbox.id + '/document/' + file.name)
        try {
            await uploadBytes(fileRef, file)
        } catch (error) {
[...]

PR #1585, opened 2025-04-04
feat: `change` event allows getting a list of the changes made

nperez0111:
This implements a `getChanges` API which will derive the changes made within a transaction as a list of: inserts, updates and deletions of blocknote blocks

To test this you can update an example like so:

```diff
diff --git i/examples/01-basic/01-minimal/App.tsx w/examples/01-basic/01-minimal/App.tsx
index a3b92bafd..d5c43ac2b 100644
--- i/examples/01-basic/01-minimal/App.tsx
+++ w/examples/01-basic/01-minimal/App.tsx
@@ -8,5 +8,12 @@ export default function App() {
   const editor = useCreateBlockNote();
 
   // Renders the editor instance using a React component.
-  return <BlockNoteView editor={editor} />;
+  return (
+    <BlockNoteView
+      editor={editor}
+      onChange={(t, ctx) => {
+        console.log("changes", ctx.getChanges());
+      }}
+    />
+  );
 }

```

Things left:
 - [x] add tests
 
 Resolves https://github.com/TypeCellOS/BlockNote/issues/1541 https://github.com/TypeCellOS/BlockNote/issues/482 https://github.com/TypeCellOS/BlockNote/issues/1566


packages/core/src/api/nodeUtil.ts
```diff
@@ -36,3 +53,192 @@ export function getNodeById(
...
+      type: transaction.getMeta("history$").redo ? "redo" : "undo",
+    };
+  } else if (transaction.getMeta("y-sync$")) {
+    source = { type: "remote" };
+  }
+
+  const changes: BlocksChanged<BSchema, ISchema, SSchema> = [];
+  // TODO when we upgrade to Tiptap v3, we can get the appendedTransactions which would give us things like the actual inserted Block IDs.
+  // since they are appended to the transaction via the unique-id plugin
+  const combinedTransaction = combineTransactionSteps(transaction.before, [
```
YousefED, line 156: I don't completely understand this. How do we get the block ids now then, if we don't yet support the uniqueid transactions?

packages/core/src/api/nodeUtil.test.ts
```diff
@@ -0,0 +1,361 @@
...
+
+const getEditor = setupTestEnv();
+
+describe("Test getBlocksChangedByTransaction", () => {
+  let transaction: Transaction;
+  let editor: BlockNoteEditor;
+  let originalDispatch: typeof editor.dispatch;
+
+  beforeEach(() => {
+    transaction = undefined as unknown as Transaction;
```
YousefED, line 17: not easier to test the onchange event directly instead of patching like this?

packages/core/src/api/nodeUtil.test.ts
```diff
@@ -0,0 +1,361 @@
...
+
+import { setupTestEnv } from "./blockManipulation/setupTestEnv.js";
+import { getBlocksChangedByTransaction } from "./nodeUtil.js";
+import { Transaction } from "prosemirror-state";
+import { BlockNoteEditor } from "../editor/BlockNoteEditor.js";
+import { Step } from "prosemirror-transform";
+
+const getEditor = setupTestEnv();
+
+describe("Test getBlocksChangedByTransaction", () => {
```
YousefED, line 11: Let's add test-cases for child blocks (add / update / delete child blocks). We can debate whether this should trigger a change on the parent or not, but at least it should be tested / documented imo

YousefED (commented):
this is so awesome, and implementation actually cleaner / clearer than expected! really cool

packages/core/src/api/nodeUtil.ts
```diff
@@ -36,3 +53,192 @@ export function getNodeById(
...
+  editor: BlockNoteEditor<BSchema, ISchema, SSchema>
+): BlocksChanged<BSchema, ISchema, SSchema> {
+  let source: BlockChangeSource = { type: "local" };
+
+  if (transaction.getMeta("paste")) {
+    source = { type: "paste" };
+  } else if (transaction.getMeta("uiEvent") === "drop") {
+    source = { type: "drop" };
+  } else if (transaction.getMeta("history$")) {
+    source = {
```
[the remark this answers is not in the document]
nperez0111: This works only for the history plugin. But, I think y-prosemirror tells us if an undo happened (unsure if locally or remote or both), so maybe I can do it for that too

packages/core/src/api/nodeUtil.ts
nperez0111: Tiptap V2 gives the [root tr that was applied](https://github.com/ueberdosis/tiptap/blob/develop/packages/core/src/Editor.ts#L501) which does not include any appended transactions (since it is only the root tr).
Tiptap V3 gives both root tr & any appended transaction [in that callback](https://github.com/nperez0111/tiptap/blob/8de8e13b085e42d540f337e44b8439f96e9f2c7e/packages/core/src/Editor.ts#L593-L594) which means you can see the appended transactions & re-combine them to get the "actual" changed ranges.

Technically, the editor state is up to date by the time we are reading this, but we can't know what transactions have been appended so our transaction ranges can be incorrect.

The reason that we are getting ids out of this right now is that nodeToBlock will generate an ID if it does not find one. But, those IDs are not the same ID that actually would have been persisted into the document.

packages/core/src/api/nodeUtil.test.ts
nperez0111: Likely easier, but the test is not for the change event, but to see if a transaction can give us the changes. I'd want it isolated

packages/core/src/api/nodeUtil.test.ts
nperez0111: agreed should be in the tests

packages/core/src/api/nodeUtil.ts
matthewlipski: What do we actually need to need the UniqueId transactions for? Since this is more for tracking user changes, so giving a new block an ID doesn't seem like smth we need to listen for. So is the issue just that if a new block is created, when we try to get it with `nodeToBlock`, it will generate a random ID as the UniqueId transaction is not detected here?

packages/core/src/api/nodeUtil.ts
```diff
@@ -36,3 +53,204 @@ export function getNodeById(
...
+
+  // All blocks that are newly removed (since they did not exist in the next state)
+  const removedBlockIds = Array.from(prevBlockIds).filter(
+    (id) => !nextBlockIds.has(id)
+  );
+
+  removedBlockIds.forEach((blockId) => {
+    changes.push({
+      type: "delete",
+      block: prevAffectedBlocks.find((block) => block.id === blockId)!,
```
matthewlipski, line 235: Shouldn't `prevBlock` be set instead of `block`? Since the block previously existed and no longer does

examples/07-collaboration/01-partykit/App.tsx
```diff
@@ -30,5 +30,12 @@ export default function App() {
   });
 
   // Renders the editor instance.
-  return <BlockNoteView editor={editor} />;
+  return (
+    <BlockNoteView
+      editor={editor}
+      onChange={(e, { getChanges }) => {
```
matthewlipski, line 36: TODO: revert?

packages/core/src/api/nodeUtil.test.ts
```diff
@@ -0,0 +1,787 @@
...
+    const blocksChanged = getBlocksChangedByTransaction(transaction, editor);
+    expect(blocksChanged).toEqual([]);
+  });
+
+  it("should return blocks inserted by a transaction", () => {
+    editor.insertBlocks([{ type: "paragraph" }], "paragraph-0", "after");
+
+    const blocksChanged = getBlocksChangedByTransaction(transaction!, editor);
+
+    expect(blocksChanged).toEqual([
```
matthewlipski, line 48: Do we wanna use `toEqual` in these tests? We generally use `toMatchFileSnapshot` to verify JSON objects, and it would make this file a lot smaller

matthewlipski: + the JSON to compare with is generated automatically so writing the tests is a bit less cumbersome

packages/core/src/api/nodeUtil.ts
nperez0111: I thought of doing that too. But felt that if someone just wanted, "list of all changed blocks", it'd be just `getChanges().map(c=>c.block)`, whereas branching on the type is more cumbersome.

I think an argument can be made for either way (i.e. the type is 'deleted' so obviously `block` is the deleted block)

examples/07-collaboration/01-partykit/App.tsx
nperez0111: Yea whoops

packages/core/src/api/nodeUtil.test.ts
nperez0111: Habit, but, yea I guess that is slightly better.

nperez0111:
Wouldn't let me reply so I'll reply here @matthewlipski 

> What do we actually need to need the UniqueId transactions for? Since this is more for tracking user changes, so giving a new block an ID doesn't seem like smth we need to listen for. So is the issue just that if a new block is created, when we try to get it with nodeToBlock, it will generate a random ID as the UniqueId transaction is not detected here?

The reason the uniue-id transaction is important is that if you do `getChanges`, any inserted block technically hasn't been assigned an ID yet, it just generates one at random. So, you can't actually trust that ID to do lookups anywhere (in the document or otherwise). So, in the future, we can actually append that transaction to get the correct ID and not have that issue (the ID reported by `getChanges` would be accurate to what is actually in the document.

packages/core/src/api/nodeUtil.ts
nperez0111: The reason the uniue-id transaction is important is that if you do `getChanges`, any inserted block technically hasn't been assigned an ID yet, it just generates one at random (in `nodeToBlock` as you described). So, you can't actually trust that ID to do lookups anywhere (in the document or otherwise). So, in the future, we can actually append that transaction to get the correct ID and not have that issue (the ID reported by `getChanges` would be accurate to what is actually in the document.

packages/core/src/api/nodeUtil.ts
matthewlipski: Yep fair, I don't have a strong opinion one way or another, just wanted to make sure it was intentional

packages/core/src/api/nodeUtil.ts
YousefED: > The reason that we are getting ids out of this right now is that nodeToBlock will generate an ID if it does not find one. But, those IDs are not the same ID that actually would have been persisted into the document.

is this a blocker for the PR? it makes the "insert" changes not very useful then, right?

nperez0111: It depends on what they are using this for. If this is just to inspect the content to do things like remove image uploads, then it won't matter.

I'm uncertain of what you'd need the insert changes for, much less whether it is a hard requirement to have the ID be correct. I get that correctness is important, in general. But, I hope that Tiptap V3 is around the corner and alleviates us of this.

Thinking out loud: One way to get around this would be to get the previous editor state & apply the transaction again (which would run the plugins again). But, the ID would still be incorrect since it was randomly generated. So, not helpful either.
Another way would be to try to find the node afterward, but technically any plugin could append a transaction that makes the position no longer valid. So not helpful in the general case, but probably fine just for blocknote.

nperez0111: I think I've solved it. We just override the Tiptap `dispatchTransaction` method & re-implement it for ourselves.

I made this into a new event type to not disrupt any callers (even though it only adds a new field to the context). It also makes it clear that we should clean this up when we get off of Tiptap V2.

The switch to v3 then becomes only having to switch which event type to listen for, simple!

https://github.com/TypeCellOS/BlockNote/pull/1585/commits/06b4306b50e9521120465545f90c1edac7e540f9

PR #1590, opened 2025-04-07
docs: add docs for the release workflow

nperez0111:
This PR adds some documentation for how to make a BlockNote release. Let's also use this PR to discuss things like:
 - The CHANGELOG.md format, do we need it? Should we configure it, what should it include?
 - Should GitHub Releases be done automatically?
 - Should we run tests before attempting to publish the package?
 - What should the process for a release be? Do we need a checklist?

Is this the right process? It is the easiest we can get away with `nx` right now, but should we think about doing other things like:
 - release branches
 - GitHub Actions manual triggers for starting a release

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,15 +55,18 @@ This diagram illustrates the release workflow for the BlockNote monorepo.
 
 Essentially, when the maintainers have decided to release a new version of BlockNote, they will:
 
- 1. Check that the `main` branch is in a releasable state (i.e. all tests are passing, etc.)
+ 1. Check that the `main` branch is in a releasable state:
+    - CI status of main branch is green
+    - Builds are passing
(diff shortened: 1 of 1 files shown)


YousefED (commented):
Nice! see inline. Re your discussion points:

> The CHANGELOG.md format, do we need it? Should we configure it, what should it include?

I think it's great as Release notes in github "lock us in" to github - so better to have this available directly in the git repo

> Should GitHub Releases be done automatically?
> Should we run tests before attempting to publish the package?
> What should the process for a release be? Do we need a checklist?
> Is this the right process? It is the easiest we can get away with nx right now, but should we think about doing other things like: release branches / github actions

I'd say it'd be great to have some stronger checks (e.g.: unit tests) automatically when doing a release, and maybe even make the workflow simpler with an Action button, but tbh I don't think it's worth optimizing this further at this moment. You've already streamlined the process a lot; let's go with this and learn+adapt as we go

PR #1591, opened 2025-04-07
fix: allow opening another suggestion menu if another is triggered #1473

### LINKED ISSUE #1473: / menu not working after a colon
If a sentence ends with ':', the '/' menu stops working.

Image

nperez0111:
This resolves #1473 by allowing the suggestion plugin to switch from one menu to another if it encounters another trigger character

https://github.com/user-attachments/assets/48765b2c-0cce-46d8-9432-e7490d704a4f

This also resolves another issue I ran into personally.

If you type something that ends with `:` that menu is carried over to the new block, even though it shouldn't here is a video before & after.

Before:

https://github.com/user-attachments/assets/8123f0e1-8107-464a-9597-a5da187948de

After:

https://github.com/user-attachments/assets/276d380c-1a17-4a3c-b877-80715a76a1b5

### DIFF
--- a/packages/core/src/extensions/SuggestionMenu/SuggestionPlugin.ts
+++ b/packages/core/src/extensions/SuggestionMenu/SuggestionPlugin.ts
@@ -260,7 +260,11 @@ export class SuggestionMenuProseMirrorPlugin<
             transaction.getMeta("pointer") ||
             // Moving the caret before the character which triggered the menu should hide it.
             (prev.triggerCharacter !== undefined &&
-              newState.selection.from < prev.queryStartPos())
+              newState.selection.from < prev.queryStartPos()) ||
+            // Moving the caret to a new block should hide the menu.
+            !newState.selection.$from.sameParent(
+              newState.doc.resolve(prev.queryStartPos())
+            )
           ) {
             return undefined;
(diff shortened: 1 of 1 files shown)


matthewlipski (commented):
Hmm, I'm wondering while we're at it if we want it to be possible to have multiple suggestion menus to be open, instead of closing the existing one when a new one is opened. Currently, the emoji picker and slash menu clash if they're open together, so I think this would improve the UX.

nperez0111:
Can two be opened at the same time? Should they both be open at the same time?
Can you show me how you got it to do that?

My understanding is only one should ever be active, and I think that is ideal from a UX standpoint.

matthewlipski:
If you type `:sm` it'll open up the emoji picker, and typing `/` after will open the slash menu without closing the emoji picker. I've also just noticed that the emoji picker remains open after deleting the query, so this is definitely just a bug.

nperez0111:
Ah good catch @matthewlipski I'll try to resolve that

nperez0111:
Think I got it, abusing the fact that transactions are synchronous

YousefED (approved):
nice fixes!

PR #1592, opened 2025-04-07
feat: add `undo` and `redo` methods to editor API

nperez0111:
This adds `undo` and `redo` commands onto the editor instance for ease of access


YousefED (commented):
Nice! let's add to docs as well?

nperez0111:
> Nice! let's add to docs as well?

I can't find the right spot to put this. Block Manipulation? I feel like this would be in a reference instead or something

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1056,6 +1056,20 @@ export class BlockNoteEditor<
     return replaceBlocks(this, blocksToRemove, blocksToInsert);
   }
 
+  /**
+   * Undo the last action.
+   */
+  public undo() {
+    this._tiptapEditor.commands.undo();
+  }
```
nperez0111, line 1064: This might not work with collaboration. Look into using the editor.exec API instead too

PR #1595, opened 2025-04-07
feat: Unit testing setup restructure

matthewlipski:


packages/core/src/api/tests/formatConversion/export/getExportTestCases.ts
```diff
@@ -0,0 +1,1521 @@
...
+  TestBlockSchema,
+  TestInlineContentSchema,
+  TestStyleSchema,
+} from "../../testSchema.js";
+
+export type ExportTestCase<
+  // At some point we probably want to only have one HTML format that is both
+  // lossless when converting to/from conversionType, content, in which case we will only need
+  // "html" test cases and can remove "blockNoteHTML".
+  ConversionType extends "blocknoteHTML" | "html" | "markdown" | "nodes"
```
YousefED, line 12: I'm not sure the generic here and below adds anything. you could just say `converstionType: "blocknoteHTML" | "html" | "markdown" | "nodes"`, right?

packages/core/src/api/tests/formatConversion/export/testExport.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    );
+  } else {
+    throw new UnreachableCaseError(testCase.conversionType);
+  }
+};
+
+describe("Export tests", async () => {
+  const getEditor = setupTestEditor();
+
+  for (const testCase of [
```
YousefED, line 65: We can choose to group the test further in a "describe" call. We can do this either:
- by format (all tests with the same format are grouped)
- by test case (all tests are grouped by test case, and within that, there are tests for each format)

Currently, for snapshots you already group them by format. I don't think I have a clear preference at this point

packages/core/src/api/tests/formatConversion/export/testExport.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  editor: BlockNoteEditor<
+    TestBlockSchema,
+    TestInlineContentSchema,
+    TestStyleSchema
+  >,
+  testCase: ExportTestCase<ConversionType>
+) => {
+  (window as any).__TEST_OPTIONS.mockID = 0;
+  if (testCase.conversionType === "blocknoteHTML") {
+    await expect(
```
YousefED, line 30: I think there are two approaches we can take here:
a) run all tests for all formats from a single `testExports` file, like we're doing now
b) colocate these tests (like regular unit tests) with the code we're testing (e.g.: markdown tests should live next to the markdownexporter). This file could then still be very minimal and just import the shared test case.

I think there are good arguments for either ((a) is maybe clearer, but (b) is more consistent with how we organize other unit tests)

packages/core/src/api/tests/formatConversion/exportParseEquality/testExportParseEquality.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        )
+      )
+    ).toStrictEqual(
+      partialBlocksToBlocksForTesting(editor.schema, testCase.content)
+    );
+  } else {
+    throw new UnreachableCaseError(testCase.conversionType);
+  }
+
+  expect(true).toBeTruthy();
```
YousefED, line 64: what's this?

packages/core/src/api/tests/formatConversion/exportParseEquality/testExportParseEquality.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+describe("Export/parse equality tests", async () => {
+  const getEditor = setupTestEditor();
+
+  for (const testCase of [
+    ...getExportParseEqualityTestCases("blocknoteHTML"),
+    ...getExportParseEqualityTestCases("nodes"),
+  ]) {
+    it(`${testCase.name}`, async () => {
+      await testExportParseEqualityTest(getEditor(), testCase);
```
YousefED, line 75: I think you could just write the test cases here instead of using a `testExportParseEqualityTest` function, because `testExportParseEqualityTest` doesn't actually have any shared logic between the different `conversionType` cases

packages/core/src/api/tests/formatConversion/exportParseEquality/testExportParseEquality.test.ts
```diff
@@ -0,0 +1,78 @@
...
+} from "../formatConversionTestUtil.js";
+import {
+  ExportParseEqualityTestCase,
+  getExportParseEqualityTestCases,
+} from "./getExportParseEqualityTestCases.js";
+
+// Test for verifying that exporting blocks to another format, then importing
+// them back results in the same blocks as the original. Used broadly to ensure
+// that exporting and importing blocks does not result in any data loss.
+const testExportParseEqualityTest = async (
```
YousefED, line 25: should we not have this for markdown or html? or only for some of those cases?

packages/core/src/api/tests/formatConversion/parse/getParseTestCases.ts
```diff
@@ -547,17 +461,113 @@ With Hard Break</p>
...
+  },
+  {
+    name: "codeBlocks",
+    conversionType: "html",
+    content: `<pre><code>console.log("Should default to JS")</code></pre>
+<pre><code data-language="typescript">console.log("Should parse TS from data-language")</code></pre>
+<pre><code class="language-python">print("Should parse Python from language- class")</code></pre>
+<pre><code class="language-ruby" data-language="typescript">console.log("Should prioritize TS from data-language over language- class")</code></pre>`,
+  },
+  {
```
YousefED, line 474: I get the idea, but I'm not entirely sure if it makes sense markdown / html of these cases should live in the same file

packages/core/src/api/tests/formatConversion/exportParseEquality/testExportParseEquality.test.ts
YousefED: also, wdym with `Used broadly`?

packages/core/src/api/tests/formatConversion/export/testExport.test.ts
YousefED: btw, better organizing in "describe" makes the tests show up clearer in vitest ui / test runners (so that would be the main goal with this)

packages/core/src/api/tests/clipboard/clipboardTestUtil.ts
```diff
@@ -0,0 +1,93 @@
+import { Node, Slice } from "@tiptap/pm/model";
+import { EditorView } from "@tiptap/pm/view";
+import * as pmView from "@tiptap/pm/view";
+
+// Helper function to get the position of a text node with given text content.
+// By default, returns the position just before the node, but can be just after
+// instead if `after` is set to true.
+export const getPosOfTextNode = (
```
YousefED, line 8: nice, this is new right :)?

packages/core/package.json
```diff
@@ -94,6 +94,7 @@
     "@tiptap/pm": "^2.11.5",
     "emoji-mart": "^5.6.0",
     "hast-util-from-dom": "^5.0.1",
+    "htmlfy": "^0.6.7",
```
YousefED, line 97: should be a devdep?

packages/react/src/tests/formatConversion/exportParseEquality/testExportParseEquality.test.ts
```diff
@@ -0,0 +1,19 @@
...
+import { testSchema } from "../../testSchema.js";
+import { getExportParseEqualityTestCases } from "./getExportParseEqualityTestCases.js";
+
+describe("Export/parse equality tests", () => {
+  const getEditor = setupTestEditor(testSchema);
+
+  for (const testCase of [
+    ...getExportParseEqualityTestCases("blocknoteHTML"),
+    ...getExportParseEqualityTestCases("nodes"),
+  ]) {
```
nperez0111, line 14: Agree with Yousef, these should be their own describe blocks which help isolate tests to groups that are relevant to what you are doing

nperez0111:
This looks great. Lot of clean up here. Thanks @matthewlipski for taking this on 💪!

packages/core/src/api/tests/formatConversion/parse/getParseTestCases.ts
matthewlipski: Changed this so that tests for different formats are in separate arrays, and get their own `describe`. Imo we don't need to also separate to different files.

packages/core/src/api/tests/formatConversion/exportParseEquality/testExportParseEquality.test.ts
matthewlipski: Added tests for HTML in #1605, though up until now HTML and Markdown wasn't really a concern since the export is lossy, so we didn't really expect the result after export + re-import to be the same as the original blocks.

matthewlipski: Also updated the comment

PR #1604, opened 2025-04-11
fix: switch to only using transactions instead of prosemirrorState

nperez0111:
This attempts to implement a transaction system which only builds up the current transaction (without lying about editor state).

The drawback to this approach though, is that `appendedTransaction`s do not apply until after transaction is actually dispatched (`state.apply(tr)`).

One example where this comes up is that we have the `unique-id` plugin, which appends ids to anything that does not already have an ID attribute.

So, when reading from `tr.doc`, some newly inserted content can have no ID attached to it, when run within a transaction (because the ID has not yet been assigned).

I've included a test case here which would demonstrate the problem:

### DIFF
--- a/packages/core/src/api/blockManipulation/commands/insertBlocks/__snapshots__/insertBlocks.test.ts.snap
+++ b/packages/core/src/api/blockManipulation/commands/insertBlocks/__snapshots__/insertBlocks.test.ts.snap
@@ -7,6 +7,62 @@ exports[`Test insertBlocks > Insert multiple blocks after 1`] = `
     "content": [
       {
         "styles": {},
+        "text": "Inserted paragraph 1",
+        "type": "text",
+      },
+    ],
+    "id": "0",
+    "props": {
+      "backgroundColor": "default",
+      "textAlignment": "left",
+      "textColor": "default",
+    },
+    "type": "paragraph",
+  },
+  {
+    "children": [],
+    "content": [
+      {
+        "styles": {},
+        "text": "Inserted paragraph 2",
+        "type": "text",
+      },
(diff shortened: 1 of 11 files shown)


YousefED:
I suppose a possible fix could be to run the uniqueid logic when calling `dispatch`? (either just the uniqueid plugin, or a generic solution for all plugins)? How do you feel about that?

nperez0111:
> I suppose a possible fix could be to run the uniqueid logic when calling `dispatch`? (either just the uniqueid plugin, or a generic solution for all plugins)? How do you feel about that?

Possible, but to do that (at least for any possible plugin), you'd have to generate an editor state to run it's logic on. Which is essentially the same as the other approach.

nperez0111:
I've added that implementation, and it ends up being almost the same thing as what we had before, but does not keept this "new editor state" around for reading from.

It ends up being the same implementation as what we have in the other PR 😂 

https://github.com/TypeCellOS/BlockNote/pull/1604/commits/ffc4a5b83d28df9f7532a28d266ab79bb48809f5

YousefED:
> I've added that implementation, and it ends up being almost the same thing as what we had before, but does not keept this "new editor state" around for reading from.
> 
> It ends up being the same implementation as what we have in the other PR 😂
> 
> [ffc4a5b](https://github.com/TypeCellOS/BlockNote/commit/ffc4a5b83d28df9f7532a28d266ab79bb48809f5)

Nice. Do you think that's a safe fix? Or will it cause issues when the transaction is _actually_ applied again? because then everything will run through `applyTransaction` again right? (incl the appended transactions)

PR #1605, opened 2025-04-11
feat: Better parsing of external HTML attributes & inline styles

matthewlipski:
This PR adds tests for parsing various HTML attributes & inline styles, and provides fixes so that these tests pass. Namely, these tests include:

- Parsing numbered list `start` props from `start` attribute.
- Parsing image `width` props from `width` attribute.
- Parsing `bold`, `italic`, `underline`, and `strike` marks from various tags & inline styles. This includes e.g. `b` tags and `font-weight` inline styles for `bold` marks as well as `i` tags and `font-style` inline styles for `italic` marks.
- Parsing `textColor` and `backgroundColor` marks from inline styles.
- Parsing `textColor` and `backgroundColor` props from inline styles.
- Parsing `textAlignment` props from inline styles.

To ensure all tests pass, several changes have also been made to the code:

- Parse rules have been added to `textColor`, `backgroundColor`, and `textAlignment` props for inline styles. 
- The `textColor` and `backgroundColor` props have been changed, so that they're now applied as attributes to `blockContent` nodes instead of `blockContainer` nodes. This is because we would only attempt to parse these props when a `blockContainer` node is found, i.e. when a `div` with `[data-node-type="blockContainer"]` is found. So we would never even attempt to parse these props from external HTML. The CSS has been updated also to account for this change. Also, all other block props are already stored on the `blockContent` node, so this makes the `textColor` and `backgroundColor` props also consistent with that.
- Small fixes have been made to numbered list item and quote blocks.

Additionally, code for color props & marks has been cleaned up slightly.

The large negative diff in this PR is also because a bunch of redundant React unit test snapshots have been removed.


nperez0111 (approved):
👍 this looks exactly as I'd have expected it to.

Great work @matthewlipski

packages/core/src/extensions/TextColor/TextColorMark.ts
```diff
@@ -25,8 +21,8 @@ const TextColorMark = Mark.create({
             return false;
           }
 
-          if (element.hasAttribute("data-text-color")) {
-            return { stringValue: element.getAttribute("data-text-color") };
+          if (element.hasAttribute("data-text-color") || element.style.color) {
+            return {};
```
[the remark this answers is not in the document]
matthewlipski: Yeah we don't need to parse the color because it's already handled in the `parseHTML` function of the `stringValue` attribute.

packages/core/src/extensions/BackgroundColor/BackgroundColorMark.ts
```diff
@@ -1,18 +1,13 @@
...
   addAttributes() {
     return {
-      stringValue: {
-        default: undefined,
-        parseHTML: (element) => element.getAttribute("data-background-color"),
-        renderHTML: (attributes) => ({
-          "data-background-color": attributes.stringValue,
-        }),
-      },
+      stringValue: getBackgroundColorAttribute("stringValue"),
```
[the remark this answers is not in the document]
matthewlipski: It doesn't really make a difference, there's not really a use case where you would set a color mark with a default value. Selecting a default text/bg color in the color picker removes the mark instead, and this is how it should always be done, so the default value for the mark should never be used.

packages/core/src/editor/Block.css
```diff
@@ -437,97 +437,111 @@ NESTED BLOCKS
...
 
-[data-background-color="pink"] {
+[data-background-color="pink"],
+.bn-block:has(> .bn-block-content[data-background-color="pink"]) {
   background-color: #f4dfeb;
 }
 
 /* TEXT ALIGNMENT */
 [data-text-alignment="left"] {
-  justify-content: flex-start !important;
```
YousefED, line 515: do you know why this was here / why has it been removed?

packages/core/src/extensions/BackgroundColor/BackgroundColorExtension.ts
```diff
@@ -1,32 +1,25 @@
...
+
+import { getAttributeFromDefaultProps } from "../../blocks/defaultProps.js";
 
 export const BackgroundColorExtension = Extension.create({
   name: "blockBackgroundColor",
 
   addGlobalAttributes() {
     return [
       {
-        types: ["blockContainer", "tableCell", "tableHeader"],
```
YousefED, line 10: I suppose not, but does this break anything for colors / alignment on custom blocks (as you're changing this from blockcontainer to explicitly naming the block types)

packages/core/src/extensions/TextColor/TextColorExtension.ts
```diff
@@ -1,29 +1,26 @@
...
+import { getAttributeFromDefaultProps } from "../../blocks/defaultProps.js";
 
 export const TextColorExtension = Extension.create({
   name: "blockTextColor",
 
   addGlobalAttributes() {
     return [
       {
-        types: ["blockContainer", "tableCell", "tableHeader"],
+        types: [
```
YousefED, line 11: same question as above; does this break anything for custom blocks?

packages/core/src/extensions/TextColor/TextColorMark.ts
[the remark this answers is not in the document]
YousefED: can't we omit line 25 / 26 then entirely? it's still a bit confusing right? (maybe tiptap issue)

YousefED (commented):
Looks good!

Just realized one caveat; for yjs stored documents the block color data will probably be lost when upgrading (or the document might break entirely, but I doubt that).

Perhaps you can see with Nick if there's a fix for that possible (or a migration path for people upgrading)

packages/core/src/editor/Block.css
matthewlipski: Yeah, this should have already been removed in #1553 since we changed the inline content there to span the whole block width (would previously only span the content itself). Since I was making changes to CSS here I just removed this too.

packages/core/src/extensions/BackgroundColor/BackgroundColorExtension.ts
matthewlipski: Still works as the API (specifically `propsToAttributes`) converts the props from the `propSchema` to TipTap attributes. This extension is just overrides those attributes so we can add unique parse rules.

packages/core/src/extensions/TextColor/TextColorMark.ts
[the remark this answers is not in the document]
matthewlipski: No, because return {} tells TipTap that the element should get parsed as a textColor mark, and to just use the default attribute values for that mark (here it is equivalent to return { textColor: defaultProps.textColor.default }). The parseHTML function of the stringValue attribute then does its thing, parsing the actual value of stringValue and replacing the default one. Returning false instead means the mark never gets parsed at all.

This is a kind of weird TipTap thing especially with return {}, but it does mostly make sense.

tests/src/unit/core/formatConversion/export/__snapshots__/html/hardbreak/multiple.html
```diff
@@ -1,4 +1,9 @@
-<p>
+<p
+  class="bn-block-content"
+  data-node-type="blockOuter"
```
YousefED: should this get added in this PR?

matthewlipski: Looks like you were reviewing only the changes from commit `13acc7e`, since that's what GitHub shows when I click the link to see the full diff. Has since been fixed👍

matthewlipski:
Changes from this PR and more were updated and added in #1991, #1998, and #2004

PR #1608, opened 2025-04-14
fix: update y-prosemirror to fix #1462

### LINKED ISSUE #1462: Can't delete a line where the is cursor
**Describe the bug**
I can't delete a line when another user's cursor is on it

**To Reproduce**
https://github.com/user-attachments/assets/a71e59e9-3839-4243-bc64-7d772f3523e5

**Misc**
- Node version:
- Package manager:
- Browser: 

nperez0111:
This only updates y-prosemirror to the latest patch version which includes this fix for selection handling: https://github.com/yjs/y-prosemirror/pull/181

Which resolves #1462

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -111,7 +111,7 @@
     "remark-stringify": "^11.0.0",
     "unified": "^11.0.5",
     "uuid": "^8.3.2",
-    "y-prosemirror": "^1.3.1",
+    "y-prosemirror": "^1.3.3",
     "y-protocols": "^1.0.6",
     "yjs": "^13.6.15"
   },
--- a/packages/server-util/package.json
+++ b/packages/server-util/package.json
@@ -60,7 +60,7 @@
     "@tiptap/core": "^2.7.1",
     "@tiptap/pm": "^2.7.1",
     "jsdom": "^25.0.1",
-    "y-prosemirror": "^1.3.1",
+    "y-prosemirror": "^1.3.3",
     "y-protocols": "^1.0.6",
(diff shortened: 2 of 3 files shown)


YousefED:
Great to see this fixed!! I'm not sure if we can prevent this, but I do think you can now force one's cursor to jump to the next block (the gif below is not super clear - can jump on a call to discuss)

cursors

nperez0111:
> Great to see this fixed!! I'm not sure if we can prevent this, but I do think you can now force one's cursor to jump to the next block (the gif below is not super clear - can jump on a call to discuss)

Unfortunately, I think this is an inherent limitation of y-prosemirror representing content as an XML tree: https://github.com/yjs/y-prosemirror/issues/44#issuecomment-844522138 If this were a flat tree, then the character position would be stable in an operation like this.

The node that the cursor is referencing is deleted so the relative position points to a node that no longer exists, but did exist after the first node, so it guesses that it was closer to the end of the content. TextSelection.between then moves that in-between position to the first position in the node since it's bias is unspecified. You could try to bias it going backward (i.e. pick the position before so the end of the node content), but that still is just as unexpected and would mess with the logic in the other direction so it makes sense to keep the current default.

If y-prosemirror applied minimal diffs, then we could map the selection using normal prosemirror behavior instead of having to use a relative position. But, that would still require work on the y-prosemirror side to resolve.

nperez0111:
Along the way of looking into this. I rewrote the selection handling to be much more robust, so I guess it wasn't completely wasted effort! https://github.com/yjs/y-prosemirror/pull/182

PR #1609, opened 2025-04-14
refactor: blocknote API uses lowest possible level prosemirror primitives

nperez0111:
This refactors all of the blocknote APIs to use low-level prosemirror APIs rather than passing around editor instances and dispatching transactions themselves.
Now, block manipulation functions will only update the current transaction it was provided rather than trying to dispatch it. This allows these commands to be chained without having to resort to any smarts for the transaction handling.

supercedes: https://github.com/TypeCellOS/BlockNote/pull/1147 https://github.com/TypeCellOS/BlockNote/pull/1604 https://github.com/TypeCellOS/BlockNote/pull/1584

Resolves: https://github.com/TypeCellOS/BlockNote/issues/829


packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.ts
```diff
@@ -23,15 +33,36 @@ import {
...
     I extends InlineContentSchema,
     S extends StyleSchema
   >(
-    editor: BlockNoteEditor<BSchema, I, S>,
+    pmSchema: Schema,
+    schema: BlockNoteSchema<BSchema, I, S>,
     posBeforeBlock: number,
-    block: PartialBlock<BSchema, I, S>
+    block: PartialBlock<BSchema, I, S>,
+    blockCache?: BlockCache
```
YousefED, line 65: do we still need this below?
```
({
    tr,
    dispatch,
  }: {
```

packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
```diff
@@ -154,22 +169,31 @@ function flattenColumns(
  * reference block.
  */
 export function moveSelectedBlocksAndSelection(
-  editor: BlockNoteEditor<any, any, any>,
+  tr: Transaction,
```
YousefED, line 172: looks like we could go either way here, right?
- operate on a `transaction`, like you did, and the lower-level (internal) BlockNote methods
- vs. `editor.transact` and use higher level BlockNote methods. If I see it correctly, this is also valid option because we don't use lower-level prosemirror operations for `moveBlocks` related methods

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1073,18 +1076,28 @@ export class BlockNoteEditor<
     targetBlock: BlockIdentifier,
     placement: "start" | "end" = "start"
   ) {
-    setTextCursorPosition(this, targetBlock, placement);
+    return this.transact((tr) =>
```
YousefED, line 1079: why are some of the methods like this one wrapped in `transact`?

YousefED (commented):
Cool! Do you think you have a preference after these explorations? Specifically looking at which codebase / architecture would be the most maintainable / strongest to work with down the line?

packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
nperez0111: Yep, both are valid. I think the code is clearer as the high-level methods. So, I'll put it back to those

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: `transact` is convenient in that you don't have to create a transaction & then dispatch it. It also is clear that it would work within a blocknote transaction.

If everything were to use this pattern, then you wouldn't even need to have `editor.transaction` hold the `activeTransaction` state, it would just be passed around through the `transact` calls. So, it ends up mapping more closely to what actually happens

nperez0111:
> Cool! Do you think you have a preference after these explorations? Specifically looking at which codebase / architecture would be the most maintainable / strongest to work with down the line?

I prefer the explicitness of what we have here, but I think the tradeoff that it makes are methods that need to take a ton of arguments, which is a code smell imo.

Take a look at this method: https://github.com/TypeCellOS/BlockNote/blob/34ebafadae7034d0e9e1c3e6281a1376b7226832/packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.ts#L67-L78

It's annoying that it needs to take 2 schemas (blocknote & prosemirror) and the blockCache. I think that this could be aleviated though. The prosemirror schema has a [cached](https://prosemirror.net/docs/ref/#model.Schema.cached) object which can store the blockCache & blocknote schema inside of it. This would make most methods only ever need a `tr` and `schema`

YousefED:
> I prefer the explicitness of what we have here, but I think the tradeoff that it makes are methods that need to take a ton of arguments, which is a code smell imo.

ok! let's go for this approach then 👍

> It's annoying that it needs to take 2 schemas (blocknote & prosemirror) and the blockCache. I think that this could be aleviated though. The prosemirror schema has a [cached](https://prosemirror.net/docs/ref/#model.Schema.cached) object which can store the blockCache & blocknote schema inside of it. This would make most methods only ever need a tr and schema

We could also just create a type `{ schema, pmSchema, blockCache }`? we could even pass in `this` from the editor if we'd like (or we create a wrapper object). Not sure if pm `cached` has any benefits to this (and downside is that it's untyped)

packages/core/src/editor/BlockNoteEditor.ts
YousefED: Ah that makes sense! I was looking at it from the angle of "only call `transact` if you're doing multiple BlockNote API calls in a method.

But you're right, we could also use it for convenience + make the methods consistent. I suppose in that case we need to do it for all usages of `dispatch`, and perhaps even make `dispatch` private?

nperez0111:
> We could also just create a type `{ schema, pmSchema, blockCache }`? we could even pass in `this` from the editor if we'd like (or we create a wrapper object). Not sure if pm `cached` has any benefits to this (and downside is that it's untyped)

I think I like this better. It turns out that from a transaction's document, you can get a schema. So storing it into the schema makes it really convenient to reach for values you may need. Just as a POC, I cleaned up the insertBlocks method with this. I think the API is a lot cleaner, and nodeToBlock & blockToNode become much simpler too: 2980c6c2541594d57dfe77f2a5fd150c4fb03a7e

YousefED:
> > We could also just create a type `{ schema, pmSchema, blockCache }`? we could even pass in `this` from the editor if we'd like (or we create a wrapper object). Not sure if pm `cached` has any benefits to this (and downside is that it's untyped)
> 
> I think I like this better. It turns out that from a transaction's document, you can get a schema. So storing it into the schema makes it really convenient to reach for values you may need. Just as a POC, I cleaned up the insertBlocks method with this. I think the API is a lot cleaner, and nodeToBlock & blockToNode become much simpler too: [2980c6c](https://github.com/TypeCellOS/BlockNote/commit/2980c6c2541594d57dfe77f2a5fd150c4fb03a7e)

ok! Let's go and see how this works in practice :)

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: Good call on standardizing it, will refactor to that pattern. I think it will come out nice

PR #1615, opened 2025-04-17
feat: expose prosemirrorState again?

nperez0111:


YousefED (commented):
looks good; anything specifically that triggered you to re-add this?

nperez0111:
It was triggered by this: https://github.com/TypeCellOS/BlockNote/blob/5edf2991e8bb76020e45ec1b0f2ff92a757f48ea/packages/core/src/api/positionMapping.ts#L98

To get a plugin state, you need the editor state. So, at some level it needs to be exposed. I figured it better to expose as a top-level with an error on incorrect usage than not having it at all

YousefED (approved):

Issue #1158, opened 2024-10-16
Support for H4 and H5 heading levels?


kaitakami:
**Is your feature request related to a problem? Please describe.**
Currently, BlockNote only supports H1, H2, and H3 headings. This limitation prevents the creation of more detailed document structures that require H4 and H5 headings, hindering the ability to properly format documents with deeper hierarchical structures.

**Describe the solution you'd like**
Extend BlockNote to support H4 and H5 heading levels. This would include:
1. Adding H4 and H5 options to the heading level selector in the editor toolbar.
2. Implementing rendering and styling for H4 and H5 headings in the editor view.
3. Ensuring correct export of H4 and H5 headings when converting to other formats.

**Describe alternatives you've considered**

**Additional context**
Supporting H4 and H5 would allow for a wider range of use cases, including technical documentation and complex reports.

**Bonus**
[ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

LucianoTrujillo:
I have the same urgency, I really need this 🙏

FerreroManuel:
Same problem here... 🙏🏻🙏🏻🙏🏻

YousefED:
Thanks for the request, would love to add this but also focussing on some
other requests atm.

Are all of you working on the same product? What kind of product are you
working on?

LucianoTrujillo:
Thanks Yousef! Yes! 
We've already used this amazing tool for creating an AI educational platform in the past (for o, where teachers could create engaging theory documents in a notion style assisted with AI.

Now we are building an editor for creating blog posts to wordpress.

Your abstractions are the best!

esteban389:
> Thanks Yousef! Yes! We've already used this amazing tool for creating an AI educational platform in the past (for our thesis at university, where teachers could create engaging theory documents in a notion style assisted with AI.
> 
> Now we are building an editor for creating blog posts to wordpress.
> 
> Your abstractions are the best!

Hola, la verdad me pareció interesante tu proyecto, de casualidad no será Open source? 👀

ClementEXWiki:
Hi, would be useful for me too :)

This can be implemented pretty easily as custom blocks, but having native support inside BlockNote would be better.

hylin:
Need this feature too. Can someone process the related PR? @YousefED @nperez0111

nperez0111:
Will be in the next release. Configurable, not enabled by default

[...]

PR #1623, opened 2025-04-22
Support h4 and h5

DrewJohnsonGT:
Related issue [1158](https://github.com/TypeCellOS/BlockNote/issues/1158)

https://github.com/user-attachments/assets/2fe5e866-2d9f-4f56-983c-02f4e5bc2a49

Figure it doesn't make sense to add these as default/suggested blocks but adding general support will satisfy a lot of use cases

First contribution here, so let me know if I missed anything!

### DIFF
--- a/tests/src/unit/core/formatConversion/parse/parseTestInstances.ts
+++ b/tests/src/unit/core/formatConversion/parse/parseTestInstances.ts
@@ -360,6 +360,7 @@ export const parseTestInstancesHTML: TestInstance<
 <h3>Heading 3</h3>
 <h4>Heading 4</h4>
 <h5>Heading 5</h5>
+<h6>Heading 6</h6>
 <p>Paragraph 1</p>
 <p>Nested Paragraph 1</p>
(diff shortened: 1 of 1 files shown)


packages/core/src/editor/Block.css
```diff
@@ -127,6 +127,12 @@ NESTED BLOCKS
 [data-content-type="heading"][data-level="3"] {
   --level: 1.3em;
 }
+[data-content-type="heading"][data-level="4"] {
+  --level: 1em;
+}
+[data-content-type="heading"][data-level="5"] {
+  --level: 0.9em;
+}
```
DrewJohnsonGT, line 135: Used https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements
as reference for appropriate sizes

milanpoliak:
Ah this is nice, just in time for what I need as well! I think it would make sense to include H6 already as well, I suppose not too long until somebody will need that too 😆

PR #1626, opened 2025-04-22
fix(code-block): handle unknown languages better

nperez0111:
This resolves a bug where if a language is a language that is not within supportedLanguages, it would throw an error rather than silently letting you continue to use it as a plaintext code block 

Sentry bug: https://blocknote-js.sentry.io/issues/37666656/?environment=vercel-production&project=4508925646078032&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=24

### DIFF
--- a/packages/core/src/blocks/CodeBlockContent/CodeBlockContent.ts
+++ b/packages/core/src/blocks/CodeBlockContent/CodeBlockContent.ts
@@ -126,7 +126,10 @@ const CodeBlockContent = createStronglyTypedTiptapNode({
             return null;
           }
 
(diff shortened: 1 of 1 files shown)


YousefED (approved):
nice. does it make sense to log a warning? it's a bit suspicious when a block has a language that doesn't have a highlighter, right?

nperez0111:
> nice. does it make sense to log a warning? it's a bit suspicious when a block has a language that doesn't have a highlighter, right?

The user can add a block like that by typing ```anytext

So, I don't think it is on the dev at that point. I like the fallback to no highlighting at all

Issue #1588, opened 2025-04-07
The wrong cursor position was inserted when clicking at the end of the paragraph


sailxjx:
**Describe the bug**
When click the end of the paragraph, if the click area is in the editor, the position is right, if the area is in the padding area, the cursor will be inserted in the head of the paragraph, which I think need to be improved, because the user can't figure out where is the padding area.

I've recorded a video of this bug

https://github.com/user-attachments/assets/b08fefb8-5bed-4114-86ae-ec37b7893414

**To Reproduce**
<clear steps to reproduce are super helpful! Best is to provide an online sandbox, [click to create one](https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/01-basic/01-minimal?file=App.tsx)>

**Misc**
- Node version:
- Package manager:
- Browser: 
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

NitanJana:
I think it's resolved in #1553

nperez0111:
I think so too

sailxjx:
The problem still exist, I've upgraded to v0.28. What I mean is to click on the padding area of the editor, not inside the editor. This area is visually indistinguishable from the inside of the editor, and the mouse behaves the same way. Please check my video, you can reproduce this issue on the official website. @nperez0111 @NitanJana

matthewlipski:
Interesting, the issue also isn't there on Safari so seems the root cause is how Chrome handles mouse clicks/selection changes in element padding (Chrome always seems to move the selection to the text start).

One way to at least make the behaviour at least consistent across browsers is to move the background & padding to the parent `.bn-container` element instead of the `.bn-editor` element. This makes clicking the padded area clear the selection instead of moving it, as the event is not captured by the editor itself.

Alternatively, we could just remove the padding on the right. It's there so that the editor appears centered while leaving space for the side menu, but it's a fairly minor thing.

Any thoughts on these 2 solutions? I think it's a bit overkill to override Chrome's default click/selection handling for this one case so I'd rather not go down that route.

Also indeed #1553 fixes a slightly different issue, as previously there was additional padding within blocks that would also cause this behaviour.

nperez0111:
I would try to move around the padding so that it works, but not spend too much time on it. I'd say any resolution to thise should be CSS only and don't think it is a big deal if we can't find a resolution to it

YousefED:
[...]

PR #1629, opened 2025-04-22
fix: Selection handling when clicking editor padding

matthewlipski:
In #731, we broke some things with block styling & selection handling on click, which we didn't fix 100% correctly. One issue that still remains from these changes is the selection moving to the start of the block instead of the end, when clicking the padded area on the right of the area (see #1588 for more detail). The issue is specific to Chrome only.

The reason for this bug is due to the `:before` pseudo-element on each `.bn-block-content` element. Previously,  the `.bn-block-content` element was `display: block`, and so it's `:before` pseudo-element took on `display: inline`. However, we then changed `.bn-block-content` to be `display: flex`, forcing the `:before` pseudo-element to take on `display: block`. Because it was being displayed as a block, and therefore had the same height as the block, this confused Chrome's selection handling behaviour on click.

This PR reverts the `.bn-block-content` element to be `display: block`, which fixes the issue. I've made sure that all vanilla blocks still look identical, and that at least the custom blocks from our examples are all good too. Additionally, I've fixed some minor issues that we didn't notice before due to the `display: flex`.

### DIFF
--- a/examples/06-custom-schema/04-pdf-file-block/styles.css
+++ b/examples/06-custom-schema/04-pdf-file-block/styles.css
@@ -1,10 +1,4 @@
-[data-content-type="pdf"] .bn-file-block-content-wrapper,
-[data-content-type="pdf"] .bn-file-and-caption-wrapper {
-    height: 800px;
-    width: 100%;
-}
-
 [data-content-type="pdf"] embed {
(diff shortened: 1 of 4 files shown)


nperez0111 (approved):
lgtm

PR #1630, opened 2025-04-22
fix: Formatting toolbar regression

### LINKED ISSUE #1569: Toolbar placement shifts unexpectedly on text alignment change
**Describe the bug**
When changing the alignment of a text block (e.g., left, center, right), the floating toolbar shifts its position in a jarring way. This behavior makes the experience feel inconsistent compared to how smooth it is when aligning images, which do not trigger the same issue.

**To Reproduce**
Go to the editor of blocknote: https://www.blocknotejs.org/
- Type some text into the BlockNote editor.
- Select the text.
- Change its alignment using the toolbar (e.g., click "Right").
- Observe the toolbar "jumping" or misplacing itself.

**Misc**
- Node version: 22.13.0
- Package manager: npm-10.9.2
- Browser: Firefox-136.0.1 (64-bit)

**Video of the issue**

https://github.com/user-attachments/assets/aec0149b-a837-473b-8218-feb14bf06d1d

matthewlipski:
This PR fixes a regression from #1285 that made it so we weren't updating the formatting toolbar if the document changed but the selection stayed the same.

Closes #1569. However, the toolbar positioning is still janky because it doesn't account for animations. This is a known issue though.

### DIFF
--- a/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
+++ b/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
@@ -140,11 +140,9 @@ export class FormattingToolbarView implements PluginView {
     // Wrapping in a setTimeout gives enough time to wait for the blur event to
     // occur before updating the toolbar.
     const { state, composing } = view;
-    const { selection } = state;
+    const { doc, selection } = state;
     const isSame =
-      oldState &&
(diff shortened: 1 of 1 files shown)


nperez0111 (approved):

nperez0111 (commented):
lgtm

PR #1634, opened 2025-04-23
Support h4, h5, and h6

DrewJohnsonGT:
Related issue [1158](https://github.com/TypeCellOS/BlockNote/issues/1158)

image

https://github.com/user-attachments/assets/8db37a0a-ebcd-406d-a3f7-446a6f36331c

Figure it doesn't make sense to add these as default/suggested blocks but adding general support will satisfy a lot of use cases

First contribution here, so let me know if I missed anything!

Re-opened with H6 as well per https://github.com/TypeCellOS/BlockNote/pull/1623#issuecomment-2824719094


DrewJohnsonGT:
@YousefED @nperez0111 is there anything else I should do here?
I know y'all are swamped, but any feedback would be really appreciated

nperez0111:
Sorry for taking so long on this. Yea, this is mostly good, the one thing we would need to finish this up would be support in the export formats (e.g. PDF, ODT)

DrewJohnsonGT:
@nperez0111 No worries! Took a quick stab at that just now
PDF is solid
image

ODT and Word defaults are alright I think?
ODT
image

Word
image

hylin:
Any updates?

nperez0111:
Sorry, I'm waiting for the [toggle blocks](https://github.com/TypeCellOS/BlockNote/pull/1707) to settle before trying to integrate this in.

For full transparency, I will also probably make this configurable, and have the defaults set to just 3 heading levels like it was before this change.

MetaMmodern:
@nperez0111 @DrewJohnsonGT can I turn this off now on schema level?:) h1-h3 were enough for me.

I know I can customize formatting toolbar and slash menu, I'm asking about more high level approach like with the rest of block types.

PR #1635, opened 2025-04-23
fix: Lossy html conversion equality

matthewlipski:
This PR adds tests for HTML export/parse equality, and provides fixes so that these tests pass. Namely, these tests include checking export/parse equality of:

- Default blocks.
- Default blocks with props.
- Default inline content.
- Default styles.
- Nested list items.
- Advanced tables.

To ensure all tests pass, some changes have also been made to the code:

- Exporting of list items has been changed so that attributes for props is now on the `li` element rather than its first child. Without this, we would be unable to parse props like `textColor` from list items in exported HTML.
- The `partialBlockToBlockForTesting` function has been fixed to better support partial table content.


packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -117,28 +117,40 @@ function serializeBlock<
...
+      attr.name !== "data-file-block" &&
+      attr.name !== "data-node-view-wrapper" &&
+      attr.name !== "data-node-type" &&
+      attr.name !== "data-id" &&
+      attr.name !== "data-index" &&
+      attr.name !== "data-editable"
+  );
+
+  if (ret.dom.classList.contains("bn-block-content")) {
+    if (!listType) {
```
YousefED, line 144: let's add a comment here why we're not doing this for listType

YousefED (approved):
nice! cool that we're fixing the lists as these were definitely not top of mind for me. great to see the test driven approach

PR #1636, opened 2025-04-24
feat: Inline style props in external HTML

matthewlipski:
This PR makes props for default props on default blocks get rendered out to inline styles instead of `data-*` attributes when converting to lossy HTML. This is done in a somewhat hacky way as it only happens for default blocks, though after discussing with @YousefED we decided this was ok for now, and will be fixed when changing our default blocks to use the custom blocks API.

No tests have been added in this PR, as the ones we need are already in #1605 and #1635, which this PR branches off of.


packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -144,6 +144,15 @@ function serializeBlock<
     // We wrap the output in an `li` element for list items, and so we want to
     // add the attributes to that element instead as it is the "root".
     if (!listType) {
+      // TODO: This is specifically for default blocks, as default props get
```
YousefED, line 147: can you explain more what the code does? You're now explaining "why" it's here, but what does it do? why do we move styles from ret.dom to ret.dom.firstchild?

packages/core/src/blocks/defaultBlockHelpers.ts
```diff
@@ -90,6 +93,27 @@ export const defaultBlockToHTML = <
     );
   }
 
+  // TODO: This is obviously pretty hacky - will need to revisit this when we
```
YousefED, line 96: minor feedback; I read "TODO" as something that needs to be fixed before merging.

I would rephrase to first explaining "what's happening" and then adding a note for future recommendations (instead of immediately going ahead and "apologizing" so to say). Suggestion:

> when exporting to external HTML, we convert from `data-` attributes to `style` properties which can be understood by external applications.
> Note: this is a bit hacky to do this here as we're hardcoding this for some default props. We should revisit this when we migrate internal blocks to use the BlockNote API

tests/src/unit/react/formatConversion/export/__snapshots__/html/contextParagraph/basic.html
```diff
@@ -1 +1 @@
-<div data-editable="">React Context Paragraph</div>
\ No newline at end of file
+<div data-editable="" style="white-space: normal;">React Context Paragraph</div>
```
YousefED, line 1: why is the white-space added?

tests/src/unit/core/formatConversion/export/__snapshots__/html/complex/misc.html
```diff
@@ -1,12 +1,12 @@
-<h2 data-text-color="yellow" data-background-color="blue" data-text-alignment="right" data-level="2">
+<h2 style="background-color: blue; color: yellow; text-align: right;" data-level="2">
```
YousefED, line 1: do we need to output data-level?

packages/core/src/blocks/defaultBlockHelpers.ts
```diff
@@ -90,6 +93,27 @@ export const defaultBlockToHTML = <
...
+  //  convert default blocks to use the custom block API.
+  if (external) {
+    const dom = renderSpec.dom as HTMLElement;
+
+    if (dom.hasAttribute("data-background-color")) {
+      dom.style.backgroundColor = dom.getAttribute("data-background-color")!;
```
YousefED, line 102: nice to have (wouldn't spend too much time on this):

pseudocode that adds this inline style:
```style: {
--blocknote-red: #colorcode;
color: var(blocknote-red);
```
and then detect `var(blocknote-red)` when parsing

This way everything should keep working with color schemes, and external apps would still have the nice color (instead of default css "red")

tests/src/unit/react/formatConversion/export/__snapshots__/html/contextParagraph/basic.html
matthewlipski: This is something that's added automatically to the React node view wrapper. Do we need to get rid of it?

matthewlipski: Ah, yeah - since we render some props out to inline styles for external HTML, I changed it so inline styles are also preserved (previously only attributes were). The React node view wrapper adds that white space style automatically, so it also gets preserved.

PR #1639, opened 2025-04-24
feat: pausing & resuming of a ydoc

nperez0111:
This implements pausing and resuming of a ydoc, switching to a new ydoc that can be written to, and then switching back.

https://github.com/user-attachments/assets/bd4536e6-39a2-4198-9852-6cd9961a578e


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -934,6 +943,96 @@ export class BlockNoteEditor<
...
+    const nextFragment = this.findTypeInOtherYdoc(prevFragment, doc);
+
+    this.yjsState = {
+      prevFragment,
+      nextFragment,
+    };
+    this._tiptapEditor.unregisterPlugin(yCursorPluginKey);
+    this._tiptapEditor.unregisterPlugin(yUndoPluginKey);
+    this._tiptapEditor.unregisterPlugin(ySyncPluginKey);
+    this._tiptapEditor.registerPlugin(ySyncPlugin(nextFragment));
```
YousefED, line 1016: should we also reregister yUndo? or not needed?

YousefED (commented):
awesome!!! Looks solid.

> It does not yet implement merging the changes back into the old ydoc, but it should be fairly straight-forward to do.

I might be missing something here; but changes are merged right?

nperez0111:
> I might be missing something here; but changes are merged right?

Whoops outdated comment, yep. Changes get merged with Kevin's help

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: Doesn't hurt, can add.

YousefED (commented):
super cool, including the demo!

- maybe extract the demo to a separate example?
- can we add unit tests?
- do you see good ways to organize the better? my concern is that `BlockNoteEditor` will grow in size infinitely if keep adding this kind of logic there

YousefED:
@nperez0111 is there a good reason for you to occasionally use force-push? We had the recent issue with lost data on matthew's branch, but for example when reviewing this PR I also run into this:

image

nperez0111:
I use a rebase workflow https://git-scm.com/book/en/v2/Git-Branching-Rebasing rather than a merge workflow for consistent histories. 

I've never seen that message before. But this branch was based on another branch which got Squash merged. So this branch would have been conflicting with main. So a rebase removes the old commits and re-bases them on to the most recent main (which already moved). 

I can see the diff just fine in my GitHub view though 🤷

Side note I could've recovered Matthew's branch I just wasn't at my computer at the time

YousefED:
> I can see the diff just fine in my GitHub view though 🤷

it's because I select "changes since last review" (which you probably don't have on your side, right?)

YousefED (approved):

Issue #1637, opened 2025-04-24
Pass the blockId to the uploadFile function of the editor


QuentinNativel:
**Is your feature request related to a problem? Please describe.**
I'm building a CMS with blocknote. I have a custom image block on which i want to save the width and the height of the uploaded file in the props

In order to do so i wrote a custom `uploadFile` function that i pass to the editor.
my uploadFile function does 2 things:
- save the image to my cdn
- get the height/width of the image and save it to my block props

The upload function that i submitted takes 2 argument a file and the blockId as described by the typescript type but if i log the blockId that my function receives it is `undefined`.

**Describe the solution you'd like**
The issue comes from the **UploadTab.tsx l.53** where the block.id is not passed to the editor.uploadFile function
I would like to change 
```
let updateData = await editor.uploadFile(file);
```
to the following
```
let updateData = await editor.uploadFile(file, block.id);
```

I tested it and it solved my problem. 
More generally speaking, I also think it is kind of the expected behaviour to also pass the id of the block for which the file panel is opened so the function can have access to it and can modify it if needed

Thanks for all your nice work !

PS: I tried to open a PR but i don't have the right to create branches so i couldn't

nperez0111:
>  I tried to open a PR but i don't have the right to create branches so i couldn't

You should be able to make a fork of the repo, push your branch there, and then PR that into our repo. See here for more info: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request?tool=cli

Happy to accept this change

QuentinNativel:
@nperez0111 Thanks for the answer
I open the PR

PR #1641, opened 2025-04-25
Pass blockId to uploadFile in UploadTab

QuentinNativel:
Implements change proposed in #1637

### Changes
- Inside the `UploadTab`, pass the block id as parameter to the `editor.uploadFile` function as expected in the function doc

- I also suggested a small variable renaming inside the uploadFule function of the editor because in the rest of the code and docs, `bloc` referres to an editor bloc which is an object but uploadFile expects a string as an input the blockId
```
/**
   * The `uploadFile` method is what the editor uses when files need to be uploaded (for example when selecting an image to upload).
   * This method should set when creating the editor as this is application-specific.
   *
   * `undefined` means the application doesn't support file uploads.
   *
   * @param file The file that should be uploaded.
   * @returns The URL of the uploaded file OR an object containing props that should be set on the file block (such as an id)
   */
  uploadFile: (
    file: File,
    blockId?: string
  ) => Promise<string | Record<string, any>>;
  ```

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -625,15 +625,15 @@ export class BlockNoteEditor<
 
     if (newOptions.uploadFile) {
       const uploadFile = newOptions.uploadFile;
-      this.uploadFile = async (file, block) => {
+      this.uploadFile = async (file, blockId) => {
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):
Thanks @QuentinNativel! This looks good to me

PR #1642, opened 2025-04-25
feat: add a ghost writer demo

nperez0111:
This adds a demo which can simulate multiple users within the same document.
It uses partykit to connect to the same server, but generates a random room.
It allows instantiating multiple instances of the ghost writer.


examples/07-collaboration/06-ghost-writer/.bnexample.json
```diff
@@ -0,0 +1,10 @@
+{
+  "playground": true,
+  "docs": true,
```
YousefED, line 3: I don't think this should be listed in docs - it's not a useful code snippet to learn from, right?

YousefED (changes requested):
Nice! I think good solution with the iframes.

- can we display the iframe url somewhere to make it easy to open a clean window with 1 client?
- I don't see awareness (cursors) of the ghost writers, is that missing?
- the preview environment (https://blocknote-git-simulate-multiple-users-typecell.vercel.app/collaboration/ghost-writer) seems to have a styling issue

examples/07-collaboration/06-ghost-writer/.bnexample.json
nperez0111: Didn't know that was possible!

PR #1644, opened 2025-04-25
fix: do not close the menu on content/selection change

nperez0111:
tbh, I'm not entirely sure why this was added in the first place with: https://github.com/TypeCellOS/BlockNote/pull/1304/files

But, the issue with this is that during collaboration, the link menu is closed on every content update made by a collaborator (closing this menu continously making a frustrating link experience).

Could use your insight here @matthewlipski


matthewlipski (approved):
I think the reason for this is because the link creation dropdown would stay open when closing the formatting toolbar and re-opening it somewhere else. I'm not sure why exactly it was a problem then and isn't now (works fine in the preview), but looks ok👍

PR #1656, opened 2025-05-02
perf: this caches the transaction diffing to not have to recompute

nperez0111:


YousefED (approved):
nice! curious though what triggered this PR and the need for a cache. Maybe we can add a comment in the code for others & our future self to remember why the cache is useful?

PR #1657, opened 2025-05-02
chore: add style exports for all stylesheets

RobinClowers:
Adds style exports for all stylesheets, so they can be imported by postcss / tailwind.

When upgrading to tailwind 4, we ran into an issue where our tailwind utility classes suddenly couldn't override blocknote styles anymore. This turns out to be related to css layers. It's easy to solve by importing the styles in css (where we can assign them a layer), but unfortunately it's not possible without these style exports.

See https://github.com/postcss/postcss-import for more details.


nperez0111:
Hi @RobinClowers I missed this one.

Could you expand a bit on why you need it to be part of the `package.json#exports` it wasn't obvious to me from the postcss-import docs.

In my understanding, you should be able to point to the file you need directly as a relative path, are you maybe trying to have it be a bit more pleasant by allowing it to just be `@import 'module-name';`? If so, it isn't obvious to me that `./styles` would do that by default or anything

RobinClowers:
@nperez0111 no problem, thanks for getting back to me.

Specifically this line from the postcss-import docs is relevant:
> When importing a module, it will look for index.css or file referenced in package.json in the style or main fields. 

I want to be able to do this:

```css
@import "@blocknote/core/style.css" layer(components);
```

Which scopes the styles in the components layer, allowing tailwind utilities to override them (because of the layer order, the utility layer has a higher precedence than components).

To your point, I suppose I could import directly from the path, e.g. `node_modules/@blocknote/core/dist/style.css`, but that's more brittle. 

Is there some downside to exposing these exports?

nperez0111:
Ah, that is what I missed. Alright, I'm on board with this change. 

I know this is minor but, could move the `style` export down? `package.json#exports` are actually order-dependent so I don't want this to cause issues for others. I'll get the workflow running so you can test this in your project with the pkg.pr.new install. (may be worth rebasing to the current main if you want the latest version)

RobinClowers:
Sure thing, rebased on the latest main and moved the exports to the end.

RobinClowers:
When I tried to install the PR package, I ran into this:

```
$ bun i https://pkg.pr.new/TypeCellOS/BlockNote/@blocknote/core@1657
bun add v1.2.11 (cb6abd21)
error: Package "@blocknote/core@https://pkg.pr.new/TypeCellOS/BlockNote/@blocknote/core@1657" has a dependency loop
  Resolution: "@blocknote/core@0.26.0"
  Dependency: "@blocknote/core@https://pkg.pr.new/TypeCellOS/BlockNote/@blocknote/core@1657"
error: An internal error occurred (DependencyLoop)
````

Not sure if this is something specific to bun, I've never used pgk.pr.new.

nperez0111:
Yea, with the pkg.pr.new, since the packages depend on each other, you may need to force the versions to resolve to the pkg.pr.new versions instead using an override.

RobinClowers:
@nperez0111 I was able to test this and it worked for me, it's good to merge from my perspective.

nperez0111 (approved):

PR #1661, opened 2025-05-06
fix: Parsing/pasting on `prosemirror-model: 1.25.1`

### LINKED ISSUE #1643: Copy-pasting content gives weird experience
**Describe the bug**
When copying and pasting the content inside the editor, the copied content looks weird. Take a look at the following screen recording.

https://github.com/user-attachments/assets/70f4991e-9cc3-4ea9-a97c-865a2ce643db

**To Reproduce**
Just to open the official basic example at [https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/01-basic/01-minimal?file=App.tsx](https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/01-basic/01-minimal?file=App.tsx), and you will be able to reproduce.

**Misc**
- Node version: Online demo that I don't know the node version
- Package manager: NPM
- Browser: Chrome with Version 135.0.7049.96 (Official Build) (arm64)

### LINKED ISSUE #1645: Lists with markdown -> block has degenerate behaviour when run locally, but not in online playground
**Describe the bug**
Lists in markdown are not being parsed correctly (See this blocknote [example](https://github.com/TypeCellOS/BlockNote/tree/main/examples/05-interoperability/04-converting-blocks-from-md)). On playground or the blocknotejs website, the parser seems to behave with intended behaviour, but when I pull and run locally (npm install, npm start in directory), the parse generates a degenerate block. Could someone on the team tell me what version of blocknote is running on the playground? Images of degenerate and proper output attached

degen local
Image

correct playground
Image

**To Reproduce**
I am using test str:
```
### Steps to Make Cheese

Making cheese involves a series of steps that transform milk into a solid, flavorful product. Here's a basic guide to making a simple cheese:

1. **Gather Ingredients and Equipment:**

   * Fresh milk (preferably whole milk)

   * Rennet (a coagulant)

   * Cheese culture (specific bacteria for cheese-making)

   * Salt

   * Cheesecloth

   * Large pot

   * Thermometer

   * Cheese mold

2. **Prepare the Milk:**

   * Pour the milk into a large pot and slowly heat it to the required temperature (usually around 86°F or 30°C). Stir gently to ensure even heating.
```
Try it when running locally with any blocknote version listed in Misc- this outputs a degenerate output. Then try it in the online playground. Both [playground](https://blocknote-main.vercel.app/interoperability/converting-blocks-from-md) and [blocknotejs.com](https://www.blocknotejs.org/examples/interoperability/converting-blocks-to-md) output correctly.

**Misc**
- Node version: v23.11.0
- Blocknote version: tried 28.0,29.0,29.1
- Package manager: tried both yarn and npm
- Browser: firefox

matthewlipski:
# Issue

This PR addresses breaking changes to parsing introduced by `prosemirror-model: 1.25.1`. Prior to this update, the `DOMParser` would drop any nodes parsed that were not valid in the schema, based on the parent node. For example, take a look at the following HTML:

```html
<ul>
  <li>
    <p>
      ...
    </p>
  </li>
</ul>
```

When parsing this before the update, the `li` element would get parsed as a `bulletListItem` node. Within it, the `p` element should also get parsed as a `paragraph` node. However, because the schema dictates that `paragraph` nodes can't be within `bulletListItem` nodes, the `paragraph` is dropped altogether and ignored.

After the update, the `paragraph` node will no longer get dropped, and the `DOMParser` will instead attempt to insert it somewhere so that it is valid in the schema. Since it can't be a child of the `bulletListItem` node, it instead gets wrapped in a `blockContainer` and `blockGroup`, which then gets inserted into the `bulletListItem`'s parent `blockContainer`. In this scenario, we actually want to ignore the `p` tag and just parse its content.

Overall, the changes in `prosemirror-model: 1.25.1` mean we have to be more diligent when writing parse rules. Paragraphs especially appear in many places, e.g. external HTML paragraphs, internal HTML paragraphs, internal HTML list items, and internal HTML table cells.

# Parse rule changes

## Internal HTML

After updating `prosemirror-model`, external HTML parse rules were being triggered when parsing `blockContent` nodes. This is because `blockContent` nodes can have all kinds of HTML inside, which we actually don't care about for parsing. To fix this, all default `blockContent` nodes have received the following change in their parse rules:

**Before:**

```
{
  tag: "div[data-content-type=" + this.name + "]",
},
```

**After:**

```
{
  tag: "div[data-content-type=" + this.name + "]",
  contentElement: ".bn-inline-content",
},
```

This change was made following a suggestion by Marijn [here](https://github.com/ProseMirror/prosemirror/issues/1526).

When parsing a `blockContent` element, this now tells the `DOMParser` to ignore all descendant elements except the one with the `bn-inline-content` class, and only parse its content.

Additionally, any `bn-inline-content` elements with the `data-editable` attribute have had this attribute removed.

## External HTML

There are a few edge cases that needed to be accounted for, one of which you can see with the `p` inside `li` element example. Here are the specific cases that were fixed:

- `p` elements inside `li` elements parsed as separate `listItem` and `paragraph` nodes.
- `input` elements inside `li` elements causing `checkListItem` nodes to be parsed twice.
- `p` elements inside `th`/`td` elements parsed as separate `tableHeader`/`tableCell` and `paragraph` nodes.
- Media elements (`embed`, `img`, `audio`, and `video`) inside `figure` elements causing their respective `blockContent` nodes to be parsed twice.

Closes #1643
Closes #1645


nperez0111 (commented):
This feels wrong, could you give a write up on why this change is needed.

It feels like it litters the code everywhere, I may just be misunderstanding something here

matthewlipski:
> This feels wrong, could you give a write up on why this change is needed.
> 
> It feels like it litters the code everywhere, I may just be misunderstanding something here

Yep you were right, I went a bit overboard to make sure parse rules weren't being triggered accidentally and pretty much all of the `getAttrs` checks for `element.closest("[data-content-type]")` were redundant.

nperez0111 (approved):

PR #1663, opened 2025-05-06
fix: Force pasting plain text in code block

### LINKED ISSUE #1594: Paste in code block don't work properly
**Describe the bug**
I copy a code, I create a code block, I paste : only the first line is in the code block, the rest is outside the code block.

**To Reproduce**
- copy a code (with ``` and enter) 
- create a code block
- paste into the code block

https://github.com/user-attachments/assets/3c25484e-f54b-483b-b341-abd7d0f574e0

I have tested on the website and on docs.numerique.gouv.fr .

**Misc**
- Node version:
- Package manager:
- Browser: Brave Version 1.77.95 Chromium: 135.0.7049.52 (Official Build) (arm64) / Firefox 137.0 (aarch64)

matthewlipski:
This PR forces pasting plain text from the clipboard inside code blocks, as it cannot be formatted anyway and can otherwise result in unexpected behaviour.

Note: No unit test for this has been added as our you have to explicitly choose the clipboard data format for paste tests, which doesn't support this scenario. I think this is fine as it's an explicit edge case anyway.

Closes #1594

### DIFF
--- a/packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
+++ b/packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
@@ -5,10 +5,6 @@ import type {
   BlockNoteEditor,
   BlockNoteEditorOptions,
 } from "../../../editor/BlockNoteEditor";
-import {
-  getBlockInfoFromResolvedPos,
-  getBlockInfoFromSelection,
-} from "../../getBlockInfoFromPos.js";
 import { isMarkdown } from "../../parsers/markdown/detectMarkdown.js";
 import {
   BlockSchema,
@@ -32,21 +28,20 @@ function defaultPasteHandler({
 }) {
   // Special case for code blocks, as they do not support any rich text
   // formatting, so we force pasting plain text.
-  const selection = editor.prosemirrorView?.state.selection;
-  if (selection) {
-    const blockInfo = getBlockInfoFromSelection(editor.prosemirrorView.state);
+  const isInCodeBlock = editor.transact(
+    (tr) =>
+      tr.selection.$from.parent.type.spec.code &&
+      tr.selection.$to.parent.type.spec.code
+  );
 
-    const selectionInCodeBlock =
-      blockInfo.isBlockContainer && blockInfo.blockContent.node.type.spec.code;
+  if (isInCodeBlock) {
+    const data = event.clipboardData?.getData("text/plain");
 
-    if (selectionInCodeBlock) {
-      const data = event.clipboardData?.getData("text/plain");
-      if (data) {
-        editor.pasteText(data);
-      }
-
-      return true;
+    if (data) {
+      editor.pasteText(data);
     }
+
+    return true;
   }
 
   let format: (typeof acceptedMIMETypes)[number] | undefined;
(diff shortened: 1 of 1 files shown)


packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -26,6 +30,25 @@ function defaultPasteHandler({
   prioritizeMarkdownOverHTML: boolean;
   plainTextAsMarkdown: boolean;
 }) {
+  // Special case for code blocks, as they do not support any rich text
+  // formatting, so we force pasting plain text.
+  const selection = editor.prosemirrorView?.state.selection;
```
nperez0111, line 35: This API is discouraged now with the introduction of BlockNote transactions, you can instead do this:

```ts
  // Special case for code blocks, as they do not support any rich text
  // formatting, so we force pasting plain text.
  const isInCodeBlock = editor.transact((tr) => {
    const blockInfo = getBlockInfoFromTransaction(tr);

    const selectionInCodeBlock =
      blockInfo.isBlockContainer && blockInfo.blockContent.node.type.spec.code;

    return selectionInCodeBlock;
  });

  if (isInCodeBlock) {
    const data = event.clipboardData?.getData("text/plain");
    if (data) {
      editor.pasteText(data);
      return true;
    }
  }
```

packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -26,6 +30,25 @@ function defaultPasteHandler({
...
+  if (selection) {
+    const blockInfo = getBlockInfoFromSelection(editor.prosemirrorView.state);
+
+    const selectionInCodeBlock =
+      blockInfo.isBlockContainer && blockInfo.blockContent.node.type.spec.code;
+
+    if (selectionInCodeBlock) {
```
nperez0111, line 42: Hm, can't we just check what the selection is pointing to, like:

```
tr.selection.$from.parent.type.spec.code&&
    tr.selection.$to.parent.type.spec.code
```

It makes it explicit that a selection may be pointing at different nodes, I'm unsure the right thing to do when only one of them is code. Whereas the getBlockInfoFromSelection kind of hides that.

nperez0111 (changes requested):

packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
```diff
@@ -26,6 +26,24 @@ function defaultPasteHandler({
...
+  );
+
+  if (isInCodeBlock) {
+    const data = event.clipboardData?.getData("text/plain");
+
+    if (data) {
+      editor.pasteText(data);
+    }
+
+    return true;
```
nperez0111, line 44:
```suggestion
    const data = event.clipboardData?.getData("text/plain");

    if (data) {
      editor.pasteText(data);
      return true;
    }
```

Sorry, realized this afterward

matthewlipski: Are you sure about this? Imo if there is no plain text we shouldn't paste anything, since otherwise it'll paste all wonky like in the original issue

matthewlipski: Honestly not sure, I think both cases are not ideal. I think your solution is better for files though come to think of it, since they should get pasted regardless.

nperez0111 (approved):

PR #1664, opened 2025-05-06
feat: File `previewWidth` prop defaults to `undefined`

### LINKED ISSUE #1600: Inserted images have a fixed 512px width
**Describe the bug**
Images are always inserted with a width of 512px. 

**To Reproduce**
- open a blocknote editor with an empty content
- create a small (e.g., 30x30px) image locally
- insert an image block, upload the small image
- observe the image

Expected:
- the image is 30x30px
Actual
- the image is resized to 512px, leady to badly upscaled and pixalted images

Image

**Misc**

matthewlipski:
This PR makes it so file blocks with a `previewWidth` prop have it default to `undefined`. This makes it so that when an image is inserted into the document, it retains its original resolution, or spans the editor width, whichever is lower.

Closes #1600


nperez0111 (commented):
When the width is undefined like this. What is the behaviour for very large, and very small images?

matthewlipski:
> When the width is undefined like this. What is the behaviour for very large, and very small images?

Very large images are max. the width of the editor, set in CSS.

Very small images will just be very small resizing and stuff should still work, though we have a min-width when resizing of 64px, otherwise the resize handles don't have enough space. If an image is less than 64px wide, it will get resized to 64px immediately when dragging the resize handles. I think it's a pretty minor edge case that someone uploads an image or video <64px wide, as even things like icons are not usually smaller than that, so imo no need to add extra logic for this.

nperez0111 (approved):

PR #1668, opened 2025-05-08
feat(i18n): add zh-TW locale files

linyiru:
Contributing `zh-TW` locale for Tranditional Chinese for the project.


nperez0111:
Thanks for the PR!

PR #1674, opened 2025-05-09
feature: AI

YousefED:


eyesfivemserver:
Impressive project structure! The architecture really demonstrates solid design principles.

packages/xl-ai/src/components/AIMenu/PromptSuggestionMenu.tsx
```diff
@@ -0,0 +1,135 @@
...
+          variant={"large"}
+          icon={<RiSparkling2Fill />}
+          value={promptTextToUse || ""}
+          autoFocus={true}
+          placeholder={props.placeholder}
+          disabled={props.disabled}
+          onKeyDown={handleKeyDown}
+          onChange={handleChange}
+          autoComplete={"off"}
+          //  TODO: loader or error
```
YousefED, line 112: @matthewlipskicould you implement this (for mantine / ariakit / shadcn of course). My idea was to use `rightSection` for this

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -488,3 +492,353 @@ export function nodeToBlock<
...
+ *     </blockContainer>
+ *    <blockContainer> (child block 2)
+ *       <p, heading, etc.>
+ *     </blockContainer>
+ *   </blockContainer>
+ *  </blockGroup>
+ * </blockGroup>
+ *
+ */
+export function prosemirrorSliceToSlicedBlocks<
```
YousefED, line 712: @matthewlipski I think this code is in part similar to some of the logic you wrote in getSelection

Can you take an extra look at this and see if it makes sense - or even whether we should combine the two somehow? (don't spend > 1 hour investigating)

packages/core/src/api/blockManipulation/selections/selectioncut.test.ts
```diff
@@ -0,0 +1,314 @@
+import { TextSelection } from "prosemirror-state";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { PartialBlock } from "../../../blocks/defaultBlocks.js";
+import { BlockNoteEditor } from "../../../editor/BlockNoteEditor.js";
+
+// describe("Test getSelectionCutBlocks", () => {
+  const initialContent: PartialBlock[] = [
```
YousefED, line 7: @matthewlipski could you make these tests in line with your new testing setup?

I think they were originally copied from existing tests - so you can probably reuse existing test document / selection data

shared/formatConversionTestUtil.ts
```diff
@@ -1,3 +1,5 @@
+// TODO: remove duplicate file
```
YousefED, line 1: @matthewlipski this file now lives in 3 spots. If the code is exactly the same - I think best to import the one from @shared?

packages/xl-ai/src/components/AIMenu/PromptSuggestionMenu.tsx
```diff
@@ -0,0 +1,135 @@
...
+    [handleEnter, handler, items.length],
+  );
+
+  // Resets index when items change
+  useEffect(() => {
+    setSelectedIndex(0);
+  }, [promptTextToUse, setSelectedIndex]);
+
+  return (
+    <div className={"bn-combobox"}>
```
YousefED, line 96: @matthewlipski does this follow the correct pattern with classnames, etc?

packages/shadcn/src/suggestionMenu/SuggestionMenuItem.tsx
```diff
@@ -24,7 +24,7 @@ export const SuggestionMenuItem = forwardRef<
 
     const overflow = elementOverflow(
       itemRef.current,
-      document.querySelector(".bn-suggestion-menu")!
+      document.querySelector(".bn-suggestion-menu, #ai-suggestion-menu")! // TODO
```
YousefED, line 27: @matthewlipski this still seems nasty, can't we pass an `elementOverflow` into the `SuggestionMenuItem` to for example? Or find another way to do this at a higher level without `querySelector`?

packages/xl-ai/src/components/AIMenu/AIMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+    <PromptSuggestionMenu
+      onManualPromptSubmit={
+        props.onManualPromptSubmit || onManualPromptSubmitDefault
+      }
+      items={items}
+      promptText={prompt}
+      onPromptTextChange={setPrompt}
+      placeholder={
+        aiResponseStatus === "thinking"
+          ? "Thinking..."
```
YousefED, line 95: @matthewlipski could you extract these strings and clean unused strings from the dictionary?

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,321 @@
...
+  //   );
+  // }
+  // withSuggestChanges(fakeDispatch).bind({
+  //   get state() {
+  //     return editor.prosemirrorState;
+  //   },
+  // })(tr);
+  // disableSuggestChanges(editor.prosemirrorState, editor.dispatch);
+
+  // TODO: errors thrown here are not shown in UI / console
```
YousefED, line 319: @YousefED TODO

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -216,3 +219,47 @@ export function setSelection(
   //  restriction that the start/end blocks must have content.
   tr.setSelection(TextSelection.create(tr.doc, startPos, endPos));
 }
+
+export function getSelectionCutBlocks(tr: Transaction) {
```
nperez0111, line 223: This is likely over-complicated. There is probably a way to do this via shared parent depths

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1099,12 +1132,26 @@ export class BlockNoteEditor<
...
+  /**
+   * Gets a snapshot of the current selection. This contains all blocks (included nested blocks)
+   * that the selection spans across.
+   *
+   * If the selection starts / ends halfway through a block, the returned block will be
+   * only the part of the block that is included in the selection.
+   */
+  public getSelectionCutBlocks() {
+    return this.transact((tr) => getSelectionCutBlocks(tr));
+  }
```
nperez0111, line 1153: I think the distinction here is selectionAround vs. selectionWithin

getSelection is getSelectionAround
getSelectionCutBlocks is getSelectionWithin

packages/xl-ai/src/AIExtension.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  /**
+   * Accept the changes made by the LLM
+   */
+  public acceptChanges() {
+    applySuggestions(this.editor.prosemirrorState, (tr) => {
+      // TODO: @Nick, I don't think there's currently a cleaner way to do this?
+      this.editor._tiptapEditor.dispatch(tr);
+    });
```
nperez0111, line 163: Can do this instead:

```ts
this.editor.exec(applySuggestions)
```

Because this follows the prosemirror command pattern

packages/core/src/api/exporters/markdown/markdownExporter.ts
```diff
@@ -16,7 +16,12 @@ import { addSpacesToCheckboxes } from "./util/addSpacesToCheckboxesRehypePlugin.
 
 // Needs to be sync because it's used in drag handler event (SideMenuPlugin)
 // Ideally, call `await initializeESMDependencies()` before calling this function
-export function cleanHTMLToMarkdown(cleanHTMLString: string) {
+export function cleanHTMLToMarkdown(
```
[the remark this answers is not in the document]
YousefED: ignore for now as parsing is not implemented

packages/xl-ai/src/AIExtension.ts
```diff
@@ -0,0 +1,293 @@
...
+        throw new Error("not implemented");
+        // ret = await llm.markdown.call(this.editor, options);
+      } else if (options.dataFormat === "html") {
+        ret = await llm.html.call(this.editor, {
+          ...options,
+          onStart: () => {
+            this.setAIResponseStatus("ai-writing"); // TODO: also apply to other formats
+          },
+          onBlockUpdate: (blockId: string) => {
+            // NOTE: does this setState with an anon object trigger unnecessary re-renders?
```
YousefED, line 253: @nperez0111 this does cause unnecessary renders (try adding a log in BlockNoteAIAU)

We could solve it by lifting blockId and status to the root, but then we lose Typescript help (see comment on line 26 of file). any insights?

packages/xl-ai/src/streamTool/streamTool.ts
```diff
@@ -0,0 +1,96 @@
+import { DeepPartial } from "ai";
+import type { JSONSchema7 } from "json-schema";
+
+export type InvalidOrOk<T> =
+  | {
+      result: "invalid";
+      reason: string;
+    }
+  | { result: "ok"; value: T };
```
nperez0111, line 9: This is more commonly referred to as an `Result` type:

```ts
export type Result<T, E = Error> =
  | { ok: true; value: T }
  | { ok: false; error: E };
```

packages/xl-ai/src/streamTool/streamTool.ts
```diff
@@ -0,0 +1,96 @@
...
+  parameters: JSONSchema7;
+  /**
+   * Validates the input of the tool call
+   *
+   * @param operation - The operation to validate.
+   * This can be a partial object as the LLM may not have completed the operation yet.
+   * The object is not guaranteed to be of type DeepPartial<T> as the LLM might not conform to the schema correctly - so this needs to be validated.
+   */
+  validate: (
+    operation: DeepPartial<T>, // TODO: maybe `unknown` is better?
```
nperez0111, line 39: Agreed, `unknown` is better

packages/xl-ai/src/streamTool/streamTool.ts
```diff
@@ -0,0 +1,96 @@
...
+  execute: (
+    operationsStream: AsyncIterable<{
+      operation: StreamToolCall<StreamTool<any>[]>;
+      isUpdateToPreviousOperation: boolean;
+      isPossiblyPartial: boolean;
+    }>,
+  ) => AsyncIterable<{
+    operation: StreamToolCall<StreamTool<any>[]>;
+    isUpdateToPreviousOperation: boolean;
+    isPossiblyPartial: boolean;
```
nperez0111, line 59: I don't know about these names, but I can't think of anything better.

isUpdateToPreviousOperation -> amend?
isPossiblyPartial -> isComplete?

IDK

packages/xl-ai/src/streamTool/streamTool.ts
```diff
@@ -0,0 +1,96 @@
...
+   * @returns the stream of operations that have not been processed (and should be passed on to execute handlers of other StreamTools)
+   */
+  execute: (
+    operationsStream: AsyncIterable<{
+      operation: StreamToolCall<StreamTool<any>[]>;
+      isUpdateToPreviousOperation: boolean;
+      isPossiblyPartial: boolean;
+    }>,
+  ) => AsyncIterable<{
+    operation: StreamToolCall<StreamTool<any>[]>;
```
nperez0111, line 57: This always resolves to an `any` is that intentional? I prefer `unknown`

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+};
+
+/**
+ * Takes an array of ReplaceSteps and applies them as a human-like agent.
+ * Every step is split into 3 phases:
+ * - select the text to be replaced (1 transaction per ReplaceStep)
+ * - replace the text with the first character of the replacement (if any) (1 transaction per ReplaceStep)
+ * - insert the replacement character by character (strlen-1 transactions per ReplaceStep)
+ */
+export function getStepsAsAgent(doc: Node, pmSchema: Schema, steps: Step[]) {
```
nperez0111, line 29: This makes more sense to operate on a whole transform, and give a transform back rather than an array of steps:

```ts
export function getStepsAsAgent(trToTransform: Transform) {
  const pmSchema = getPmSchema(trToTransform);
  const { modification } = pmSchema.marks;

  const agentSteps: AgentStep[] = [];

  const tr = new Transform(trToTransform.doc);

  for (const step of trToTransform.steps) {
    ...

return tr;
```

packages/xl-ai/src/api/formats/base-tools/createAddBlocksTool.ts
```diff
@@ -0,0 +1,290 @@
...
+            return {
+              result: "invalid",
+              reason: "referenceId must end with $",
+            };
+          }
+
+          referenceId = referenceId.slice(0, -1);
+        }
+
+        const block = editor.getBlock(referenceId);
```
nperez0111, line 142: This is always pulling from the current editor state. I would probably have an in-memory version of the document & pull from that instead. Probably more of an issue for collab mode than anything

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+  if (step.type === "select") {
+    await new Promise((resolve) => setTimeout(resolve, 100));
+  } else if (step.type === "insert") {
+    await new Promise((resolve) => setTimeout(resolve, 10));
+  } else if (step.type === "replace") {
+    await new Promise((resolve) => setTimeout(resolve, 200));
+  } else {
+    throw new UnreachableCaseError(step.type);
+  }
+}
```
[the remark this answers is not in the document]
nperez0111: I liked this result:

```ts
export async function delayAgentStep(step: AgentStep) {
  const jitter = Math.random() * 0.3 + 0.85; // Random between 0.85 and 1.15
  if (step.type === "select") {
    await new Promise((resolve) => setTimeout(resolve, 100 * jitter));
  } else if (step.type === "insert") {
    await new Promise((resolve) => setTimeout(resolve, 10 * jitter));
  } else if (step.type === "replace") {
    await new Promise((resolve) => setTimeout(resolve, 200 * jitter));
  } else {
    throw new UnreachableCaseError(step.type);
  }
}
```

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+          })
+          .addToSet(marks);
+      }
+
+      const attrNames = new Set([
+        ...Object.keys(newNode.attrs),
+        ...Object.keys(oldNode.attrs),
+      ]);
+      for (const attr of attrNames) {
+        if (newNode.attrs[attr] !== oldNode.attrs[attr]) {
```
nperez0111, line 87: attributes can be objects in prosemirror, though this works for blocknote attributes

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+      const stepIndex = tr.steps.length;
+      tr.setNodeMarkup(
+        tr.mapping.map(step.from),
+        newNode.type,
+        newNode.attrs,
+        marks,
+      );
+
+      agentSteps.push({
+        prosemirrorSteps: tr.steps.slice(stepIndex),
```
nperez0111, line 108: I'm a bit confused on the `tr.steps.slice(stepIndex)`, isn't that just always going to be slicing the last one off (i.e. the `tr.setNodeMarkup` you did between)?

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+
+    let replaceEnd = tr.mapping.map(step.to);
+    const replaceFrom = tr.mapping.map(step.to);
+    let first = true;
+
+    // Note that below, we don't actually delete content, but we mark it as deleted
+    // with a deletion mark. Similarly, we mark inserted content with an insertion mark.
+    // It might be cleaner to;
+    // a) make this optional
+    // b) actually delete / insert the content and let prosemirror-suggest-changes handle the marks
```
nperez0111, line 168: Yea, I would consider doing b, it feels like prosemirror-suggest-changes should be doing this transformation, though I can understand wanting the fine-grained control of how it displays visually.

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+        }
+        tr.addMark($pos.pos, replaceEnd, pmSchema.mark("deletion", {}));
+        replaceEnd = tr.mapping.map(step.to);
+      }
+
+      // note, instead of inserting one charachter at a time at the end (a, b, c)
+      // we're replacing the entire part every time (a, ab, abc)
+      // would be cleaner to do just insertions, but didn't get this to work with the add operation
+      // (and this kept code relatively simple)
+      const replacement = new Slice(step.slice.content.cut(0, i), 0, 0);
```
nperez0111, line 189: This will affect how the ops are laid out in the undo/redo stack.
Here is where in prosemirror-history steps are merged together: https://github.com/ProseMirror/prosemirror-history/blob/master/src/history.ts#L237
This is a ReplaceStep's merge method: https://github.com/ProseMirror/prosemirror-transform/blob/137ff74738bd1b50d49416cd6cfdbbf52cb059ef/src/replace_step.ts#L48-L62
So, this can only merge ops which occur one after the other, not overlapping ranges like you've got here.

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+
+      tr.replace(replaceFrom, replaceEnd, replacement).addMark(
+        replaceFrom,
+        replaceFrom + replacement.content.size,
+        pmSchema.mark("insertion", {}),
+      );
```
nperez0111, line 195: This will also affect how undo/redo stack works, since an addMark step cannot be merged with a replace step.

What I've done before for this, is to add the mark to the `replacement` content separately, so it doesn't need a separate addMark step

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+        replaceFrom + replacement.content.size,
+        (node, pos) => {
+          if (
+            pos < replaceFrom ||
+            pos > replaceFrom + replacement.content.size
+          ) {
+            return true;
+          }
+          if (node.isBlock) {
+            tr.addNodeMark(pos, pmSchema.mark("insertion", {}));
```
nperez0111, line 208: Same here

packages/xl-ai/src/prosemirror/changeset.ts
```diff
@@ -0,0 +1,340 @@
...
+  type?: "mark-update" | "node-type-or-attr-update";
+};
+
+/**
+ * Adds missing changes to the changes array.
+ * This is needed because prosemirror-changeset may miss some changes,
+ * such as marks or node attributes.
+ *
+ * NOTE: we might be able to replace this code with a custom encoder
+ * (this didn't exist yet when this was written)
```
nperez0111, line 21: Yea, I think the whole custom encoder thing is just for this. Probably okay for now, but could be simplified

packages/xl-ai/src/prosemirror/changeset.ts
```diff
@@ -0,0 +1,340 @@
...
+
+  updatedDoc.descendants((node) => {
+    if (node.type.name === "tableCell") {
+      tableCellsNew.add(JSON.stringify(node.toJSON()));
+    }
+  });
+
+  const tableCells = new Set(
+    [...tableCellsOld].filter((cell) => tableCellsNew.has(cell)),
+  );
```
nperez0111, line 145: const tableCells = tableCellsOld.intersection(tableCellsNew);

packages/xl-ai/src/prosemirror/changeset.ts
```diff
@@ -0,0 +1,340 @@
...
+
+  const encoder: TokenEncoder<any> = {
+    encodeCharacter: (char) => char,
+    encodeNodeStart: (node) => {
+      if (node.type.name === "tableCell") {
+        const str = JSON.stringify(node.toJSON());
+        if (tableCells.has(str)) {
+          return str;
+        }
+        return node.type.name;
```
nperez0111, line 155: Maybe worth a comment that you are encoding it like this to "flatten" the changeset

packages/xl-ai/src/prosemirror/changeset.ts
```diff
@@ -0,0 +1,340 @@
...
+      }
+      return node.type.name;
+    },
+    encodeNodeEnd: (node) => {
+      if (node.type.name === "tableCell") {
+        const str = JSON.stringify(node.toJSON());
+        if (tableCells.has(str)) {
+          return str;
+        }
+        // return Math.random();
```
nperez0111, line 165: Haha 😂

packages/xl-ai/src/streamTool/callLLMWithStreamTools.ts
```diff
@@ -0,0 +1,283 @@
...
+          switch (chunk.type) {
+            case "object":
+              controller.enqueue(chunk.object);
+              break;
+
+            case "text-delta":
+            case "finish":
+              break;
+            case "error":
+              throw chunk.error;
```
nperez0111, line 273: Why throw on error instead of using the `controller.error` https://developer.mozilla.org/en-US/docs/Web/API/TransformStreamDefaultController/error

packages/xl-ai/src/api/formats/html-blocks/defaultHTMLPromptBuilder.ts
```diff
@@ -0,0 +1,111 @@
...
+      content: `First, determine what part of the document the user is talking about. You SHOULD probably take cursor info into account if needed.
+       EXAMPLE: if user says "below" (without pointing to a specific part of the document) he / she probably indicates the block(s) after the cursor. 
+       EXAMPLE: If you want to insert content AT the cursor position (UNLESS indicated otherwise by the user), then you need \`referenceId\` to point to the block before the cursor with position \`after\` (or block below and \`before\`).
+      
+      Prefer updating existing blocks over removing and adding (but this also depends on the user's question).`,
+    },
+  ];
+}
+
+export const defaultHTMLPromptBuilder: PromptBuilder = async (editor, opts) => {
```
YousefED, line 95: @nperez0111 as discussed on slack:

> I think it would be even nicer if you can provide a function that takes data similar to promptManipulateSelectionHTMLBlocks, but then we’d need to call getPromptData automatically and also make that optionally configurable. Now you just need to call the helpers yourself (similar to how defaultHTMLPromptBuilder does this)
> I think if we generalize this further it might get quite complicated considering the difference between selection / non selection, and different formats

packages/core/src/api/blockManipulation/selections/selection.ts
YousefED: Got it, I think an easier solution might be similar to what @matthewlipski built for `getSelection` above

packages/core/src/editor/BlockNoteEditor.ts
YousefED: I can see we'd like a better name, but I also don't find "Around" vs "Within" very intuitive tbh. Unless you've seen that pattern before?

packages/xl-ai/src/prosemirror/agent.ts
YousefED: I agree on the input, but the current return type does have additional metadata, right? So we can't really return a transform?

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: Prosemirror's replaceAround for one (it encompassing the whole node). 

Maybe it is just one getSelection, and it can return `blocks` vs `partialContent`?

packages/xl-ai/src/prosemirror/agent.ts
YousefED: Good to know. I think `prosemirror-suggest-changes` also only supports string attrs fwiw

packages/xl-ai/src/prosemirror/agent.ts
nperez0111: Yep, a transform wouldn't be enough for that.

packages/xl-ai/src/prosemirror/agent.ts
YousefED: Yeah the initial prototype used the current approach, I only later hooked up prosemirror-suggest-changes. I tried to migrate to having that handle the insertions, but was running into a few edge-cases which made it more cumbersome than I initially hoped (ofc, might have just been a single mapping / positioning error). For now I think we can keep it as is, but good to leave the note here if we run into things that the "other approach" could address, wdyt?

packages/xl-ai/src/prosemirror/agent.ts
nperez0111: Sure, I guess I was mostly confused at why tracking it indirectly, but further down the file it made more sense when it actually could've been multiple steps

packages/xl-ai/src/prosemirror/agent.ts
YousefED: Let's discuss undo / redo tomorrow morning - let's look at the zoomed out issue first because I think the way undo / redo works in prosemirror might not be applicable to our use-case at all

packages/xl-ai/src/prosemirror/agent.ts
nperez0111: Totally fine with that. If we experience a bug with it, I'd probably just throw it away and re-implement

packages/xl-ai/src/prosemirror/agent.ts
```diff
@@ -0,0 +1,325 @@
...
+    index += sel.head;
+    if (!content.isLeaf) {
+      // for regular nodes, add 1 position for the node opening
+      // (annoyingly TextSelection.atStart doesn't account for this)
+      index += 1;
+    }
+    return index;
+  }
+  return undefined;
+}
```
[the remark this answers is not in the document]
YousefED: great, thanks!! knew there must have been a better way. extracting function + adding test

packages/xl-ai/src/prosemirror/agent.ts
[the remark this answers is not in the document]
YousefED: cool let's do it!

packages/xl-ai/src/prosemirror/changeset.ts
YousefED: do you know if this consistently works or requires extra setup? I tried this but got an error at some point that `intersection` is not available

packages/xl-ai/src/prosemirror/changeset.ts
YousefED: resolving ;)

packages/xl-ai/src/prosemirror/changeset.ts
YousefED: The idea here was to give two equal table cells in before / after a unique Encoding, to nudge prosemirror-changeset to keep them intact.

I think this encoder improved the table handling a bit, but I'm still running into issues like described here: https://github.com/ProseMirror/prosemirror-changeset/issues/22

will require more research unless you have a good idea. At least, I'll add more documentation to the code here

packages/xl-ai/src/streamTool/callLLMWithStreamTools.ts
YousefED: oof not sure why I didn't do so! will test and apply, thanks

packages/xl-ai/src/streamTool/streamTool.ts
YousefED: thanks, refactored (kept `error` as `string` for now instead of a generic `E`)

packages/core/src/editor/BlockNoteEditor.ts
YousefED: @matthewlipski also had good ideas about this with maybe 1 method that exposes both values; wdyt about that?

@matthewlipski can you write down your earlier thoughts?

packages/shadcn/src/suggestionMenu/SuggestionMenuItem.tsx
YousefED: tracking here and in linear; https://github.com/TypeCellOS/BlockNote/pull/1104

packages/xl-ai/src/prosemirror/changeset.ts
nperez0111: Hm, it's a newer feature, but should be reasonably available: https://caniuse.com/?search=Set%20intersection

No, biggie yours works too

packages/xl-ai/src/streamTool/streamTool.ts
YousefED: TODO: comment

packages/xl-ai/src/components/AIMenu/PromptSuggestionMenu.tsx
matthewlipski: Looks ok imo

packages/core/src/editor/BlockNoteEditor.ts
matthewlipski: I would just have a single `getSelection`:

```ts
public getSelection(): {
    // Blocks spanned by selection, same as existing `getSelection`.
    blocks: Block<BSchema, I, S>[];
    // Fields returned by what's currently `getSelectionCutBlocks.
    cutBlocks: Block<BSchema, I, S>[];
    blockCutAtStart: string | undefined;
    blockCutAtEnd: string | undefined;
} { ... }
```

This way we don't have to have 2 separate functions (which imo is confusing for consumers) while also not introducing any breaking changes.

packages/xl-ai/src/streamTool/streamTool.ts
YousefED: documented here https://github.com/TypeCellOS/BlockNote/blob/69d06f2eb392387484aa9aec99b8e0ebcaa67f54/packages/xl-ai/src/streamTool/callLLMWithStreamTools.ts#L54

PR #1677, opened 2025-05-10
Fix import for buffer as per usage instructions

### LINKED ISSUE #1675: Docx exporter encounters error due to how it's importing buffer package
**Describe the bug**
In the course of normal usage the call to `toDocxJsDocument` does not work with the following error:  

```
@blocknote_xl-docx-e…r.js?v=258d18af:380 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'from')
    at async App.tsx:23:26
```

Seems to be that `buffer` package removed the named export recently. If change `const Buffer = (await import("buffer")).default.Buffer;` to `const Buffer = (await import("buffer")).default;` then it works correctly.

I will create a PR to fix shortly.

**To Reproduce**
1. [https://stackblitz.com/edit/github-7wt3cfcn?file=App.tsx](https://stackblitz.com/edit/github-7wt3cfcn?file=App.tsx)
2. Click the button at the top.

richardvanbergen:
Doesn't show up in tests as this is browser only issue.

I have tested in my project though using `pnpm link` and it seems to be working for me now.

Fixes #1675

### DIFF
--- a/packages/xl-docx-exporter/src/docx/docxExporter.ts
+++ b/packages/xl-docx-exporter/src/docx/docxExporter.ts
@@ -154,7 +154,10 @@ export class DOCXExporter<
       geistMonoFont instanceof Uint8Array
     ) {
       // conversion with Polyfill needed because docxjs requires Buffer
-      const Buffer = (await import("buffer")).default.Buffer;
+      // NOTE: the buffer/ import is intentional and as documented in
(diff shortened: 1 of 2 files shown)


nperez0111:
Thanks @richardvanbergen this looks correct to me

PR #1680, opened 2025-05-12
chore: reformat everything with new prettier version

nperez0111:
A few weeks ago, I hoisted prettier so that it does not have to be configured in every package, this had the side-effect of bumping to the latest version of prettier which changed some of the defaults.

Now, whenever we modify a file it is resulting in a larger diff than necessary, so this just reformats all of the files.

We can tweak the prettier config if it isn't what we want, but this is mostly the defaults that prettier applies by default


YousefED:
nice. let's do this after the AI PR is merged to avoid too many conflicts (or do it on both branches)

PR #1690, opened 2025-05-16
docs: Custom AI menu items docs

matthewlipski:
This PR adds an example for custom AI menu items, and makes a few code changes to provide an API for that.

The READMEs of all AI examples have also been updated.

TODO:

- Docs page


YousefED (approved):
Nice! Will merge it already and take it from there. I think it might make sense to just get rid of `BlockNoteAIUI` and only use the controller - wdyt?

PR #1699, opened 2025-05-19
feat: Merged `getSelection` and `getSelectionCutBlocks`

matthewlipski:
This PR merges the currently 2 separate `getSelection` methods into one.

The functionality of `getSelection` is unchanged, only the additional fields from `getSelectionCutBlocks` have been added (with `blocks` renamed to `cutBlocks`).

`getSelectionCutBlocks` is slightly different as the `blocks` field has been renamed to `cutBlocks` so as not to clash with the same field that was already in `getSelection`. Also, now returns `undefined` when the selection is collapsed.

TODO:

- Review increment selection test cases as these are erroring. @yousef are they necessary? Since we already test offsets in the get selection tests and AFAIK some of the selections represented here arent actually possible


packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -127,8 +143,55 @@ export function getSelection<
...
+        ? $startBlockBeforePos.nodeAfter!.attrs.id
+        : undefined,
+    blocksCutAtEnd:
+      endNode!.nodeSize !== cutEndNode!.nodeSize
+        ? $endBlockBeforePos.nodeAfter!.attrs.id
+        : undefined,
+    _meta: {
+      startPos: tr.selection.from,
+      endPos: tr.selection.to,
+    },
```
nperez0111, line 194: I think this is useful to have, so let's just make it `meta`

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -127,8 +143,55 @@ export function getSelection<
...
     blocks,
+    cutBlocks,
+    blocksCutAtStart:
+      startNode.nodeSize !== cutStartNode!.nodeSize
+        ? $startBlockBeforePos.nodeAfter!.attrs.id
+        : undefined,
+    blocksCutAtEnd:
+      endNode!.nodeSize !== cutEndNode!.nodeSize
+        ? $endBlockBeforePos.nodeAfter!.attrs.id
+        : undefined,
```
nperez0111, line 190: I'm still going to have to push back on this naming, I don't know what a `blocksCutAtStart` means.

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -127,8 +143,55 @@ export function getSelection<
...
 
+  const selectionContent = tr.selection.content().content;
+  selectionContent.descendants((node) => {
+    if (cutStartNode && cutEndNode) {
+      return false;
+    }
+
+    if (
+      node.type.name === startNode.type.name &&
+      node.attrs.id === startNode.attrs.id
```
nperez0111, line 154:
```suggestion
      node.eq(startNode)
```

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -127,8 +143,55 @@ export function getSelection<
...
+    ) {
+      cutStartNode = node;
+    }
+
+    if (
+      node.type.name === endNode.type.name &&
+      node.attrs.id === endNode.attrs.id
```
nperez0111, line 161:
```suggestion
      node.eq(endNode)
```

packages/core/src/api/blockManipulation/selections/selection.ts
nperez0111: Ah maybe this won't match because it is the sliced node?

packages/core/src/api/blockManipulation/selections/selection.ts
nperez0111: OK, I think I understand what cutBlocks are now from the tests. It is the slice of content that the selection range is over. imo, I don't think we should introduce another term on top of what prosemirror has already established for this (a slice).

I think it should be: `blockSlice`, I still don't understand the usefulness of blocksCutAtStart & blocksCutAtEnd, but I think it is something more like `startBlockId` or `sliceBlockId`

Issue #1687, opened 2025-05-15
Image alignment broken in latest update


nperez0111:
**Describe the bug**

<img width="846" alt="Image" src="https://github.com/user-attachments/assets/31eb52fe-c31a-4aa0-a052-04e024fb0036" />

When trying to use image alignment, it isn't being respected

**To Reproduce**
<clear steps to reproduce are super helpful! Best is to provide an online sandbox, [click to create one](https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/01-basic/01-minimal?file=App.tsx)>
Reproducible on: https://www.blocknotejs.org/examples/basic/minimal
**Misc**
- Node version:
- Package manager:
- Browser: 
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

PR #1700, opened 2025-05-19
fix: Re-added `display: flex` to block content

matthewlipski:
This style must have gotten removed somewhere.

Closes 

TODO #1687

### DIFF
--- a/packages/core/src/editor/Block.css
+++ b/packages/core/src/editor/Block.css
@@ -14,6 +14,7 @@ BASIC STYLES
 }
 
 .bn-block-content {
+  display: flex;
   padding: 3px 0;
   transition: font-size 0.2s;
   width: 100%;


YousefED:
@matthewlipski this was removed intentionally here; https://github.com/TypeCellOS/BlockNote/pull/1629

tests/src/end-to-end/static/static.test.ts-snapshots/static-rendering-chromium-linux.png
nperez0111, line 1: The current spacing looks off, could we get it back to how it was before this?

packages/core/src/editor/Block.css
```diff
@@ -214,14 +227,6 @@ NESTED BLOCKS
   text-decoration: line-through;
 }
 
-.bn-block-content[data-text-alignment="center"] {
-  justify-content: center;
-}
-
-.bn-block-content[data-text-alignment="right"] {
-  justify-content: flex-end;
-}
```
nperez0111, line 223: Is this right?

tests/src/end-to-end/static/static.test.ts-snapshots/static-rendering-chromium-linux.png
nperez0111: Like there is something with the bullet list spacing to be doing this: 
image
(where the cursor is right up to the bullet)

matthewlipski: Looks like `display: inline` is just a nightmare. The empty list item cursor is browser-specific, and the PW snapshot has an additional space between the inline elements that you can't remove without some hacks. I'll see if I can get the selection working with flexbox.

PR #1707, opened 2025-05-21
feat: Toggle blocks

### LINKED ISSUE #196: Support for folding?
In IDEs, blocks are often associated with folding. For example in VS Code:

Unfolded:
image

Folded:
image

Is there support for folding in BlockNote? Has anyone built this feature on top of BlockNote? Would it be straightforward to add?

### LINKED ISSUE #549: Spinner showing in all icons

### LINKED ISSUE #910: The ability for the blockgroup element to collapse/expand.
Currently, the blockgroup element does not collapse/expand, which is inconvenient.

Adding a simple floating button on the right side of the editor (just a button for mobile) would immediately solve the problem of long text and serve as an alternative to adding “nested blocks” in a custom block (``{ content: 'block+' }``).
This functionality can be made available in a parameter ``{ toggleBlocksVisibility: { enabled: true } }``.

Make custom tiptap extension or use pure tiptap with custom extension

**Additional context**

**Bonus**

matthewlipski:
# Toggle wrapper element/component

This PR adds a new component, `ToggleWrapper`, as well as a vanilla HTML element implementation of it, `createToggleWrapper`. These can be used to wrap existing blocks, or create new ones and allow the user to toggle showing/hiding their children.

The toggle wrapper renders a button to the left of the block to show/hide the children. Additionally, if children are shown but none exist, a button is rendered to prompt the user to add a block. Because the toggled state is only in the view, toggling the visibility of children only happens client-side when using collaboration.

A `toggledState` object can also be passed to the toggle wrapper, which includes a `get` and `set` function to be able to preserve the state on re-render and reload. By default, this is done using local storage.

# Schema changes

This PR also adds toggle headings and toggle list items to the default editor schema.

The toggle headings are implemented by adding an `isToggleable` prop, and rendering the block in a toggle wrapper when this is set to `true`.

The toggle list item is a new, separate block, which always renders a paragraph in a toggle wrapper. Like the other list items though, hitting Enter in one creates a new block of the same type, which is why I made it a separate block rather than adding `isToggleable` to the existing paragraph or list item blocks. Additionally, the block is exported differently to formats like docx, as the content gets prepended with a ">".

# TODO

Currently some `xl-ai` tests are failing, and these have been temporarily disabled.

Closes #196
Closes #549
Closes #910


packages/core/src/blocks/ToggleHeadingBlockContent/ToggleHeadingBlockContent.ts
```diff
@@ -0,0 +1,22 @@
...
+  {
+    ...headingImplementation,
+    render: (block, editor) =>
+      createToggleWrapper(
+        block,
+        editor,
+        headingImplementation.render(block, editor),
+      ),
+  },
+);
```
nperez0111: This is fantastic, composability!

nperez0111: Now, this is creating a new block type, can we instead just "wrap over" heading, and if it has a prop `isToggleable` it switches it's render implementation from the default, into using this toggle wrapper?

So that there is only one heading implementation.

matthewlipski: Should I also revert the heading block to using `createStronglyTypedTiptapNode`? Since the keyboard shortcuts and input rules won't work if we stick with `createBlockSpec`. Or do you wanna see how it looks using the `createBlockSpec` first?

nperez0111: For now, it is more important to see that the pattern can work before being caught up in those sort of details.

Once we can see that it works for headings, we can see if we can do it for something like list items

matthewlipski: Makes sense, updated the code👍

nperez0111: Great, this looks about right. Ideally it'd be "around" the heading component, but that is minor and fixable.

To explain more, like this you made this:

```tsx
function Heading(props){
function renderHeading(){}

return props.isToggleable ? <Toggle>{renderHeading()}</Toggle> : renderHeading()
}

blocks: { heading: Heading }
```

Where I'd like something more like this:

```tsx
function Heading() {

return renderHeading()
}

const WrapWithToggle=(component) => props => {
return props.isToggleable ? <Toggle>{component}</Toggle> : component
}

blocks: { heading: WrapWithToggle(heading) }
```

nperez0111: Now that we've validated this approach, I do think you have the right idea with making this a strongly typed tiptap node to keep all previous functionality.

matthewlipski: Implemented your suggestions, think it looks pretty decent. There are 2 things that aren't super nice though:

1. Headings now always need to include a node view, even if they're not togglable.
2. Because `createStronglyTypedTiptapNode` works on the TipTap API level while `createToggleWrapper` works on the BlockNote API level, we have to convert from a node to a block in the node view and do some slightly weird stuff there.

How much of a concern are those 2?

nperez0111: 1. should be fine, node views are essentially "ejecting" from normal rendering but shouldn't be different (if in pure JS not react)
2. Yea, so the ideal here is to build everything in terms of the blocknote API instead in the future, just don't wanna invest into that at the moment. May be worth extracting into something common, but not necessary for the moment either.

So, I think this is good from here. 🚦 green light to implement the lists & other ones, it seems like a sound approach for the moment.

examples/01-basic/04-default-blocks/App.tsx
```diff
@@ -32,6 +32,11 @@ export default function App() {
         type: "heading",
         content: "Heading",
       },
+      {
+        type: "heading",
+        props: { isTogglable: true },
```
YousefED: let's use isToggleable to be more inline with notion (https://developers.notion.com/reference/block#toggle-blocks)

examples/06-custom-schema/06-togglable-blocks/README.md
```diff
@@ -0,0 +1,8 @@
+# Togglable Blocks
+
+This example shows how to create blocks with a toggle button to show/hide their children, like with the default toggle heading and list item blocks. This is done using the use the `ToggleWrapper` component from `@blocknote/react`.
```
YousefED: let's make explicit (both in the example name and description) that this is about custom blocks

packages/core/src/i18n/locales/en.ts
```diff
@@ -18,12 +18,36 @@ export const en = {
...
+    },
     quote: {
       title: "Quote",
       subtext: "Quote or excerpt",
       aliases: ["quotation", "blockquote", "bq"],
       group: "Basic blocks",
     },
+    toggle_list: {
+      title: "Toggle List",
+      subtext: "List with hidable sub-items",
```
YousefED: hideable

YousefED (changes requested):
Nice! Still need to dive a bit deeper in the code but in general looks great. 

Let's make the UX a bit more polished. When comparing with other products (e.g. Notion), this catches my attention:

- let's make the arrow similar to notion (the rounded caret doesn't work well with the font), and same size regardless of heading size
- Can you pay extra attention to alignment and margins? These are some of the most impactful design concepts. For example, the nested content should be vertically aligned with the parent content (if it helps, remove the vertical border on the left)
- empty toggle list should have a placeholder "Toggle" (same as Notion)
- dragging a collapsed toggle to a different position expands it's state
- Nice to have: dropping blocks in an empty toggle

Others:
- Toggle heading doesn't appear in formatting toolbar dropdown
- We also need to think about an API of where to store "collapsed state" for the current user
- Typing in an empty toggle collapses the arrow (but does show the children placeholder)
- key down in an empty toggle doesn't work. In general, make sure to test keyboard handling

matthewlipski:
Note about enter handling - when hitting enter within a togglable block, Notion does one of 2 things:

- If the child blocks are visible, it splits the parent block and moves the content after the cursor into a new child at the start.
- If the child blocks are visible, it splits the parent block and moves the content after the cursor into a new sibling after.

Since the BlockNote API currently doesn't have a way of splitting blocks in this way (`getSelectionCutBlocks` only works if the selection is not collapsed and even then doesn't give us all the info we need), I'm skipping this for now. Though I can ofc just implement that in the ProseMirror/TipTap API if it's worth the time.

packages/core/src/blocks/HeadingBlockContent/HeadingBlockContent.ts
```diff
@@ -153,6 +156,39 @@ const HeadingBlockContent = createStronglyTypedTiptapNode({
...
+      const contentDOM = renderedElement.contentDOM;
+      const contentDOMParentElement = contentDOM.parentElement!;
+
+      const editor = this.options.editor;
+      const block = getBlockFromPos(getPos, editor, this.editor, this.name);
+
+      const toggleWrapper = createToggleWrapper(block as any, editor, {
+        dom: contentDOM,
+        contentDOM,
+      });
```
nperez0111: Might be missing something here, but this feels wrong? Why would you need to pass it in twice? Can this not just be a default if `dom` is needed? I'd expect toggleWrapper to completely wrap the element, as in, it should be the bare return here: `return toggleWrapper`

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
```diff
@@ -0,0 +1,191 @@
...
+  if (!toggledState) {
+    toggledState = {
+      set: (block, isToggled: boolean) => {
+        window.localStorage.setItem(`toggle-${block.id}`, String(isToggled));
+      },
+      get: (block) => {
+        return window.localStorage.getItem(`toggle-${block.id}`) === "true";
+      },
+    };
+  }
```
nperez0111: This is a constant, it should be defined outside of this scope, since it will be constantly re-created

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
```diff
@@ -0,0 +1,191 @@
...
+  destroy?: () => void;
+} => {
+  if ("isToggleable" in block.props && !block.props.isToggleable) {
+    return renderedElement;
+  }
+
+  if (!toggledState) {
+    toggledState = {
+      set: (block, isToggled: boolean) => {
+        window.localStorage.setItem(`toggle-${block.id}`, String(isToggled));
```
nperez0111: Would prefer explicitness

```suggestion
        window.localStorage.setItem(`toggle-${block.id}`,isToggled ? "true" : "false");
```

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
```diff
@@ -0,0 +1,191 @@
...
+  renderedElement: {
+    dom: HTMLElement;
+    contentDOM?: HTMLElement;
+    ignoreMutation?: (mutation: ViewMutationRecord) => boolean;
+    destroy?: () => void;
+  },
+  toggledState?: {
+    set: (block: Block<any, any, any>, isToggled: boolean) => void;
+    get: (block: Block<any, any, any>) => boolean;
+  },
```
nperez0111: This type should be exported for external use

nperez0111: `toggledState` doesn't feel like a great name, perhaps: `store: ToggleStore` (it is namespaced by `createToggleWrapper` so it will always be clear what store is in reference to).

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
```diff
@@ -0,0 +1,191 @@
...
+        toggleWrapper.setAttribute("data-show-children", "false");
+        toggledState.set(editor.getBlock(block)!, false);
+      }
+
+      // Remove the "add block" button as we want to hide child blocks,
+      // regardless of whether there are child blocks or not.
+      if (dom.contains(toggleAddBlockButton)) {
+        dom.removeChild(toggleAddBlockButton);
+      }
+    }
```
nperez0111, line 131: Hm, not a fan of contents of the editor subscribing to the editor state.

Can this not be achieved with the `update` API: https://prosemirror.net/docs/ref/#view.NodeView.update

Maybe not, since it seems to not give contentDom changes?

packages/core/src/editor/Block.css
```diff
@@ -174,9 +174,10 @@ NESTED BLOCKS
 }
 
 /* Ordered */
-.bn-block-content[data-content-type="numberedListItem"] {
+.bn-block-content[data-content-type="numberedListItem"]::before {
   display: flex;
-  gap: 0.5em;
+  justify-content: center;
+  width: 25px;
```
nperez0111: I don't think this (or the others) should be a fixed width, since the contents are actually variable

image

packages/core/src/extensions/SuggestionMenu/getDefaultSlashMenuItems.ts
```diff
@@ -122,6 +122,37 @@ export function getDefaultSlashMenuItems<
...
+        onItemClick: () => {
+          insertOrUpdateBlock(editor, {
+            type: "heading",
+            props: { level: 1, isToggleable: true },
+          });
+        },
+        key: "toggle_heading",
+        ...editor.dictionary.slash_menu.toggle_heading,
+      },
+      {
```
nperez0111, line 135: Is there a way to disable toggleable headings but not headings?

packages/react/src/blocks/ToggleWrapper/ToggleWrapper.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const { block, editor, children } = props;
+
+  const toggledState = useMemo(
+    () =>
+      props.toggledState || {
+        set: (block, isToggled: boolean) =>
+          window.localStorage.setItem(`toggle-${block.id}`, String(isToggled)),
+        get: (block) =>
+          window.localStorage.getItem(`toggle-${block.id}`) === "true",
+      },
```
nperez0111: Is constant, should be extracted to module scope
can remove memo too, since it will always be broken or valid

packages/react/src/blocks/ToggleWrapper/ToggleWrapper.tsx
```diff
@@ -0,0 +1,97 @@
...
+    [props.toggledState],
+  );
+
+  const [showChildren, setShowChildren] = useState(toggledState.get(block));
+  const [childCount, setChildCount] = useState(block.children.length);
+
+  useEffect(() => {
+    toggledState.set(editor.getBlock(block)!, showChildren);
+  }, [block, block.id, editor, showChildren, toggledState]);
```
nperez0111: This will be over-called, generally this pattern is problematic for performance.

Store the state when the state is transitioned, not unconditionally

packages/react/src/blocks/ToggleWrapper/ToggleWrapper.tsx
```diff
@@ -0,0 +1,97 @@
...
+    } else if (newChildCount === 0 && newChildCount < childCount) {
+      // If the last child block is removed while children are shown, hide
+      // children.
+      if (showChildren) {
+        setShowChildren(false);
+      }
+    }
+
+    setChildCount(newChildCount);
+  });
```
nperez0111: early exit if the thing being wrapped is not even toggleable, otherwise, just overhead.

packages/react/src/blocks/ToggleWrapper/ToggleWrapper.tsx
nperez0111: A nice pattern for this is `useReducer`

nperez0111 (changes requested):
Does isToggleable have to show up on every element now? Can it not just be assumed false if undefined? Or, do we always render "default values"?

packages/core/src/extensions/SuggestionMenu/getDefaultSlashMenuItems.ts
matthewlipski: I think so, you could just do `getDefaultSlashMenuItems(...).filter((item) => !item.key.startsWith("toggle_heading"))` right? So pretty much the same way you'd filter out any other items from the menu. Or do you mean in the schema itself?

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
matthewlipski: Yeah unfortunately `update` doesn't work. Even if it did trigger on `contentDOM` changes, that wouldn't help since when the nested blocks changed, it's the `contentDOM` of the sibling `blockGroup` that we actually need to listen for changes in.

nperez0111: Fair enough!

packages/core/src/extensions/SuggestionMenu/getDefaultSlashMenuItems.ts
nperez0111: I meant from the schema

matthewlipski: Ah, in that case not really but I think it's ok for now, seems like something that will get fixed either way when converting default blocks to use the custom blocks API.

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
```diff
@@ -0,0 +1,182 @@
...
+import { Block } from "../defaultBlocks.js";
+
+type ToggledState = {
+  set: (block: Block<any, any, any>, isToggled: boolean) => void;
+  get: (block: Block<any, any, any>) => boolean;
+};
+
+export const defaultToggledState: ToggledState = {
+  set: (block, isToggled: boolean) =>
+    window.localStorage.setItem(
```
YousefED, line 13: should we move this logic (whether to store to local storage) to be configurable on the editor? Or want to keep it as part of the block spec (and thus the schema)?

YousefED: cc @nperez0111

packages/core/src/blocks/ListItemBlockContent/ToggleListItemBlockContent/ToggleListItemBlockContent.ts
```diff
@@ -0,0 +1,104 @@
...
+import { createDefaultBlockDOMOutputSpec } from "../../defaultBlockHelpers.js";
+import { defaultProps } from "../../defaultProps.js";
+import { createToggleWrapper } from "../../ToggleWrapper/createToggleWrapper.js";
+import { handleEnter } from "../ListItemKeyboardShortcuts.js";
+
+export const toggleListItemPropSchema = {
+  ...defaultProps,
+} satisfies PropSchema;
+
+const ToggleListItemBlockContent = createStronglyTypedTiptapNode({
```
YousefED, line 18: Food for thought; when streamlining built-in blocks, I'm curious if we can come up with a light-weight way to make this more readable (React / jsx like), without pulling in all of react. And thus, get rid of the duplicate (vanilla + react) implementation

cc @nperez0111

packages/core/src/editor/Block.css
```diff
@@ -204,9 +206,11 @@ NESTED BLOCKS
 }
 
 /* Unordered */
-.bn-block-content[data-content-type="bulletListItem"] {
+.bn-block-content[data-content-type="bulletListItem"]::before {
```
YousefED, line 209: can you explain why this changed? why the before element that wasn't here earlier?

YousefED (commented):
Awesome, almost there I think!

> empty toggle list should have a placeholder "Toggle" (same as Notion)

I don't think this works?

Others:

- [ ] let's make the font color of the "add button" placeholder more placeholder-like (gray?)
- [ ] Is it easy to remove the vertical lines for toggled blocks? imo this still looks a bit weird

One architectural question:

I realize we're currently following notion in this design, but let's make sure having an is_toggleable prop on a heading makes more sense than having a separate ToggleHeading block. What are the pros / cons? (cc @nperez0111 )

packages/core/src/blocks/ListItemBlockContent/ToggleListItemBlockContent/ToggleListItemBlockContent.ts
nperez0111: Yes, I have ideas for this. Basically use jsx for actual document.createElement calls, without all of React. In my ideal, you wouldn't even need React.

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
nperez0111: imo, for now, part of the block spec (lowest-level possible).

In the future, this would be a "bundle" that you can configure independently

nperez0111:
> I realize we're currently following notion in this design, but let's make sure having an is_toggleable prop on a heading makes more sense than having a separate ToggleHeading block. What are the pros / cons? (cc @nperez0111 )

My argument for this was that let's say you were generating the table of contents for a page, would you consider a heading that happens to be toggle-able, not a heading? By making it a different block, you change the semantics of what it is, whereas toggling is purely a visual thing (much like setting a background color wouldn't change the type of block it is)

YousefED:
> > I realize we're currently following notion in this design, but let's make sure having an is_toggleable prop on a heading makes more sense than having a separate ToggleHeading block. What are the pros / cons? (cc @nperez0111 )
> 
> My argument for this was that let's say you were generating the table of contents for a page, would you consider a heading that happens to be toggle-able, not a heading? By making it a different block, you change the semantics of what it is, whereas toggling is purely a visual thing (much like setting a background color wouldn't change the type of block it is)

Good point. Going to challenge a bit more to make sure we're confident (I still don't think I have a preference, or maybe start to prefer the current implementation):

- Following your reasoning, does it still make sense the toggle list items are different from regular lists?
- Having separate blocks would solve things like this, right? https://github.com/TypeCellOS/BlockNote/pull/1707#discussion_r2150279943

nperez0111:
> Good point. Going to challenge a bit more to make sure we're confident (I still don't think I have a preference, or maybe start to prefer the current implementation):
> 
>     * Following your reasoning, does it still make sense the toggle list items are different from regular lists?

Good point, I'd argue that these should actually be named  `toggleBlock`s rather than a list item, and they roughly correspond to the `<details>` element in HTML: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details.

UI-wise, the naming could be different for understanding maybe, but, I think, they are functionally equivalent. List does sort of convey that hitting enter, you'd get more of them.

List items are actually the perfect demonstration of what I mean here about semantics, it is always an `li` but what it is wrapped (e.g. `<ul>`) in that determines it's display.

>     * Having separate blocks would solve things like this, right? [feat: Toggle blocks #1707 (comment)](https://github.com/TypeCellOS/BlockNote/pull/1707#discussion_r2150279943)

I think what would solve this is the ability to easily customize the schema. In this case, I'd want an option to remove `isToggable` from the prop schema, and if that prop schema is completely missing, we could hide the slash menu items as not being applicable.

packages/core/src/editor/Block.css
matthewlipski: Ah I forgot to mention this in the PR description - I adjusted offsets for list items, indented blocks, and the "add block" button of the toggle blocks to make them all the same. So all of these are now vertically in-line now, and these styles are a part of that.

matthewlipski:
To follow up with the naming convention for `toggleListItem` - worth mentioning that this is how Notion represents them too.

Also r.e. the `<details>` element, I agree this would be nice to use though I'm not 100% sure it fits within our HTML structure. Basically the hidden content, i.e. the `blockGroup`, should be inside it below the `<summary>` element, which we can't really do since it's outside of the `blockContent` element.

packages/core/src/i18n/locales/nl.ts
```diff
@@ -157,7 +157,7 @@ export const nl: Dictionary = {
   placeholders: {
     default: "Voer tekst in of type '/' voor commando's",
     heading: "Kop",
-    toggleList: "Schakelaar",
+    toggleListItem: "Schakelaar",
```
YousefED: This word means "switch" (as in, "light switch" / "light toggle").

Can we give your AI some extra context to improve the translations?

YousefED:
> List does sort of convey that hitting enter, you'd get more of them.

@nperez0111  that's also the case.

I'm starting to think that the current naming is ok - unless anyone feels otherwise after the above discussion

nperez0111:
I think the current naming is okay if it keeps the newline behavior. If someone wants to disable that, then it should be called a toggle block.

But, obviously no way for us to enforce it.

Agreed, naming is fine here.

packages/core/src/i18n/locales/nl.ts
matthewlipski: Hmm, I'm just using Copilot built into VSCode so it should at least have the context of the english translation file, any way I can add to this?

YousefED: I think you can @ mention specific files in the chat? It also helps to just chat explanation of what the toggle is (i.e.: a block that expands in a document editor)

nperez0111:
Thought about it, & decided it was fine like this:

 - If you want there to be both toggle & not toggable headings, that is the default
 - If you want only not toggable headings, we could disable the view layer from wrapping the heading with the `creatToggleWrapper`
 - If you want only toggable headings, you would similarly just enforce that the view layer to always be wrapped in a toggleable

MetaMmodern:
@nperez0111 how do I turn off togglable headings only on schema level? I see then in suggestions when highlighting text. I couldn't find anything in official documention on website about `createToggleWrapper`. That's quite low-level, no?

nperez0111:
@MetaMmodern See https://github.com/TypeCellOS/BlockNote/issues/1816 there isn't a great way to do this atm. I'd recommend just removing it from the slash menu. In the future we will make the schema configurable

PR #1716, opened 2025-05-23
fix: Selection clicking between blocks

matthewlipski:


nperez0111 (approved):
👍  this looks about right to me. Good work on this, I know it can be annoying

nperez0111:
Oof good catch @YousefED, yea given that the line-height is the majority of this fix, it looks like we might be at a stalemate here.

PR #1717, opened 2025-05-23
fix: Selection clicking editor padding

matthewlipski:
# Problem

In Chrome, clicking the editor padding on the right side of a list item block moves the selection to the block's start instead of end. This is because the list item indicators are displayed using a `:before` pseudo-element, which Chrome seems to prioritize over the block content element, despite it being closer to the cursor.

[This ProseMirror discussion](https://discuss.prosemirror.net/t/mouse-down-selection-behaviour-different-on-chrome/8426) has some more information on the issue.

# Solution

This PR makes the `:before` elements for list item indicators have a height of 0, which fixes the issue in Chrome. The elements are positioned exactly as they were before as the element's overflow is visible, and some additional rules have been added to fix the alignment that was broken from the height change.

Before:

https://github.com/user-attachments/assets/36e56bfc-4206-4f6e-ae50-890de074d580

After:

https://github.com/user-attachments/assets/dcb5a1d2-c9ba-4993-a023-9fd90ceb1839

# Caveats

Feels slightly hacky, but don't think it should cause any issues.

# Alternatives tried

Using `display: inline` instead of flexbox: Added a space to the server rendered editor that couldn't be removed without hacks. Also caused an issue when the block was empty on Firefox. See [this comment]() for both of these.

Custom selection handling logic: Was overly complex and somewhat hacky.


packages/core/src/editor/Block.css
```diff
@@ -14,23 +14,32 @@ BASIC STYLES
...
+  /* Workaround for selection issue on Chrome, see #1588 and also here:
+  https://discuss.prosemirror.net/t/mouse-down-selection-behaviour-different-on-chrome/8426
+  The :before element causes the selection to be set in the wrong place vs 
+  other browsers. Setting no height fixes this, while list item indicators are 
+  still displayed fine as overflow is not hidden. */
+  height: 0;
+  /* Also have to center the text vertically to ensure alignment at different 
+  line heights (which we change for `.bn-inline-content` below). */
+  display: flex;
+  align-items: center;
```
nperez0111, line 40: Let's also explicitly set the overflow to show. There are many CSS resets out there so if you rely on a default behavior it may not work in some projects.

nperez0111 (approved):

PR #1724, opened 2025-05-29
refactor: zod poc

### LINKED ISSUE #444: Props complexity for custom blocks
I currently try to benchmark BlockNote for my use cases, including complex custom blocks, currently, the type system is only allowing props to be `string`, `number` or `boolean`, and for example, my use case is a block that have an array of object containing a title and some text, each element of the array is a possible display of the block, this doesn't look possible (or with many `@ts-ignore`?) maybe I'm missing something or do something wrong.

Being able to have complex props types

Encode my objects/array to JSON, but it's messy and really not great

**Additional context**

YousefED:
This is a POC to use Zod instead of our custom schema language for block and inline content props. Let's discuss if this is a direction we want to take. If so, we should:

- [ ] add tests for more complex schemas than we currently support
- [ ] use the same pattern for styles?
- [ ] fix the other packages / examples. Atm at least all tests in `core` and `tests` pass!
- [ ] validate input / output types are still the same. E.g.: can a heading only take the configured levels. And does the output type of a document with a heading always have the level prop?
- [ ] (optional) create a backwards compatible API for custom blocks (we could translate the old schema passed to `createReactBlockSpec` to a Zod schema
- [ ] check compatibility with yjs

thanks @aeplay for mentioning this in your talk 😂

closes https://github.com/TypeCellOS/BlockNote/issues/444


packages/core/src/extensions/BackgroundColor/BackgroundColorExtension.ts
```diff
@@ -10,15 +10,15 @@ export const BackgroundColorExtension = Extension.create({
         types: ["blockContainer", "tableCell", "tableHeader"],
         attributes: {
           backgroundColor: {
-            default: defaultProps.backgroundColor.default,
+            default: defaultProps.shape.backgroundColor._zod.def.defaultValue,
```
nperez0111, line 13: I don't really like reaching into what feels like an internal value for this, is there another way of doing it that is part of the public API?

PR #1729, opened 2025-05-30
fix: Restrict auto list conversion to paragraphs

### LINKED ISSUE #1710: Typing 1. or + and hitting space changes block type.
I'm having an issue where if I write 1. or "+" inside of a custom block and then hit space, it will change the block type to a list. This behavior is present in the [alert block with full dx example](https://www.blocknotejs.org/examples/custom-schema/alert-block-full-ux).

I'm not sure if there is a way to create an API for this other than listening on tiptap events manually. Looking for guidance.

I tried to check prosemirror and tiptap docs but couldn't find anything helpful. I also read through existing issues and couldn't find anything similar.

**Bonus**

Mulgyeol:
## Description
When users type list markers ("1." or "-/+/*") at the beginning of a block, BlockNote automatically converts it into a list.
Previously, this conversion would trigger in any block with inline content, including headings and custom blocks.
This was disruptive when users wanted to start headings or other blocks with numbers or list markers.

This change restricts the auto-conversion to list blocks to paragraph blocks only,
preserving these inputs in other block types as plain text.

Fixes #1710

## Changes
- Modified the input rule condition in NumberedListItemBlockContent.ts to check for paragraph blocks specifically
- Modified the input rule condition in BulletListItemBlockContent.ts to check for paragraph blocks specifically

## Testing
The following scenarios should be tested:

1. In a heading block:
   - Type "1." at the start
   - Expected: Text remains as "1." without converting to a numbered list
   - Type "-" or "+" at the start
   - Expected: Text remains as is without converting to a bullet list

2. In a paragraph block:
   - Type "1." at the start
   - Expected: Block converts to a numbered list item
   - Type "-" or "+" at the start
   - Expected: Block converts to a bullet list item

3. In a custom block (like alert block):
   - Type "1." or list markers (-/+/*)
   - Expected: Text remains as is without converting to any type of list

### DIFF
--- a/packages/core/src/blocks/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
+++ b/packages/core/src/blocks/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
@@ -29,7 +29,12 @@ const BulletListItemBlockContent = createStronglyTypedTiptapNode({
         find: new RegExp(`^[-+*]\\s$`),
         handler: ({ state, chain, range }) => {
           const blockInfo = getBlockInfoFromSelection(state);
-          if (blockInfo.blockNoteType !== "paragraph") {
+          if (
+            !blockInfo.isBlockContainer ||
+            blockInfo.blockContent.node.type.spec.content !== "inline*" ||
+            blockInfo.blockNoteType === "bulletListItem" ||
+            blockInfo.blockNoteType === "heading"
+          ) {
             return;
           }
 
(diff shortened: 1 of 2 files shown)


Mulgyeol:
@nperez0111 

I see your point — restricting auto-conversion in non-paragraph blocks might cause inconvenience for users who expect that feature.
So, I reverted the previous approach and now restrict auto list conversion only in heading blocks, allowing it for other block types for more flexibility.

I figured that heading blocks are explicitly intended for titles, and users starting with "1." or "-" in those contexts are typically not trying to create a listso this felt like a safe and targeted change to apply directly.

PR #1736, opened 2025-06-03
fix: Typing improvements

matthewlipski:
This PR adds a few minor improvements to typing:

- Type guards have been added to check if an arbitrary block is in an editor's schema.
- Some utility types for inline content have been added, that already exist for blocks.
- The `render` function for custom react inline content now provides the editor in its props.


nperez0111 (approved):
Yep, this is better, thanks for extracting it

PR #1741, opened 2025-06-09
fix: AI system messages should always be at start of prompt

YousefED:
Anthropic / Gemini would give an error "Multiple system messages that are separated by user/assistant messages' functionality not supported" when alternating between system and user messages


packages/xl-ai/src/api/LLMRequest.ts
```diff
@@ -163,7 +163,18 @@ export async function doLLMRequest(
...
+    previousMessages = previousResponse.messages.map((m) => {
+      // Some models, like Gemini and Anthropic don't support mixing system and user messages.
+      // Therefore, we convert all user messages to system messages.
+      // (also see comment below on a possibly better approach that might also address this)
+      if (m.role === "user" && typeof m.content === "string") {
+        return {
+          role: "system",
+          content: m.content,
+        };
+      }
```
nperez0111: Hm, this feels off to me, this seems more likely to be "jail-broken" because the llm treats system messages as higher priority no?

Shouldn't future system messages just be filtered out? like:

```
>system: do the thing
>user: I want XYZ
>llm response: do XYZ thing
>user: No, more like ABC
>llm response: do ABC thing
```

YousefED: That would be nice, but the issue is we want to add system messages to explain the "iteration" to the LLM ([here](https://github.com/TypeCellOS/BlockNote/blob/fecdf1b0a7ccf2f2304cc9e442ea910d0a21e53b/packages/xl-ai/src/api/formats/html-blocks/defaultHTMLPromptBuilder.ts#L104). These messages imo should be at least as "strong" or "stronger" than the previous iteration.

Good point re "jailbreaking", so it definitely is not ideal. However, afaik LLM prompts can always be jailbroken even when using user messages.

Do you have any ideas for a better solution? (I don't atm, except maybe rewriting it entirely using tool responses as mentioned in the comments)

nperez0111: I see, hm, then I think you are right that this is likely better, maybe what we can do is prefix the content with something like:

```suggestion
      if (m.role === "user" && typeof m.content === "string") {
        return {
          role: "system",
          content: `USER_MESSAGE: ${m.content}`,
        };
      }
```

YousefED: good idea!

PR #1742, opened 2025-06-09
fix(website): log in bug fixes

nperez0111:
- **fix: do not send email on use of a magic link**
- **fix: on subscription.active, check subscription status**

### DIFF
--- a/docs/auth.ts
+++ b/docs/auth.ts
@@ -223,25 +223,34 @@ export const auth = betterAuth({
         // webhooks have to be publicly accessible
         // ngrok http http://localhost:3000
         secret: process.env.POLAR_WEBHOOK_SECRET as string,
-        onSubscriptionUpdated: async (payload) => {
-          const authContext = await auth.$context;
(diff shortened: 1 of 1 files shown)


docs/auth.ts
```diff
@@ -254,11 +263,7 @@ export const auth = betterAuth({
   },
   hooks: {
     after: createAuthMiddleware(async (ctx) => {
-      if (
-        ctx.path === "/magic-link/verify" ||
-        ctx.path === "/verify-email" ||
-        ctx.path === "/sign-in/social"
-      ) {
+      if (ctx.path === "/verify-email" || ctx.path === "/sign-in/social") {
```
YousefED, line 266: @nperez0111 wouldn't this still break for new sign ins with social login?

I also saw https://github.com/better-auth/better-auth/issues/1512

nperez0111: Yep, you are right. Couldn't get the database hooks working so will just use the createdAt field

PR #1756, opened 2025-06-13
feat: API design feedback

nperez0111:
This is to open some decisions for discussion around some core parts of the blocknote API.

There are a few proposals in here that can be seen in the ADR directory as markdown files. With some Typescript files to support how they might actually look like in an implementation.

### DIFF
--- a/adr/2025_06_13-document-transforms.md
+++ b/adr/2025_06_13-document-transforms.md
@@ -1,0 +1,20 @@
+# Document Transforms
+
+A core part of the BlockNote API is the ability to make changes to the document, using BlockNote's core design of blocks, this is much easier to do programmatically than with other editors.
+
+We've done pretty well with our existing API, but there are a few things that could be improved.
+
+Referencing content within the document is either awkward or non-existent. Right now we essentially really only have an API for referencing blocks by their id with no further level of granularity.
+
+## Locations
+
+[Looking at Slate](https://docs.slatejs.org/concepts/03-locations) (highly recommend reading the docs), they have the concept of a `Location` which is a way to reference a specific point in the document, but it does not have to be so specific as positions, it has levels of granularity.
+
+This gives us a unified way to reference content within the document, allowing for much more granular editing. Take a look at the `Location.ts` file for more details around this.
+
+## Transforms separation of concerns
+
+Right now all transformation functions are defined directly on the `Editor` instance, this is not ideal because it only further muddles the API.
+
+Instead, we should have a separate `Transform` class which defines methods that operate on the editor's document to make changes to it. This will also be very useful for doing server-side transformations.
+
--- a/adr/2025_06_13-document-transforms.md
+++ b/adr/2025_06_13-document-transforms.md
@@ -18,3 +18,48 @@ Right now all transformation functions are defined directly on the `Editor` inst
 
 Instead, we should have a separate `Transform` class which defines methods that operate on the editor's document to make changes to it. This will also be very useful for doing server-side transformations.
 
+```ts
+editor.transform.insertBlocks(ctx:{
+  at: Location,
+  blocks: Block | Block[]
+});
+
+editor.transform.replaceBlocks(ctx: {
+  at: Location,
+  with: Block | Block[]
+})
+
+editor.transform.replaceContent(ctx: {
+  at: Location,
+  with: InlineContent | InlineContent[]
+})
+
+editor.transform.deleteContent(ctx: {
+  at: Location,
+})
+```
+
+## References
+
+Currently, we do not have a good way to reference things about content within a document, except by block id. With Locations, we can be much more granular & more powerful.
+
+I think one application of this would be to introduce the concept of "references" to content within a document. For example, we currently do not store comments into the blocknote json, taking the position that it really is not part of the document, but rather metadata about the document.
+
+With references, we could store comments with references to the blocks that they are about. And map those references through changes if they are ever modified. For example, if someone commented on the text within block id `123`, and then the block was moved to a new location, we could update the comment to reference the new block id.
+
+So, this would allow a comment to still be a separate entity, but be able to "hydrate" within the editor and keep track of the content that it was about.
+
+```ts
+type Reference<Metadata extends Record<string, any>> = {
+  to: Location,
+  metadata: Metadata
+}
+
+type Comment = Reference<{
+  threadId: string
+}>
+
+editor.references.add(reference: Reference, onUpdate: (reference: Reference) => void);
+```
+
+This is inspired by [bluesky's concept of facets for rich text content](https://docs.bsky.app/docs/advanced-guides/post-richtext) which is a great example of how to make block note content more inter-operable between different applications.
--- a/adr/2025_06_13-extensions.md
+++ b/adr/2025_06_13-extensions.md
@@ -39,3 +39,11 @@ Zustand is a popular library, but allowing partial states makes it somewhat unsa
 
 Jotai is probably my second choice, but it makes it a bit awkward to update states because it relies on a separate store instance rather than the "atom" being able to update itself <https://jotai.org/docs/guides/using-store-outside-react>.
 </details>
+
+## Exposing extension methods
+
+Not everything can be communicated through just state, so we need to be able to expose additional methods to the application.
+
+I propose that we have a `extensions` property on the `BlockNoteEditor` instance, which is a map of the extension key to the extension instance, but filtered out to only include non-blocknote methods (as defined by the `ExtensionMethods` type).
+
+This will allow the application to access the extension methods, and also allows us to keep the extension instance private to the editor (type-wise).
--- a/adr/2025_06_13-extensions.md
+++ b/adr/2025_06_13-extensions.md
@@ -1,0 +1,41 @@
+# BlockNote Extension API
+
+It is definitely more designed by accident than intention, so let's put some thought into the sort of API we would want people to build extensions with.
+
+## Core Requirements
+
+What I identified was:
+
+- The ability to "eject" to the prosemirror API, if we don't provide something good enough. This is a core choice, and one I don't think we would ever need to walk back on. Fundamentally, we do not want to expose absolutely everything that Prosemirror can do, but also do not want to stop those with the know-how to actually get stuff done.
+- A unified store API, to make consuming extension state homogenous, this will be discussed further below
+- Life-cycle event handlers, by providing hooks for `create`, `mount`, and `unmount` we allow extensions to attach event handlers to DOM elements, and have a better general understanding of the current state of the editor. The `onCreate` handler will be very handy to guarantee access to the editor instance before anything is called.
+- Editing event handlers, by providing hooks for `change`, `selectionChange`, `beforeChange`, and `transaction` we give extensions access to the fundamental primitives of the editor.
+
+## State Management
+
+What I had the most trepidation about was deciding on whether we should prescribe a state management library, and if so, which one.
+
+I think the answer is _yes_, we should prescribe a state management library, and I think the answer is @tanstack/store.
+
+<details>
+<summary>Why @tanstack/store? And not something else?</summary>
+It comes with a few benefits:
+
+- It gives a single API for all state management, which is more convenient/consistent for consumers
+
+As for which library to use, I think we should use @tanstack/store.
+
+- The store is very simple, and can easily be re-implemented if needed
+- There is already bindings for most major frameworks, but can be used without them
+- It seems that anything tanstack does will be widely adopted so it should be a pretty safe bet
+
+What I had trouble with is there are a few different use-cases for state management, events like we have now aren't great because they put the burden on the consumer to manage the state. Or, they can emit an event (e.g. `update`), but then have to round-trip back to the extension to get the state (and somehow store it on their side again).
+
+Something like observables have a nicer API for this, but they are for pushing data (i.e. multiple readers), not for pulling it (i.e. any writers). They also have the same problem of putting the burden on the consumer.
+
+Signals are a nice middle ground, being that they are for both pushing & pulling data. The problem is that there are many implementations, and not super well-known in the React ecosystem.
+
+Zustand is a popular library, but allowing partial states makes it somewhat unsafe in TypeScript.
+
+Jotai is probably my second choice, but it makes it a bit awkward to update states because it relies on a separate store instance rather than the "atom" being able to update itself <https://jotai.org/docs/guides/using-store-outside-react>.
+</details>
--- a/adr/2025_06_13-extensions.md
+++ b/adr/2025_06_13-extensions.md
@@ -8,8 +8,7 @@ What I identified was:
 
 - The ability to "eject" to the prosemirror API, if we don't provide something good enough. This is a core choice, and one I don't think we would ever need to walk back on. Fundamentally, we do not want to expose absolutely everything that Prosemirror can do, but also do not want to stop those with the know-how to actually get stuff done.
 - A unified store API, to make consuming extension state homogenous, this will be discussed further below
-- Life-cycle event handlers, by providing hooks for `create`, `mount`, and `unmount` we allow extensions to attach event handlers to DOM elements, and have a better general understanding of the current state of the editor. The `onCreate` handler will be very handy to guarantee access to the editor instance before anything is called.
-- Editing event handlers, by providing hooks for `change`, `selectionChange`, `beforeChange`, and `transaction` we give extensions access to the fundamental primitives of the editor.
+- Life-cycle event handlers, by providing hooks for `mount` and `unmount` we allow extensions to attach event handlers to DOM elements, and have a better general understanding of the current state of the editor.
 
 ## State Management
 
@@ -47,3 +46,15 @@ Not everything can be communicated through just state, so we need to be able to 
 I propose that we have a `extensions` property on the `BlockNoteEditor` instance, which is a map of the extension key to the extension instance, but filtered out to only include non-blocknote methods (as defined by the `ExtensionMethods` type).
 
 This will allow the application to access the extension methods, and also allows us to keep the extension instance private to the editor (type-wise).
+
+# BlockNote Bundles
+
+Somewhat related, to extensions, we also need a way for bundling sets of editor elements in a single packaging. This will be a much higher-level API, which will aim to provide a single import for adding an entire set of functionality to the editor (e.g. adding multi-column support, or comments, etc.)
+
+## Core Requirements
+
+- A way to add blocks, inline content, and styles to the editor
+- A way to add extensions to the editor
+- A way to add to the dictionary of locales to the editor
+- A way to add to the slash menu
+- A way to add to the formatting toolbar
--- a/adr/2025_06_13-packages.md
+++ b/adr/2025_06_13-packages.md
@@ -1,0 +1,30 @@
+# BlockNote Packages & Sub-Packages
+
+## Core
+
+The `@blocknote/core` package should contain everything needed to build a pure JS editor instance. It has no opinion on the UI layer, and is only concerned with the core editor functionality.
+
+It contains sub-packages for:
+
+- `@blocknote/core/blocks` - Contains the default blocks, inline content, styles and their implementations (excluding things like blockContainer, blockGroup, etc which are core to the editor but should not be part of the schema).
+  - `@blocknote/core/blocks/header` - Contains the header block.
+  - `@blocknote/core/blocks/code-block` - Contains the code block block.
+  - etc...
+- `@blocknote/core/extensions` - Re-exports the extension packages below.
+  - `@blocknote/core/extensions/comments` - Contains the comments extension.
+- `@blocknote/core/exporter` - Contains the exporter package.
+  - `@blocknote/core/exporter/html` - Contains the html exporter package.
+  - `@blocknote/core/exporter/markdown` - Contains the markdown exporter package.
+- `@blocknote/core/server-util` - Contains the server util package.
+- `@blocknote/core/locales` - Contains the locales package.
+
+## React
+
+The `@blocknote/react` package should contain everything needed to build a React editor instance. It has no opinion on the core editor functionality, and is only concerned with the React UI layer.
+
+It contains sub-packages for:
+
+- `@blocknote/react/editor` - Contains the editor package.
+  - `@blocknote/react/comments` - Contains the comments package.
+  - `@blocknote/react/table-handles` - Contains the table handles package.
+  - etc... (need to figure out what needs to be exported vs. available and how coupled this all is)
--- a/adr/2025_06_13-packages.md
+++ b/adr/2025_06_13-packages.md
@@ -28,3 +28,9 @@ It contains sub-packages for:
   - `@blocknote/react/comments` - Contains the comments package.
   - `@blocknote/react/table-handles` - Contains the table handles package.
   - etc... (need to figure out what needs to be exported vs. available and how coupled this all is)
+
+## Editor instance
+
+The editor instance has been growing in complexity, and handles too many different concerns. To focus it, we should split it up into something like: the `BlockNoteEditor.ts` file.
+
+The goal would be to group the functionality of the editor instance in such a way that it is easier to navigate and understand.
--- a/adr/2025_06_13-schema.md
+++ b/adr/2025_06_13-schema.md
@@ -1,0 +1,21 @@
+# BlockNote Schema
+
+Right now it is overly burdensome to have to pass around 3 different types to the editor, and it is also not very type-safe (when you just end up with `any` everywhere).
+
+The idea is to have a single type that is a union of the 3 types, and then make type predicates available to check if accessed properties are valid (and likely just assertions too).
+
+You'll see some of what I came up with in the `Schema.ts` file.
+
+You'll also notice that the default blocks, inline content, styles, and groups are all defined in the `@blocknote/core/blocks` package. This is assuming that we have already moved them to the blocknote API and out of the core package.
+
+## Groups
+
+In a somewhat similar vein, I think there might be use for having an indirection layer for referring to specific blocks, inline content, styles, etc. Reason being that it allows callers to refer to a group of things with a single identifier, and allow customizing that membership by just modifying that group.
+
+Examples include:
+
+- Keyboard shortcuts can refer to a group of blocks/inline-content/styles without modification to the handler
+- Relationships between blocks/inline-content/styles can be defined (e.g. allow for a todo block to only have todo item children)
+- Properties of blocks/inline-content/styles can be defined (e.g. adding `heading` to the `toggleable` group)
+
+This may or may not be useful, but it is a thought.
--- a/adr/2025_06_16-configuration.md
+++ b/adr/2025_06_16-configuration.md
@@ -1,0 +1,28 @@
+# BlockNote Configuration
+
+Editors are widely different across applications, often with users having opposing requirements. Most editors explode with complexity when trying to support all of these use cases. So, there needs to be a guide for the different ways to configure so that there is not just a single overwhelming list of options.
+
+Fundamentally, there are few different kinds of things to be configured:
+
+- **block-configuration**: The configuration for a specific kind of block
+  - e.g. the `table` block might have toggles for enabling/disabling the header rows/columns
+- **schema-configuration**: The configuration of what blocks, inline content, and styles are available in the editor
+  - e.g. whether to include the `table` block at all
+- **extension-level-configuration**: The configuration of the extension itself
+  - e.g. what ydoc should the collaboration extension use
+- **extension-configuration**: The configuration of the extensions that are available in the editor
+  - e.g. whether to add collaboration to the editor
+- **editor-view-configuration**: The configuration of the editor views
+  - e.g. whether to show the sidebar, the toolbar, etc.
+- **editor-configuration**: The configuration of the editor itself
+  - e.g. how to handle paste events
+
+These forms of configuration are not mutually exclusive, and can be combined in different ways. For example, knowing that the editor has collaboration enabled, might change the what the keybindings do for undo/redo.
+
+In an ideal world, these configurations would be made at the "lowest possible level", like configuring the number of levels of headings would be configured when composing the schema for the editor, rather than at the editor level.
+
+Configuration should be publicly accessible, so that mixed combinations can be created (i.e. different behaviors for an editor with or without collaboration).
+
+## TODO
+
+- Describe how you configure at the block level, then compose that into a schema, the compose that into an editor, and then compose that into a view.
--- a/adr/2025_06_16-nested-blocks.md
+++ b/adr/2025_06_16-nested-blocks.md
@@ -1,0 +1,155 @@
+# BlockNote Nested Blocks
+
+There are two separate problems when it comes to "nested blocks" support in BlockNote:
+
+- **nested-blocks** The first is the ability for a block to contain other blocks within it (e.g. a table cell having not just inline content, but actual other blocks inside it)
+- **content-fields** The second is the ability for a block to contain multiple pieces of content within it (e.g. an alert block having a title & description field (which contain inline content))
+
+Let's start with the first problem, nested blocks. By describing existing block relationships:
+
+## Block with inline content
+
+This is the simplest case, and the only one that is currently supported by BlockNote.
+
+```json
+{
+  "type": "block-type",
+  "content": [
+    {
+      "type": "text",
+      "text": "Hello, world!"
+    }
+  ],
+}
+```
+
+There is a 1:1 relationship between the block type and it's content. And, no restrictions of the inline content allowed within the block.
+
+> TODO come up with a description of the sorts of keybinds & cursor behavior that should be supported for this block type
+
+## Block with nested blocks
+
+This is a proposal for how to support nested blocks.
+
+```json
+{
+  "type": "custom-block-type",
+  "props": {
+    "abc": 123
+  },
+  "content": null,
+  "children": [
+    {
+      "type": "nested-block",
+      "content": [
+        {
+          "type": "block",
+          "content": [
+            {
+              "type": "text",
+              "text": "Hello, world!"
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
+```
+
+This would completely enclose all nested blocks within the `custom-block-type` block. And, works the same way as the multi-column example.
+
+> TODO come up with a description of the sorts of keybinds & cursor behavior that should be supported for this block type
+
+## Block with structured inline content fields
+
+This is a proposal for how to support multiple inline content fields within a block.
+
+```json
+{
+  "type": "alert",
+  "content": null,
+  "children": [
+    {
+      "type": "alert-title",
+      "content": [
+        {
+          "type": "text",
+          "text": "Hello, world!"
+        }
+      ]
+    },
+    {
+      "type": "alert-content",
+      "content": [
+        {
+          "type": "text",
+          "text": "Hello, world!"
+        }
+      ]
+    }
+  ]
+}
+```
+
+The core idea is that the `parent` block restricts what content is allowed within it's children.
+
+> TODO come up with a description of the sorts of keybinds & cursor behavior that should be supported for this block type
+
+## Examples
+
+### Rebuilding tables with nested blocks
+
+Using this new structure, we can rebuild tables if we had this new API:
+
+```json
+{
+  "type": "table",
+  "content": null,
+  "props": {
+    "backgroundColor": "default",
+    "textColor": "default",
+    "columnWidths": [100, 100, 100],
+    "headerRows": 1,
+    "headerCols": 1,
+  },
+  "children": [
+    {
+      "type": "table-row",
+      "content": null,
+      "children": [
+        {
+          "type": "table-cell",
+          "content": null,
+          "children": [
+            {
+              "type": "paragraph",
+              "content": [
+                {
+                  "type": "text",
+                  "text": "Hello, world!"
+                }
+              ]
+            }
+          ]
+        },
+        {
+          "type": "table-cell",
+          "content": null,
+          "children": [
+            {
+              "type": "paragraph",
+              "content": [
+                {
+                  "type": "text",
+                  "text": "Hello, world!"
+                }
+              ]
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
+```
--- a/adr/2025_06_17-formats.md
+++ b/adr/2025_06_17-formats.md
@@ -1,0 +1,55 @@
+# BlockNote Formats
+
+Right now, there are several formats supported by BlockNote:
+
+- internal html (serialized html from the editor's schema) for things like clipboard handling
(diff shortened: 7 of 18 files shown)


packages/core/src/editor/Location.ts
```diff
@@ -0,0 +1,148 @@
...
+ * A block identifier is a unique identifier for a block, it is either a string, or can be object with an id property (out of convenience).
+ */
+export type BlockIdentifier = { id: BlockId } | BlockId;
+
+/**
+ * A path is a list of block identifiers, describing a path to a block within a document.
+ * Each level of the path is a child of the previous level.
+ * The entire document can be described by the path [].
+ */
+export type Path = BlockIdentifier[];
```
YousefED: Because a block has a unique id, I'm not sure whether `Path` is useful? For example, in `Point` below, we could just use a blockIdentfier instead of `path`

adr/2025_06_13-document-transforms.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Transforms separation of concerns
+
+Right now all transformation functions are defined directly on the `Editor` instance, this is not ideal because it only further muddles the API.
+
+Instead, we should have a separate `Transform` class which defines methods that operate on the editor's document to make changes to it. This will also be very useful for doing server-side transformations.
+
+```ts
+editor.transform.insertBlocks(ctx:{
+  at: Location,
```
YousefED, line 23: Is it useful to insertblocks at a `Location` vs the current API (before / after / inside an existing block)?

adr/2025_06_13-document-transforms.md
```diff
@@ -0,0 +1,65 @@
...
+  with: InlineContent | InlineContent[]
+})
+
+editor.transform.deleteContent(ctx: {
+  at: Location,
+})
+```
+
+## References
```
YousefED, line 42: Nice, currently things like comments can't be represented in the BlockNote API, and indeed, this would enable us to separate them from the document data definitely feels like the "right approach"

packages/core/src/editor/Location.ts
```diff
@@ -0,0 +1,148 @@
...
+
+/**
```
YousefED, line 11: How would this work for things like table cells?

adr/2025_06_13-document-transforms.md
```diff
@@ -0,0 +1,65 @@
+# Document Transforms
+
+A core part of the BlockNote API is the ability to make changes to the document, using BlockNote's core design of blocks, this is much easier to do programmatically than with other editors.
+
+We've done pretty well with our existing API, but there are a few things that could be improved.
+
+Referencing content within the document is either awkward or non-existent. Right now we essentially really only have an API for referencing blocks by their id with no further level of granularity.
```
YousefED, line 7: Overall, I think the concept of a `Location` is useful to have - but if we start implementing this I think we should have more sight on practical use-cases. i.e. which use-cases /  APIs do we want to "unlock" with this?

adr/2025_06_13-document-transforms.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Locations
+
+[Looking at Slate](https://docs.slatejs.org/concepts/03-locations) (highly recommend reading the docs), they have the concept of a `Location` which is a way to reference a specific point in the document, but it does not have to be so specific as positions, it has levels of granularity.
+
+This gives us a unified way to reference content within the document, allowing for much more granular editing. Take a look at the `Location.ts` file for more details around this.
+
+## Transforms separation of concerns
+
+Right now all transformation functions are defined directly on the `Editor` instance, this is not ideal because it only further muddles the API.
```
YousefED, line 17: Cool. I think maybe this should be a separate work-item ("cleaning up `BlockNoteEditor`"). Doing this hand-in-hand with updating documentation probably gives us a lot of insights into what we think would be a better organization of functions

adr/2025_06_13-extensions.md
```diff
@@ -0,0 +1,49 @@
...
+Something like observables have a nicer API for this, but they are for pushing data (i.e. multiple readers), not for pulling it (i.e. any writers). They also have the same problem of putting the burden on the consumer.
+
+Signals are a nice middle ground, being that they are for both pushing & pulling data. The problem is that there are many implementations, and not super well-known in the React ecosystem.
+
+Zustand is a popular library, but allowing partial states makes it somewhat unsafe in TypeScript.
+
+Jotai is probably my second choice, but it makes it a bit awkward to update states because it relies on a separate store instance rather than the "atom" being able to update itself <https://jotai.org/docs/guides/using-store-outside-react>.
+</details>
+
+## Exposing extension methods
```
YousefED, line 42: We'll need to expose them in a type-safe way imo. What do you think of the pattern with `getAIExtension` atm?

adr/2025_06_13-extensions.md
```diff
@@ -0,0 +1,49 @@
...
+
+As for which library to use, I think we should use @tanstack/store.
+
+- The store is very simple, and can easily be re-implemented if needed
+- There is already bindings for most major frameworks, but can be used without them
+- It seems that anything tanstack does will be widely adopted so it should be a pretty safe bet
+
+What I had trouble with is there are a few different use-cases for state management, events like we have now aren't great because they put the burden on the consumer to manage the state. Or, they can emit an event (e.g. `update`), but then have to round-trip back to the extension to get the state (and somehow store it on their side again).
+
+Something like observables have a nicer API for this, but they are for pushing data (i.e. multiple readers), not for pulling it (i.e. any writers). They also have the same problem of putting the burden on the consumer.
```
YousefED, line 33: do you have an example of what you mean with `but they are for pushing data (i.e. multiple readers), not for pulling it (i.e. any writers)`?

adr/2025_06_13-extensions.md
```diff
@@ -0,0 +1,49 @@
...
+- A unified store API, to make consuming extension state homogenous, this will be discussed further below
+- Life-cycle event handlers, by providing hooks for `create`, `mount`, and `unmount` we allow extensions to attach event handlers to DOM elements, and have a better general understanding of the current state of the editor. The `onCreate` handler will be very handy to guarantee access to the editor instance before anything is called.
+- Editing event handlers, by providing hooks for `change`, `selectionChange`, `beforeChange`, and `transaction` we give extensions access to the fundamental primitives of the editor.
+
+## State Management
+
+What I had the most trepidation about was deciding on whether we should prescribe a state management library, and if so, which one.
+
+I think the answer is _yes_, we should prescribe a state management library, and I think the answer is @tanstack/store.
```
YousefED, line 17: I think the best way to validate this is to create a PR that migrates the AI extension from zustand to tanstack. Let's create a task for this?

adr/2025_06_13-extensions.md
```diff
@@ -0,0 +1,49 @@
+# BlockNote Extension API
+
+It is definitely more designed by accident than intention, so let's put some thought into the sort of API we would want people to build extensions with.
```
YousefED, line 3: Before diving into technical details below, let's zoom back out and ask ourselves what an Extension actually is / should be used for.

When should people use extensions? And what for? Why do we want to introduce the concept of an extension?

Having some clarity to these questions will help us to both better explain the concept to users, and help us create a better API design

adr/BlockNoteExtension.ts
```diff
@@ -0,0 +1,188 @@
...
+    },
+  ) => boolean | void;
+  public onTransaction?: (
+    context: ExtensionContext<BSchema> & {
+      tr: Transaction;
+    },
+  ) => void;
+}
+
+export interface BlockNoteExtension<State, BSchema extends Schema = Schema> {
```
YousefED: Allow me to challenge some of the ideas. Goal of my questions is to see if we can make things even _simpler_.

- What are the use-cases of an extension before an editor is creation? i.e.: do we need `onCreate`?
- Why expose event handlers like `onChange`, `onSelectionChange` etc. here? The user should already be able to subscribe to them via the editor instance, correct?
- Same for `keyboardShortcuts`. Could there be an API for adding keyboardshortcuts without an extension?

adr/2025_06_13-packages.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Core
+
+The `@blocknote/core` package should contain everything needed to build a pure JS editor instance. It has no opinion on the UI layer, and is only concerned with the core editor functionality.
+
+It contains sub-packages for:
+
+- `@blocknote/core/blocks` - Contains the default blocks, inline content, styles and their implementations (excluding things like blockContainer, blockGroup, etc which are core to the editor but should not be part of the schema).
+  - `@blocknote/core/blocks/header` - Contains the header block.
+  - `@blocknote/core/blocks/code-block` - Contains the code block block.
```
YousefED, line 11: except the code-block will probably stay in a different package right?

adr/2025_06_13-packages.md
```diff
@@ -0,0 +1,30 @@
...
+  - etc...
+- `@blocknote/core/extensions` - Re-exports the extension packages below.
+  - `@blocknote/core/extensions/comments` - Contains the comments extension.
+- `@blocknote/core/exporter` - Contains the exporter package.
+  - `@blocknote/core/exporter/html` - Contains the html exporter package.
+  - `@blocknote/core/exporter/markdown` - Contains the markdown exporter package.
+- `@blocknote/core/server-util` - Contains the server util package.
```
YousefED, line 18: not sure if `server-util` should be part of core as it might pull in some other deps?

adr/2025_06_13-packages.md
```diff
@@ -0,0 +1,30 @@
+# BlockNote Packages & Sub-Packages
```
YousefED, line 1: Looks good overall. What's your reasoning for leaning heavy into subpackages? I don't think this is very common, right?

adr/2025_06_13-schema.md
```diff
@@ -0,0 +1,21 @@
...
+
+In a somewhat similar vein, I think there might be use for having an indirection layer for referring to specific blocks, inline content, styles, etc. Reason being that it allows callers to refer to a group of things with a single identifier, and allow customizing that membership by just modifying that group.
+
+Examples include:
+
+- Keyboard shortcuts can refer to a group of blocks/inline-content/styles without modification to the handler
+- Relationships between blocks/inline-content/styles can be defined (e.g. allow for a todo block to only have todo item children)
+- Properties of blocks/inline-content/styles can be defined (e.g. adding `heading` to the `toggleable` group)
+
+This may or may not be useful, but it is a thought.
```
YousefED, line 21: cool, let's keep in mind 👍

adr/2025_06_13-schema.md
```diff
@@ -0,0 +1,21 @@
+# BlockNote Schema
+
+Right now it is overly burdensome to have to pass around 3 different types to the editor, and it is also not very type-safe (when you just end up with `any` everywhere).
+
+The idea is to have a single type that is a union of the 3 types, and then make type predicates available to check if accessed properties are valid (and likely just assertions too).
```
YousefED, line 5: would be great if we can get this to work!

adr/2025_06_16-configuration.md
```diff
@@ -0,0 +1,28 @@
...
+- **extension-configuration**: The configuration of the extensions that are available in the editor
+  - e.g. whether to add collaboration to the editor
+- **editor-view-configuration**: The configuration of the editor views
+  - e.g. whether to show the sidebar, the toolbar, etc.
+- **editor-configuration**: The configuration of the editor itself
+  - e.g. how to handle paste events
+
+These forms of configuration are not mutually exclusive, and can be combined in different ways. For example, knowing that the editor has collaboration enabled, might change the what the keybindings do for undo/redo.
+
+In an ideal world, these configurations would be made at the "lowest possible level", like configuring the number of levels of headings would be configured when composing the schema for the editor, rather than at the editor level.
```
YousefED, line 22: great principle 🙌

adr/2025_06_16-nested-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+There is a 1:1 relationship between the block type and it's content. And, no restrictions of the inline content allowed within the block.
+
+> TODO come up with a description of the sorts of keybinds & cursor behavior that should be supported for this block type
+
+## Block with nested blocks
+
+This is a proposal for how to support nested blocks.
+
+```json
+{
```
YousefED, line 35: this is technically what we already support to some level, right?

adr/2025_06_16-nested-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+## Block with structured inline content fields
+
+This is a proposal for how to support multiple inline content fields within a block.
+
+```json
+{
+  "type": "alert",
+  "content": null,
+  "children": [
+    {
```
YousefED, line 73: I'm not sure if this will work out. now suddenly, `children` can not only contain blocks anymore, but also inline content. I'm afraid this will break quite a bit (both technically and in terms of explainability)

YousefED: or wait, alert-title and alert-content are considered blocks here as well?

YousefED (commented):
Nice! Lots of great stuff in here. Besides discussing the open comments, how do we move forward from here?

My suggestion would be to extract isolated work items and prioritize them. Maybe first identify which other substantial core improvements are missing, the following come to mind:

- UI elements cleanup
- Streamline Rendered content, "external HTML", markdown, exporters (i.e.: topics discussed in Berin)
- Zod props
- unified style / theming API
- ...?

adr/2025_06_16-nested-blocks.md
nperez0111: Yep, considered blocks. Would need to introduce the concept of a "nested block" which is not valid on it's own but only in the context of being part of another block.

This is also shown with the table example below

adr/2025_06_16-nested-blocks.md
nperez0111: Yep, modeled on the multi-column, but generic for others to implement

adr/2025_06_13-packages.md
nperez0111: Leaning heavy into sub-packages isn't super common for editors, because they normally split on the package level, and most editors are fairly old (i.e. pre-subpackages).

I've always hated how many packages there are just to get an editor going. But, the main trade-off of splitting at the package level is that it makes the install size smaller and feels "composable", users can get into weird states (e.g. a user using a new version that depends on a core feature that didn't exist on their version). This sort of a approach, completely side-steps package version drift, while also not making the top-level API just a huge list of options. It also is nice for code-splitting and general "organization", we can treat sub-packages as a proper top-level package and design it's API to be more sensible than a huge export of just everything we have.

adr/2025_06_13-packages.md
nperez0111: Worthwhile to look into, I should mention that the main reason to split to another package is around deps & install size (or being a very different concern of course)

adr/2025_06_13-packages.md
nperez0111: There is the `code-block` default block, and there is the code-block package which would be for things like the shiki languages and all that. Probably worthwhile to revisit the naming there.

adr/2025_06_13-extensions.md
nperez0111: I was leaning more towards something like `editor.extension.ai` since the editor can know what extensions it has been given.

Though this can get awkward if you don't have the consumer's editor instance. So we likely would have a helper similar to that. But, I wanna see how that shakes out type-wise

adr/2025_06_13-document-transforms.md
nperez0111: Working on that 😉

adr/2025_06_13-document-transforms.md
nperez0111: insertContentAt(at: Location, content: Block | PartialBlock | string)
getSelection(): {range: Range}

A number of the low-level Tiptap API uses for things like deleting some chars

This will also be very useful for server-side editing ops like rewriting a paragraph

packages/core/src/editor/Location.ts
nperez0111: Likely would use a Point to refer to table-cells as it is right now.

If [a table-cell was implemented as nested block](https://github.com/TypeCellOS/BlockNote/blob/a80259fe21992cd6842150540e8111d84a8c2609/adr/2025_06_16-nested-blocks.md#L122), then it would just be a Path

adr/2025_06_13-document-transforms.md
nperez0111: Toss-up for `insertBlocks`, but for `insertContent(ctx: {at: Location; content: string })`

packages/core/src/editor/Location.ts
nperez0111: Fair.

a `Path` can be useful for being able to compare depths of things based on their content alone.

For example:

```ts
const a = ['abc', '123']
const b = ['abc', '345']
```

Based on this alone, you know they share parents, and therefore siblings of some kind, but I guess you would always have to check with the actual document to see that they really exist or not, so you are probably right

adr/2025_06_13-extensions.md
nperez0111: An Observable is a one-way stream, like an event emitter.

From [RXJS Observable](https://rxjs.dev/guide/observable):
image

So, they still don't fix the current problem we have with events because they are still putting the burden on consumers.

TanStack & Zustand are "push-pull" systems where you get a snapshot (i.e. a pull), and also be notified of changes (i.e. a push).

adr/2025_06_13-extensions.md
nperez0111: An extension is a way to extend the editor with additional functionality, in a way that is integrated with the editor API.

Examples of when to use an extension:

You want to add behaviors to the editor that are reactive to the editor's current state such as:
 - menus and toolbars (e.g. cmd-k for AI ops)
 - keyboard shortcuts (e.g. adding a shortcut for inserting a custom block)
 - side-effects (e.g. collaborative editing with another lib)

These extensions have APIs of their own, which should be accessible across the UI, (e.g. a button may open a menu).

We want to introduce the concept of an extension for 2 reasons:

 - We already are using an extension system internally, prosemirror plugins, which are difficult to reason about because of the level of abstraction they operate at.
 - We can expose an API for users to extend the functionality of the editor at a higher-level API, which is then even sharable across projects (e.g. we may not want to introduce a step custom node, but it'd be nice for the community to be able to add one, or open project tasks, etc.)

adr/BlockNoteExtension.ts
nperez0111: **onCreate**: I went back and forth on this one, I settled on it because:
You need some point where you give the extension the `editor` instance, but also want to be able to support an extension being configurable conveniently.

So you could define an extension to be:
```ts
type FunctionExtension = (options: Options) => (editor: Editor) => ExtensionInstance;
```
To get around this, but then it makes the typing a slight bit more difficult to extract and infer properly.

Open to removing it as long as we can define a good way about solving this


**onChange, onSelectionChange**: the editor instance I believe is doing too much by providing these methods, it has to track these instances and each extension has to remember to unsubscribe to them. We should be providing a higher-level API that gives you what you'd likely need.

**keyboard shortcuts**: Right now, keyboard shortcuts are pretty difficult to understand, since many things can register them, if there was a unified API, you can do things like explain which extension's keyboard shortcut was triggered.
Also, if you need an editor instance to register a keyboard shortcut, that feels backward to me, you'd have to re-initialize the state of the editor to register this new keyboard shortcut rather than having it at creation time.
Maybe you mean a separate API for keyboard shortcuts (like schema is now), but often it is the case where a keyboard shortcut is tied to an extension's state, which would feel awkward to be separated from.

nperez0111:
>     * UI elements cleanup
Agreed, I feel like this is a bit larger than I'm willing to take on at the moment though.


>     * Streamline Rendered content, "external HTML", markdown, exporters (i.e.: topics discussed in Berin)

Agreed, I will add this to the list.

>     * Zod props

Will add to the list

>     * unified style / theming API

I think this is the same as the ui elements cleanup

>     * ...?
I feel like these are the most pressing at the moment, but let me know what you think!

YousefED:
Forgot this one:

- streamline default blocks

adr/2025_06_13-extensions.md
nperez0111: Based on our discussion here, I think I've come around to something. I think that there are 2 separate needs here like you pointed out:

 - There is a need for a higher-level API than what prosemirror affords. This would be for adding low-level plugins to the editor. Examples would include things like menus, toolbars, and things that need access to editor state like table of contents.
 
 - There is also a need for a higher-level packaging that "bundles" editor functionality into a single distributable packaging. It may include things like adding blocks to the schema, adding lower-level plugins for exposing new methods to the editor (e.g. `editor.extensions.ai.acceptChanges()`)
 
I'll propose the naming of: **extension** (low-level) vs. **bundle** (high-level)

adr/2025_06_17-formats.md
```diff
@@ -0,0 +1,55 @@
+# BlockNote Formats
+
+Right now, there are several formats supported by BlockNote:
```
YousefED, line 3: I would even consider the "rendered" HTML another format (i.e.: what you see in the editor)

adr/BlockNoteBundle.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Add a dictionary of locales to the editor
+   */
+  dictionary?: Record<string, Locale>;
+
+  /**
+   * Items that are available to the slash menu
+   */
+  slashMenuItems?: (
```
YousefED, line 25: Let's cross this bridge when we get there, but I think we can come up with something more generic / scalable

adr/BlockNoteEditor.ts
```diff
@@ -0,0 +1,131 @@
...
+  // low-level operations
+  public exec: (command: Command) => void;
+  public canExec: (command: Command) => boolean;
+  public transact: <T>(callback: (transform: Transform) => T) => T;
+
+  // current state
+  public get document(): Block[];
+  public set document(document: Block[]);
+  public get selection(): Location; // likely more complex than this
+  public set selection(selection: Location);
```
YousefED, line 18: Thinking about naming, my association with a "transform" is that it would change something. I wouldn't expect things like `selection`, `getBlock` etc. to be "transforms". Should these be something else, or do we need a better name?

adr/2025_06_13-extensions.md
YousefED: I doubt this will work type-wise, but curious what you come up with!

adr/2025_06_13-extensions.md
YousefED: Ah got it, coming from RX I think you're right. I was thinking about MobX observables which (if I'm not mistaken) are more like (deep) signals

adr/2025_06_13-extensions.md
YousefED: Cool. I like the separation, I think when we work on this it will become clear if the separation will be needed and / or whether we need to expose both to consumers

adr/BlockNoteEditor.ts
nperez0111: I toyed with calling this `editor.state` but it felt overly generic, but maybe that is the right way to go.

My way for explaining transforms was that it is all about transitions of state & what you might need to go from A -> B

adr/BlockNoteBundle.ts
nperez0111: 100% was just a concept of cross-cutting concerns

adr/2025_06_17-formats.md
nperez0111: Yep, I did miss that one

adr/2025_06_13-packages.md
YousefED: > not making the top-level API just a huge list of options

Cool, seems like this is the main reason. Note to self to summarize:

- packages come with a lot of bundling / dependency / organizational overhead
- just using a strict folder structure (which imo would be the other alternative) doesn't achieve the goal of cleanly organizing the API (unless we export using namespaces)

Also agree with the second point? I.e.: better to use subpackages vs namespaces? One benefit of namespaces I suppose would be that it's more discoverable (e.g.: when you do `import *` you easily get to see everything that's available)

(note that with namespaces I mean that for example, instead of exporting all blocks top-level, we can export a "blocks" object which encapsulates the default blocks. i.e. `import { blocks } from "@blocknote/core"`. In your design this would be in subpackage `@blocknote/core/blocks`)

nperez0111: Yep, I think a strict folder structure with namespaces is conceptually very similar.
Also sub-packages are more fine-grained for package resolution, you don't need to tree-shake what was never imported.

adr/2025_06_13-extensions.md
nperez0111: Yep, overloaded term, mobx is based on signals since it is "push-pull"

adr/2025_06_13-document-transforms.md
```diff
@@ -0,0 +1,65 @@
...
+
+We've done pretty well with our existing API, but there are a few things that could be improved.
+
+Referencing content within the document is either awkward or non-existent. Right now we essentially really only have an API for referencing blocks by their id with no further level of granularity.
+
+## Locations
+
+[Looking at Slate](https://docs.slatejs.org/concepts/03-locations) (highly recommend reading the docs), they have the concept of a `Location` which is a way to reference a specific point in the document, but it does not have to be so specific as positions, it has levels of granularity.
+
+This gives us a unified way to reference content within the document, allowing for much more granular editing. Take a look at the `Location.ts` file for more details around this.
```
matthewlipski, line 13: Do we want to necessarily use the `Location` concept? I like the granularity but it feels like we could achieve something similar by just using block/inline content ID references. I.e.:

- Low granularity: Point to the start/end of a block by its ID.
- Mid granularity: Point to the start/end of inline content by its ID.
- High granularity: Point to the start of inline content with an offset by its ID.

I feel like Slate's `Location` is necessary due to the fact that you can only reference nodes by their positions. Since we set IDs, I think we should make use of that since imo it'll result in a simpler API. So our `Location` could look more like:

```ts
type Location = {
  id: string;
  side?: "start" | "end";
  offset?: number
}
```

adr/2025_06_13-document-transforms.md
matthewlipski: Yeah I like this idea a lot, feels more general than ProseMirror's decorations. In theory it would also make it much nicer for our UI architecture, since instead of writing plugins you could just use references. E.g. for the formatting toolbar, you could listen to selection changes and create a reference pointing to the selection when you want to show it.

adr/2025_06_13-extensions.md
```diff
@@ -0,0 +1,60 @@
...
+
+I think the answer is _yes_, we should prescribe a state management library, and I think the answer is @tanstack/store.
+
+<details>
+<summary>Why @tanstack/store? And not something else?</summary>
+It comes with a few benefits:
+
+- It gives a single API for all state management, which is more convenient/consistent for consumers
+
+As for which library to use, I think we should use @tanstack/store.
```
matthewlipski, line 25: Duplicate

adr/2025_06_13-extensions.md
```diff
@@ -0,0 +1,60 @@
...
+Zustand is a popular library, but allowing partial states makes it somewhat unsafe in TypeScript.
+
+Jotai is probably my second choice, but it makes it a bit awkward to update states because it relies on a separate store instance rather than the "atom" being able to update itself <https://jotai.org/docs/guides/using-store-outside-react>.
+</details>
+
+## Exposing extension methods
+
+Not everything can be communicated through just state, so we need to be able to expose additional methods to the application.
+
+I propose that we have a `extensions` property on the `BlockNoteEditor` instance, which is a map of the extension key to the extension instance, but filtered out to only include non-blocknote methods (as defined by the `ExtensionMethods` type).
```
matthewlipski, line 46: I'm curious to see what you have in mind for how extensions should be defined too. Right now, to make something resembling an extension using the BlockNote API only, you generally listen to `editor.onChange`/`useEditorChange` and go from there, whether it's making changes to the editor, or doing stuff outside the editor.

So I'm wondering if all we really need an extension to be is a state + listeners, sort of like this:

```ts
function defineExtension<State extends Record<string, any>>({
  initialState: State;
  listeners?: Partial<{
    onCreate: (editor: BlockNoteEditor) => State,
    onContentChange: (editor: BlockNoteEditor) => State,
    onSelectionChange: (editor: BlockNoteEditor) => State,
    onDestroy: (editor: BlockNoteEditor) => State,
  }>
}) {...}
```

adr/2025_06_16-nested-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+- **nested-blocks** The first is the ability for a block to contain other blocks within it (e.g. a table cell having not just inline content, but actual other blocks inside it)
+- **content-fields** The second is the ability for a block to contain multiple pieces of content within it (e.g. an alert block having a title & description field (which contain inline content))
+
+Let's start with the first problem, nested blocks. By describing existing block relationships:
+
+## Block with inline content
+
+This is the simplest case, and the only one that is currently supported by BlockNote.
+
+```json
```
matthewlipski, line 14: Worth mentioning that the current inline content object shape is a mess. You have `StyledText` which is a different shape to `Link`, which is also different to custom inline content. This makes trying to do basic things like reading plain text kind of a nightmare, and is something that we definitely need to address.

adr/2025_06_13-document-transforms.md
nperez0111: > inline content by its ID.

Inline content doesn't have an ID does it? And, I don't think it should, it'd be too granular and require wrapping things in marks everywhere.

I'm willing to remove `Path` as pointed out here: https://github.com/TypeCellOS/BlockNote/pull/1756#discussion_r2151572243, and, I've now updated the typescript file to represent that.

I don't think your proposal for a `Location` captures everything that Slate's could though, such as representing a `Range` which is extremely useful.

Using your language, I think we can achieve this with:

```ts
/**
 * A block id is a unique identifier for a block, it is a string.
 */
export type BlockId = string;
/**
 * A block identifier is a unique identifier for a block, it is either a string, or can be object with an id property (out of convenience).
 */
export type BlockIdentifier = { id: BlockId } | BlockId;
/**
 * A point has an id and an offset, it is used to identify a specific position within a block.
 */
export type Point = {
  id: BlockId;
  offset: number;
};
/**
 * A range is a pair of points, it is used to identify a range of blocks within a document.
 */
export type Range = {
  anchor: Point;
  head: Point;
};
/**
 * A location is an id, point, or range, it is used to identify positions within a document.
 */
export type Location = BlockId | Point | Range;
```

adr/2025_06_16-nested-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+
+The core idea is that the `parent` block restricts what content is allowed within it's children.
+
+> TODO come up with a description of the sorts of keybinds & cursor behavior that should be supported for this block type
+
+## Examples
+
+### Rebuilding tables with nested blocks
+
+Using this new structure, we can rebuild tables if we had this new API:
```
matthewlipski, line 103: This makes sense, but 2 questions:

- The consumer will have to specify when creating a block, which blocks it can be a part of or which blocks it can contain, right? And for the alert example, there will have to be some way of enforcing that each alert block must have 1 alert title followed by 1 alert description. I assume we would use something like the ProseMirror `group` and `content` fields for this?
- If you wanted to have nested blocks in the table block, would you just append `nested-block` at the end of the children? If so, is it possible to have e.g. `children: [{ type: "block1", ... }, { type: "nested-block", ... }, { type: "block2", ... }]`?

adr/2025_06_13-extensions.md
nperez0111: Described in the TS file here: https://github.com/TypeCellOS/BlockNote/blob/9684a7d3ce180bc028e6a147f57f6265459e3046/adr/BlockNoteExtension.ts#L9

I also thought it'd be good to have hooks for these sort of event's, but Yousef pointed out they were already available on the editor instance.

adr/2025_06_16-nested-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+```json
+{
+  "type": "custom-block-type",
+  "props": {
+    "abc": 123
+  },
+  "content": null,
+  "children": [
+    {
+      "type": "nested-block",
```
matthewlipski, line 43: `nested-block` is analagous to the current `blockGroup` right?

adr/2025_06_16-nested-blocks.md
nperez0111: Good thing to add, I don't have an answer for this atm. But feel free to write a proposal

adr/2025_06_17-formats.md
```diff
@@ -0,0 +1,59 @@
...
+type Docx = Buffer;
+function toDocx(model: Model): Docx;
+
+type PDF = Buffer;
+function toPDF(model: Model): PDF;
+```
+
+You'll notice that the formats are all derived from the model.
+
+This gives us a unidirectional flow of data. Any other direction is potentially lossy (e.g. markdown -> model).
```
matthewlipski, line 41: Isn't `Model -> ***` also potentially lossy if the target format doesn't have mappings to aspects of the model (e.g. nested blocks in Markdown)?

nperez0111: Sure, that is true too. Depends how good your parse/render is (e.g. model to html **should be** lossless)

adr/BlockNoteBundle.ts
```diff
@@ -0,0 +1,80 @@
...
+    | (() => any)
+  )[];
+}
+
+export type Locale = Record<string, string | Record<string, string>>;
+
+/**
+ * This is an example of what an extension might look like
+ */
+export function multiColumnExtension(extensionOptions: {
```
matthewlipski, line 54: I noticed this is quite a different approach to how we currently do e.g. the AI extension, as here everything is v much bundled together whereas for AI you add stuff like slash menu items manually.

This seems like another case of convenience vs composability, as it's much nicer to add an extension but can be annoying if e.g. you don't want to include the extension's keyboard handlers. What made you lean towards this approach instead?

adr/2025_06_13-document-transforms.md
nperez0111: >  E.g. for the formatting toolbar, you could listen to selection changes and create a reference pointing to the selection when you want to show it.

🤯 what a great idea! If it were robust enough for that, this could simplify a ton of things!

adr/BlockNoteEditor.ts
matthewlipski: Yeah I'm also not 100% convinced about `Transform` - I see the reasoning it just feels like it should only include stuff which does indeed modify the editor state, whereas the getters do not.

adr/BlockNoteBundle.ts
nperez0111: This is meant to be a higher-level API than extensions are, and would be useful for packaging sets of functionality (e.g. for 3rd-party packages).

There is always a trade-off when it comes to flexibility vs convenience. This is definitely the most half-baked of the APIs, but is mostly to point out that there isn't any single "BlockNote Extension API". There has to be something low-level like a prosemirror plugin, and something high-level like a "Add comments to the editor".

adr/2025_06_13-document-transforms.md
matthewlipski: Ah good point about ranges, missed that. I'm wondering if there's a use case for pointing to a specific inline content that we would miss by just having block + offset, maybe for extensions which work with links?

adr/BlockNoteEditor.ts
nperez0111: I also toyed around with trying to separate reads from writes, but it too felt awkward for example:

```ts
// feels like mixing levels of the API
editor.transform.replaceBlocks(editor.state.getCurrentSelection().blocks, [])

// feels about right, but state is overloaded with prosemirror
editor.state.replaceBlocks(editor.state.selection.blocks, [])

// Feels awkward
Transform.replaceBlocks(editor, editor.state.getCurrentSelection().blocks, [])

// Maybe? But, split now
editor.commands.replaceBlocks(editor.state.getCurrentSelection().blocks, [])
```

What I'm trying to stay away from is just throwing it on the top-level for the editor instance, things which operate on it's state should be different than things like `editor.focus`, but maybe I've gone too far in the other direction.

adr/2025_06_13-document-transforms.md
nperez0111: To point to a specific inline content, you could have a `Point` which points at the start of the inline content, or a `Range` that encapsulates the whole inline content range:

```ts
{id: 'abc', content:[{type:'link', text: 'ABC'}]}

const point = { id: 'abc', offset: 0 };

const range: { anchor: { id: 'abc', offset: 0 }, head: { id: 'abc', offset: 3 } }
```

adr/2025_06_16-nested-blocks.md
nperez0111: > The consumer will have to specify when creating a block, which blocks it can be a part of or which blocks it can contain, right? And for the alert example, there will have to be some way of enforcing that each alert block must have 1 alert title followed by 1 alert description. I assume we would use something like the ProseMirror group and content fields for this?

Correct. We'd have to have prosemirror enforce this, we can decide how best to let the consumer describe this structure. I proposed groups partially for this reason: https://github.com/TypeCellOS/BlockNote/blob/52ecefbdfbdef8249be2ea389d4f5bba36148ab5/adr/2025_06_13-schema.md#L11-L21

> If you wanted to have nested blocks in the table block, would you just append nested-block at the end of the children?

Tables would be defined to only allow `table-cell`s as children, so no other blocks would be allowed within a `table` block. However, `table-cell`s would be allowed to have arbitrary blocks (or, not if they were configured not to).

matthewlipski: I see, so this is more of a change of concept of what a nested blocks actually is I suppose. Because right now, it's an inherent property of all blocks, that they can have nested blocks that are indented below. Whereas in this new design, nested blocks are a lot more flexible, and the indented nested blocks are just a generic implementation.

Am I understanding that right?

nperez0111: I tried to define my terms here: https://github.com/TypeCellOS/BlockNote/blob/a80259fe21992cd6842150540e8111d84a8c2609/adr/2025_06_16-nested-blocks.md#L5-L6

But, I see that "nested-blocks" is indeed overloaded, take it to mean "blocks as content", maybe.

adr/2025_06_16-nested-blocks.md
```diff
@@ -0,0 +1,155 @@
...
+      "type": "alert-title",
+      "content": [
+        {
+          "type": "text",
+          "text": "Hello, world!"
+        }
+      ]
+    },
+    {
+      "type": "alert-content",
```
nperez0111, line 83: These sub-blocks are a new content type, which are only valid within elements which explicitly allow them. So an alert block would have to declare the shape of the data. How we do this is still up for grabs, in some cases you want a rigid schema (i.e. only ever title & content, no other combinations are valid vs. something like tables that can have any number of rows)

PR #1766, opened 2025-06-17
feature: custom ai executor

YousefED:
This adds support for a custom AI executor instead of always relying on the Vercel AI SDK to execute requests.

See [this example code](https://github.com/TypeCellOS/BlockNote/compare/feature/ai-executors?expand=1#diff-b9a12b96e90df4ac66171c63c2441fcd7c4863489bfab5a41aeb35103c1a6f69R68) for usage.

_You can test this using `npm i https://pkg.pr.new/TypeCellOS/BlockNote/@blocknote/xl-ai@1766`_

### TODO:
- Implement an easier way to create LLMResponse from a custom streaming response (for now, use `LLMResponse.fromArray` for testing
- We could move "model", "stream", "maxRetries" etc. to become responsibilities of the executor

### Alternatives

This PR basically kind of lets you "opt out" from using the AI SDK as layer to call underlying LLMs.

We can also explore:
a) (lean in on AI SDK) making it easy to implement a custom Model / Provider which you can then use as "executor". (i.e.: we consider the Model to be the point that defines how LLMs are called
b) (lean "out" of AI SDK on client side): instead of using the AI SDK heavily on the client and a lightweight proxy on the server, we could follow their default design more closely and recommend users to expose REST endpoints that can then in turn use the AI SDK


nperez0111:
I would prefer for the client to do less work, and given that backends will have a lot more context than a frontend would, I'd personally ease more toward moving the work into the backend & making that easier.

Also worth documenting what the format should look like over the wire even if only examples

BetterAndBetterII:
I don't know if the author has experienced AI IDE, such as Cursor. When editing a segment, Cursor supports **streaming** very well. According to the reproduction of some open source AI IDEs, such as void.

This is the form of applying code modification in void: https://github.com/voideditor/void/blob/main/VOID_CODEBASE_GUIDE.md#apply

This method can be easier to implement streaming edit.

I think this modification method can make the model output more accurate and promote the generation of more new content. (I have observed that the current form of tool calling makes it easy for the model output content to be short)

RafaelRMaier-CA:
This would probably help my use case a lot. I am trying to build a canvas style app, but I have my own backend with defined personas, graphs, etc. The model is defined inside the CanvasPersona, so I don't wanna expose this info to the frontend. I have also set up /stream and /astream endpoints with Langserve, so I would like to call those.

YousefED:
moved to #1903

PR #1768, opened 2025-06-18
XL React Email Renderer 

jmarbutt:
## Overview 
This pull request introduces a new example for converting BlockNote documents into React Email documents, along with setup and dependencies for the `@blocknote/xl-react-email-exporter` package. The changes include adding a complete example implementation, updating configurations, and creating the necessary package files for React Email integration.

## Example
I have setup the example in the examples folder to test with. 

## Implemented Rendering for `react-email`
This implementation is a first pass allowing exporting to react-email components to send via email.  Most of the implementation was done originally in the branch https://github.com/TypeCellOS/BlockNote/tree/feature/react-email-exporter

- **Paragraph** - This is unchanged from the first pass in the original branch.
- **Bullet List** - I implemented this using standard '<ul>...</ul>', from what I can tell this is completely compatible with the react-email rendering
- **Numbered List** - Followed the pattern in the `pdfExporterBlocks` to create the numbered list  correctly
- **Check List** - Setup the Check list with SVGs, I am not 100% this will work in every email client but this is a first pass. My original pass just used a simple block but did not render well. I need to test this in more email clients to find the compatibility. 
- **Heading** - Unchanged since original implementation
- **Code Block** - This utilizes the built in `react-email` codeblock component which renders pretty nicely in the email.
- **Audio, Video, Files** - I have these return blank for now because they are really not compatible with email.
- **Image** - Unchanged 
- **Table** - Table utilizes the `section` component from `react-email` as their recommendation. But one thing you will find is it renders 1 table per row, which this created an issue with the alignment of columns. So I am currently setting a `defaultColWidth` but a future enhancement may be to make that better
- **Quote** - Simple Implementation, not sure if it needs something else
- **Page Break** - Unchanged


examples/05-interoperability/08-converting-blocks-to-react-email/App.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+  useEffect(() => {
+    onChange();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const slashMenuItems = useMemo(() => {
+    return combineByGroup(
+      getDefaultReactSlashMenuItems(editor),
+      getPageBreakReactSlashMenuItems(editor),
```
YousefED: I'm not sure page break makes sense for email. It's good that we support it in the exporter, but maybe remove from the demo to make things simpler?

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
```diff
@@ -0,0 +1,164 @@
...
+        );
+    },
+
+    codeBlock: (block) => {
+        const textContent = (block.content as StyledText<any>[])[0]?.text || "";
+        
+        return <CodeBlock
+            code={textContent}
+            fontFamily="'CommitMono', monospace"
+            language="javascript"
```
YousefED: we should be able to get the language from the props

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
```diff
@@ -0,0 +1,164 @@
...
+  React.ReactElement<any>,
+  React.ReactElement<typeof Link> | React.ReactElement<HTMLSpanElement>
+> = {
+    paragraph: (block, t) => {        
+        return <Text>{t.transformInlineContent(block.content)}</Text>;
+    },
+    bulletListItem: (block, t) => {
+        // Use <ul> and <li> with Tailwind classes via className (supported by react-email)
+        return (
+            <ul className="list-disc pl-6 mb-2">
```
YousefED: we might need find a way to combine adjacent list items into one list, right?

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
```diff
@@ -0,0 +1,164 @@
...
+        
+        return <CodeBlock
+            code={textContent}
+            fontFamily="'CommitMono', monospace"
+            language="javascript"
+            theme={dracula}
+        />
+        
+    },
+    audio: () => {
```
YousefED: We could probably use a similar approach to pdf / docx where we render these as an icon + link

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { testDocument } from "@shared/testDocument.js";
+interface KoalaWelcomeEmailProps {
+  userFirstname: string;
+}
+
+const baseUrl = process.env.VERCEL_URL
+  ? `https://${process.env.VERCEL_URL}`
+  : "";
+
+export const KoalaWelcomeEmail = ({
```
YousefED: is this used or can we remove this?

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+  it("should export a document", async () => {
+    const exporter = new ReactEmailExporter(
+      BlockNoteSchema.create(),
+      reactEmailDefaultSchemaMappings
+    );
+
+    const html = await exporter.toReactEmailDocument(testDocument as any);
+
+    // eslint-disable-next-line no-console
+    console.log(html);
```
YousefED: you should be able to do a snapshot test here

(similar to [here](https://github.com/TypeCellOS/BlockNote/blob/2040f865c189be7cc473df037f7ca1c040a0d978/tests/src/unit/shared/formatConversion/export/exportTestExecutors.ts#L26). Maybe you don't need the prettify part if this is built in to react-email)

Or, we could snapshot the React Email component instead of the HTML (like we do in the pdf exporter), not sure what's better

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
```diff
@@ -0,0 +1,188 @@
...
+    public renderFonts() {
+      return (
+        <>
+          <Font
+            fallbackFontFamily="Helvetica"
+            fontFamily="Inter"
+            fontStyle="normal"
+            fontWeight={400}
+            webFont={{
+              url: "https://fonts.gstatic.com/s/inter/v18/UcCO3FwrK3iLTeHuS_nVMrMxCp50SjIw2boKoduKmMEVuLyfAZ9hiA.woff2",
```
YousefED: is there a good way to do this without relying on google fonts?

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
```diff
@@ -0,0 +1,188 @@
...
+          />
+        </>
+      );
+    }
+    public async toReactEmailDocument(blocks: Block<B, I, S>[]) {
+
+      const transformedBlocks = await this.transformBlocks(blocks);
+      return renderEmail(
+        <Html>
+        <Head>{this.renderFonts()}</Head>
```
YousefED: we can expose some options to configure the Head / Preview, similar to PDF: https://www.blocknotejs.org/docs/editor-api/export-to-pdf#customizing-the-pdf.

Maybe we should also expose a way to customize a footer / header? (this should be a pretty common requirement I think)

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.test.tsx
YousefED: Might also be cool to run the result through the compatibility checker?

https://resend.com/blog/react-email-4#compatibility-checker

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
```diff
@@ -0,0 +1,164 @@
...
+    },
+    image: (block) => {
+        return (
+            <Img
+                src={block.props.url}
+                width={block.props.previewWidth}
+                alt={block.props.caption} />
+        );
+    },
+    table: (block, t) => {
```
YousefED: > Table utilizes the section component from react-email as their recommendation. But one thing you will find is it renders 1 table per row, which this created an issue with the alignment of columns. So I am currently setting a defaultColWidth but a future enhancement may be to make that better

This sounds a bit funky, it's weird that they don't expose a table component, right? I doubt this approach will work with large (wide) tables, merged cells, etc. Maybe we can just use regular tables?

Resources:

- https://github.com/resend/react-email/issues/503#issuecomment-1445450510
- maybe check implementation of https://github.com/LLazyEmail/react-email-table/?

I think it's quite weird react-email doesn't expose this. I'd expect it to export a table component so that it can add the right styling / attributes to make things consistent across clients. Seems like mjml takes a nicer approach here: https://documentation.mjml.io/#mj-table (perhaps we can also learn from their implementation?)

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
jmarbutt: I had left this in from your implementation, I will try removing it

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
jmarbutt: I got this working, this probably can be simplified but it was more complicated then I expected to get the styling to apply correctly

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.test.tsx
jmarbutt: I have put the test in but I am having some issues running all the tests, I get a number issues around the `pnpm test` is there something else I need to do to set that up?

I say we put the compatibility checker as an issue and get this in first

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
jmarbutt: I think we just want to setup the react-email's `preview` component. https://react.email/docs/components/preview

I am not sure there a real concept as a footer in email? Would we just want to render the content

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.test.tsx
jmarbutt: Ok I have the first unit test and snap shot working now

packages/xl-react-email-exporter/src/react-email/__snapshots__/reactEmailExporter.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`react email exporter > should export a document (HTML snapshot) 1`] = `Promise {}`;
```
YousefED: hmm this seems wrong. Are we missing an await in the test?

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
```diff
@@ -91,52 +114,44 @@ export const reactEmailBlockMappingForDefaultSchema: BlockMapping<
...
-                                    }}
-                                >
-                                    {t.transformInlineContent(cell.content)}
-                                </Column>
-                            );
-                        })}
-                    </Row>
-                ))}
-            </Section>
+            <table style={{ borderCollapse: 'collapse', width: '100%', margin: '16px 0', border: '1px solid #ddd', borderRadius: 4, overflow: 'hidden' }}>
```
YousefED: mjml adds this:

` border="0" cellpadding="0" cellspacing="0"`, do you think it's necessary? see https://mjml.io/try-it-live/components/table (and view HTML tab)

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
```diff
@@ -70,92 +68,166 @@ export class ReactEmailExporter<
...
-        const self = await this.mapBlock(b as any, nestingLevel, numberedListIndex) as any; // TODO: any
-
+        const self = await this.mapBlock(b as any, nestingLevel, 0) as any;
         const style = this.blocknoteDefaultPropsToReactEmailStyle(b.props as any);
-
         ret.push(
           <React.Fragment key={b.id}>
             <Section style={style}>{self}</Section>
-            {children.length > 0 && (
-              <Section style={{ marginLeft: 10, ...style }}>{children}</Section>
```
YousefED: I think we lost the support for indentation of nested elements with this change

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
jmarbutt: Yeah I am sure that is needed, I remember something about that when I have done other tables now that you mention it.

packages/xl-react-email-exporter/src/react-email/__snapshots__/reactEmailExporter.test.tsx.snap
jmarbutt: Oh thats weird, I will check.

packages/xl-react-email-exporter/src/react-email/defaultSchema/blocks.tsx
```diff
@@ -0,0 +1,196 @@
...
+        }
+        const headerRowsCount = (table.headerRows as number) ?? 0;
+        const headerColsCount = (table.headerCols as number) ?? 0;
+
+        return (
+            <table style={{ borderCollapse: 'collapse', width: '100%', margin: '16px 0', border: '1px solid #ddd', borderRadius: 4, overflow: 'hidden' }} border={0} cellPadding={0} cellSpacing={0}>
+                <tbody>
+                    {table.rows.map((row: any, rowIndex: number) => (
+                        <tr key={'row-' + rowIndex}>
+                            {row.cells.map((cell: any, colIndex: number) => {
```
nperez0111: We could make this mapping a little better by using `mapTableCell` which would conform a table cell into a standard interface

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
```diff
@@ -0,0 +1,272 @@
...
+    Head,
+    Html,
+    Link,
+    Preview,
+    Section,
+    Tailwind  } from "@react-email/components";
+  import { render as renderEmail } from "@react-email/render";
+
+import React from "react";
+  import { CSSProperties } from "react";
```
nperez0111: This is odd, does it pass the lint?

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
```diff
@@ -0,0 +1,272 @@
...
+        }
+        listItems.push(
+          <React.Fragment key={block.id}>
+            {liContent}
+            {nestedList.length > 0 && nestedList}
+          </React.Fragment>
+        );
+        j++;
+        itemIndex++;
+      }
```
nperez0111: This seems overly complicated, wouldn't it be more standard as a `for` loop

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    private async renderNestedLists(
+      children: Block<B, I, S>[],
+      nestingLevel: number,
+      parentId: string
+    ): Promise<React.ReactElement<any>[]> {
+      const nestedList: React.ReactElement<any>[] = [];
+      let i = 0;
+      while (i < children.length) {
```
nperez0111: same here

packages/xl-react-email-exporter/src/react-email/reactEmailExporter.tsx
jmarbutt: yes I have used it in css props

Issue #1745, opened 2025-06-10
PDF Export of Multi Column Block


thedivac:
When I add a multi column block to my editor, the PDF export no longer works. I assume this is because column and columnList are not implemented in the PDF scheme mapping.

Did I maybe miss something? I can't find anything in the docs or in the source code.

YousefED:
You're right, this hasn't been implemented yet - we're planning to add support for this within the next few weeks!

thedivac:
> You're right, this hasn't been implemented yet - we're planning to add support for this within the next few weeks!

Awesome, thanks!

PR #1781, opened 2025-06-23
feat: Support for multi-column block in PDF, DOCX & ODT exporters

areknawo:
This PR adds support for the multi-column block inside PDF, DOCX & ODT exporters based, using:
- Flexbox inside PDF exporter;
- Tables with invisible borders inside DOCX & ODT exporters;

closes https://github.com/TypeCellOS/BlockNote/issues/1745


packages/xl-odt-exporter/src/odt/odtExporter.tsx
```diff
@@ -126,20 +126,35 @@ export class ODTExporter<
...
-        numberedListIndex,
-      );
-
-      ret.push(content);
-      if (children.length > 0) {
-        ret.push(...children);
+      if (["columnList", "column"].includes(block.type)) {
+        const children = await this.transformBlocks(
+          block.children,
+          nestingLevel,
```
YousefED: should this be 0 instead?

areknawo:
> Great! Couple of small things:
> 
> * docx: word on mac gives me a warning
> * odt: libreoffice on mac shows borders
> 
> Also, the lockfile has quite some changes / upgrade. @nperez0111 what's a good policy for this? rather exclude these from this PR?
> 
> Somewhat unrelated, I noticed the heading spacing / font size for PDF is off (see "Heading Right"). Would it be possible to add a fix for that as well?

@YousefED I've fixed the headings in PDFs with added line-height, applied the feedback and cleaned-up the lockfile. The two exporter issues need some more research/discussion

nperez0111:
Looks like maybe `pnpm update ran` honestly it is better for upkeep to update this occasionally. Especially, for our actual dependencies (the ones our consumer will depend on). No preference in when this should happen, so I'm fine with it so long as our build & tests pass it.

Issue #1788, opened 2025-06-24
Can't copy paste multiple times


virgile-dev:
**Describe the bug**<br>If you copy paste this in the doc and keep hitting `ctrl` + `V` you'll be able to do it only one time.

```
- .

  .
- .
```

**To Reproduce**<br>If you delete the text and try to the same thing again you won't be able to.<br>https://github.com/user-attachments/assets/93509a37-eaaa-4d5d-a01d-e6a5be4afd22<br>**Misc**

- Tested here : [https://www.blocknotejs.org/docs](<https://www.blocknotejs.org/docs>)
- Browser: Version 137.0.7151.103 (Build officiel) snap (64 bits)
- [X] I'm a [sponsor](<https://www.blocknotejs.org/pricing>) and would appreciate if you could look into this sooner than later 💖

woutervroege:
@virgile-dev Cannot reproduce this issue as of now.

Could you confirm/verify this particular issue is still happening? And if so, could you provide details on the environment used (OS, Browser)?

Issue #1653, opened 2025-04-30
code block cannot be moved (example page)


EddyVinck:
**Describe the bug**
See the video. Cannot drag and move the code block or the title near the code block.

**To Reproduce**

https://github.com/user-attachments/assets/c75bd176-1356-4c10-88b9-cd08711272a4

**Misc**
- Node version:
- Package manager:
- Browser: 
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

bigscotia10:
I have the same issue, happens on all my blocks. It's also happening on your actual site as well.

![Image](https://github.com/user-attachments/assets/3e186618-a794-4ef0-8f34-7b91c73bf18c)

bryszard:
As far as I can tell this is a general issue - I can reproduce it both in my app and on the BlockNote's documentation page:

https://github.com/user-attachments/assets/e7347e11-68cf-4a2b-a368-b7d66ed7e3cb

bigscotia10:
@matthewlipski any luck on this one or any update?

nperez0111:
Please don't ping, or add more "me too" comments, we know this is a problem.

Issue #1610, opened 2025-04-16
Dragging text around copies the text instead of moving it


Matthieu-Gorgias:
## Dragging text around copies it instead of moving it

Hi, I’ve noticed a strange behavior when testing BlockNote (e.g. in the [Basic Setup](https://www.blocknotejs.org/examples/basic/minimal) example): if I select some word in the editor and try to drag it to move it elsewhere (on the same line or another block), instead of moving the word, it is duplicated.

Cf. example gif below:

![Image](https://github.com/user-attachments/assets/c4f1d4e8-47ed-447f-9036-5f855f5804e4)

**To Reproduce**

Drag some text around, it will be copied.

**Misc**
 - Tested on online examples and also locally with node v14.21.3 + packages `@blocknote/...`  v0.28.0
 - Browser: occurs on firefox and chrome and probably any browser

Thank you :)

tolypash:
Experiencing the same issue!

VladYuskiv:
Have the same issue!

kayvanhuisseling:
Same here!

YousefED:
Thanks, we'll look into this

Issue #1452, opened 2025-02-21
Custom blocks change type when dragging.


ralf-al:
**Describe the bug**
When I drag custom blocks and try to reorder them I get a strange issue.
Lets say I have 2 custom blocks. when I drag one of them to reorder to place it underneath another the custom block will morph into the custom block that was above,

lets say I have 

**To Reproduce**
Add 2 custom blocks of different types to the editor. drag the custom block right below the other. occasionally (quite often) the custom block that I dragged will turn into the other,

**Misc**
- Node version: 0.24.1
- Package manager: npm
- Browser: arc

Aichnerc:
Same here. If I drag custom blocks to another block, it sometimes becomes an exact clone of this block.

nperez0111:
Can either of you share a reproducible example? I think this might just be related to html serialization of the custom block (perhaps your parsers are matching the same element).

PR #1790, opened 2025-06-25
refactor: serialize clipboard to json instead of html

### LINKED ISSUE #1693: Dragging doesn't work on mobile
**Describe the bug**<br>I was trying the Blocknote editor in my project on my phone, but I noticed the dragging did not work on mobile.

I can see the drag handle, but I can't get it to work.

I then tried one of the official examples on my phone which also had the same problem.

**To Reproduce**

1. Go to one of the examples on your phone: [https://www.blocknotejs.org/examples/basic/default-blocks](<https://www.blocknotejs.org/examples/basic/default-blocks>)
2. Try dragging an element

**Misc**

- Node version:
- Package manager:
- Browser: Chrome on iPhone 13 Mini
- [ ] I'm a [sponsor](<https://github.com/sponsors/YousefED>) and would appreciate if you could look into this sooner than later 💖

### LINKED ISSUE #1694: Dragging custom block show Error: block type does not match
**Describe the bug**
If two different types of custom blocks are one after another, dragging one of them is causing the error 'Block type does not match'

**To Reproduce**
Here's a [demo](https://stackblitz.com/edit/github-1queg8p1?file=example.js) sandbox, try to drag the bottom one beside the top one

https://github.com/user-attachments/assets/7dc17f12-a330-4bc4-ba82-9b92d882c27e

**Misc**

- Node version: 22.13.1
- Package manager: pnpm(10.10.0), npm(11.1.0)
- Browser: Zen(firefox), chrome

nperez0111:
Instead of trying to serialize everything into `blocknote/html`, we can instead serialize into `blocknote/json` whatever slice of data we want to be on the clipboard or drag & drop.

This should resolve a number of issues related to drag & drop, and simplifies things significantly.

HTML cannot encode the same sort of information that json can, for example, that a slice has different depths is not representable by html & has to be inferred, whereas just representing it as a slice, just works.
 
 Does not resolve:
 - https://github.com/TypeCellOS/BlockNote/issues/1205
 - https://github.com/TypeCellOS/BlockNote/issues/1610 need to handle a drag like that as a sort of cut & paste
 - Likely the same https://github.com/TypeCellOS/BlockNote/issues/1653
   - this likely needs to be re-worked: https://github.com/TypeCellOS/BlockNote/blob/0eb3bceae6a52cd9612c7d52a1715d2f24400949/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts#L301-L335
 - Needs more info: https://github.com/TypeCellOS/BlockNote/issues/1452
 - Did not check, but unlikely to resolve: https://github.com/TypeCellOS/BlockNote/issues/1693
 - Did not check, but unlikely to resolve: https://github.com/TypeCellOS/BlockNote/issues/1694
 -  Don't know what is happening here: https://github.com/TypeCellOS/BlockNote/issues/1788
 
 ### What is left?

 - [ ] We should add some tests around this to make sure that copy & paste results in the correct document (rather than just testing clipboard serialization)
 - [ ] There is some work to do when it comes to drag & drop behavior (i.e. when to cut & paste)

### DIFF
--- a/packages/core/src/api/clipboard/fromClipboard/acceptedMIMETypes.ts
+++ b/packages/core/src/api/clipboard/fromClipboard/acceptedMIMETypes.ts
@@ -1,6 +1,6 @@
 export const acceptedMIMETypes = [
   "vscode-editor-data",
-  "blocknote/html",
+  "blocknote/json",
   "text/markdown",
   "text/html",
   "text/plain",
--- a/packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
+++ b/packages/core/src/api/clipboard/fromClipboard/pasteExtension.ts
@@ -14,6 +14,8 @@ import {
 import { acceptedMIMETypes } from "./acceptedMIMETypes.js";
 import { handleFileInsertion } from "./handleFileInsertion.js";
 import { handleVSCodePaste } from "./handleVSCodePaste.js";
+import { Slice } from "prosemirror-model";
+import { getPmSchema } from "../../pmUtil.js";
 
 function defaultPasteHandler({
   event,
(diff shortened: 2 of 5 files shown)


YousefED:
@nperez0111 following up from our discussion yesterday. What I missed is that the existing serializer is also used for `blocksToFullHTML`; which is used (for scenarios like server-side rendering) to get the HTML structure of a document that's the same (as much as possible) as what's rendered in the editor.

I think the workarounds you pointed to are mostly in place for that, not for the clipboard handling.

Also, in the preview environment I can also not move the PDF block (#1205).

The idea (json on clipboard) makes sense of course, but as far as I can see now it doesn't resolve the open issues. This makes me think it becomes more of an architectural decision which we should explore when we get to streamlining the exporters / rendering - wdyt?

nperez0111:
Yep, that's fine. I renamed the PR to reflect that.

PR #1794, opened 2025-06-26
fix: `bn-inline-content` class name getting duplicated

### LINKED ISSUE #1727: Custom Blocks - Duplicate Class Bug
When typing inside a custom inline block, the div containing the text adds a bn-inline-content class on keydown, creating many repeated bn-inline-content classes.

**To Reproduce**
Go to the [Alert Block](https://www.blocknotejs.org/examples/custom-schema/alert-block-full-ux) example page and type within the alert block. Open the inspector and observe many bn-inline-content classes being added to the wrapper.

Image

matthewlipski:
This PR fixes the `bn-inline-content` class name being added to React custom block's inline content element whenever its content changes.

Closes #1727

### DIFF
--- a/packages/react/src/schema/ReactBlockSpec.tsx
+++ b/packages/react/src/schema/ReactBlockSpec.tsx
@@ -189,7 +189,10 @@ export function createReactBlockSpec<
                   editor={editor as any}
                   contentRef={(element) => {
                     ref(element);
-                    if (element) {
+                    if (
+                      element &&
+                      !element.classList.contains("bn-inline-content")
+                    ) {
                       element.className = mergeCSSClasses(
                         "bn-inline-content",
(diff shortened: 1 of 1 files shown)


packages/react/src/schema/ReactBlockSpec.tsx
```diff
@@ -189,7 +189,10 @@ export function createReactBlockSpec<
                   editor={editor as any}
                   contentRef={(element) => {
                     ref(element);
-                    if (element) {
+                    if (
+                      element &&
+                      !element.classList.contains("bn-inline-content")
+                    ) {
```
nperez0111: Shouldn't this be handled by the `mergeCSSClasses` fn?

PR #1798, opened 2025-06-30
fix: Duplicate `formatConversionTest` files

matthewlipski:
This PR removes the duplicate `formatConversionTest` files, and only preserves the one in the `@shared` directory. This is still not ideal as really, all the unit tests should be within `/tests/`, but that is a pretty big refactor that imo is for another time The `xl-tests-refactor` branch has an early WIP for that, when I was checking how much work it is.


tests/src/unit/shared/formatConversion/formatConversionTestUtil.ts
```diff
@@ -1,199 +1,5 @@
...
-    addIdsToBlocks(block.children);
-  }
-}
-
-export function addIdsToBlocks(blocks: PartialBlock<any, any, any>[]) {
-  for (const block of blocks) {
-    addIdsToBlock(block);
-  }
-}
+export * from "@shared/formatConversionTestUtil.js";
```
YousefED: let's for now remove this file altogether and directly reference `@shared/formatConversionTestUtil`. This makes the codebase consistent and less "in limbo" state.

If we want to prioritize moving all unit tests, we should discuss this with 3 of us. Atm I'm not immediately convinced; afaik we decided on moving the formatting tests to a shared directory, because they shared similar logic - but we didn't settle on a rule to "have all unit tests" in `/tests`

YousefED (changes requested):

PR #1799, opened 2025-06-30
feat: Input rule scoping

### LINKED ISSUE #1710: Typing 1. or + and hitting space changes block type.

### LINKED ISSUE #1789: I can't add numbered headings
**Describe the bug**
My title si converted into a numbered block
**https://github.com/user-attachments/assets/ec99e959-a821-480d-a483-1c88988062a0**

**To Reproduce**
1. Create a heading like "My heading"
2. Try to edit it as "1. My heading"
3. The heading block is converted to a numbered list item

**Misc**
- Node version: tested on blocknotejs.org
- Package manager:
- Browser: Version 138.0.4 (64 bits)

matthewlipski:
This PR adds an editor option which allows for scoping on input rules, i.e. commands which change the block type after a string is typed ("# " for headings, "- " for bullet lists, etc). The editor option, `inputRules`, can be set to `"allBlocks"`, `"paragraphs"`, or `"none"` (defaults to `"allBlocks"`). This makes it so that the input rules can be triggered in all blocks, paragraphs only, or never respectively.

I think until we have a proper input rule/keyboard shortcut API, this is the approach that makes most sense.

Closes #1729 
Closes #1710
Closes #1789

### DIFF
--- a/packages/core/src/blocks/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
+++ b/packages/core/src/blocks/ListItemBlockContent/BulletListItemBlockContent/BulletListItemBlockContent.ts
@@ -31,7 +31,8 @@ const BulletListItemBlockContent = createStronglyTypedTiptapNode({
           const blockInfo = getBlockInfoFromSelection(state);
           if (
             !blockInfo.isBlockContainer ||
-            blockInfo.blockContent.node.type.spec.content !== "inline*"
+            blockInfo.blockContent.node.type.spec.content !== "inline*" ||
+            blockInfo.blockNoteType === "heading"
           ) {
             return;
           }


nperez0111:
I'm not sure about this...

It makes all input rules an all or nothing thing. I think what we should aim for is to mimic whatever notion does right now (which I suspect is similar behavior to inputRule: paragraph in this PR like this pr too: https://github.com/TypeCellOS/BlockNote/pull/1729).

The thing here is that as a user it is doing something that the user does not expect, in a heading, it should not change to a list item just because you typed `1. XYZ` since that is a normal heading name.

So ideally, we'd minimize configuration, because I don't see why anyone would configure this. This is similar to the sidemenu configuration, a stop gap that does not really solve the problem, I'd like to avoid adding more of that into the project.

matthewlipski:
Yeah that's kind of the weird thing, is that Notion's behaviour is the same as what we have now, i.e. hitting "- " at the start of a heading block does change it into a bullet list. I also thought it would have the same behaviour as `inputRules: "paragraphs"` in this PR, but seems like what people are requesting is different to what Notion does. That's why I decided to make it configurable.

I do agree though that it's a stopgap measure and that we ideally we would just let the consumer configure input rules and keyboard shortcuts for each block, but we don't currently have an API for that.

nperez0111:
You are right that notion does change a heading to a list item, but it does not change a `1. Heading` to a numbered list. So there are some more rules involved here. I'll say this much, I've never seen a "real heading" start with `- Heading`, but I've seen plenty with `1. Heading`, so I think what notion is doing is right here.

I still stand by not making this configurable and instead just encode "better rules" rather than making this configurable and offload the work of doing the right thing to the developer of the app.

Also separately, when I undo after an input rule is executed my selection moves to the wrong spot:


https://github.com/user-attachments/assets/0f1c331c-def5-43cc-8fb4-dc37176f4cab

matthewlipski:
You're right, after doing some further testing it seems like Notion literally just makes an exception for the numbered list input rules ("1. " and other numbers) in headings, so I've removed the editor options and changed the behaviour to that.

nperez0111:
This looks good to me, though I'm also going to make the change for list items, given that we allow custom blocks, I think that it can be in conflict if another custom block implements a list type functionality & is replaced by a list item.

PR #1800, opened 2025-06-30
docs: AI playground styling fixes

matthewlipski:
This PR scopes the Mantine styles used for the model settings panel in the AI playground demo, as they were causing issues with the editor styling. Additionally, the panel now correctly uses dark mode when the editor does.

Since it's a similar fix, this PR also makes the playground Mantine components (AppShell, sidebar) also correctly use dark mode depending on system theme like the editor. Just a small QoL change.


YousefED (approved):
awesome, nice you fixed darkmode as well

PR #1801, opened 2025-06-30
feat: FloatingUI options prop for `BlockPositioner`

matthewlipski:
This PR adds an `options` prop to `BlockPositioner`, which lets you pass in FloatingUI options that get passed on to `useUIElementPositioning`.

This was requested by a client, but also the `BlockPositioner` is generally useful for custom UI elements and this gives it more flexibility.


packages/xl-ai/src/components/AIMenu/BlockPositioner.tsx
```diff
@@ -11,6 +18,9 @@ export const BlockPositioner = (props: {
     reason: OpenChangeReason,
   ) => void;
   canDismissViaOutsidePress?: boolean;
+  options?: Partial<
```
YousefED: let's call this `floatingOptions` inline with other parts of the codebase

PR #1820, opened 2025-07-07
fix: Improve React Re-render Loops

justsml:
Adjusted state structurally to allow for simpler useMemo/useCallback usage.

### DIFF
--- a/packages/react/src/editor/BlockNoteView.tsx
+++ b/packages/react/src/editor/BlockNoteView.tsx
@@ -152,7 +152,26 @@ function BlockNoteViewComponent<
 
   // We set defaultUIProps and editorProps on a different context, the BlockNoteViewContext.
   // This BlockNoteViewContext is used to render the editor and the default UI.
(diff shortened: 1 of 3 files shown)


nperez0111 (approved):
This all looks correct to me, thanks for the PR. Did you run into a react re-rendering issue that needs this?

justsml:
> This all looks correct to me, thanks for the PR. Did you run into a react re-rendering issue that needs this?

Yes. 😺 

This is the 1st of (hopefully) a couple of targeted PRs to address excess re-render issues.

Excess = 10-100's or re-renders a second.

My goal is to improve the performance story around Custom toolbar UX, animating editor height (effect), programmatic select, custom slash UX, etc.

nperez0111:
Hm, 10s or 100s sounds a bit excessive, I'm not sure that we see that in our test harness so maybe there is something in your blocknote integration that is overly rendering.

Either way, would happily accept more PRs for performance updates

Issue #1825, opened 2025-07-08
Invalid regular expression in Safari < 16.4


gbasto:
**Describe the bug**
Lookbehind regular expressions are being used to bold and code markdown text ([detectMarkdown.ts](https://github.com/TypeCellOS/BlockNote/blob/906a4559623a59e8f177c5af745021e94e657ae4/packages/core/src/api/parsers/markdown/detectMarkdown.ts#L5-L11)).
Lookbehind is only supported from 16.4 above ([caniuse](https://caniuse.com/js-regexp-lookbehind)), making any page that uses BlockNote unable to be used by Safari browsers bellow 16.4.

**To Reproduce**
Open any page with BlockNote v0.33.0 in a Safari prior 16.0 and you get a blank page with a console error:
`Invalid regular expression: invalid group specifier name`
<clear steps to reproduce are super helpful! Best is to provide an online sandbox, [click to create one](https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/01-basic/01-minimal?file=App.tsx)>

**Misc**
This was also an [issue in Tiptap](https://github.com/ueberdosis/tiptap/issues/5889) before, fixed in the meantime.
The issue #1760 is related with this, but it was misunderstood that the problem was related with named capture groups. The error is related with lookbehinds but Safari < 16.4 misinterprets them with a named capture groups hence the error message.

- Node version:
- Package manager:
- Browser: Safari < 16.4
- [ ] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

nperez0111:
Would be happy to accept a PR that resolves this.

gbasto:
> Would be happy to accept a PR that resolves this.

Sure! Adding first unit test in #1826. I'll follow-up with a PR to update the regular expressions.

gbasto:
Fixed in #1827.

PR #1826, opened 2025-07-08
Add unit test for isMarkdown

gbasto:
As part of #1825, let's add unit tests to this function to make sure following changes to the regular expressions don't break the current logic.


nperez0111:
I just merged the main one, thanks for this though!

PR #1827, opened 2025-07-08
No lookbehind

gbasto:
Solves the issue when running BlockNote on Safari version prior to 16.4. See #1825 for more details.
Unit tests in #1826 guarantee regex compatibility.


nperez0111 (approved):
Looks good to me, thanks for the PR!

PR #1830, opened 2025-07-09
fix: re-evaluate side-menu on scroll

nperez0111:
This attempts to resolve a bug where scrolling with a mouse over an element will keep the block side menu attached to the previous block rather than re-evaluating to the new hovered block.

This makes it more inline with normal hover behavior

### DIFF
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -192,6 +192,12 @@ export class SideMenuView<
       true,
     );
 
+    // Update state when the editor scrolls
+    this.pmView.root.addEventListener(
+      "scroll",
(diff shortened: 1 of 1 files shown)


packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -192,6 +192,12 @@ export class SideMenuView<
       true,
     );
 
+    // Update state when the editor scrolls
+    this.pmView.root.addEventListener(
```
matthewlipski: Makes sense, though we already attach an existing `onScroll` handler a few lines below - wouldn't it make more sense to call `this.updateStateFromMousePos` there instead of creating an additional listener?

nperez0111: I didn't see that one! Very true

PR #1845, opened 2025-07-11
fix: resolve several drag & drop issues

nperez0111:
This adjusts the logic for drag & drop (especially between editor instances) to handle things better. It is surprisingly complicated to get drag & drop to work nicely across editor instances.

The problem is that each editor instance is listening for events on it's own HTMLElement, but we want to support _dragging near_ an element, and allowing the drop to occur anyway (despite not actually dropping on the element, just near it). To do this, we have to fiddle with the event handling (since we still want for each editor instance to only listen for drag & drop events on it's own element) such that we create synthetic events to trick the editor into doing what the user intended, but did not actually do.

We need to fire these synthetic events under very specific circumstances:
 - When the drag-over or drop event is close to an editor element (within a range, right now 250px), but not within an editor

But, now that we are sometimes triggering these synthetic events, Prosemirror does not have the full-scope of what we've implemented so when dragging & dropping across different editor instances we need to either:
 - When the editor is the drop point, but not the drag point, we need to collapse the selection
   - When the content is dragged outside of editor, but should be dropped into the editor, we need to fire a synthetic event
 - When the editor is the drag point, but not the drop point, we need to delete the selection

This leads to some very complex logic & several edge cases. For example, there are differences between dragging text vs. using the side menu handle.

~~I also want to fix the drag previews in this PR~~ (This seems to not be possible in firefox)

- Closes https://github.com/TypeCellOS/BlockNote/issues/1653
- Closes https://github.com/TypeCellOS/BlockNote/issues/1610
- Closes https://github.com/TypeCellOS/BlockNote/issues/1851

## TODO
 - [x] I want to revisit the purpose of the sideMenuDetection, now that it is not used for the drag & drop of content

### DIFF
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -440,14 +440,7 @@ export class SideMenuView<
     const sideMenuIsBeingDragged = Boolean(this.isDragOrigin);
     // Tells us that the current editor instance has a drag ongoing (either text or side menu)
     const isDragOrigin = textContentIsBeingDragged || sideMenuIsBeingDragged;
-    // Tells us which editor instance the drag event is happening in (or null if it's not in any editor like when dragging on another part of the page)
-    const eventEditorParentElement =
-      event.target instanceof Node
-        ? (event.target instanceof HTMLElement
-            ? event.target
-            : (event.target as any)
-          )?.closest(".bn-editor") || null
-        : null;
+
     // Tells us which editor instance is the closest to the drag event (whether or not it is actually reasonably close)
     const closestEditor = this.findClosestEditorElement(event);
 
@@ -464,7 +457,7 @@ export class SideMenuView<
     const isDropPoint = closestEditor.element === this.pmView.dom;
     // We check if the current editor instance is the same as the editor instance that the drag event is happening within
     const isDropWithinEditorBounds =
-      isDropPoint && this.pmView.dom === eventEditorParentElement;
+      isDropPoint && closestEditor.distance === 0;
 
     // We never want to handle drop events that are not related to us
     if (!isDropPoint && !isDragOrigin) {
@@ -620,9 +613,9 @@ export class SideMenuView<
 
   private dispatchSyntheticEvent(event: DragEvent) {
     const evt = new Event(event.type as "dragover", event) as any;
-    const dropPointBoundingBox = this.pmView.dom
-      .querySelector(".bn-block-group")!
-      .getBoundingClientRect();
+    const dropPointBoundingBox = (
+      this.pmView.dom.firstChild as HTMLElement
+    ).getBoundingClientRect();
     evt.clientX = event.clientX;
     evt.clientY = event.clientY;
 


packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -411,29 +323,242 @@ export class SideMenuView<
...
+        ? (event.target instanceof HTMLElement
+            ? event.target
+            : (event.target as any)
+          )?.closest(".bn-editor") || null
+        : null;
+    // Tells us which editor instance is the closest to the drag event (whether or not it is actually reasonably close)
+    const closestEditor = this.findClosestEditorElement(event);
+
+    // We arbitrarily decide how far is "too far" from the closest editor to be considered a drop point
+    if (!closestEditor || closestEditor.distance > 250) {
```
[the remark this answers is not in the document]
nperez0111: @matthewlipski not sure what the best thing to do here would be. Likely a percentage like you are using elsewhere in the file?

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -488,11 +613,9 @@ export class SideMenuView<
...
-  private createSyntheticEvent(event: DragEvent) {
-    const evt = new Event(event.type, event) as any;
-    const editorBoundingBox = (
-      this.pmView.dom.firstChild as HTMLElement
-    ).getBoundingClientRect();
+  private dispatchSyntheticEvent(event: DragEvent) {
+    const evt = new Event(event.type as "dragover", event) as any;
+    const editorBoundingBox = this.pmView.dom.getBoundingClientRect();
     evt.clientX = event.clientX;
     evt.clientY = event.clientY;
```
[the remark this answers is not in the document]
nperez0111: I tried to refactor this, but there is something finicky with drag events that they drop the data transfer

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -411,29 +323,242 @@ export class SideMenuView<
...
+   * and is trying to handle dragover events for all editors.
+   *
+   * It specifically is trying to handle the following cases:
+   *  - If the dragover event is within the bounds of any editor, then it does nothing
+   *  - If the dragover event is outside the bounds of any editor, but close enough (within 250px) to the closest editor,
+   *    then it dispatches a synthetic dragover event to the closest editor (which will trigger the drop-cursor to be shown on that editor)
+   *  - If the dragover event is outside the bounds of the current editor, then it will dispatch a synthetic dragleave event to the current editor
+   *    (which will trigger the drop-cursor to be removed from the current editor)
+   *
+   * The synthetic event is a necessary evil because we do not control prosemirror-dropcursor to be able to show the drop-cursor within the range we want
```
YousefED, line 389: would it be a lot easier if we fork dropcursor?

normally I'd not suggest this, but given we already have an [alternative](packages/xl-multi-column/src/extensions/DropCursor/MultiColumnDropCursorPlugin.ts) dropcursor plugin for mult-column, it might be a cleaner solution?

packages/core/package.json
```diff
@@ -93,7 +93,7 @@
     "@tiptap/pm": "^2.12.0",
     "emoji-mart": "^5.6.0",
     "hast-util-from-dom": "^5.0.1",
-    "prosemirror-dropcursor": "^1.8.1",
+    "prosemirror-dropcursor": "^1.8.2",
```
YousefED, line 96: we'll probably need to update our multi-column dropcursor as well (see comment below)

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -279,95 +279,6 @@ export class SideMenuView<
...
-        // handle the event.
-        setTimeout(
-          () => this.pmView.dispatch(this.pmView.state.tr.deleteSelection()),
-          0,
-        );
-      }
-    }
-
-    if (
-      this.sideMenuDetection === "editor" ||
```
YousefED, line 339: did we drop `sideMenuDetection`? then we also need to remove it from editor options / docs / etc

YousefED (changes requested):
Great!! The comments make things so much clearer :)

Some UX glitches I encountered when dragging in a single editor:

see gif, in order of appearance (I think all of these don't occur on the live version):

1) dragging to side+below an editor moves drag line to top (instead of at bottom)
2) dropping a block opens the formatting toolbar
3) dropping a block to the side doesn't seem to work

dragging

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
nperez0111: Only for the drag and drop, it is still used for the actual side menu positioning

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
nperez0111: It would be cleaner to implement it like that... I'm just not sure it is worth the effort right now.

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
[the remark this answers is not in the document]
matthewlipski: We have a similar issue with the multi-column blocks, in order to check when a block should be dropped normally vs in a new column, see here:

https://github.com/TypeCellOS/BlockNote/blob/558a71e41a94ac5e41bf44728c36f16fae63626e/packages/xl-multi-column/src/extensions/DropCursor/MultiColumnDropCursorPlugin.ts#L12

I think here it's ok to just define a constant in px honestly, doing a percentage would be weird for UX imo. The logic for dropping above/below a block when hovering it is handled by PM, so we don't have any additional constants we could reuse there.

PR #1860, opened 2025-07-19
fix: preserve leading whitespace in PDF output using nbsp

Martinrsts:
Small fix to make leading whitespaces in pdf to render correctly.

### DIFF
--- a/packages/xl-pdf-exporter/src/pdf/__snapshots__/example.jsx
+++ b/packages/xl-pdf-exporter/src/pdf/__snapshots__/example.jsx
@@ -869,7 +869,7 @@
             Styled Text
           </TEXT>
           <TEXT style={{}}>
-            {' '}
+             
           </TEXT>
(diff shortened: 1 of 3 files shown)


nperez0111:
Thanks for the PR @Martinrsts, I'm not entirely sure what situations this fix would be relevant in. Can you describe a scenario that was broken in the previous behavior that works in this new behavior?

Martinrsts:
Yes no problem. I use whitespaces at the beggining of new lines to represent indentation or align text. Without the change, if there is ' ' characters at the beginning of the line, they are all replaced by only one ' '. Here is an example:

Screenshot of the editor:
Captura de pantalla 2025-07-21 112540

Screenshot of the PDF exportation without this PR:
Captura de pantalla 2025-07-21 112705

Screenshot of the PDF exportation with this PR
 Captura de pantalla 2025-07-21 112548

PR #1877, opened 2025-07-25
Fix HTML to Block conversion of blockquotes

### LINKED ISSUE #1762: tryParseMarkdownToBlocks parses blockquote as empty with content moved outside
Using editor.tryParseMarkdownToBlocks, when parsing a Markdown blockquote like:

> Indented content
the parser creates an empty blockquote block, and the "Indented content" text is moved to the next line as a separate paragraph outside the blockquote.

This issue also occurs in your official BlockNote demo — you can try the same input there to reproduce the problem.

Expected: the entire line should be parsed as a blockquote containing "Indented content".

Actual: empty blockquote plus separate paragraph with content outside.

Environment: @blocknote/core@0.31.2, React + BlockNoteView.

Please advise if there is a fix or workaround. Happy to provide minimal repro if needed.

bgreenlee:
Fixes #1762


nperez0111:
@bgreenlee thanks for this, this is great. I tried to push some changes to your branch, but it seems like your branch is not allowed to be modified. So I just merged it straight onto our branch.

Thanks again!

Issue #1882, opened 2025-07-25
Alt text not preserved when converting a Markdown image to a Block


bgreenlee:
**Describe the bug**

Given the following Markdown:

```
![Sample image](https://picsum.photos/200)
```

The "Sample image" is the alt text and should be preserved as such. However the Block HTML for this looks like:

```
<img class="bn-visual-media" src="https://picsum.photos/200" alt="BlockNote image" contenteditable="false" draggable="false">
```

**To Reproduce**

This can be reproduced in the [BlockNote Playground](https://blocknote-main.vercel.app/interoperability/converting-blocks-from-md)

PR #1883, opened 2025-07-26
Use image alt text for image block name

bgreenlee:
This fixes the issue (#1882) where Markdown images that are converted to blocks and back again lose their alt text, having it replaced with the default alt text ("BlockNote image").

An alternative fix would be to use the alt text as the caption, but given that the caption is rendered as a separate element, that seemed less appropriate.


nperez0111 (approved):
This is great @bgreenlee Thanks for the PR

PR #1885, opened 2025-07-28
Remove extra blank lines between list items in markdown export

### LINKED ISSUE #1881: Converting ordered and unordered lists to Markdown adds newlines.
**Describe the bug**

Given the following Markdown:

```
Ordered list:
1. one
2. two
3. three

Unordered list:
* one
* two
* three

Checklist:
* [ ] one
* [x] two
* [ ] three
```

Converting this to blocks with `tryParseMarkdownToBocks` and then back to Markdown with `blocksToMarkdownLossy` produces:

```
Ordered list:

1. one

2. two

3. three

Unordered list:

* one

* two
 
* three

Checklist:

* [ ] one
* [x] two
* [ ] three
```

(Note the checklist is rendered to Markdown correctly.)

While this does import back into BlockNote correctly, they are not valid Markdown lists.

Image

**To Reproduce**

This can be reproduced in the Playground: https://blocknote-main.vercel.app/interoperability/converting-blocks-to-md

bgreenlee:
Fixes #1881

### DIFF
--- a/packages/core/src/api/exporters/markdown/markdownExporter.ts
+++ b/packages/core/src/api/exporters/markdown/markdownExporter.ts
@@ -37,7 +37,12 @@ export function cleanHTMLToMarkdown(cleanHTMLString: string) {
     })
     .processSync(cleanHTMLString);
 
-  return markdownString.value as string;
+  let result = markdownString.value as string;
+
+  // Remove extra blank lines between list items
+  result = result.replace(/\n\n(?=\d+\.|-|\*)/g, "\n");
+
+  return result;
 }
 
 export async function blocksToMarkdown<
--- a/packages/server-util/src/context/__snapshots__/ServerBlockNoteEditor.test.ts.snap
+++ b/packages/server-util/src/context/__snapshots__/ServerBlockNoteEditor.test.ts.snap
@@ -132,7 +132,6 @@ exports[`Test ServerBlockNoteEditor > converts to and from markdown (blocksToMar
 "## **Heading&#x20;***~~2~~*
 
 Paragraph
-
 * list item
 
 ![Example](exampleURL)
--- a/tests/src/unit/core/formatConversion/export/__snapshots__/markdown/complex/misc.md
+++ b/tests/src/unit/core/formatConversion/export/__snapshots__/markdown/complex/misc.md
@@ -1,5 +1,4 @@
 ## **Heading&#x20;***~~2~~*
 
 Paragraph
-
 *
--- a/tests/src/unit/core/formatConversion/export/__snapshots__/markdown/lists/basic.md
+++ b/tests/src/unit/core/formatConversion/export/__snapshots__/markdown/lists/basic.md
@@ -1,10 +1,6 @@
 * Bullet List Item 1
-
 * Bullet List Item 2
-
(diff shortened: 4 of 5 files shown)


nperez0111 (changes requested):
I don't think that this solution is sufficient enough, you'll see that in your test cases it also strips the new line between lists (e.g. in `markdown/lists/basic.md` the bullet list to numbered list boundary).

I think there are two solutions here:
 1. Make the markdown parser smarter (we have a somewhat unusual structure of `<li><p>Text content` which I'm sure is why the parser is tripping up)
 2. Simplify the the HTML output so that the markdown parser doesn't have to be so smart (by removing that inner paragraph element & unpacking it's children into the `<li>` element)

I took a stab at the second approach and got something that worked better:

```diff
diff --git i/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts w/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
index f74757c8d..9dc25da9a 100644
--- i/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
+++ w/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
@@ -172,6 +172,14 @@ function serializeBlock<
       fragment.append(list);
     }
     const li = doc.createElement("li");
+    //unpack p nodes into their children
+    const childNodes = Array.from(elementFragment.childNodes);
+    for (const child of childNodes) {
+      if (child.nodeName === "P") {
+        const children = Array.from(child.childNodes);
+        child.replaceWith(...children);
+      }
+    }
     li.append(elementFragment);
     fragment.lastChild!.appendChild(li);
   } else {
diff --git i/packages/core/src/api/exporters/markdown/markdownExporter.ts w/packages/core/src/api/exporters/markdown/markdownExporter.ts
index 812f20ffb..ce82f56f9 100644
--- i/packages/core/src/api/exporters/markdown/markdownExporter.ts
+++ w/packages/core/src/api/exporters/markdown/markdownExporter.ts
@@ -25,6 +25,8 @@ export function cleanHTMLToMarkdown(cleanHTMLString: string) {
     );
   }
 
+  console.log("cleanHTMLString", cleanHTMLString);
+
   const markdownString = deps.unified
     .unified()
     .use(deps.rehypeParse.default, { fragment: true })
@@ -37,10 +39,7 @@ export function cleanHTMLToMarkdown(cleanHTMLString: string) {
     })
     .processSync(cleanHTMLString);
 
-  let result = markdownString.value as string;
-
-  // Remove extra blank lines between list items
-  result = result.replace(/\n\n(?=\d+\.|-|\*)/g, "\n");
+  const result = markdownString.value as string;
 
   return result;
 }
```

Before going with this solution though, we'd need to double check that things like background colors still work for the HTML output, I don't remember at this moment whether they are set on the paragraph or higher. But, if we can be sure that the output isn't affected by this, then unpacking paragraphs may end up being the better solution since it is more HTML-like & easier to parse

bgreenlee:
Make sense. Let me know if I can help.

nperez0111:
Yea, @bgreenlee feel free to go further with the diff that I have above, as long as we can be sure that it doesn't affect things like background color, then I think we can merge it in

PR #1887, opened 2025-07-28
docs: add analytics back to website

nperez0111:


YousefED:
I see 1 request per page nav to vercel analytics, so that seems ok.

I see multiple requests to Sentry though (either 2 or 3 requests per nav), there might be a bug there?

nperez0111:
Will address separately if it is an issue, but it looks about right to me

PR #1888, opened 2025-07-28
docs: Fixes after fumadocs migration

matthewlipski:


nperez0111 (approved):
Thanks, this is great

PR #1892, opened 2025-07-29
fix: Overflow on table blocks

### LINKED ISSUE #1737: Scroll appears in table when content becomes too long
**Describe the bug**
Scroll appears in table when content becomes too long and pushes the table to be as wide as the editor

**To Reproduce**

1. Open homepage playground or stackblitz sandbox
2. Create a new empty table
3. Type (or paste) some very long text in one of the cells until it makes the table at least as wide as the editor and the text overflows on a second line inside its cell

Image

Image

This is also broken the same way in our application that uses Blocknote.

**Misc**

- Node version:
- Package manager:
- Browser: Chrome latest

matthewlipski:
Closes #1737

### DIFF
--- a/packages/core/src/editor/editor.css
+++ b/packages/core/src/editor/editor.css
@@ -148,19 +148,17 @@ Tippy popups that are appended to document.body directly
   transition: all 0.2s;
 }
 
-/* .tableWrapper {
-  padding
-} */
-
 .bn-editor [data-content-type="table"] .tableWrapper {
+  /* Size of + buttons to add rows/columns. */
+  --bn-table-widget-size: 22px;
   overflow-y: hidden;
-  padding-bottom: 22px;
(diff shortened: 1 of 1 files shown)


packages/core/src/editor/editor.css
```diff
@@ -153,20 +153,14 @@ Tippy popups that are appended to document.body directly
...
 
 .bn-editor [data-content-type="table"] .tableWrapper {
-  position: relative;
-  top: -16px;
-  left: -16px;
-  /* padding: 16px; */
-  min-width: calc(100% + 16px);
-  padding-bottom: 16px;
   overflow-y: hidden;
+  padding-bottom: 22px;
```
nperez0111: What is 22px referring to? Maybe add a comment for it

PR #1895, opened 2025-07-30
docs: Add docs for props without default values

### LINKED ISSUE #1765: Invalid type for `previewWidth` property on image blocks
**Describe the bug**
Since the `0.31.1` release, the default value of `previewWidth` on image blocks is `undefined`:

Image

Problem is, the type of this property is still `number`, where is should be `number | undefined`.

**To Reproduce**
N/A

**Misc**

N/A

matthewlipski:
Closes #1765


nperez0111 (approved):
Yep, this feels like the right thing here.

Let's see if this merges with your other fumadocs changes since I think you changed that file there too

matthewlipski:
This PR merges into that one, so it's no problem

PR #1898, opened 2025-07-31
feat: call AI streamtools manually

YousefED:
This PR exposes low-level functionality to apply operations (add / update / delete blocks) suggested by an LLM to the editor manually. 

This makes it possible to completely offload LLM invocation to the client application. The client application can then (based on an LLM response), manually invoke the applicable StreamTools.

Demo @ https://blocknote-git-feature-call-streamtools-manually-typecell.vercel.app/ai/manual-execution (view [example source](https://github.com/TypeCellOS/BlockNote/blob/feature/call-streamtools-manually/examples/09-ai/05-manual-execution/src/App.tsx))

_Future work_
Note that this will not integrate with the BlockNote AI UI components (UI toolbar) and is a very low-level way where the consumer will have full responsibility of setting up the LLM call (including prompts), and BlockNote only handles the "diffing" + display of document updates.

We'll also work on a more higher level abstraction that will make it easier to integrate BlockNote AI with custom backends. The proposed abstraction levels / integration points:
- _Lowest level_: invoke streamtools directly to apply updates to an editor
- _Mid-level_: an API that allows you to hook up your own backend / LLM integration to customize LLM requests (TBD)
- _Highest level_: current version of BlockNote AI: BlockNote handles everything including calling the LLM (which is called either directly or via a Proxy)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3303,6 +3303,64 @@ importers:
         specifier: ^5.3.4
         version: 5.4.15(@types/node@22.15.2)(lightningcss@1.30.1)(terser@5.43.1)
 
+  examples/09-ai/05-manual-execution:
+    dependencies:
+      '@ai-sdk/groq':
+        specifier: ^1.2.9
+        version: 1.2.9(zod@3.25.76)
+      '@blocknote/ariakit':
+        specifier: latest
+        version: link:../../../packages/ariakit
+      '@blocknote/core':
+        specifier: latest
+        version: link:../../../packages/core
+      '@blocknote/mantine':
+        specifier: latest
+        version: link:../../../packages/mantine
+      '@blocknote/react':
+        specifier: latest
+        version: link:../../../packages/react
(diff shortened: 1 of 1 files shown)


packages/xl-ai/src/streamTool/StreamToolExecutor.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+    await writer.close();
+  }
+
+  /**
+   * Accepts a single chunk and processes it using the same logic.
+   *
+   * (alternative to writing to the writable stream using {@link writable})
+   */
+  async executeOne(chunk: StreamToolCall<T>): Promise<void> {
```
nperez0111, line 137: Should be noted that running this once, will close the stream, so you can't call it in a loop or anything (since the .execute method closes the writer once it is done).

imo, this is a bit unexpected, I'd expect to be able to apply multiple executeOne calls, but I see how this is overlapping with just a writer.write.

PR #1899, opened 2025-07-31
fix: Suggestion menu closing when clicking scroll bar

### LINKED ISSUE #1130: Unable to scroll on suggestion menu controller.
I scroll on the homepage demo editor, im using brave browser, i can't scroll, once i click to scroll, it gets hidded/destroyed.

NOTE: i tryed to fix it/debug, but have no idea why this happens, i tryed to remove the onblur event 'OnCustomHandle' attached to the .bn-editor and it works, but i don't know how to keep it that way. tryed to mess with index.js which has this event, and no luck. reinstalled everything, no luck.

here's the code i run in the brave browser console to fix... but doesn't last...
export default function removeBlur() {
const editorElement = document.querySelector('.bn-editor');
if (!editorElement); // Ensure the element exists
const listeners = getEventListeners(editorElement).blur; 
console.log('REMOVED', listeners[0].listener)// Get blur listeners 
editorElement.removeEventListener('blur', listeners[0].listener); // Remove second listener
}

i afraid adding this could result in more bugs for other browsers... i would really appreaciate a response. i just can't change blocknotejs with any other editor, including, tiptap, lexical, yoopto... wethever it is. blocknotejs is the future.

**Misc**
- Node version: v21.6.2
- Package manager: npm
- Browser: brave

### LINKED ISSUE #1833: SuggestionMenu collapses when you try to use scrollbar with your mouse
**Describe the bug**

When you click scrollbar in suggestion menu it will close instantly.

https://github.com/user-attachments/assets/4c8f8954-99c1-4091-bb8f-60f21698c532

**To Reproduce**
https://www.blocknotejs.org/examples/basic/minimal

**Misc**

- Node version: 22
- Package manager: pnpm
- Browser: firefox, safari
- It appears the issue exists for some time now, at least since 0.30 version
- There is also the similar issue with trying to click heading in the suggestion menu: #1043

matthewlipski:
Closes #1130
Closes #1833

### DIFF
--- a/packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
+++ b/packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
@@ -147,7 +147,13 @@ export function SuggestionMenuController<
   }
 
   return (
-    <div ref={ref} style={style} {...getFloatingProps()}>
+    <div
+      ref={ref}
+      style={style}
+      {...getFloatingProps()}
+      // Prevents editor blurring when clicking the scroll bar.
+      onMouseDown={(e) => e.preventDefault()}
+    >
       <SuggestionMenuWrapper
         query={state.query}
         closeMenu={callbacks.closeMenu}


nperez0111 (approved):
I also ran into this with table cell buttons. I wonder if we just need to have better control of when the editor blurs. It feels like a mess to have to constantly capture these mouse events on every element related to the editor. I wonder of another approach here

YousefED:
> I also ran into this with table cell buttons. I wonder if we just need to have better control of when the editor blurs. It feels like a mess to have to constantly capture these mouse events on every element related to the editor. I wonder of another approach here

Good point! maybe add to the UI elements refactor ticket?

PR #1878, opened 2025-07-25
docs: website update

YousefED:
- add license + sla
- update home / about page to make it reflect better the current state of the project
- add items to pricing FAQ
- move pages to /legal

Open items:
- possible switch to GPL for XL packages (@YousefED )
- fix font colors of purple "buy now" button on pricing page and links in text (i.e.: links to commercial license in the benefit list) (@matthewlipski )

PR #1901, opened 2025-07-31
docs: Fixed pricing page colors 

matthewlipski:
Fixes bullet point outlined in #1878


YousefED:
Thanks. @matthewlipski, on the homepage there's also a link to pricing page (under `Is BlockNote really free?`). Can you add that as well?

PR #1903, opened 2025-07-31
feat: custom AI executors

YousefED:
Instead of the current solution where BlockNote makes direct calls to LLMs from the client using the Vercel AI SDK (or via a proxy), you can now define custom executors and route calls via your own backend.

On your backend, you can then invoke your LLM (potentially adding more context, tools, etc). While a bit more work to set up, this architecture is more inline with the Vercel SDK:
- clients do a POST request to the backend
- the backend calls `streamObject`, `generateObject`, `streamText` or `generateText`
- the responses are then passed to the client as a "datastream"

I think this will be a significant architectural improvement, but consider this a first draft / RFC

- Example which calls backend via POST request ([source](https://github.com/TypeCellOS/BlockNote/tree/feature/ai-executors/examples/09-ai/06-server-execution/src/App.tsx#L46))
- Example backend implementation (only `streamObject` and `generateObject` are implemented for now) ([source](https://github.com/TypeCellOS/BlockNote/tree/feature/ai-executors/packages/xl-ai-server/src/routes/vercelAiSdk.ts#L55))

_(at this moment the live demo doesn't work yet - it will require a deploy of the server (which is based on `main`))_

### TODOs
- [ ] Add support for tool calling + `streamText` / `generateText` instead of object generation
- [ ] Add examples to integrate with other backends (e.g.: langchain / openai directly)
- [ ] update docs + tests

### To discuss

_Messages configurability_
The current implementation still composes the `messages` array on the client (configurable via a PromptBuilder). Of course this can still be modified on the server. Alternatively, we could choose to pass all relevant information from the editor (document, cursor position, etc) to the backend, and compose the messages array there.

--> fixed in #2007

_Response handling_
The current demo (see server code) passes in the schema from BlockNote and directly responses to the client as a datastream. Real-world use-cases might also require mixing with other (app-specific) tools which would change the schema. We'll need some feedback on whether this needs better support in the API. Of course, it's always possible to construct your own schema (input) and datastream (output), but this might not be the ideal DX

--> fixed by using tool calling in #2007

### DIFF
--- a/examples/09-ai/01-minimal/src/App.tsx
+++ b/examples/09-ai/01-minimal/src/App.tsx
@@ -64,7 +64,7 @@ export default function App() {
     // Register the AI extension
     extensions: [
       createAIExtension({
-        model,
+        executor: model as any, // TODO
       }),
     ],
     // We set some initial content for demo purposes
--- a/examples/09-ai/02-playground/src/App.tsx
+++ b/examples/09-ai/02-playground/src/App.tsx
@@ -31,7 +31,6 @@ import {
 import { en as aiEn } from "@blocknote/xl-ai/locales";
 import "@blocknote/xl-ai/style.css";
 import { Fieldset, MantineProvider, Switch } from "@mantine/core";
-import { LanguageModelV1 } from "ai";
 import { useEffect, useMemo, useState } from "react";
 import { useStore } from "zustand";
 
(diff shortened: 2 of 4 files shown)


jongse7:
Really appreciate all the work on this!  
I was wondering if there’s a rough timeline for when this PR could be merged.  
I understand it’s still a draft, just asking as a user who’s interested in trying it out. :smiley_cat:

mawnir:
any example with langchain?

Issue #1658, opened 2025-05-04
Adding toExternalHTML to ReactCustomInlineContentImplementation


mbresson:
# Why

Currently, for custom block types, it is possible to implement `toExternalHTML` as part of their `ReactCustomBlockImplementation`, so that the HTML can be output to e.g. a custom tag (e.g. a `<video>` tag for a custom Video block). This is useful especially when parsing back this HTML to blocks, thanks to the `parse` function.

However, we do not have the same ability for custom inline content. Let’s say that we have a `Mention` custom inline content (as in the docs 🙂) then when BlockNoteJS exports this to HTML, it uses its `render` method, which can be inconvenient.
A good example is when you want the HTML to be parsed or reused by other tools, it would be nice, in the example of our `Mention` inline content, to be able to export it to a  `<span class="mention" user="steve"></span>`.

## Ideal solution

Example with the `Mention` inline content from the docs:

```
export const Mention = createReactInlineContentSpec(
  {
    type: "mention",
    propSchema: {
      user: {
        default: "Unknown",
      },
    },
    content: "none",
  },
  {
    render: (props) => (
      <span style={{ backgroundColor: "#8400ff33" }}>
        @{props.inlineContent.props.user}
      </span>
    ),
    toExternalHTML: (props) => (
      <span className="mention" user={props.user}></span>
    ),
  }
);
```

YousefED:
Thanks for the input - it's on our radar to review and improve this - likely this summer

nperez0111:
Inline content now supports parsing & exporting to html

Issue #1625, opened 2025-04-22
Custom inline Content parse rules


rzylber:
Is it possible to parse this HTML Element: 
```HTML
<span class="linkex" data-id="1234">My Link</span>
```
To a Custom inline content as:
{ type: "linkex", props: { ref: "1234", title: "My Link" } }
?

nperez0111:
Inline content now can specify parse rules & export to html

PR #1904, opened 2025-08-01
refactor: move all blocks to use the custom blocks API

nperez0111:
- **feat: wip**
- **feat: topological sorting, input rules and keyboard shortcuts**
- **feat: move code block over too**


packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -111,6 +117,37 @@ export function getParseRules(
 
         return props;
       },
+      getContent:
```
matthewlipski, line 70: Do you think it would be possible to return the content element in the `parse` function, instead of having a separate `parseContent`? Because I know they're split in the PM API, but I think it would be more intuitive to merge them into 1 function for the custom block API.

packages/core/src/blks/NumberedListItem/IndexingPlugin.ts
```diff
@@ -0,0 +1,175 @@
...
+        pos,
+        pos + node.nodeSize,
+        (deco: DecoSpec) =>
+          deco.index === index &&
+          deco.isFirst === isFirst &&
+          deco.hasStart === hasStart,
+      );
+
+      if (existingDecorations.length === 0) {
+        // Create a widget decoration to display the index
```
matthewlipski: Comment needs updating

packages/core/src/blocks/defaultBlockTypeGuards.ts
```diff
@@ -32,8 +32,8 @@ export function checkDefaultBlockTypeInSchema<
   S
 > {
   return (
-    blockType in editor.schema.blockSchema &&
-    editor.schema.blockSchema[blockType] === defaultBlockSchema[blockType]
+    blockType in editor.schema.blockSchema //&&
```
matthewlipski: This is just a temp change while the updated blocks are not yet in the default schema right?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -798,31 +804,99 @@ export class BlockNoteEditor<
...
+              !ext.inputRules
+            ) {
+              return undefined;
+            }
+            // "blocknote" extensions (prosemirror plugins)
+            return Extension.create({
+              name: key,
+              priority: ext.priority,
+              addProseMirrorPlugins: () => ext.plugins,
+              // TODO maybe collect all input rules from all extensions into one plugin
```
matthewlipski, line 826: I think it's more intuitive to keep logic for a block/extension within the same extension

packages/core/src/editor/BlockNoteSchema.ts
```diff
@@ -105,3 +105,6 @@ export class BlockNoteSchema<
     this.styleSchema = getStyleSchemaFromSpecs(this.styleSpecs) as any;
   }
 }
+
```
matthewlipski: TODO: Remove?

packages/core/src/schema/blocks/types.ts
```diff
@@ -21,6 +21,8 @@ export type BlockNoteDOMAttributes = Partial<{
   [DOMElement in BlockNoteDOMElement]: Record<string, string>;
 }>;
 
+// TODO we should remove FileBlockConfig, and only use BlockConfig
+// Ideally something like this would be represented via `groups: ["file"]` or similar
```
matthewlipski: Could we add a `file` field to the new `meta` property in the config?

packages/core/src/util/topo-sort.ts
```diff
@@ -0,0 +1,154 @@
+// Based on https://github.com/n1ru4l/toposort/blob/main/src/toposort.ts (MIT)
```
matthewlipski: Could you add a comment for why this file was added?

packages/core/src/blks/BulletListItem/definition.ts
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            updateBlockTr(tr, blockInfo.bnBlock.beforePos, {
+              type: "bulletListItem",
+              props: {},
+            });
+            return true;
+          }),
+      },
+      inputRules: [
```
YousefED: cool that there's a BN-style API for this. Can we do the same for keyboardshortcuts you think?

packages/core/src/schema/blocks/playground.ts
```diff
@@ -0,0 +1,204 @@
...
+   * The blocks that this block should run before.
+   * This is used to determine the order in which blocks are rendered.
+   */
+  runsBefore?: string[];
+
+  /**
+   * Advanced parsing function that controls how content within the block is parsed.
+   * This is not recommended to use, and is only useful for advanced use cases.
+   */
+  parseContent?: (options: { el: HTMLElement; schema: Schema }) => Fragment;
```
YousefED: would be nice to do this in a way that we don't expose Prosemirror APIs (i.e.: not expose Fragment) - not sure this is possible

packages/core/src/schema/blocks/playground.ts
```diff
@@ -0,0 +1,204 @@
...
+  render: (
+    /**
+     * The custom block to render
+     */
+    block: Block<Record<TName, BlockConfig<TName, TProps>>>,
+    /**
+     * The BlockNote editor instance
+     */
+    editor: BlockNoteEditor<Record<TName, BlockConfig<TName, TProps>>>,
+  ) => {
```
YousefED: These functions (dom / contentDOM / ignoreMutation) are very prosemirror specific. For now out of scope (as they're also in the existing API and this PR is about moving blocks over), but don't think this is ideal

packages/core/src/blks/BulletListItem/definition.ts
```diff
@@ -0,0 +1,100 @@
...
+      div.appendChild(el);
+
+      return {
+        dom: div,
+        contentDOM: el,
+      };
+    },
+  }),
+  () => [
+    createBlockNoteExtension({
```
YousefED: What was the reasoning again to define keyboardshortcuts / inputrules in a separate argument + `createBlockNoteExtension`?

YousefED (commented):
nice. posted some first thoughts / questions!

packages/core/src/schema/blocks/createSpec.ts
nperez0111: Yea, I thought about this too. Right now, it is pretty easy to do a mapping of HTML -> BlockNote node. But what isn't clear is how to do that mapping for parsing of inner content. We've been going through Prosemirror DOMParser & doing this low-level. I tried a couple of different approaches to do it more high-level & couldn't figure out anything satisfactory.

So, I went with the approach of a "good default", (merge paragraphs to preserve content within the parsed element), and left the rest as an exercise for later. I think what will be useful in this is when we separate out the HTML parsing & exporting from the editor instance, then maybe it will be easier for consumers to parse HTML into the shape that they need 🤷.  I marked it as an "advanced" API so that it is clear we are not happy with it. Right now the only time that any of the default blocks need to use parseContent is the list item types, because we added better handling of HTML -> list items.

packages/core/src/schema/blocks/types.ts
nperez0111: I'm not sure on this. It feels wrong to me that files are somehow different. Like why is there a file field & not a listItem field? I am not in favor of adding more properties than needed, since it adds complexity to each custom block that people can add.

I don't fully understand why this is even required at this point

packages/core/src/blks/BulletListItem/definition.ts
nperez0111: I thought about what that would look like for keyboard shortcuts, but the problem is that keyboard shortcuts can sort of just do anything. Some usecases I thought of were:
 - bringing up some menu items
 - triggering some action (e.g. stop AI generation)
 - inserting a new element
 - modifying an existing block
 
These sorts of cases are too disparate to really "unify" into a sensible API. Fundamental tension between composability & centralization.

packages/core/src/schema/blocks/playground.ts
nperez0111: https://github.com/TypeCellOS/BlockNote/pull/1904#discussion_r2250594827

Yea I couldn't find a great way to do this. Especially given that the only case that really _needs_ this are list items

packages/core/src/schema/blocks/playground.ts
nperez0111: Yep, agreed, but not looking to rock that boat right now

packages/core/src/blks/BulletListItem/definition.ts
nperez0111: This introduces keyboard shortcuts & inputrules to the extension API because we didn't have a way of doing them before.

I considered exposing them to both APIs (e.g. a custom block has an input rule API & so does the extension API), but that only added complexity.

Allowing extensions to be included to the block API is useful in that some things may require an extension to function (e.g. code block syntax highlighting, numbered list decorations).

packages/core/src/blocks/defaultBlocks.ts
```diff
@@ -3,58 +3,85 @@ import Code from "@tiptap/extension-code";
...
+  table: BlockConfig<
+    "table",
+    {
+      textColor: {
+        default: "default";
+      };
+    },
+    "table"
+  >;
+};
```
nperez0111: Table is treated differently, since it is the only one that couldn't be mapped to the custom block API

tests/src/unit/core/formatConversion/export/__snapshots__/blocknoteHTML/codeBlock/contains-newlines.html
```diff
@@ -1,9 +1,16 @@
...
+        class="bn-block-content"
+        data-content-type="codeBlock"
+        data-language="javascript"
+      >
+        <div>
+          <select></select>
+        </div>
         <pre>
-          <code class="bn-inline-content language-javascript" data-language="javascript">const hello ='world';console.log(hello);</code>
+          <code class="bn-inline-content">const hello ='world';console.log(hello);</code>
```
nperez0111: This doesn't look right to me @matthewlipski

examples/05-interoperability/05-converting-blocks-to-pdf/src/App.tsx
```diff
@@ -35,7 +36,13 @@ export default function App() {
 
   // Creates a new editor instance with some initial content.
   const editor = useCreateBlockNote({
-    schema: withMultiColumn(withPageBreak(BlockNoteSchema.create())),
+    schema: BlockNoteSchema.create().extend({
+      blockSpecs: {
+        pageBreak: createPageBreakBlockSpec(),
```
YousefED: Two things:
1. This seems inconsistent (why not `createColumnBlockSpec` or `PageBreak`) (see also my other comment)
2. `.extend` looks great. The only pro of `withMultiColumn` would be that you don't need to add both `ColumnBlock` and `ColumnListBlock`. Adding one of them would be invalid which makes it a bit weird that this is possible. Perhaps this is more the subject of a (tbd) extension API. Where 1 extension can register multiple block types it needs.

examples/06-custom-schema/react-custom-blocks/src/App.tsx
```diff
@@ -41,8 +41,8 @@ export const alertBlock = createReactBlockSpec(
       textAlignment: defaultProps.textAlignment,
       textColor: defaultProps.textColor,
       type: {
-        default: "warning",
-        values: ["warning", "error", "info", "success"],
+        default: "warning" as const,
```
YousefED: is the `as const` necessary now where before it wasn't? it's not ideal imo (the `as const` syntax is probably not understood by basic TS devs, so if we can avoid it that'd be better ofc)

packages/xl-multi-column/src/blocks/schema.ts
```diff
@@ -23,21 +23,10 @@ export const withMultiColumn = <
 >(
   schema: BlockNoteSchema<B, I, S>,
 ) => {
-  return BlockNoteSchema.create({
+  return schema.extend({
```
YousefED, line 26: very cool

examples/06-custom-schema/react-custom-blocks/src/App.tsx
nperez0111: It was necessary before too, this example just was not typed properly from the get go. So it isn't worse, it just isn't any better. TS infers an object within an object like this by default. I think we can just `as const` the whole object though so it can be simpler.

packages/xl-odt-exporter/src/odt/defaultSchema/blocks.tsx
```diff
@@ -170,9 +170,9 @@ const wrapWithLists = (
 };
 
 export const odtBlockMappingForDefaultSchema: BlockMapping<
-  DefaultBlockSchema &
-    typeof pageBreakSchema.blockSchema &
-    typeof multiColumnSchema.blockSchema,
+  DefaultBlockSchema & {
+    pageBreak: ReturnType<typeof createPageBreakBlockConfig>;
```
YousefED, line 174: minor: would be nice to have this consistent between pagebreak / multicolumn

examples/05-interoperability/05-converting-blocks-to-pdf/src/App.tsx
nperez0111: Yea, I left `withMultiColumn` as an API. But this makes it clearer what is happening with a single API. I'll probably just use the withMultiColumn again, since that is easier for consumers

packages/core/src/blocks/Paragraph/block.ts
```diff
@@ -0,0 +1,67 @@
...
+export const createParagraphBlockConfig = createBlockConfig(
+  () =>
+    ({
+      type: "paragraph" as const,
+      propSchema: defaultProps,
+      content: "inline" as const,
+      meta: {
+        isolating: false,
+      },
+    }) as const,
```
YousefED, line 19: fyi, as we now use `createBlockSpec` we use custom NodeViews for all blocks, right?

I asked Marijn a long time ago if this can have negative consequences; https://discuss.prosemirror.net/t/pros-cons-of-custom-nodeviews-vs-regular-nodespec/5999

Do we need specific update logic? what's the default update logic for non-nodeview nodes?

packages/core/src/schema/blocks/types.ts
```diff
@@ -321,3 +446,87 @@ export type PartialBlockFromConfig<
...
+        dom: HTMLElement | DocumentFragment;
+        contentDOM?: HTMLElement;
+      }
+    | undefined;
+
+  /**
+   * Parses an external HTML element into a block of this type when it returns the block props object, otherwise undefined
+   */
+  parse?: (el: HTMLElement) => Partial<Props<TProps>> | undefined;
+
```
YousefED, line 474: doesn't this mostly affect the extensions / event handlers? Not the "order they are rendered"?

packages/core/src/blocks/PageBreak/getPageBreakSlashMenuItems.ts
```diff
@@ -6,19 +6,21 @@ import {
...
+  {
+    pageBreak: ReturnType<typeof createPageBreakBlockConfig>;
+  },
+  I,
+  S
+> {
+  return "pageBreak" in editor.schema.blockSchema;
 }
 
 export function getPageBreakSlashMenuItems<
```
YousefED, line 26: future work; there should be an extension point / bundle API that automatically adds this to the slash menu, instead of relying on the consumer to manually call `getPageBreakSlashMenuItems`

packages/core/src/blocks/PageBreak/block.ts
```diff
@@ -0,0 +1,45 @@
...
+      ) {
+        return {};
+      }
+
+      return undefined;
+    },
+    render() {
+      const pageBreak = document.createElement("div");
+
+      pageBreak.className = "bn-page-break";
```
YousefED: do we need both bn-page-break and data-page-break?

packages/core/src/blocks/Quote/block.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      return {
+        dom: quote,
+        contentDOM: quote,
+      };
+    },
+  },
+  [
+    createBlockNoteExtension({
+      key: "quote-block-shortcuts",
```
YousefED, line 58: Let's think about a higher level (not transaction) specific API for the keyboard shortcuts. I see I commented this earlier, but let's discuss

packages/core/src/schema/blocks/types.ts
nperez0111: yep comment is wrong

packages/core/src/blocks/Quote/block.ts
```diff
@@ -0,0 +1,75 @@
...
+      return undefined;
+    },
+    render() {
+      const quote = document.createElement("blockquote");
+
+      return {
+        dom: quote,
+        contentDOM: quote,
+      };
+    },
```
YousefED, line 45: I'd be interested to see what we can come up with if we think about this API further. Things I think are within reach:
- decouple this further from block specs / schema
- make it easy to change / disable shortcuts / input rules

Related to this comment: https://blocknotejs.slack.com/archives/C01G3C47338/p1756719142481359?thread_ts=1756716444.502609&cid=C01G3C47338

maybe brainstorm together this week at office? good to research how the (de)coupling (between editor behavior vs schemas vs rendering) works in other editors?

packages/core/src/blocks/ListItem/NumberedListItem/IndexingPlugin.ts
```diff
@@ -0,0 +1,175 @@
...
+      );
+
+      if (existingDecorations.length === 0) {
+        // Create a widget decoration to display the index
+        decorationsToAdd.push(
+          // move in by 1 to account for the block container
+          Decoration.node(pos + 1, pos + node.nodeSize - 1, {
+            "data-index": index.toString(),
+            // TODO figure out start? is this needed?
+            // "data-start": hasStart ? index.toString() : undefined,
```
YousefED: TODO (just commenting here to make sure we don't miss it)

packages/core/src/blocks/BlockNoteSchema.ts
```diff
@@ -0,0 +1,53 @@
...
+  StyleSchemaFromSpecs,
+  StyleSpecs,
+} from "../schema/index.js";
+import {
+  defaultBlockSpecs,
+  defaultInlineContentSpecs,
+  defaultStyleSpecs,
+} from "./defaultBlocks.js";
+
+export class BlockNoteSchema<
```
YousefED, line 19: why do we have both BlockNoteSchema and CustomBlockNoteSchema?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -798,31 +779,96 @@ export class BlockNoteEditor<
...
+                                  .content === "inline*"
+                              ) {
+                                updateBlockTr(
+                                  state.tr,
+                                  blockInfo.bnBlock.beforePos,
+                                  replaceWith,
+                                  range.from,
+                                  range.to,
+                                );
+                                return undefined;
```
YousefED, line 862: agree, if we can remove the tiptap layer for this that would be preferred

packages/core/src/exporter/mapping.ts
```diff
@@ -1,4 +1,4 @@
-import { BlockNoteSchema } from "../editor/BlockNoteSchema.js";
+import { BlockNoteSchema } from "../blocks/BlockNoteSchema.js";
```
YousefED, line 1: I'm not sure `BlockNoteSchema` should be in the `blocks` directory. we should have a directory (or maybe even subpackage?) that solely contains block implementations (and not other fundamentals like `BlockNoteSchema`)

packages/core/src/schema/blocks/types.ts
```diff
@@ -21,92 +24,130 @@ export type BlockNoteDOMAttributes = Partial<{
...
+  /**
+   * The properties that the block supports
+   * @todo will be zod schema in the future
+   */
+  readonly propSchema: PS;
+  /**
+   * The content that the block supports
+   */
+  content: C;
+  // TODO: how do you represent things that have nested content?
```
YousefED, line 82: this meta information is unrelated to the schema. All other items in `BlockConfig` are related to the schema. It seems to me that `meta` is more related to what we call the block implementation?

packages/core/src/schema/blocks/types.ts
```diff
@@ -21,92 +24,130 @@ export type BlockNoteDOMAttributes = Partial<{
...
-    };
-    name: {
-      default: "";
-    };
+export interface BlockConfigMeta {
+  /**
+   * Defines which keyboard shortcut should be used to insert a hard break into the block's inline content.
+   * @default "shift+enter"
+   */
+  hardBreakShortcut?: "shift+enter" | "enter" | "none";
```
YousefED, line 32: does it make sense that this is specified on the block?

packages/core/src/blocks/defaultBlockTypeGuards.ts
```diff
@@ -1,201 +1,170 @@
...
+                type: Props[PN];
+                values?: any[];
+              };
+            }
+          >
+        : BlockConfig<BT, PropSchema>;
+  },
+  any,
+  any
 > {
```
YousefED, line 36: this function is quite lengthy. Is it a deep compare or sth more complex than that? can a standard js library provide this?

packages/core/src/schema/propTypes.ts
```diff
@@ -9,6 +9,7 @@ export type PropSpec<PType extends boolean | number | string> =
   | {
       // We infer the type of the prop from the default value
       default: PType;
+      type?: "string" | "number" | "boolean";
```
YousefED: sure this is needed? previously this was inferred from the default I think

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -111,6 +69,37 @@ export function getParseRules(
...
+          ? (node, schema) => {
+              if (customParseContentFunction) {
+                return customParseContentFunction({
+                  el: node as HTMLElement,
+                  schema,
+                });
+              }
+
+              if (config.content === "inline") {
+                // Parse the blockquote content as inline content
```
YousefED: should not be specific to blockquote

packages/react/src/components/Comments/schema.ts
```diff
@@ -1,28 +1,16 @@
...
-  ),
-  // disable default props on paragraph (such as textalignment and colors)
-  {},
-);
 
 // remove textColor, backgroundColor from styleSpecs
 const { textColor, backgroundColor, ...styleSpecs } = defaultStyleSpecs;
 
 // the schema to use for comments
 export const schema = BlockNoteSchema.create({
```
YousefED: todo?

packages/react/src/schema/ReactBlockSpec.tsx
```diff
@@ -106,182 +115,151 @@ export function BlockContentWrapper<
...
+                      }}
+                    />
+                  </BlockContentWrapper>
+                );
+              },
+              {
+                className: "bn-react-node-view-renderer",
+              },
+            )(this.props!) as ReturnType<BlockImplementation["render"]>;
+          } else {
```
YousefED, line 308: when does this get called? wouldn't this be covered by toExternalHTML or am I missing a scenario?

packages/react/src/schema/ReactBlockSpec.tsx
```diff
@@ -106,182 +115,151 @@ export function BlockContentWrapper<
...
               </BlockContentWrapper>
             );
-          },
-          {
-            className: "bn-react-node-view-renderer",
-          },
-        )(props) as NodeView;
-
-        if (blockConfig.isSelectable === false) {
-          applyNonSelectableBlockFix(nodeView, this.editor);
```
YousefED, line 210: I suppose it's ok that this is gone because it's in the core createSpec now?

packages/core/src/blocks/Paragraph/block.ts
nperez0111: I spent a good amount of time looking into this.

This is the [default implementation of a NodeView's update method](https://github.com/ProseMirror/prosemirror-view/blob/94797bca00bcb5e6a817ec3b7f73d65012a44638/src/viewdesc.ts#L990-L1005), It checks that the node type is the same as the currently rendered one - and only if it isn't, will it recreate the node view.

I think we should hold off on this until we have an actual use case, before trying to come up with a node view that does anything more than this. You would need this method for a nodeview which is carrying state that you may want to preserve (if the node is updating often). In the default blocks, this would really only be the code block extension to save a re-instantiation of the highlighter in specific situations. But, if we don't update when we should have, the nodeview can become out of sync with the document state.

So, I think we can safely hold off on this for now & come back to it when it is a problem

packages/core/src/blocks/PageBreak/getPageBreakSlashMenuItems.ts
nperez0111: Yep, there should be. Out of scope for now

packages/core/src/blocks/PageBreak/block.ts
nperez0111: I'll leave that decision to @matthewlipski

YousefED:
- Just noticed this example is broken: http://localhost:5173/custom-schema/react-custom-blocks (try changing alert type). Can you or @matthewlipski also add a test for this?
- Let's double check the React block and React ic / styles for rendering performance. In my experience this is easy to break when changing the addNodeView related code

packages/core/src/blocks/Quote/block.ts
nperez0111: I don't know that there is one. Keyboard shortcuts can do pretty much anything. There could be an easier API for doing things like adding a new block of a certain type or updating a block to a certain type. But I'd argue that is useful outside of keyboard shortcuts so it is better to leave them be top-level things (e.g. `editor.addBlock({ type: string; props: Record<string, unknown> })`)

If you have something specific you'd like to see, feel free to mention it!

packages/react/src/schema/ReactBlockSpec.tsx
nperez0111: Yep, exactly

packages/core/src/blocks/BlockNoteSchema.ts
nperez0111: `BlockNoteSchema` is a thin wrapper over `CustomBlockNoteSchema`. `CustomBlockNoteSchema` is the actual implementation & does not set any of the default blocks & requires the schema to be fully specified.

`BlockNoteSchema` will pre-load all of the default blocks too, this separation is to allow `BlockNoteSchema` to be tree-shaken off if it is not needed by someone. They can just use `CustomBlockNoteSchema` and provide everything they will actually include in their editor.

packages/core/src/exporter/mapping.ts
nperez0111: This is related to this: https://github.com/TypeCellOS/BlockNote/pull/1904#discussion_r2314010495

I want for blocks to be a sub-package, & the default implementation of `BlockNoteSchema` can be pulled from there if you want all the default blocks, but otherwise they would not be used by the editor internally (which should reference the actual implementation `CustomBlockNoteSchema`).

packages/core/src/schema/blocks/types.ts
nperez0111: Sure, it can be moved to the implementation. This was to not rock the boat on changes, but it is already breaking.

packages/core/src/schema/blocks/types.ts
nperez0111: It doesn't but again was not trying to rock the boat. I don't know what to do with such a detailed thing.

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -134,128 +123,259 @@ export function getParseRules(
...
+
+/**
+ * Helper function to create a block definition.
+ * Can accept either functions that return the required objects, or the objects directly.
+ */
+export function createBlockSpec<
+  TName extends string,
+  TProps extends PropSchema,
+  TContent extends "inline" | "none",
+  TOptions extends Record<string, any> | undefined = undefined,
```
YousefED: not sure, but can we do this with signature overloading? might make it more readable

packages/core/src/schema/blocks/internal.ts
```diff
@@ -232,51 +224,44 @@ export function createStronglyTypedTiptapNode<
 
 // This helper function helps to instantiate a blockspec with a
 // config and implementation that conform to the type of Config
-export function createInternalBlockSpec<T extends BlockConfig>(
+export function createTypedBlockSpec<T extends BlockConfig>(
```
YousefED: Is there anything in this file that you feel could simplify things? e.g. maybe the whole `StronglyTypedTipTapNode` is overengineered / code could be easier to follow if we remove it.

Curious to your thoughts as you probably went deep into it; and (even though it's not the main goal of this PR) we should be open to clearing things up

packages/core/src/blocks/ListItem/NumberedListItem/IndexingPlugin.ts
nperez0111: @matthewlipski can you look at this? I think it may need to be output on the html? This is supposed to be that a list starts at a specific index right?

matthewlipski: Not needed here - unlike `index`, numbered list items have a `start` prop which already gets rendered out to a `data-start` attribute.

packages/core/src/blocks/PageBreak/block.ts
matthewlipski: Not used in any stylesheets, so looks like it's redundant

packages/core/src/schema/blocks/internal.ts
nperez0111: I thought it was over engineered, it was easy to refactor it out though, so I did that: [`b16a9ac` (#1904)](https://github.com/TypeCellOS/BlockNote/pull/1904/commits/b16a9ac0734881f823c2c3eb861c0f9785ab81e1)

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -53,16 +54,63 @@ export function serializeInlineContentExternalHTML<
     throw new UnreachableCaseError(blockContent.type);
   }
 
-  // We call the prosemirror serializer here because it handles Marks and Inline Content nodes nicely.
-  // If we'd want to support custom serialization or externalHTML for Inline Content, we'd have to implement
-  // a custom serializer here.
-  const dom = serializer.serializeFragment(Fragment.from(nodes), options);
+  // Check if any of the nodes are custom inline content with toExternalHTML
+  const doc = options?.document ?? document;
```
YousefED, line 58: nice to see this implemented, do we have unit tests for this?

also, would this close;
https://github.com/TypeCellOS/BlockNote/issues/1625
and https://github.com/TypeCellOS/BlockNote/issues/1658
?

packages/core/src/api/exporters/html/util/serializeBlocksInternalHTML.ts
```diff
@@ -74,19 +129,14 @@ function serializeBlock<
   }
 
   const impl = editor.blockImplementations[block.type as any].implementation;
-  const ret = impl.render?.call({}, { ...block, props } as any, editor as any);
-
-  if (block.type === "numberedListItem") {
```
YousefED, line 79: was this moved somewhere / do we still support this?

packages/core/src/blocks/PageBreak/block.ts
```diff
@@ -43,3 +49,20 @@ export const createPageBreakBlockSpec = createBlockSpec(
     },
   },
 );
+
+/**
+ * Adds multi-column support to the given schema.
```
YousefED: page break

packages/react/src/schema/ReactBlockSpec.tsx
```diff
@@ -117,11 +117,74 @@ export function BlockContentWrapper<
   );
 }
 
+/**
```
YousefED, line 119: specific reason for adding this? doesn't it add a lot of extra complexity?

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
nperez0111: Does have unit tests (new test cases & snapshots added).

Yea, it'd close both. Will add to the description.

packages/core/src/schema/blocks/types.ts
YousefED: agree. Maybe mark as TODO in code, but agree it's out of scope.

packages/react/src/schema/ReactBlockSpec.tsx
nperez0111: Allows inference of the shape of the config based on the options provided. So it is useful complexity

packages/core/src/blocks/BlockNoteSchema.ts
YousefED: got it. The naming is a bit confusing maybe but I suppose that's a consequence of having `BlockNoteSchema` set up defaults. What would the ideal naming be if you could pick from scratch?

YousefED:
Seems like we're almost there! Left some small comments, let's discuss following on Monday as well before we merge:
- keyboard shortcuts API
- two items you mention in the description

packages/core/src/api/exporters/html/util/serializeBlocksInternalHTML.ts
nperez0111: It is no longer needed, since list indexes are derived from the decoration now.

PR #1908, opened 2025-08-01
fix: Formatting toolbar getting wrong bounding box when updating React inline content

### LINKED ISSUE #965: Toolbar loses its position when custom item is selected
**Describe the bug**
Toolbar loses its posiition when custom items are selected

**To Reproduce**
Select alert from dropdown
(https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/03-ui-components/03-formatting-toolbar-block-type-items?file=README.md)>

**Misc**
- Node version:
- Package manager:
- Browser: 

matthewlipski:
Closes #965


nperez0111 (approved):
a hack, but needed

PR #1911, opened 2025-08-04
refactor: editor monolith

nperez0111:
- **refactor: split out managers from the editor instance**
- **refactor: split out collaboration & extensions to their own manager**
- **refactor: move location**


packages/core/src/editor/managers/Location.ts
nperez0111, line 1: This is not yet there, but will be the basis for the location api

packages/core/src/editor/BlockNoteTipTapEditor.ts
```diff
@@ -303,9 +303,11 @@ export class BlockNoteTipTapEditor extends TiptapEditor {
     if (!element) {
       this.destroy();
       this.isInitialized = false;
+      this.emit("unmount", { editor: this });
```
matthewlipski, line 306: Is this smth relevant to splitting up the editor, or just a small fix that was added by the way?

matthewlipski (approved):
Looks good!

PR #1914, opened 2025-08-04
fix: Updated resizable file wrappers

matthewlipski:
This PR updates the component/function for the resizable file wrapper. As well as some simplifying and renaming, the main change is adding an invisible element on resize. This fixes janky behaviour that can be seen in the PDF example, and other blocks using `embed`, `iframe`, and other, similar elements. More information about what this does is commented in the code.


packages/core/src/blocks/FileBlockContent/helpers/render/createResizableFileWithCaption.ts
```diff
@@ -33,6 +33,15 @@ export const createResizableFileBlockWrapper = (
...
   rightResizeHandle.style.right = "4px";
 
+  // This element ensures `mousemove` and `mouseup` events are captured while
+  // resizing when the cursor is over the wrapper content. This is because
+  // embeds are treated as separate HTML documents, so if the content is an
+  // embed, the events will only fire within that document.
+  const eventCaptureElement = document.createElement("div");
+  eventCaptureElement.style.position = "absolute";
+  eventCaptureElement.style.height = "100%";
+  eventCaptureElement.style.width = "100%";
```
nperez0111: I think for this to work properly it depends on the parent being relative. Can we force the parent to have position relative

nperez0111 (commented):
My only question on this is why rename it?

matthewlipski:
I ended up reverting most of the changes and only keeping the fix - felt like it wasn't worth restructuring and renaming everything.

PR #1919, opened 2025-08-05
feat: More generic type guards

matthewlipski:


packages/core/src/extensions/SuggestionMenu/getDefaultSlashMenuItems.ts
```diff
@@ -87,7 +94,11 @@ export function getDefaultSlashMenuItems<
 >(editor: BlockNoteEditor<BSchema, I, S>) {
   const items: DefaultSuggestionItem[] = [];
 
-  if (checkDefaultBlockTypeInSchema("heading", editor)) {
+  if (
+    editorHasBlockWithTypeAndProps(editor, "heading", {
```
nperez0111: Can we make the props an optional param on `editorHasBlockWithType`

packages/core/src/extensions/SuggestionMenu/getDefaultSlashMenuItems.ts
```diff
@@ -315,7 +326,14 @@ export function getDefaultSlashMenuItems<
     });
   }
 
-  if (checkDefaultBlockTypeInSchema("heading", editor)) {
+  if (
+    editorHasBlockWithTypeAndProps(editor, "heading", {
+      level: defaultBlockSpecs["heading"].config.propSchema.level,
+      isToggleable: {
```
nperez0111: Let's think ahead here & try to figure out how this would work for both the current schema system & a zod-based one. I'm not sure that we should have to pass a propschema into this function.

I guess there are two types of checks we may want:

 - checking that a prop exists in the schema (i.e. does this editor even support toggle blocks)
 - checking that a prop matches a shape we expect for this function (i.e. is `isToggleable` a boolean)

How might we express that with zod?

I think in zod, whether a prop exists is easy, is there a zod validator for this key. (I also have the feeling this is the more common check).

Then to actually check if it fits what we expect, we would have to run the validator?

So maybe these checks should be de-coupled, i.e. separate checks?

packages/core/src/blocks/defaultBlockTypeGuards.ts
```diff
@@ -2,53 +2,179 @@ import { CellSelection } from "prosemirror-tables";
...
+  any
+> {
   return (
-    blockType in editor.schema.blockSchema &&
-    editor.schema.blockSchema[blockType] === blockConfig
+    editorHasBlockWithType(editor, blockType, props) && block.type === blockType
   );
 }
 
+// TODO: Only used once in the emoji picker - is it even needed? If so, should
```
matthewlipski: @nperez0111, any thoughts on this btw?

nperez0111: I would move it to just the emoji picker. If nothing else needs it right now, let's keep it simple

PR #1930, opened 2025-08-08
fix: List item HTML parsing edge cases

matthewlipski:
This PR changes `getListItemContent` to fix some edge cases, but sacrifices some UX. A better fix is possible, but would require significant changes, which should probably not be implemented before at least #1904 is merged. Full write up here:

https://linear.app/blocknote/issue/BLO-300/list-item-external-html-parsing


nperez0111:
We are going to hold off on this for now, it does not seem like the benefits out-weight the cons at the moment.

Issue #1924, opened 2025-08-07
`onChange` triggers, but `getChanges` array is empty when moving blocks


rmolinamir:
**Describe the bug**
Dragging/moving blocks in the editor triggers `onChange`, but the array returned by `getChanges` is empty. 

See attached video, notice how the changes are returned when an indentation is triggered, but the arrays are empty otherwise.

https://github.com/user-attachments/assets/98306f04-2034-40c2-b36c-3488b188c306

**To Reproduce**
1. Go to this sandbox: https://stackblitz.com/edit/github-thjchykw?file=package.json
2. Open the console
3. Drag a block
4. Notice that the array is empty

**Misc**

- Node version: v22.14.0
- Package manager: pnpm v10.12.4
- Browser: Chromium Engine Version 138.0.7204.185
- [ ] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

nperez0111:
Hm, yea re-ordering means the nodes are the same in content, but the order is not checked. So, yea this is a bug with the getChanges detection. right now [only indents are considered for changes](https://github.com/TypeCellOS/BlockNote/blob/5cd50e75d743c7e9efb76c5a60ac112b8c79cb9c/packages/core/src/api/nodeUtil.ts#L266-L281) would need to add ordering there too.

Feel free to contribute if this is important to you

rmolinamir:
> Hm, yea re-ordering means the nodes are the same in content, but the order is not checked. So, yea this is a bug with the getChanges detection. right now [only indents are considered for changes](https://github.com/TypeCellOS/BlockNote/blob/5cd50e75d743c7e9efb76c5a60ac112b8c79cb9c/packages/core/src/api/nodeUtil.ts#L266-L281) would need to add ordering there too.
> 
> Feel free to contribute if this is important to you

Thanks for pointing out the logic @nperez0111, I'll open a PR this weekend with a fix for this ❤️

PR #1931, opened 2025-08-09
fix(nodeUtil): moved blocks now included in the editor changes

rmolinamir:
# Context

#1924

## Summary

Previously, only parent changes emitted a "move" (e.g, indentations). Reordering blocks within the same parent was missed. Now, we capture each block’s sibling index and emit a "move" when either parent or index changes.

## Changes

- `nodeUtils.ts`: On top of the parent IDs, we're now keeping track of each block’s index and emitting a "move" when either parent or index changes.
- `nodeUtils.test.ts`: TODO

### DIFF
--- a/packages/core/src/api/nodeUtil.ts
+++ b/packages/core/src/api/nodeUtil.ts
@@ -260,6 +260,8 @@ export function getBlocksChangedByTransaction<
       });
     });
 
+  const draggedBlockId = transaction.getMeta("draggedBlockId");
+
   // Handle updated, moved, indented, outdented blocks
   Object.keys(nextBlocks)
     .filter((id) => id in prevBlocks)
@@ -267,9 +269,9 @@ export function getBlocksChangedByTransaction<
       const prev = prevBlocks[id];
       const next = nextBlocks[id];
       const isParentDifferent = prev.parentId !== next.parentId;
-      const isIndexDifferent = prev.index !== next.index;
+      const isDraggedBlock = prev.index !== next.index && draggedBlockId === id;
 
-      if (isParentDifferent || isIndexDifferent) {
+      if (isParentDifferent || isDraggedBlock) {
         changes.push({
           type: "move",
           block: next.block,
--- a/packages/core/src/api/nodeUtil.ts
+++ b/packages/core/src/api/nodeUtil.ts
@@ -184,6 +184,7 @@ function collectAllBlocks<
   {
     block: Block<BSchema, ISchema, SSchema>;
     parentId: string | undefined;
+    index: number;
   }
 > {
   const blocks: Record<
@@ -191,15 +192,17 @@ function collectAllBlocks<
     {
       block: Block<BSchema, ISchema, SSchema>;
       parentId: string | undefined;
+      index: number;
     }
   > = {};
   const pmSchema = getPmSchema(doc);
-  doc.descendants((node, pos) => {
+  doc.descendants((node, pos, _, index) => {
     if (isNodeBlock(node)) {
       const parentId = getParentBlockId(doc, pos);
       blocks[node.attrs.id] = {
         block: nodeToBlock(node, pmSchema),
         parentId,
+        index,
       };
     }
     return true;
@@ -264,8 +267,9 @@ export function getBlocksChangedByTransaction<
       const prev = prevBlocks[id];
       const next = nextBlocks[id];
       const isParentDifferent = prev.parentId !== next.parentId;
+      const isIndexDifferent = prev.index !== next.index;
 
-      if (isParentDifferent) {
+      if (isParentDifferent || isIndexDifferent) {
         changes.push({
           type: "move",
           block: next.block,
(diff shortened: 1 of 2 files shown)


packages/core/src/api/nodeUtil.ts
```diff
@@ -264,8 +267,9 @@ export function getBlocksChangedByTransaction<
       const prev = prevBlocks[id];
       const next = nextBlocks[id];
       const isParentDifferent = prev.parentId !== next.parentId;
+      const isIndexDifferent = prev.index !== next.index;
```
rmolinamir: Hey @nperez0111 👋 

This solution tracks the previous index and registers a change if it changes. This results in every affected block being included in the changes (not just the one that was dragged), see the attached video for reference.

LMK if you like this approach. If it's good, I'll update the snapshots.

https://github.com/user-attachments/assets/34d47545-210a-49b0-9e39-9d3877ab6906

nperez0111: Hm, ideally the only block that would be considered "moved" would be the block which actually changed. I'm sure that will be a little more difficult to figure out, but imo, it is unexpected that other blocks between would be considered moved

rmolinamir: Yeah it's too much noise. I'm thinking at most we'd include the moved blocks plus any blocks that might've a new parent. Detecting the moved block with just indexes won't do it though because when rotating arrays we might detect multiple "moved" elements (e.g. when switching places with another item, both technically moved so it's hard to detect which one moved). Any ideas? I'm looking to see if it's possible to mark them as "dragged" by the time we determine changes.

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -706,6 +709,23 @@ export class SideMenuProsemirrorPlugin<
...
+            return false;
+          },
+        },
+        filterTransaction: (tr) => {
+          if (tr.getMeta("uiEvent") === "drop" && this.draggedBlockId) {
+            tr.setMeta("draggedBlockId", this.draggedBlockId);
+            this.draggedBlockId = undefined;
+          }
+          return true;
+        },
```
rmolinamir, line 728: Ended up attaching the `draggedBlockId` to the transaction, I'm unsure if this is the ideal approach though 😅 

https://github.com/user-attachments/assets/e37787fb-08d7-4b6b-8ff8-993b847f5112

nperez0111: No, I don't think this is the right approach. What should happen here is using the order of the blocks only, derive which block has actually moved

rmolinamir: That's what I meant [here](https://github.com/TypeCellOS/BlockNote/pull/1931#discussion_r2264969798). It's not possible with just indexes.

Say you have an editor with these values: `[1,2,3]`. If you drag `2` to the first slot, you end up with `[2,1,3]`, but the user could also have dragged `1` into the second slot ending up with `[2,1,3]`, this is why it's impossible to do it with indexes alone. Knowing which block was dragged is necessary if you want to filter the rest of the blocks.

nperez0111: Hm, yea. This is a good argument for over reporting on all of the indexes that have changed.

I don't want for dragged blocks to be handled specially in some sort of way, since that will mean having to maintain every possible way the editor may change (to get the most "fine-grained" update), when this should be a general solution.

Sorry to make you do more work here. I didn't see your comment since it was marked as resolved.

rmolinamir: No worries at all, that's fair. In that case, would you rather go with the original approach (i.e., only check for blocks with different indexes)?

nperez0111: @rmolinamir yea, that might be the best way forward here.

Sorry, I've been on holiday & didn't answer this. Did you mean to close this PR?

rmolinamir: Hey man!

Hope the holidays were good, glad to see you back. 

I was working on a different approach and closed it, but didn't get a chance to push, sorry!

The new approach isn't leveraging indexes; instead, it's checking the siblings of the blocks and comparing the current siblings against the previous siblings. If any of the siblings are different, then I consider that a moved block.

LMK what you think of that approach, if you like it I'll work on it this weekend.

nperez0111: That sounds like it might just work!

nperez0111: @rmolinamir I had a go at this with https://github.com/TypeCellOS/BlockNote/pull/1960 and I think I've got something which should work well enough. Appreciate you bring it up & working on this

rmolinamir: Thanks for the good work! I'll keep an eye out for anything else and contribute. Love this library ❤️

PR #1942, opened 2025-08-14
feat(core,docs): add AccessibleImageBlock (alt=) and wire into homepa…

Ovgodd:
**Title**

Add AccessibleImageBlock for decorative images (alt="", role="presentation", aria-hidden, tabindex)

**Why**
Avoiding screen reader to read `alt`,
[RGAA 1.2 criterion](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/)

**Description**

This PR introduces a reusable opt‑in image block, AccessibleImageBlock, exported from @blocknote/core. It renders images as decorative by default (alt="") and sets role="presentation", aria-hidden="true", and tabindex="-1". The homepage demo is updated to use this block via a schema override.

**Details**

- Added AccessibleImageBlock to core:
- File: packages/core/src/blocks/ImageBlockContent/AccessibleImageBlock.ts
- Exported via packages/core/src/index.ts
- Implementation wraps imageBlockConfig with a custom render that:
- Sets alt="" (decorative image per WCAG 1.1.1)
- Sets role="presentation" and aria-hidden="true" so assistive tech ignores the image
- Sets tabindex="-1" so the image isn’t focusable
- Reuses existing imageParse and imageToExternalHTML

Demo integration:
docs/app/(home)/hero/DemoEditor.tsx overrides the schema to use AccessibleImageBlock for the image block type

Usage : 

Override the image block in your schema:
```
import { BlockNoteSchema, defaultBlockSpecs, AccessibleImageBlock } from '@blocknote/core';

  blockSpecs: {
    ...defaultBlockSpecs,
    image: AccessibleImageBlock, // alt="" by default
  },
});
```
React:
```
import { useCreateBlockNote } from '@blocknote/react';
const editor = useCreateBlockNote({ schema });
```
Vanilla:
React:
```
import { useCreateBlockNote } from '@blocknote/core';
const editor = useCreateBlockNote({ schema });
```

### DIFF
--- a/docs/app/(home)/hero/DemoEditor.tsx
+++ b/docs/app/(home)/hero/DemoEditor.tsx
@@ -1,8 +1,10 @@
 import {
   BlockNoteSchema,
   combineByGroup,
+  defaultBlockSpecs,
   filterSuggestionItems,
   uploadToTmpFilesDotOrg_DEV_ONLY,
+  AccessibleImageBlock,
 } from "@blocknote/core";
 import * as locales from "@blocknote/core/locales";
 import "@blocknote/core/fonts/inter.css";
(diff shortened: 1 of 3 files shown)


nperez0111:
@Ovgodd, any reason that we would not just apply this to the image block instead of creating a separate block for this?

I would think that these are pretty sane defaults for an image block, in the future we can add optional props to make these configurable

PR #1947, opened 2025-08-14
fix: Table column widths not being set in exported HTML

### LINKED ISSUE #1574: blocksToFullHTML and blocksToHTMLLossy produces broken tables
**Describe the bug**
The lack of `tableWrapper`, `tableWrapper-inner`, and inline styling (such as `<table style="--default-cell-min-width: 120px; min-width: 494px;">`) in the output HTML results in unstyled tables, which are entirely different from what we have in the editor.

**To Reproduce**
1. Create any table. 

Image

2. Output it using blocksToFullHTML or blocksToHTMLLossy 

Image

[Code](https://stackblitz.com/edit/github-qgqgumrb?file=App.tsx)

matthewlipski:
Closes #1574

TODO: Empty paragraphs should also include a line break for correct row height.

### DIFF
--- a/packages/core/src/blocks/TableBlockContent/TableBlockContent.ts
+++ b/packages/core/src/blocks/TableBlockContent/TableBlockContent.ts
@@ -34,16 +34,45 @@ export const TableBlockContent = createStronglyTypedTiptapNode({
     ];
   },
 
-  renderHTML({ HTMLAttributes }) {
-    return createDefaultBlockDOMOutputSpec(
+  renderHTML({ node, HTMLAttributes }) {
+    const domOutputSpec = createDefaultBlockDOMOutputSpec(
       this.name,
       "table",
       {
-        ...(this.options.domAttributes?.blockContent || {}),
+        ...(this.options.domAttributes?.blockContent || { fesfes: "fesfes" }),
         ...HTMLAttributes,
       },
       this.options.domAttributes?.inlineContent || {},
     );
+
+    // Need to manually add column widths.
+    const cols: HTMLTableColElement[] = [];
+    for (const tableCell of node.children[0].children) {
+      const colWidths: null | (number | undefined)[] =
+        tableCell.attrs["colwidth"];
+
+      if (colWidths) {
+        for (const colWidth of tableCell.attrs["colwidth"]) {
+          const col = document.createElement("col");
+          if (colWidth) {
+            col.style = `width: ${colWidth}px`;
+          }
+
+          cols.push(col);
+        }
+      } else {
+        cols.push(document.createElement("col"));
+      }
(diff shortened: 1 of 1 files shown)


packages/core/src/blocks/TableBlockContent/TableBlockContent.ts
```diff
@@ -34,16 +34,45 @@ export const TableBlockContent = createStronglyTypedTiptapNode({
...
 
-  renderHTML({ HTMLAttributes }) {
-    return createDefaultBlockDOMOutputSpec(
+  renderHTML({ node, HTMLAttributes }) {
+    const domOutputSpec = createDefaultBlockDOMOutputSpec(
       this.name,
       "table",
       {
-        ...(this.options.domAttributes?.blockContent || {}),
+        ...(this.options.domAttributes?.blockContent || { fesfes: "fesfes" }),
```
YousefED: what's this?

packages/core/src/blocks/TableBlockContent/TableBlockContent.ts
```diff
@@ -146,8 +175,8 @@ const TableParagraph = createStronglyTypedTiptapNode({
     ];
   },
 
-  renderHTML({ HTMLAttributes }) {
-    return ["p", HTMLAttributes, 0];
+  renderHTML({ node, HTMLAttributes }) {
+    return ["p", HTMLAttributes, node.childCount ? 0 : ["br"]];
```
nperez0111, line 177: Why a break?

packages/core/src/blocks/TableBlockContent/TableBlockContent.ts
```diff
@@ -34,16 +34,45 @@ export const TableBlockContent = createStronglyTypedTiptapNode({
...
+        }
+      } else {
+        cols.push(document.createElement("col"));
+      }
+    }
+
+    const colGroup = document.createElement("colgroup");
+    for (const col of cols) {
+      colGroup.appendChild(col);
+    }
```
nperez0111:
```suggestion
    // Need to manually add colgroup element
    const colGroup = document.createElement("colgroup");
    for (const tableCell of node.children[0].children) {
      const colWidths: null | (number | undefined)[] =
        tableCell.attrs["colwidth"];

      if (colWidths) {
        for (const colWidth of tableCell.attrs["colwidth"]) {
          const col = document.createElement("col");
          if (colWidth) {
            col.style = `width: ${colWidth}px`;
          }

          colGroup.appendChild(col);
        }
      } else {
        colGroup.appendChild(document.createElement("col"));
      }
    }
```

Just makes it a bit clearer what is happening here without the extra insertion loop

PR #1950, opened 2025-08-19
[Fix] block dragging errors between editors with multiple column extension

Hector-Zhuang:
This PR fixes an error that occurs when dragging a block from one editor to another while using the multiple column extension.

## How Has This Been Tested?
- Dragging blocks within editor configured with multiple column extension
- Dragging blocks from one editor to another

## Video

https://github.com/user-attachments/assets/d008bdd7-bebf-4e91-ba59-5db23e0885b7

## Root Cause

When a block is dragged from an editor, its ID is regenerated. After the block is dropped, the target editor attempts to delete a block with the same ID. However, since that ID does not exist in the target editor, this leads to an error.

image

image

## FIx
The fix ensures that the target editor checks whether the block exists before attempting to delete it.


Hector-Zhuang:
Also, there's a miscalculation of the cursor position, as you can see in the video—the cursor should only appear in a single editor

PR #1954, opened 2025-08-22
fix: add data attribute for correct react rendering

### LINKED ISSUE #1844: "Cursor" placed incorrectly when adding custom block
**Describe the bug**
I am using this very simple custom block:

```typescript
export const Title = createReactBlockSpec(
    {
        type: 'title',
        propSchema: {
            type: {
                default: 'type-a',
                values: ['type-a', 'type-b'],
            },
        },
        content: 'inline',
        hardBreakShortcut: 'none',
    },
    {
        render: (props) => {
            return (
                <div className="title-block" data-title-type={props.block.props.type}>
                    <div className="inline-content" ref={props.contentRef}/>
                </div>
            );
        },
    },
);
```

The CSS is something like this:

```css
.title-block {
    text-align: center;
} 
```

When the block is added, the cursor is blinking on the left edge of the added block div (hardly visible in the image). 

Image

As soon as I start typing the cursor moves to the center (as per `text-align: center`) and everything looks OK. 

Image

If I inspect the DOM however, the following is created:

```html
<div class="title-block" data-title-type="title">
    dfvvdfvdf
    <div class="bn-inline-content inline-content"></div>
</div>
```

The text is not added to the `.inline-content` element that holds the `contentRef` but before that node. Also the document content is not updated and the block has an empty `content` property. And therefore the UI also considers the block **empty** replacing it when you try to add a block with the `+` button on said custom block.

It works however, if I do not start typing immediately after adding the block:

* add custom block
* click anywhere into another block (if there's no other block than this does not work)
* click back into the newly added custom block
* start typing
* DOM looks good.

**To Reproduce**

* add custom block
* start typing after the block has been added

The issue is reproducible here: https://stackblitz.com/edit/github-scugpd2y?file=App.tsx

**Misc**

- Node version: v24.4.0
- Package manager: npm 10.9.1 
- Browser: Google Chrome
- React: 18.3.1

### LINKED ISSUE #1897: copy and paste issue on custom block note
copy and paste issue on custom block note

We are using BlockNote in our React application and have created a custom "callout" block in the editor. While the block renders and functions correctly initially, we encounter a problem when copying and pasting it.

After pasting, the editor renders the wrapper and its text content separately. Additionally, when we attempt to edit the pasted content, the text unexpectedly moves back into the parent frame!

We reproduced this behavior using your official example for custom blocks:
https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/06-custom-schema/01-alert-block?file=src%2FApp.tsx

The issue persists across both version 0.33.0 and 0.35.0.
Would you kindly review this and let us know if it's a bug or if there's a workaround?

---------------------------------------------------------------------------------------------------------------------------
```
import {
  insertOrUpdateBlock,
  BlockNoteEditor,
  BlockSchema,
  defaultProps,
} from "@blocknote/core";
import { createReactBlockSpec } from "@blocknote/react";
import { LightbulbIcon } from "lucide-react";

export const CalloutBlockSpec = createReactBlockSpec(
  {
    type: "callout",
    propSchema: {
      backgroundColor: defaultProps.backgroundColor,
    },
    content: "inline",
  },
  {
    render: ({ block, contentRef }) => {
      return (
        <div
          className="bg-gray-50 border-[1px] min-h-[68px] border-e123 rounded-[8px] flex items-center gap-3 px-4"
          data-block-id={block.id}
          data-node-type="callout"
        >
          <LightbulbIcon
            className="h-5 w-5 flex-shrink-0 text-gray-500"
            aria-hidden="true"
          />

          <p
            className="text-sm font-normal text-gray-600 inline-content"
            ref={contentRef}
          />
        </div>
      );
    },
  },
);
export default function customCalloutBlock<BSchema extends BlockSchema>(
  editor: BlockNoteEditor<BSchema>,
) {
  return {
    title: "Callout",
    onItemClick: () => {
      insertOrUpdateBlock(editor, {
        type: "callout",
        props: {
          textAlignment: "left",
          textColor: "gray-600",
        } as any,
      });
    },
    aliases: ["callout"],
    group: "Section 1",
    key: "bi_callout",
    icon: <LightbulbIcon size={18} />,
  };
}

```

- Node version: {
   "@blocknote/core": "^0.33.0",
    "@blocknote/mantine": "^0.33.0",
    "@blocknote/react": "^0.33.0",
    "react": "^19.0.0",
    "react-dom": "^19.0.0"
}
- Package manager: pnpm
- Browser: all
- 
 I would appreciate if you could look into this sooner than later 💖

nperez0111:
This was caused by a change to Tiptap in a more recent version: https://github.com/ueberdosis/tiptap/pull/6600 which specifically looks for an attribute which Tiptap expects on an element to exist to be considered as part of the contentDOM.

BlockNote does not use the `NodeViewContent` React component (instead using the BlockNote `contentRef`). So, to address this issue, we now explicitly set the data-attribute which Tiptap's React integration expects on the element.

This element is handed off to ProseMirror anyway so React should not blow it away on render.

While we are at it, updated all `@tiptap/*` packages to their latest versions. These are likely the last versions before we switch to v3. We may want to consider always pulling the latest tiptap versions to catch these sorts of issues sooner

Fixes #1844
Fixes #1897
Resolves Linear Ticket: https://linear.app/blocknote/issue/BLO-301/react-blocks-broken-tiptap-2250


MetaMmodern:
Thank you!

alibiyabani:
thanks bro

Issue #1938, opened 2025-08-12
Video disappears when parsing markdown back to blocks after page refresh


iyansr:
## Bug Description
When uploading a video to the editor, it renders correctly and exports to markdown as expected. However, when refreshing the page and parsing the same markdown back to blocks, the video disappears and only the thumbnail image remains.

## Steps to Reproduce
1. Upload a video file to the BlockNote editor
2. Export content using `editor.blocksToMarkdownLossy(editor.document)`
3. Refresh the page
4. Parse the markdown back using `editor.tryParseMarkdownToBlocks(content)`
5. Replace blocks with `editor.replaceBlocks(editor.document, blocks)`

## Expected Behavior
The video should be preserved and rendered correctly after parsing markdown back to blocks, maintaining the same functionality as before the page refresh.

## Actual Behavior
The video disappears and only the thumbnail image is displayed. The markdown link reference to the video URL is lost during the parsing process.

## Code Samples

**Export (works correctly):**
```javascript
const markdown = await editor.blocksToMarkdownLossy(editor.document)
```

**Import (loses video):**
```javascript
const blocks = await editor.tryParseMarkdownToBlocks(content);
editor.replaceBlocks(editor.document, blocks);
```

## Markdown Output Comparison

**After initial upload (correct):**
```markdown
asdasd

[](https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-576p.mp4)![1lpV49uSTaxnGNRCY99wXkQ.png](https://xxxxxxx.cloudfront.net/d52a08cb-1ede-44a1-bb57-4eb72bc329b0/files/7ea09d03-4273-467e-ab03-b86d8bb18c1d.png)
```

<img width="822" height="930" alt="Image" src="https://github.com/user-attachments/assets/015ce948-cd98-4b1e-b0c2-06146f07a90f" />

**After page refresh and re-import (incorrect):**
```markdown
asdasd

![BlockNote image](https://xxxxx.cloudfront.net/d52a08cb-1ede-44a1-bb57-4eb72bc329b0/files/7ea09d03-4273-467e-ab03-b86d8bb18c1d.png)
```

<img width="894" height="576" alt="Image" src="https://github.com/user-attachments/assets/7e338b9d-ed58-4795-b357-d41db0801176" />

## Environment
- BlockNote version: `^0.34.0`
- Browser: Arc Browser
- OS: Mac OS Sonoma

anggapur:
[...]

PR #1955, opened 2025-08-22
Fix: video parsing and Markdown export issues

Hector-Zhuang:
## Description:
This PR addresses two issues related to `<video>` elements in the editor:

### Video parsed as an image
Previously, `<video>` elements were being processed by the image handler in `remarkRehype`, causing videos to be treated as images.

### Video exported as a link in Markdown
When exporting content containing videos to Markdown, the videos were incorrectly serialized as standard link Markdown:

[](https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-576p.mp4)

## Fixes implemented:

Updated the image handler to detect videos by extension (mp4, webm, etc.) and process them as `<video>` elements with the correct attributes: src, data-url, data-name and so on.

Added a `convertVideoElementToMarkdown` plugin to transform `<video>` elements into a dedicated video markdown with url and title.

#1938


iyansr:
@matthewlipski @nperez0111 @YousefED 
Any possibilities to merge this ? We really need

Hector-Zhuang:
> @matthewlipski @nperez0111 @YousefED Any possibilities to merge this ? We really need

Try to maintain or fork a repository of this project. This is what I'm doing. I've been fixing bugs while integrating this with my work

packages/core/src/api/exporters/markdown/util/convertVideoToMarkdownRehypePlugin.ts
```diff
@@ -0,0 +1,18 @@
...
+  return (tree: any) => {
+    visit(tree, "element", (node, index, parent) => {
+      if (node.tagName === "video") {
+        const src = node.properties?.src || node.properties?.["data-url"] || "";
+        const name =
+          node.properties?.title || node.properties?.["data-name"] || "";
+        parent.children[index!] = {
+          type: "text",
+          value: `![${name}](${src})`,
+        };
```
nperez0111: Based on the test cases, I don't know that this is working entirely correctly.

It is being rendered to a link here: tests/src/unit/core/formatConversion/export/__snapshots__/html/video.html which I don't fully get why. So, it makes sense why this parse would not actually pick it up.

This may have to do with the way we output internal html vs external HTML?

I think we should hold off on merging this until we change up the default block API in https://github.com/TypeCellOS/BlockNote/pull/1904

nperez0111:
@iyansr feel free to use the pkg pr new versions that were generated above. They'll include this and some unreleased changes

iyansr:
> @iyansr feel free to use the pkg pr new versions that were generated above. They'll include this and some unreleased changes

@nperez0111 I got this when i install from pkg pr 
CleanShot 2025-09-05 at 11 56 33

nperez0111:
@iyansr 
This error happens when you have multiple versions of @blocknote/core installed in the same repo. You should update all of the versions to be pkg.pr.new versions, not just one package. Since they depend on each other.

iyansr:
> @iyansr This error happens when you have multiple versions of @blocknote/core installed in the same repo. You should update all of the versions to be pkg.pr.new versions, not just one package. Since they depend on each other.

@nperez0111 Should I install all with the pkg.pr.new ? 
CleanShot 2025-09-05 at 19 05 12@2x

nperez0111:
Yep, that should be it @iyansr

iyansr:
> Yep, that should be it @iyansr

The issue still exist unfortunately 🥲

nperez0111:
You likely have duplicates somewhere, look around for duplicate versions using `npm ls @blocknote/core` or inspect your package-lock, this is your project so I can't help you with it anymore than this

iyansr:
> You likely have duplicates somewhere, look around for duplicate versions using `npm ls @blocknote/core` or inspect your package-lock, this is your project so I can't help you with it anymore than this

Sure, I will try, thanks for helping 🙏

packages/core/src/util/string.ts
```diff
@@ -13,3 +13,24 @@ export function filenameFromURL(url: string): string {
   }
   return parts[parts.length - 1];
 }
+
+export function isVideoUrl(url: string) {
```
matthewlipski, line 17: @nperez0111 how do you feel about checking if a file is a video based on a hard coded list of file endings? The alternative would be to get it from the `meta.fileBlockAccept` field in the video block's implementation. It would be nice to use this instead of a hard coded list, but I'm wary of what happens if someone replaces the default video block with their own. Wdyt?

matthewlipski (approved):

packages/core/src/util/string.ts
nperez0111: I think hard-coding is fine for now, this is what the markdown parser would've done anyway.

nperez0111 (approved):

nperez0111:
Thanks @Hector-Zhuang for this!

PR #1956, opened 2025-08-22
[CRITICAL] Fix: Infinite render loop when selecting all content (Ctrl+A / Cmd+A)

Hector-Zhuang:
## Reproduction Steps

Select all content manually, or use the shortcut Ctrl+A / Cmd+A.

The editor enters an infinite render loop and eventually crashes the page.

## Root Cause

In the `FormattingToolbarView` update event handler, the logic attempts to check the selection boundary to determine whether the formatting toolbar is rendered.
If the toolbar isn’t rendered, it forces the view to update.

When all content is selected, posToDOMRect (from Tiptap) reports a selection width of 0.
This results in the handler continuously scheduling forced updates, creating an infinite render loop.

```
// Why width = 0?
| <- Start (left = 0, right = 0)
Content Content Content Content
| <- End (left = 0, right = 0)

const width = right - left; // width = 0
```

image

#965 
#1908

### DIFF
--- a/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
+++ b/packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
@@ -206,15 +206,14 @@ export class FormattingToolbarView implements PluginView {
       // individually in each button.
       const newReferencePos = this.getSelectionBoundingBox();
 
-      // Workaround to ensure the correct reference position when rendering 
+      // Workaround to ensure the correct reference position when rendering
       // React components. Without this, e.g. updating styles on React inline
-      // content causes the formatting toolbar to be in the wrong place. We 
-      // know the component has not yet rendered if the reference position has 
+      // content causes the formatting toolbar to be in the wrong place. We
+      // know the component has not yet rendered if the reference position has
       // zero dimensions.
       if (
         newReferencePos.x === 0 ||
         newReferencePos.y === 0 ||
-        newReferencePos.width === 0 ||
         newReferencePos.height === 0
       ) {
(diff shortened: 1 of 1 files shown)


Hector-Zhuang:
@matthewlipski 

I think the update handler is firing too frequently right now, which could become a performance bottleneck. Would there be a better approach to handle this? For example, could we leverage something like `useEffect` inside `createReactBlockSpec` or other custom element APIs to trigger the format bar update in a more controlled way?

image

nperez0111 (approved):
Thanks @Hector-Zhuang this does indeed resolve the issue.

@matthewlipski can you take @Hector-Zhuang's feedback & either track it in another issue or PR. There is definitely some valid stuff in there, but I want to make a release today, so I'll take the current change as is

matthewlipski:
> @matthewlipski
> 
> I think the update handler is firing too frequently right now, which could become a performance bottleneck. Would there be a better approach to handle this? For example, could we leverage something like `useEffect` inside `createReactBlockSpec` or other custom element APIs to trigger the format bar update in a more controlled way?
> 
> image

I think you're right in that `update` runs too often, as it gets called on each transaction made to the editor, whereas we are only looking for selection changes for the formatting toolbar. However, I don't think this has a meaningful impact on performance, as checking whether the selection has changed is relatively inexpensive. More expensive methods like `shouldShow` and `getSelectionBoundingBox` are only called when necessary.

We are planning to at some point (likely within the next few months) refactor how we handle UI element state, since the current system is pretty convoluted. For that, it's definitely a good suggestion to have better React integration, with hooks to manage the state things like the formatting toolbar. I'll reference this PR in the internal ticket we have for that👍

Issue #1248, opened 2024-11-14
Using shadcn components in editor results in overridden styles 


amosbastian:
**Describe the bug**
For example, I'm creating a custom block that uses shadcn components, e.g. the `Card` component. For some reason the styles are being overridden and so it doesn't have a border. Same for e.g. the `Input` component, but there even the padding is off

**To Reproduce**
For example, using this block:

```tsx
import { createReactBlockSpec } from "@blocknote/react";
import { Card, CardContent, CardDescription, CardFooter, CardHeader, CardTitle } from "~/components/ui/card";
import { Input } from "~/components/ui/input";

export const TextPromptBlock = createReactBlockSpec(
  {
    type: "textPrompt",
    propSchema: {
      value: {
        default: "",
      },
    },
    content: "inline",
  },
  {
    render: (props) => {
      return (
        <Card>
          <CardHeader>
            <CardTitle>Card Title</CardTitle>
            <CardDescription>Card Description</CardDescription>
          </CardHeader>
          <CardContent>
            <Input
              type="text"
              placeholder="Type something..."
              onChange={(event) => {
                props.editor.updateBlock(props.block, {
                  type: "textPrompt",
                  props: { value: event.currentTarget.value },
                });
              }}
            />
          </CardContent>
          <CardFooter>Hello</CardFooter>
        </Card>
      );
    },
  },
);
```
results in this
![CleanShot 2024-11-14 at 16 13 15@2x](https://github.com/user-attachments/assets/b91243e8-c80c-4d7e-9074-ac899cb2fb6b)

**Misc**
- Node version: 20
- Package manager: pnpm
- Browser: Arc
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

YousefED:
Thanks. could you provide a stackblitz sandbox so we can easily reproduce it?

amosbastian:
> Thanks. could you provide a stackblitz sandbox so we can easily reproduce it?

https://stackblitz.com/edit/vitejs-vite-4dsvbw?file=src%2Fblock.tsx

dccarmo:
This is happening because, importing CSS from shadcn, it adds an inline style that overrides styles such as border color (see screenshot where I'm trying to set the border color and it's being overwritten by this: https://github.com/TypeCellOS/BlockNote/blob/main/packages/shadcn/src/style.css#L69).

[...]

PR #1959, opened 2025-08-25
chore: Tailwind v4 + ShadCN update

matthewlipski:
This PR updates TailwindCSS and ShadCN to the newest versions, and also changes the overall setup for the `@blocknote/shadcn` package.

Currently, we import Tailwind styles in `@blocknote/shadcn/src/style.css`. This generates CSS from all of the utility classes that Tailwind detects in the code, and that has multiple issues:

- Creates duplicate CSS rules in applications which already import Tailwind styles themselves.
- BlockNote doesn't respect the Tailwind/ShadCN theme of the parent application.
- We have to scope the BlockNote Tailwind CSS to prevent conflicts (which is why we have things like the `bn-` prefix).

In practice, this doesn't make sense. We expect `@blocknote/shadcn` to be used in an app that already uses Tailwind, and probably ShadCN. Therefore, it doesn't make sense to include Tailwind at all in the package, and should instead rely on the parent application to handle importing the Tailwind styles. This way, all of the above issues are resolved.

There is one caveat to this - Tailwind only generates CSS based on the utility classes it finds in your code, which is why we previously needed to include that generated CSS in the built files. Since the `@blocknote/shadcn` package is buried inside `node_modules`, Tailwind doesn't check the code there, and there is no way to guarantee that the parent app would use all of the utility classes that BlockNote uses to generate all the necessary CSS. However, Tailwind v4 introduced the [`@source`](https://tailwindcss.com/docs/functions-and-directives#source-directive) directive for this exact purpose, so the parent app can now point Tailwind to the code in `@blocknote/shadcn` and generate all the additional CSS needed.

So this PR removes the Tailwind dependency from the `@blocknote/shadcn` package, and the parent app is instead expected to handle it. Even if the parent app doesn't use ShadCN and only uses Tailwind, the worst case is that you have to paste some CSS from [here](https://ui.shadcn.com/docs/installation/manual#configure-styles) to set up the theme. In our case, the ShadCN example is responsible for importing Tailwind styles and setting up the ShadCN theme.


YousefED:
nice! @matthewlipski does this close any open issues?

matthewlipski:
Good point! Closes #1248

PR #1974, opened 2025-08-30
Fix documentation formatting for `foundations/document-structure` page

VoloDziu:
Remove extra code block ticks that prevent proper doc page rendering (see screenshot below)

Screenshot 2025-08-29 at 9 01 37 PM


nperez0111:
Thanks for this @VoloDziu !

PR #1975, opened 2025-09-01
fix: Suggestion menu positioning

### LINKED ISSUE #1042: Inconsistent position of menu for a block close to bottom of the viewport
**Describe the bug**
When opening the menu for a block close to the bottom of the viewport the positioning is inconsistent. 

**To Reproduce**

Reproduced in https://www.blocknotejs.org/examples/basic/minimal

Position content and editor so the empty block is close to the bottom of the viewport (but not too close)
image

Click on a plus once, the menu is positioned above the empty block:
image

Click on same plus second time, the menu is positioned below and very small:
image

**Misc**
- Node version: 22.2.0
- Package manager: yarn
- Browser: Firefox 129.0.2

### LINKED ISSUE #1291: suggestion menu opens in wrong direciton and cause page to overflow
**Describe the bug**
I have a chat app and I'm trying to use the <BlockNoteView> as a message composer. When opening the suggestion menu for the **second time** (the first time it works fine) it opens on the wrong direction (downwards) and since there is not enough space it causes the page to overflow vertically

https://github.com/user-attachments/assets/e9cd30ef-17a7-4f47-9bb3-ede9ff4c403f

**To Reproduce**
I don't know but I can attach a video

**Misc**
- Node version: 20.10
- Package manager: pm
- Browser: chrom

matthewlipski:
This PR fixes 2 issues with the positioning of suggestion menus:

- The menu does not always flip to the other side when near the bottom of the viewport.
- The menu sometimes uses an outdated reference position

Closes #1042
Closes #1291


nperez0111 (approved):
This looks good to me

PR #1981, opened 2025-09-03
fix: File block resize handles not working with touch input

### LINKED ISSUE #1035: Images can't be resized on touch screen devices
### Images can't be resized on mobile devices / touch screen devices

**To Reproduce**
Embed an image URL and try to use the lateral selector to resize the image.
I have tried this both on Chrome browser mobile view, both on my phone. I haven't tried resizing videos.

**Misc**
- Package manager: npm
- Browser: Chrome on Linux, Chrome on Android

matthewlipski:
Closes #1035

### DIFF
--- a/packages/core/src/blocks/FileBlockContent/helpers/render/createResizableFileBlockWrapper.ts
+++ b/packages/core/src/blocks/FileBlockContent/helpers/render/createResizableFileBlockWrapper.ts
@@ -73,7 +73,7 @@ export const createResizableFileBlockWrapper = (
     let newWidth: number;
 
     const clientX =
-      event instanceof TouchEvent ? event.touches[0].clientX : event.clientX;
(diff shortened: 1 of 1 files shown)


nperez0111 (approved):
Yep this is good

YousefED (commented):
@matthewlipski @nperez0111 can't we streamline this by using pointer events?

nperez0111:
I wasn't aware of this API, seems like it should do it: https://developer.mozilla.org/en-US/docs/Web/API/Element/pointermove_event

YousefED:
> @matthewlipski @nperez0111 can't we streamline this by using pointer events?

I see this was released; are we following up on the pointer events separately or was there an issue with that?

nperez0111:
@matthewlipski looked into it, and the API is different enough that it did not simplify things.

PR #1982, opened 2025-09-03
fix: doLLMRequest fails when deleting a non-existent block

Hector-Zhuang:
In some cases, which are hard to reproduce, `doLLMRequest` attempts to delete a non-existent block. This could be caused by a retry.


nperez0111 (approved):
Thanks @Hector-Zhuang this looks good to have anyway

PR #1984, opened 2025-09-04
fix: Improve setting text for custom file blocks

matthewlipski:
This PR makes some small but useful changes for how text in the add block button & file panel is changed for file blocks. It's now done entirely through the dictionary, whereas it was previously split. The file panel text was set throught the dictionary, but the add block button text was set using a `buttonText` prop. The PDF block example has also been updated with this change.

### DIFF
--- a/examples/06-custom-schema/04-pdf-file-block/src/App.tsx
+++ b/examples/06-custom-schema/04-pdf-file-block/src/App.tsx
@@ -4,6 +4,7 @@ import {
   filterSuggestionItems,
   insertOrUpdateBlock,
 } from "@blocknote/core";
+import { en } from "@blocknote/core/locales";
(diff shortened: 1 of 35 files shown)


nperez0111 (approved):
Good change!

matthewlipski:
@YousefED since the PDF block is a pro example, should we also add an example e.g. changing the text for image blocks to be “Add/Upload/Embed Photo”?

YousefED:
@matthewlipski we don't need a separate example.

Thanks for this! Note that while the PR is a good improvement, this way of changing the dictionary doesn't really scale (i.e.: you can change some strings, but not icons etc.). I don't think this is a priority now, but good to keep in mind when we revisit UI Components

PR #1985, opened 2025-09-05
fix: properly validate link content in AI JSON manipulation

Hector-Zhuang:
Currently, AI fails to handle link inline components correctly because the validation function does not iterate over the content array inside a link node.

image


nperez0111:
@Hector-Zhuang can we add a test case that proves that this fixes the intended use case?

Hector-Zhuang:
@nperez0111 I can do it. How do you use for generating these snapshots, or its by manual?

image

nperez0111:
@Hector-Zhuang I think this is where the tests should execute: https://github.com/TypeCellOS/BlockNote/blob/838a52d19175ec688007958d49dfe2fee9b762ee/packages/xl-ai/src/api/formats/json/tools/jsontools.test.ts#L79-L122

Here is an example of one of the files which would probably need to be added onto to test that the validation is properly working: https://github.com/TypeCellOS/BlockNote/blob/7e5b354bfa1c02da9eb303f77fce8fe291315918/packages/xl-ai/src/testUtil/cases/addOperationTestCases.ts#L1-L102


Thanks for this!

Hector-Zhuang:
@nperez0111 The unit test is added.

nperez0111:
@Hector-Zhuang the test doesn't pass though?

PR #1989, opened 2025-09-05
docs: Update exporter examples

matthewlipski:
This PR unifies and cleans up the code across the export/import examples, and overhauls the styling so that each of them have a clean & matching look.

### DIFF
--- a/examples/05-interoperability/05-converting-blocks-to-pdf/src/App.tsx
+++ b/examples/05-interoperability/05-converting-blocks-to-pdf/src/App.tsx
@@ -25,7 +25,7 @@ import {
   withMultiColumn,
 } from "@blocknote/xl-multi-column";
 import { pdf, PDFViewer } from "@react-pdf/renderer";
-import { JSX, useEffect, useMemo, useState } from "react";
+import { JSX, useEffect, useMemo, useReducer, useState } from "react";
 
 import "./styles.css";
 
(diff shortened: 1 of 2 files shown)


nperez0111:
I'm not sure about this part of the change:

> I've also changed the PDF and Email examples to just have a download file button

It demos well and shows the output immediately which is very useful for comparison, I'd prefer to keep it as it was.

nperez0111 (commented):
I'm not sure that I get the motivation behind this change?

nperez0111:
Sorry to do this to you again, but why is this not just the side-by-side diff? I feel like I'm missing why the change was made.

matthewlipski:
If you look at the website currently, there's not really enough space to do a side-by-side view properly in the demos. We could make them stacked vertically instead (the editor and email/pdf view) but again it's not great because of the limited space. So I think having a toggle makes the most sense. Ofc for the playground, side-by-side would be better as we have almost the whole viewport to work with.

nperez0111:
Ok, yea I get what you mean, I just worry about the discovery of it...

Is there an in-between here, where if we have the space it does a side-by-side? I'm not sure what the best thing to do is here.

matthewlipski:
Hmm I think maybe if we switch between vertically stacked and side-by-side depending on available space that could work

matthewlipski:
@nperez0111 I ended up doing the suggestion from my previous comment, and ended up overhauling the styling for all import/export examples since they were ugly as all hell. The stacked view is still really cramped for the PDF and email demos in the docs section, but everything looks a lot better now in the examples section.

examples/05-interoperability/05-converting-blocks-to-pdf/src/App.tsx
```diff
@@ -381,47 +387,78 @@ export default function App() {
...
-        </PDFViewer>
+      <div className="view-wrapper">
+        <div className="view-label">
+          PDF Output
+          <span className="view-label-download" onClick={onDownloadClick}>
+            Download
+          </span>
+        </div>
+        <div className="view">
+          <PDFViewer height={"100%"} width={"100%"}>
```
nperez0111: The key was there for an issue I had when you edit the document. The pdf renderer dies, is it not an issue anymore?

Issue #1849, opened 2025-07-16
Text disapears when composing


nadjitan:
**Describe the bug**
When the **font size** of a text is changed and then the user start **composing** (e.g. Japanese characters), the paragraph/content **disappears**.
When user goes back to english keyboard, the previously typed content with different font size reappears and the newly typed content goes to a newline.

https://github.com/user-attachments/assets/9c703c38-c021-491c-a8b7-1a439f1b2b76

This bug is related to my past issue https://github.com/TypeCellOS/BlockNote/issues/1611
`data-mark-view-*` elements might be the issue when composing a span element with font size styles

**To Reproduce**
1. Run the Blocknote repository using `pnpm start`
2. Go to `/custom-schema/react-custom-styles`

**Misc**

- Node version:
- Package manager: pnpm
- Browser: Google Chrome
- [ ] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

pltoledo:
I'm having the same issue with characters with accents (ã, á, à, etc) while changing the font family

nperez0111:
I believe this has been fixed in the latest version, feel free to let me know if it hasn't, try it out here: https://blocknote-git-main-typecell.vercel.app/custom-schema/react-custom-styles

PR #1991, opened 2025-09-08
fix: External HTML parse/export cases

matthewlipski:
## Implemented Changes

This PR adds the non-breaking changes from #1605, updating them to work with the default blocks refactor. These are:

### Styles:

- Adding text/background color style parsing from inline styles.

### Props:

- Adding text/background color prop exporting to inline styles.
- Adding text alignment prop parsing from inline styles.
- Adding text alignment prop exporting to inline styles.
- Fixing `start` prop parsing for ordered list items.
- Moving responsibility of parsing text alignment from proprietary TipTap extension to custom blocks API. This is done so that it behaves like other props, as it's currently inconsistent.

### Tests:

- Added unit tests for new text/background parsing logic.
- Updated snapshots for HTML export tests.

## Missing Changes

There are still  changes that are relevant for this PR that are either breaking or cannot be implemented atm:

- Text/background color styles cannot currently be exported to inline styles. This is because we create TipTap marks for these, and create BlockNote styles from those marks. When we did smth similar creating default blocks from TipTap nodes, this was fine as we could separate the "internal" and "external" HTML using `addNodeView` and `render`. But we can't do this for styles as `addMarkView` is not supported in TipTap v2, so we'd need to upgrade to v3. Also, if we wanted to use the custom style API instead to create the text & background color styles, this is also a problem as `toExternalHTML` is not in the API. Adding this would also require `addMarkView`. There is a hack in the React package that we use to add a mark view, but this seems to have its own issues (see #1849).
- Text/background color props cannot currently be parsed from inline styles. This is because the TipTap attributes for these props are set on the `blockContainer` node, meaning it's responsible for parsing the attributes from HTML. However, when parsing external HTML, we never actually parse anything as a `blockContainer`, so the text/background color attributes can never be parsed either. To fix this, we'd need to move the attributes to be on content nodes (`paragraph`, `heading`, etc) like text alignment currently is. The issue with this is that it changes the underlying BlockNote schema, making this a breaking change for any apps that store the TipTap/PM state underlying BlockNote (notably YDocs for collaboration).

## Next steps

- Update TipTap to v3 for mark view support, so that the text/background color styles can be exported to inline styles in "external" HTML. Ideally, we would use this to  add `toExternalHTML` to the custom styles API, and redefine text/background color using the custom styles API directly rather than porting TipTap marks.
- Figure out how to handle migrating collaborative docs to a new underlying TipTap/PM schema. This way, we can move the text/background color attributes out of `blockContainer` nodes and fix parsing them from inline styles in external HTML, without it being a breaking change.
- Advanced table cells can also have text/background colors. Since tables are still created using `createBlockSpecFromTiptapNode`, handling exporting/parsing to/from inline styles must be done differently to the other default blocks. I think this is doable, but haven't looked super far into this yet.


packages/core/src/blocks/defaultProps.ts
nperez0111, line 1: Why use a CSS variable? Can't BlockNote use the data-text-color attribute, and everything else can just get the raw color that is set?

tests/src/unit/core/formatConversion/parse/__snapshots__/html/orderedListStart.json
```diff
@@ -0,0 +1,54 @@
...
+    "content": [
+      {
+        "styles": {},
+        "text": "List Item 2",
+        "type": "text",
+      },
+    ],
+    "id": "1",
+    "props": {
+      "backgroundColor": "default",
```
nperez0111, line 13: Do we need to emit defaults? I would think that we would only set the value if it is not the default?

matthewlipski: In this case we should, as `tryParseHTMLToBlocks`/`tryParseMarkdownToBlocks` have a return type `Block[]`. So we should include all non-optional props in order for the type to be correct.

nperez0111: Ah okay

packages/core/src/blocks/defaultProps.ts
matthewlipski: Fixed this in #1998

packages/core/src/editor/defaultColors.ts
```diff
@@ -35,7 +35,7 @@ export const COLORS_DEFAULT = {
     text: "#ad1a72",
     background: "#f4dfeb",
   },
-};
+} as Record<string, { text: string; background: string }>;
```
nperez0111, line 38: Intentionally not an exact type?

matthewlipski: Yep correct I was running into an issue with that somewhere, IIRC it was parsed color strings from HTML were just regular strings and the default colors keys were a union of string literals, so there was a mismatch.

Issue #1994, opened 2025-09-10
Comment mark disappear


AntoLC:
**Describe the bug**
We have users with different roles, on the same document we can have users that can write / read comments (admin / owner / editor etc...), and users with role that are not allowed to see the comments.

Depend these roles, we init or not comments: https://github.com/TypeCellOS/BlockNote/blob/c3422c2b7fc4aa62c8bb2fb062bd0d9fee426de7/examples/07-collaboration/04-comments/src/App.tsx#L78-L80

By doing like that, if a user has the comments feature and another user does not have it, the line disappear.

https://github.com/user-attachments/assets/097eed67-65ad-42a7-acaa-4305600e2dd1

`ThreadStoreAuth` does not provide a way to block user to see comment:
https://github.com/TypeCellOS/BlockNote/blob/c3422c2b7fc4aa62c8bb2fb062bd0d9fee426de7/packages/core/src/comments/threadstore/ThreadStoreAuth.ts#L3-L13

In our case, we will not want to see the "mark" neither on the text if the user is not allowed to see comments.

**Misc**

- Browser: Any browser
- [ ] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

nperez0111:
We can add a permission for showing the comments (display only), but they would still exist in the document.

So for now, you can just always initialize the threadstore, and we will give you the ability to deny someone from seeing them within the document.

AntoLC:
We found a way to not store the comments in the ydoc, only REST way.

PR #1995, opened 2025-09-10
feat(comments): add ability to deny a user to view a comment #1994

nperez0111:
This PR introduces the ability to deny the user from viewing the comment information, the comment is still within the document, so this is not totally secure, just hides it from being viewable.

https://github.com/user-attachments/assets/f2691b94-3567-4af4-b4f0-d2a19c00a583


matthewlipski:
To add to this, could we e.g. add some CSS styles that make the comment highlight also invisible when the user isn't authenticated? Maybe add a HTML attribute to the editor or smth along those lines.

nperez0111:
I like the html attribute, unsure on what the styles would be (dimmed or hidden?). Would be a good add here @matthewlipski

matthewlipski:
I think just fully hidden/transparent, not really a point having them visible if the comments themselves aren't either imo

matthewlipski (approved):

PR #1998, opened 2025-09-11
fix: External HTML parse/export cases cont.

matthewlipski:
This PR continues the work from #1991, fixing parsing of `textColor` and `backgroundColor` props. To do this, the attributes representing these props have been moved in the underlying ProseMirror schema from `blockContainer` nodes to `blockContent` nodes (e.g. paragraphs and headings). The reasoning for this can also be found in #1991.

Because the ProseMirror has changed, existing collaborative documents lose any `textColor` and `backgroundColor` props that were set, as the ProseMirror editor attempts to read the attributes from a different node to where they actually are. Therefore, the `SchemaMigrationPlugin` has been added, which updates existing collaborative documents to the new schema. This plugin runs whenever a collaborative document is initially loaded and traverses it. When it finds content that doesn't align with the editor schema, it can dispatch transactions to correct any issues caused by the mismatch between the document & editor schemas. Each correction is a separate piece of code that targets a specific change in the schema, and is called a `MigrationRule`. Currently, there is just one `MigrationRule`, which moves the `textColor` and `backgroundColor` attributes as described above. Thanks to the `MigrationRule` design, it should be fairly easy to deal with any additional ProseMirror schema changes that we might have to make in the future.

Besides these major changes, there are also a few minor ones:

- List item HTML export has been slightly changed, so that we can actually put the color styles on the `li` elements rather than their children.
- Color styles no longer export a CSS variable and then read it, the color is just set directly to the value we'd previously assign to the variable.

### DIFF
--- a/packages/core/src/extensions/Collaboration/schemaMigration/SchemaMigrationPlugin.ts
+++ b/packages/core/src/extensions/Collaboration/schemaMigration/SchemaMigrationPlugin.ts
@@ -2,9 +2,8 @@ import { Plugin, PluginKey } from "@tiptap/pm/state";
 import { ySyncPluginKey } from "y-prosemirror";
 import * as Y from "yjs";
 
-import migrationRules from "./migrationRules/index.js";
 import { BlockNoteExtension } from "../../../editor/BlockNoteExtension.js";
-import { BlockNoteEditor } from "../../../editor/BlockNoteEditor.js";
+import migrationRules from "./migrationRules/index.js";
 
 // This plugin allows us to update collaboration YDocs whenever BlockNote's
 // underlying ProseMirror schema changes. The plugin reads the current Yjs


packages/core/src/extensions/Collaboration/schemaMigration/SchemaMigrationPlugin.ts
```diff
@@ -0,0 +1,74 @@
...
+          }
+
+          if (
+            transactions.length !== 1 ||
+            !transactions[0].getMeta(ySyncPlugin)
+          ) {
+            return undefined;
+          }
+
+          const fragment = editor.collaboration?.fragment;
```
nperez0111: I don't think we need to expose the collaboration options on the editor instance, two solutions:

1. Pass it in as a parameter (on creation of the extension)
2.Or,  Expose it on the ySyncPlugin extension

packages/core/src/extensions/Collaboration/schemaMigration/SchemaMigrationPlugin.ts
```diff
@@ -0,0 +1,74 @@
...
+        state: {
+          init: () => ({ migrationDone: false }),
+          apply: (tr, oldPluginState) => {
+            const newPluginState = tr.getMeta(pluginKey);
+            if (newPluginState) {
+              return newPluginState;
+            }
+
+            return oldPluginState;
+          },
```
nperez0111: No need to use a whole plugin state for this, just use a local variable for this extension instance, simpler

packages/core/src/extensions/Collaboration/schemaMigration/SchemaMigrationPlugin.ts
```diff
@@ -0,0 +1,74 @@
...
+import { BlockNoteExtension } from "../../../editor/BlockNoteExtension.js";
+import { BlockNoteEditor } from "../../../editor/BlockNoteEditor.js";
+
+// This plugin allows us to update collaboration YDocs whenever BlockNote's
+// underlying ProseMirror schema changes. The plugin reads the current Yjs
+// fragment and dispatches additional transactions to the ProseMirror state, in
+// case things are found in the fragment that don't adhere to the editor schema
+// and need to be fixed. These fixes are defined as `MigrationRule`s within the
+// `migrationRules` directory.
+export class SchemaMigrationPlugin extends BlockNoteExtension {
```
nperez0111, line 14: So, this handles the case where the attributes were stored in Y.js, what about if the user was using json or HTML as their storage?

packages/core/src/extensions/Collaboration/schemaMigration/SchemaMigrationPlugin.ts
```diff
@@ -0,0 +1,74 @@
...
+          if (pluginKey.getState(newState).migrationDone) {
+            return undefined;
+          }
+
+          const ySyncPlugin: Plugin<{ doc: Y.XmlFragment["doc"] }> | undefined =
+            (editor.extensions["ySyncPlugin"] as BlockNoteExtension | undefined)
+              ?.plugins[0];
+          if (!ySyncPlugin) {
+            return undefined;
+          }
```
nperez0111: This check could be eliminated by passing in options instead of trying to pull them.

packages/core/src/extensions/Collaboration/schemaMigration/SchemaMigrationPlugin.ts
matthewlipski: With JSON we don't need to change anything as the `props` field abstracts away having the prop attributes on `blockContainer` vs `blockContent` nodes.

With HTML, existing documents will indeed lose `textColor`/`backgroundColor` props when they get parsed under the new schema. However, we do explicitly state in the docs that you should use JSON for storing documents to the backend, and in general the use case of having HTML for document storage is not officially supported. So I don't think this is worth implementing. If it turns out there's demand for this then we can implement it. Unlike with the Yjs document migration, with HTML it can just be a simple function to transform the HTML string.

The more relevant HTML use case though is for static rendering, as we want old documents to still be displayed correctly with the new schema. Fortunately, the CSS changes in this PR are backwards compatible with the old schema, so this should be fine.

PR #2001, opened 2025-09-12
feat(core): support for Tiptap V3

nperez0111:
This PR is meant to add support for Tiptap V3, there are a couple of things which it depends on:

 - The `default-blocks` branch
 - This PR to Tiptap landing (can re-introduce the `BlockNoteTiptapEditor` if it does not happen quick enough): https://github.com/ueberdosis/tiptap/pull/6946
 - Along the way I found a bug with y-prosemirror that we can hopefully fix, I've tried fixing it before: https://github.com/yjs/y-prosemirror/pull/185


packages/xl-ai/src/testUtil/cases/combinedOperationsTestCases.ts
```diff
@@ -22,39 +22,39 @@ export const combinedOperationsTestCases: DocumentOperationTestCase[] = [
...
-        type: "update",
-        id: "ref2",
-        block: {
-          content: [{ type: "text", text: "Hallo", styles: {} }],
-        },
-      },
-    ],
-    getTestSelection: (editor: BlockNoteEditor<any, any, any>) => {
-      const posInfo = getNodeById("ref2", editor.prosemirrorState.doc)!;
+  // {
```
nperez0111: @YousefED, I looked into this & could not figure out what the root cause of this is. The update tool seems to be replacing the block when it should only be partially updating the block? I'm not entirely certain what is going on here, can you take a look at this for me?

MetaMmodern:
@nperez0111 does this update mean that blocknotejs should be able to support SSR now in environments like Next.js, just like tiptap?

nperez0111:
@MetaMmodern, Tiptap does not support SSR in Next.js & so this change has no effect on our SSR support, Next.js has very specific requirements to render in SSR which is fundamentally incompatible with rendering in ProseMirror which requires a DOM to function properly.

PR #2028, opened 2025-09-23
feat: Mantine v8 upgrade

matthewlipski:
This PR updates the Mantine version to v8. There are 2 notable changes with this:

1. The Mantine CSS styles have been updated. The previous `global.css` has now been split into several files, from which we only need [`baseline.css`](https://mantine.dev/styles/css-files-list/#global-styles). The styles from that file have been copied and scoped to `.bn-mantine`.
2. There is now proper support for submenus, so our previous workaround has been removed.


examples/09-ai/01-minimal/.bnexample.json
nperez0111, line 1: I think you need to update the .bnblock config file or whatever it is called. So that pnpm run gen doesn't overwrite these changes

packages/mantine/src/mantineStyles.css
```diff
@@ -82,62 +82,46 @@
...
+.bn-mantine button,
+.bn-mantine select {
+  text-transform: none;
 }
 
-.bn-mantine .mantine-focus-never:focus {
-  outline: none;
-}
+.bn-mantine {
+  /* margin: 0;
```
nperez0111, line 117: Purposefully commented out?

packages/mantine/src/menu/Menu.tsx
```diff
@@ -24,128 +15,26 @@ const SubMenuContext = createContext<
...
   assertEmpty(rest);
 
   if (sub) {
-    return <SubMenu {...props} />;
+    return (
+      <MantineMenu.Sub
+        transitionProps={{ duration: 250, exitDelay: 250 }}
+        withinPortal={false}
+        middlewares={{ flip: true, shift: true, inline: false, size: true }}
+        // For some reason, `onChange` is not supported in submenus yet
```
nperez0111: Maybe open an issue with them?

packages/mantine/src/mantineStyles.css
```diff
@@ -82,62 +82,46 @@
 @import url("@mantine/core/styles/Timeline.css");
 @import url("@mantine/core/styles/Title.css");
 @import url("@mantine/core/styles/Tooltip.css");
-@import url("@mantine/core/styles/TypographyStylesProvider.css");
+@import url("@mantine/core/styles/Typography.css");
+
+/* Mantine baseline styles, scoped to .bn-mantine element. */
+/* https://github.com/mantinedev/mantine/blob/7.17.8/packages/%40mantine/core/src/core/MantineProvider/baseline.css */
```
YousefED: the comment points to mantine v7, not sure that's correct?

packages/mantine/src/mantineStyles.css
```diff
@@ -82,62 +82,46 @@
 @import url("@mantine/core/styles/Timeline.css");
 @import url("@mantine/core/styles/Title.css");
 @import url("@mantine/core/styles/Tooltip.css");
-@import url("@mantine/core/styles/TypographyStylesProvider.css");
+@import url("@mantine/core/styles/Typography.css");
+
+/* Mantine baseline styles, scoped to .bn-mantine element. */
```
YousefED, line 87: are you sure we don't need the other styles from [global](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/core/MantineProvider/global.css).css?

matthewlipski: Yes, the `global.css` file does have some styling for focused elements, but they aren't necessary and in some cases where we use non-Mantine components (like regular divs with FloatingUI), they can look broken:

Default focus style:
Screenshot 2025-09-23 at 17 23 18

Mantine global focus style:
Screenshot 2025-09-23 at 17 23 34

The other styles in `global.css` are unused within BlockNote.

examples/09-ai/01-minimal/.bnexample.json
matthewlipski: Is it not correct as is? In the `.bnexample.json` file, I updated the Mantine package version under `dependencies`, which `pnpm gen` uses to update the dependencies in the corresponding `package.json`. Or am I misunderstanding?

packages/mantine/src/menu/Menu.tsx
matthewlipski: https://github.com/mantinedev/mantine/issues/8292

packages/mantine/src/mantineStyles.css
matthewlipski: Yes, added explanation in code

examples/09-ai/01-minimal/.bnexample.json
nperez0111: Ah maybe I missed it then

PR #2029, opened 2025-09-23
feat: Render Mantine `BlockNoteView` without `MantineProvider`

### LINKED ISSUE #1555: Making @blocknote/mantine compatible with Mantine projects
When using `@blocknote/mantine` in projects that already use Mantine, there is a conflict between variables, provider and maybe mantine versions.

`BlockNoteView` from `@blocknote/mantine` wraps the editor in `MantineProvider`. We should be able to opt-out of this, in case we're already in a Mantine codebase and have such a provider higher up.

This issue became obvious when I tried to customize Blocknote's CSS variables using mantine variables. I expected `var(--mantine-color-text)` (for example) to refer to the app's variable, but instead it refered to the local variable set up by Blocknote's Mantine provider so it didn't change the color the way I expected. This makes it hard to customize the theme using the app's mantine variables. Things like the default radius will be different from the app's theme as well.

Two additional points:
- Making mantine packages peer dependencies
- Having a CSS export that does not include Mantine CSS files, as these are usually already imported in a Mantine codebase

Unclear.

**Additional context**

**Bonus**

matthewlipski:
This PR changes the Mantine `BlockNoteView` to check if it's being rendered within an existing `MantineContext`. If it isn't, it wraps the editor in a `MantineProvider`. Otherwise, it just renders the editor.

I've changed how the Mantine active styles are reset too (using CSS instead of Mantine theme) so that in case a `MantineContext` is found, the theme set on it doesn't affect BlockNote's styles.

Closes #1555

### DIFF
--- a/docs/content/docs/getting-started/mantine.mdx
+++ b/docs/content/docs/getting-started/mantine.mdx
@@ -23,9 +23,3 @@ bun add @blocknote/core @blocknote/react @blocknote/mantine
 To use BlockNote with Mantine, you can import `BlockNoteView` from `@blocknote/mantine` and the stylesheet from `@blocknote/mantine/style.css`.
 
 <Example name="basic/minimal" />
-
-## Usage within a Mantine app
-
-By default, the [`BlockNoteView`](/docs/getting-started/editor-setup#render-the-editor) component from `@blocknote/mantine` will be wrapped in a [`MantineProvider`](https://mantine.dev/theming/mantine-provider/) context. However, if you're already using Mantine for your application UI, this context will already be provided.
-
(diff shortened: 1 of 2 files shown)


matthewlipski:
@nperez0111 any suggestions for how to name the component? I think `BlockNoteViewNoProvider` is not great but not sure what to change it with since it should still be specific in what it does imo.

nperez0111:
@matthewlipski, `BlockNoteViewRaw`, `CustomBlockNoteView` maybe?

I prefer naming things as what they are, instead of naming them as what they are not 😂

PR #2032, opened 2025-09-26
fix: disable input rules for numbered headings #1789

### LINKED ISSUE #1789: I can't add numbered headings

nperez0111:
Closes #1789


matthewlipski:
Do we want to disable the input rules for both numbered and bullet list items? Notion seems to only disable them for numbered lists.

nperez0111:
It is what we implemented here: https://github.com/TypeCellOS/BlockNote/pull/1799

PR #2039, opened 2025-09-29
feat: AI menu auto scrolling

matthewlipski:


packages/xl-ai/src/components/AIMenu/AIMenuController.tsx
nperez0111, line 1: I'm unsure that this makes sense to implement in the React UI layer rather than in the ai extension layer.

I'm also very wary of all of these effects since it is requiring React to react to state changes that we can know in the ai extension rather than having react schedule them

packages/xl-ai/src/components/AIMenu/AIMenuController.tsx
```diff
@@ -16,6 +16,82 @@ export const AIMenuController = (props: { aiMenu?: FC<AIMenuProps> }) => {
...
+    if (
+      typeof aiMenuState === "object" &&
+      "status" in aiMenuState &&
+      aiMenuState.status === "ai-writing"
+    ) {
+      setAiWriting(true);
+    } else {
+      setAiWriting(false);
+    }
+  }, [aiMenuState]);
```
nperez0111: Would prefer for this to be "derived" rather than using react effects for this

packages/xl-ai/src/AIExtension.ts
```diff
@@ -134,6 +137,51 @@ export class AIExtension extends BlockNoteExtension {
...
+        const aiMenuNonErrorState =
+          aiMenuState === "closed" ? undefined : aiMenuState;
+        if (aiMenuNonErrorState?.status === "ai-writing") {
+          const blockElement = this.editor.domElement?.querySelector(
+            `[data-node-type="blockContainer"][data-id="${aiMenuNonErrorState.blockId}"]`,
+          );
+          blockElement?.scrollIntoView({ block: "center" });
+        }
+      });
+    });
```
nperez0111, line 29: Why wrap in an `editor.onCreate`? I'm actually uncertain that `onCreate` gets invoked right now, does it? Is it on mount?

packages/xl-ai/src/AIExtension.ts
```diff
@@ -134,6 +137,51 @@ export class AIExtension extends BlockNoteExtension {
...
+        }
+
+        const aiMenuState = this._store.getState().aiMenuState;
+        const aiMenuNonErrorState =
+          aiMenuState === "closed" ? undefined : aiMenuState;
+        if (aiMenuNonErrorState?.status === "ai-writing") {
+          const blockElement = this.editor.domElement?.querySelector(
+            `[data-node-type="blockContainer"][data-id="${aiMenuNonErrorState.blockId}"]`,
+          );
+          blockElement?.scrollIntoView({ block: "center" });
```
nperez0111, line 26: I somewhat dislike using a selector for this instead of asking prosemirror for that blocks position, don't we already have utils for getting this? If not, it is common enough that we probably should

packages/xl-ai/src/AIExtension.ts
```diff
@@ -134,6 +137,51 @@ export class AIExtension extends BlockNoteExtension {
...
+      });
+    });
+
+    // Listens for `scroll` and `scrollend` events to see if a new scroll was
+    // started before an existing one ended. This is the most reliable way we
+    // have of checking if a scroll event was caused by the user and not by
+    // `scrollIntoView`, as the events are otherwise indistinguishable. If a
+    // scroll was started before an existing one finished (meaning the user has
+    // scrolled), auto scrolling is disabled.
+    document.addEventListener(
```
nperez0111: It [looks like prosemirror-view has `handleScrollToSelection` that we could then handle the scrolling ourselves](https://github.com/ProseMirror/prosemirror-view/blob/75c3c77f62bf583a00e551ca45c0aee4fcd8f4b8/src/index.ts#L240) with maybe?

It seems like it also [exports the underlying function](https://github.com/ProseMirror/prosemirror-view/blob/d9cf7a439c7e0916984139689d810f35794d0666/src/domcoords.ts#L32) that does the scrolling, so we can maybe just use that?

matthewlipski: The AI writing doesn't actually move the user selection, so `handleScrollToSelection` unfortunately isn't something we can use. `scrollRectIntoView` looks like it could work, though I need to see if it works the same as `scrollIntoView`.

nperez0111: I may not have been clear enough with this, I was expecting that we'd be calling `tr.scrollIntoView()` on each of the AI ops, rather than relying on changes or the ai state. It is just a more direct way of handling this, since it is really just a side-effect of that operation. Maybe I'm missing something for this, let's chat tmrw on it

packages/xl-ai/src/AIExtension.ts
matthewlipski: The underlying TipTap editor is undefined when the extension constructor is run, and so an error gets thrown attempting to call `tiptapEditor.on("update", ...)` (within `editor.onChange(...)`). Wrapping it in `editor.onCreate` fixes this. Maybe `onMount` would be better than `onCreate`? Any way that that we can ensure the TipTap editor is already initialized.

nperez0111: We need to fix this then, because I don't think this is a great pattern

packages/xl-ai/src/AIExtension.ts
YousefED: Interesting discussions and nice fix. Curious; what ended up to be the reason that `tr.scrollIntoView` wasn't suitable?

nperez0111: `scrollIntoView` is based on the current selection. In some cases though, running the AI does not actually update the selection (e.g. when running at the end of a block, and it decides to make a new one).

That, and it gives us less flexibility on where to actually position the scroll (e.g. start, center, end).

PR #2044, opened 2025-09-29
feat(core): introduce the `Location`s API for pin-pointed positioning

nperez0111:
This is still WIP, but the bones of the `Location` API are here.

We have the ability to convert a prosemirror position into a location, and a location into a prosemirror position. There are several new primitives with varying levels of specificity, most of the changes can be introduced in a backwards-compatible way, since we were already allowing blocknote ids or blocknote blocks in various places (`BlockIdentifier`), it has been expanded to support `Point`s & `Range`s now

 - [ ] add docs
 - [ ] see other places where we can use locations

Closes #1699 #2195 #2309 #2411


packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -1,28 +1,27 @@
...
 
 export function getSelection<
   BSchema extends BlockSchema,
   I extends InlineContentSchema,
   S extends StyleSchema,
 >(tr: Transaction): Selection<BSchema, I, S> | undefined {
-  const pmSchema = getPmSchema(tr);
   // Return undefined if the selection is collapsed or a node is selected.
   if (tr.selection.empty || "node" in tr.selection) {
     return undefined;
```
nperez0111, line 27: Why is this like this @matthewlipski? I find it strange that just because my selection isn't over a range, that I don't get back what the current selection is on. Can I change this?

packages/core/src/locations/types.ts
```diff
@@ -0,0 +1,55 @@
...
+   * @note an offset of -1 is used to indicate that this point actually represents the whole block
+   */
+  offset: number;
+};
+
+/**
+ * A range is a pair of points, it is used to identify a range of blocks within a document.
+ */
+export type Range = {
+  anchor: Point;
```
YousefED, line 27: the term anchor / head comes from prosemirror selections, where the direction of a selection can matter.

For locations, I don't think direction matters, right? In that case, would start / from be better?

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
```diff
@@ -20,25 +20,17 @@ export function insertBlocks<
 >(
   tr: Transaction,
   blocksToInsert: PartialBlock<BSchema, I, S>[],
-  referenceBlock: BlockIdentifier,
+  referenceBlock: Location,
```
YousefED, line 23: If `referenceBlock` would be a Point or Range, what's the intended behavior?

(same question for replaceBlocks / updateBlock etc)

packages/core/src/editor/managers/BlockManager.ts
```diff
@@ -171,7 +178,7 @@ export class BlockManager<
    * @param update A partial block which defines how the existing block should be changed.
    */
   public updateBlock(
-    blockToUpdate: BlockIdentifier,
+    blockToUpdate: Location,
```
YousefED, line 181: should this take a Location?

packages/core/src/editor/managers/BlockManager.ts
```diff
@@ -155,7 +162,7 @@ export class BlockManager<
    */
   public insertBlocks(
     blocksToInsert: PartialBlock<BSchema, ISchema, SSchema>[],
-    referenceBlock: BlockIdentifier,
+    referenceBlock: Location,
```
YousefED, line 165: should probably either:
- Not take a location
- rename referenceBlock to `location` or `referenceLocation`

packages/core/src/editor/managers/SelectionManager.ts
```diff
@@ -40,23 +39,12 @@ export class SelectionManager<
...
-    return this.editor.transact((tr) => getSelectionCutBlocks(tr));
-  }
-
   /**
    * Sets the selection to a range of blocks.
    * @param startBlock The identifier of the block that should be the start of the selection.
    * @param endBlock The identifier of the block that should be the end of the selection.
    */
-  public setSelection(startBlock: BlockIdentifier, endBlock: BlockIdentifier) {
+  public setSelection(startBlock: Location, endBlock: Location) {
```
YousefED, line 47: rename params + jsdoc

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -127,19 +129,84 @@ export function getSelection<
...
+      return [
+        nodeToBlock(cutStartNode),
+        ...blocks.slice(1, -1),
+        nodeToBlock(cutEndNode),
+      ];
+    },
   };
 }
 
 export function setSelection(
```
YousefED, line 203: this function only works for blocks right? not points / ranges?

packages/core/src/locations/types.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * A point is a path with an offset, it is used to identify a specific position within a block.
+ */
+export type Point = {
+  id: BlockId;
+  /**
+   * The number of characters from the start of the block.
+   * @note an offset of -1 is used to indicate that this point actually represents the whole block
+   */
+  offset: number;
```
YousefED, line 20: Are there positions that we can't represent using the current Location / Point concept?

One thing I'm thinking of atm is the end of a table cell vs the start of the next table cell, but there might be more cases

YousefED (commented):
Cool! I think this can work out great - but we should think through naming and APIs before we rush this out.

(--> Added some first comments that should clarify what I mean by that)

packages/core/src/locations/types.ts
nperez0111: I went with it like this to allow the direction to matter when it needs to, but be agnostic in most other cases. My only fear with making it be `from` & `to` is that I don't want to have to be finding the minimum between the two to make sure that the order is right all of the time.


For example:

```ts
editor.insertContentAt({ from: Math.min(resolveLocationToPm(point1), resolveLocationToPm(point2)), to: Math.max(resolveLocationToPm(point1), resolveLocationToPm(point2))

// v.s.

editor.insertContentAt({ from: point1, to: point2 })
```

Maybe, I'm thinking too much in terms of ProseMirror, and we can relax `from` being less than `to` as an invariant? That would still make sense to me, I guess, just a slight difference from ProseMirror behavior

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
nperez0111:
```ts
editor.insertBlocks(blocks, { id: 'xyz', offset: 2 }) === editor.insertBlocks(blocks, 'xyz') === editor.insertBlocks(blocks, { anchor: { id: 'xyz', offset: 2 }, head: { id: 'xyz', offset: 4 } }) 
```

It would find the block that the location is pointing to, in the case of a range, this can be ambiguous, so, currently it throws an error. The user can be more specific by just passing in the point that is in the `anchor` or `head` of the `Range`. I guess that we can disallow `Range`s in the block operations, but it can be convenient for things like the current selection as the argument.

packages/core/src/editor/managers/BlockManager.ts
nperez0111: Same as above, it can be resolved to a block.

packages/core/src/editor/managers/BlockManager.ts
nperez0111: renaming is fine

packages/core/src/api/blockManipulation/selections/selection.ts
nperez0111: This one is the block selection, the text setTextCursorPosition one handles points/ranges. I would prefer to only have one selection type, but I need an answer to https://github.com/TypeCellOS/BlockNote/pull/2044/files/ea043cd944e23cd3b52166c039eafedef1af2535#r2394892594 before I can merge them both into 1 selection interface.

I feel like you should always be able to get the current selection whether it is across multiple blocks or just a single collapsed selection within a block. But, I'm wanting to check how load-bearing this is first.

packages/core/src/locations/types.ts
nperez0111: I will need to update the jsdoc, since it is not actually the number of characters, but the number of ProseMirror positions, so table cell boundaries can be distinguished, that was the main reason for not just counting characters, which should work for most block types.

nperez0111: Two options for moving forward with the simpler API:
 - add an additional `affinity` property which can be used to tell whether you are on the left or right of a boundary (we may want to make this a number to not be boxed-in in terms of API, if there are a number of layers to have to specify this in). Perhaps this is a `depth` parameter?
 - give inline content an id (this may prove to be a larger change than we are willing to take on at this point, but worthwhile to consider)

packages/core/src/locations/types.ts
nperez0111: I think we are settled on changing `Range` to be based on `from` & `to`, where it will happen to be ordered in the case of a selection, otherwise the `Range` will just be inferred as any content _between_ the `from` & `to` ends, but not reliant on a specific ordering.

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
matthewlipski: I think ranges should still work - since `insertBlocks` takes an argument to insert before or after, you could just have it insert before the block containing the start of the range, or after the block containing the end of the range.

packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.ts
```diff
@@ -312,32 +311,18 @@ export function updateBlock<
   S extends StyleSchema = any,
 >(
   tr: Transform,
-  blockToUpdate: BlockIdentifier,
+  blockToUpdate: Location,
```
matthewlipski, line 314: Presumably this should work the same as `insertBlocks` r.e. points and ranges?

packages/core/src/api/blockManipulation/selections/selection.ts
matthewlipski: Yep, feel free! I honestly don't think there was a good reason to do it this way in the first place and agree it's just kind of annoying.

packages/core/src/editor/managers/BlockManager.ts
```diff
@@ -58,9 +59,11 @@ export class BlockManager<
    * matching block was found.
    */
   public getBlock(
```
matthewlipski, line 61: Hmm, `getBlock` has the issue of not being clear what should happen when a range is passed. Do you think it would make sense to restrict the type to `Point` only in cases like this?

packages/core/src/editor/managers/SelectionManager.ts
```diff
@@ -79,7 +67,7 @@ export class SelectionManager<
    * @param placement Whether the text cursor should be placed at the start or end of the block.
    */
   public setTextCursorPosition(
-    targetBlock: BlockIdentifier,
+    targetBlock: Location,
```
matthewlipski, line 70: Adding to above comment, probably worth doing a cmd+shift+f to find all cases that need renaming.

packages/core/src/locations/types.ts
matthewlipski: Imo the current API makes most sense. I feel like smth similar to an `affinity` property would cause more confusion than it solves. I'm also not sure inline content IDs + offset would be useful for this particular use-case vs just using block IDs + offset.

packages/core/src/locations/types.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * A block id is a unique identifier for a block, it is a string.
+ */
+export type BlockId = string;
+
+/**
+ * A block identifier is a unique identifier for a block, it is either a string, or can be object with an id property (out of convenience).
+ */
+export type BlockIdentifier = { id: BlockId } | BlockId;
```
matthewlipski, line 9: I'm kind of not a fan of `BlockIdentifier` as a type when it can very easily just be converted to a point. I understand it's for backwards compatibility, but would it be a bad idea to get rid of it altogether?

tests/src/unit/core/selection/getSelection/__snapshots__/regular/blockWithoutContent.json
```diff
@@ -1,21 +1 @@
...
-        "url": "https://ralfvanveen.com/wp-content/uploads/2021/06/Placeholder-_-Glossary.svg",
-        "caption": "",
-        "showPreview": true
-      },
-      "children": []
-    }
-  ]
-}
\ No newline at end of file
+undefined
```
matthewlipski, line 1: Is this correct?

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
nperez0111: @matthewlipski that is a great point, I think that this is all cohesive then

packages/core/src/locations/types.ts
nperez0111: @matthewlipski I think it is a little hard to explain prosemirror positioning for inline content, take the following example:

```txt
<p>ab<mention>12</mention>def</p>
   12    3    45     6    789
```

feels more complex than:

```txt
<p>ab<mention>12</mention>def</p>
   12         34          567
```

& for most operations, it can be inferred what you meant to do (e.g. replace over the range), it is just at the boundaries that it can be ambiguous like does `insertText` at position 4 get included within the mention or just after the mention?

If there was a `depth` parameter, you could think of it like:

```txt
<p>ab<mention>  </mention>def</p> : DEPTH 0
         \    12     /            : DEPTH 1
   12         34          567
```

I think this is much easier conceptually for the end user, as they can think of everything in terms of character offsets rather than wrapping elements & tree structures. Trees work well in the dom, since you can easily grab siblings, but is harder to describe in a positioning system especially when 90% of your operations are "get a node id & insert some text at the beginning or end of it".

Essentially this is making all inline content which are currently represented as nodes (which affect positioning with their start & ends) to be positioned as though they were marks (which do not affect positions)

packages/core/src/editor/managers/BlockManager.ts
nperez0111: Probably

packages/core/src/locations/types.ts
nperez0111: Are you against it being a separate type? Or, are you saying that points should be widened to allow the offset to be optional? I'm not sure what the action here should be

PR #2052, opened 2025-10-01
feat: Shortcut to delete empty table while cells are selected

### LINKED ISSUE #2038: Make it easier to delete tables
When selecting the whole table and hitting backspace you only delete the table content. 
Users who are not native to how blocks work struggle to delete the table. As now you need to over on the left, click on the block icon and select delete. 

In Gdoc or Word when you select a whole table and hit backspace the table is deleted. 
I'm wondering if we could find a sweet spot where, the first backspace deletes the content, the table remains selected and if you hit backspace a second time, the table is deleted.
What do you guys think ?

**Additional context**

**Bonus**
[x] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

matthewlipski:
Closes #2038

### DIFF
--- a/packages/core/src/blocks/Table/block.ts
+++ b/packages/core/src/blocks/Table/block.ts
@@ -1,6 +1,6 @@
 import { Node, mergeAttributes } from "@tiptap/core";
 import { DOMParser, Fragment, Node as PMNode, Schema } from "prosemirror-model";
-import { CellSelection, TableMap, TableView } from "prosemirror-tables";
+import { CellSelection, TableView } from "prosemirror-tables";
 import { NodeView } from "prosemirror-view";
 import { createBlockNoteExtension } from "../../editor/BlockNoteExtension.js";
 import {


packages/core/src/blocks/Table/block.ts
```diff
@@ -265,6 +265,59 @@ const TiptapTableParagraph = Node.create({
...
+
+          if (cellNode.firstChild && cellNode.firstChild?.content.size > 0) {
+            return false;
+          }
+        }
+
+        return editor.commands.deleteRange({
+          from: posBeforeTable,
+          to: posBeforeTable + tableNode.nodeSize,
+        });
```
nperez0111: When deleting a table, the selection is moving to the next node rather than the current node. Ideally the selection would stay in the current node

packages/core/src/blocks/Table/block.ts
nperez0111, line 1: While this definitely works, ideally we would do this as a BlockNote extension & use BlockNote APIs instead of Tiptap ones. This is because eventually we would have to migrate out of this, so it would be less to change later rather than piling on existing patterns we are trying to move away from

packages/core/src/blocks/Table/block.ts
```diff
@@ -394,6 +395,87 @@ export const createTableBlockSpec = () =>
...
+            const maxRowIndex = Math.max(anchorCellRowIndex, headCellRowIndex);
+
+            if (
+              minColIndex !== 0 ||
+              maxColIndex !== cols - 1 ||
+              minRowIndex !== 0 ||
+              maxRowIndex !== rows - 1
+            ) {
+              return false;
+            }
```
nperez0111: Did you test this with merged/split cells? I think that it will actually be a bit off since the number of table cells will not be the same as the number of `td` nodes. I suggest looking at what: `getCellSelection` does in tableHandlesPlugin, and either use or come up with something similar to that

packages/core/src/blocks/Table/block.ts
```diff
@@ -394,6 +395,87 @@ export const createTableBlockSpec = () =>
...
+
+            const selectionBlock =
+              editor.getPrevBlock(block) || editor.getNextBlock(block);
+            if (selectionBlock) {
+              editor.setTextCursorPosition(block);
+            }
+
+            editor.removeBlocks([block]);
```
nperez0111: Would prefer for this to run in an `editor.transact` so that it is only a single undo step.

packages/core/src/blocks/Table/block.ts
matthewlipski: I realized this can be done much simpler - the whole table is selected when each cell is selected, meaning we can just check if the number of selected cells is equal to the number of total cells or if it's less.

nperez0111 (approved):
Yep, this is simple enough!

PR #2054, opened 2025-10-02
fix: make validation errors recoverable by LLM

YousefED:
This makes sure when validation errors occur (and specifically if an LLM omits the `type` of an operation), this error will bubble up to the error handler correctly. This gives the LLM the right info to "retry" the operation and recover the error when the user requests this


nperez0111 (approved):

YousefED:
Most of this already landed in an earlier refactor, but still nice to remove the deprecated function + tests and improved error formatting

Issue #2055, opened 2025-10-02
Image caption isn't associated to the image in an accessible way


Ovgodd:
**Is your feature request related to a problem? Please describe.**  
When adding a caption to an image in BlockNote, the caption currently fails to meet **WCAG criterion**  
The issue is that the caption is not semantically associated with its image.  

**Expected behavior**  
The `<figure>` element should wrap both the `<img>` and its caption (`<figcaption>`), ensuring a proper semantic relationship.  

**Describe the solution you'd like**  
- Use `<figure>` and `<figcaption>` for images with captions.  
- If a caption is present:  
  - Apply the caption as the `alt` text for the image.  
  - Keep the `<figcaption>` as visible content.  
- If no caption is present:  
  - Use an empty `alt=""`.  
  - Mark the image as decorative with `role="presentation"` and `aria-hidden="true"`.  

**Describe alternatives you've considered**  
- Keeping the `<p>` caption: rejected, as it does not satisfy WCAG requirements.  
- Using `aria-describedby`: possible fallback, but less semantic than the `<figure>/<figcaption>` solution.  

**Additional context**  
Actual structure : 

<img width="635" height="198" alt="Image" src="https://github.com/user-attachments/assets/e77b3032-c813-4b5a-a186-8f8b3b08a20d" />

Example of expected structure:  

```html
<figure class="bn-file-block-content-wrapper" role="img" aria-label="Image: Example caption">
  <img src="..." alt="Example caption" />
  <figcaption class="bn-file-caption">Example caption</figcaption>
</figure>

Ovgodd:
I will propose a solution

PR #2056, opened 2025-10-02
convert to figure/figcaption if caption exists

Ovgodd:
**Purpose**
Improve accessibility and semantic HTML by applying the appropriate structure when captions are present, to comply with WCAG requirement.

issue : [2055](https://github.com/TypeCellOS/BlockNote/issues/2055)

figcap

**Proposal**

-  Detect if a caption exists in the component/template
-  Place the caption inside a `figcaption `tag
-  Wrap media elements in a `figure `tag
-  Place the caption inside a `figcaption `tag

### DIFF
--- a/packages/core/src/blocks/FileBlockContent/helpers/render/createFileBlockWrapper.ts
+++ b/packages/core/src/blocks/FileBlockContent/helpers/render/createFileBlockWrapper.ts
@@ -52,6 +52,31 @@ export const createFileBlockWrapper = (
     };
   }
 
+  // Special accessible structure for images with captions: use figure/figcaption
+  const shouldUseFigure =
+    element &&
+    block.type === "image" &&
+    block.props.caption &&
+    block.props.showPreview !== false;
+  if (shouldUseFigure) {
+    const figure = document.createElement("figure");
+    figure.className = wrapper.className;
+
+    // Move preview element inside the figure
+    figure.appendChild(element.dom);
+
+    // Create figcaption instead of a paragraph
+    const figcaption = document.createElement("figcaption");
+    figcaption.className = "bn-file-caption";
+    figcaption.textContent = block.props.caption;
+    figure.appendChild(figcaption);
+
+    // ARIA enhancements for screen readers
+    figure.setAttribute("role", "img");
+    figure.setAttribute("aria-label", `Image: ${block.props.caption}`);
+
+    return { dom: figure, destroy: element.destroy };
+  }
   const ret: { dom: HTMLElement; destroy?: () => void } = { dom: wrapper };
 
   // Show the file preview, or the file name and icon.
--- a/packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
+++ b/packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
@@ -70,7 +70,23 @@ export const imageRender = (
     image.src = block.props.url;
   }
 
-  image.alt = block.props.name || block.props.caption || "BlockNote image";
+  // Accessibility: set alt/aria based on presence of caption per RGAA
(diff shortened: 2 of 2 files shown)


packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
```diff
@@ -142,7 +165,11 @@ export const imageToExternalHTML = (
 
   if (block.props.caption) {
     if (block.props.showPreview) {
-      return createFigureWithCaption(image, block.props.caption);
+      const { dom } = createFigureWithCaption(image, block.props.caption);
+      // Enhance figure with explicit role and aria-label
+      dom.setAttribute("role", "img");
+      dom.setAttribute("aria-label", `Image: ${block.props.caption}`);
+      return { dom };
```
nperez0111, line 161: Why do this here instead of fixing it in createFigureWithCaption?

packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
```diff
@@ -130,7 +146,14 @@ export const imageToExternalHTML = (
...
     image.src = block.props.url;
-    image.alt = block.props.name || block.props.caption || "BlockNote image";
+    // Accessibility for exported HTML: prefer caption as alt when present
+    if (block.props.caption) {
+      image.alt = block.props.caption;
+    } else {
+      image.alt = "";
+      image.setAttribute("role", "presentation");
+      image.setAttribute("aria-hidden", "true");
+    }
```
nperez0111: Missing `block.props.name`, why not just:
```suggestion
	image.alt = block.props.name || block.props.caption || "";
    if (!image.alt) {
      image.setAttribute("role", "presentation");
      image.setAttribute("aria-hidden", "true");
    }
```

packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
```diff
@@ -70,7 +70,23 @@ export const imageRender = (
...
+  };
+
+  const accessibleImage = () => {
+    image.alt = "";
+    image.setAttribute("role", "presentation");
+    image.setAttribute("aria-hidden", "true");
+    image.setAttribute("tabindex", "-1");
+  };
+
+  block.props.caption ? accessibleImageWithCaption() : accessibleImage();
```
nperez0111: Same as below, just fallback when there is no alt, and this doesn't have the `block.props.name`

packages/core/src/blocks/FileBlockContent/helpers/render/createFileBlockWrapper.ts
```diff
@@ -52,6 +52,31 @@ export const createFileBlockWrapper = (
...
+    figcaption.className = "bn-file-caption";
+    figcaption.textContent = block.props.caption;
+    figure.appendChild(figcaption);
+
+    // ARIA enhancements for screen readers
+    figure.setAttribute("role", "img");
+    figure.setAttribute("aria-label", `Image: ${block.props.caption}`);
+
+    return { dom: figure, destroy: element.destroy };
+  }
```
nperez0111, line 79: I would prefer this as attributes rather than figure figcaption, this makes the code more difficult to reason about, and don't buy that it is any more accessible

Ovgodd: I understand !

packages/core/src/blocks/ImageBlockContent/ImageBlockContent.ts
Ovgodd: I removed it,role="img" isn’t needed on figure since it already has proper semantics for screen readers.

PR #2063, opened 2025-10-03
fix: allow listening to `onChange` and other events before the underlying editor is initialized

nperez0111:
# Summary
This change allows for using `editor.onChange` before the `editor.tiptapEditor` is actually assigned. This allows us to not have to wrap every event we register in extensions (the only time you'd have access to the editor instance without the tiptapEditor being initialized), be forced to be wrapped in `editor.onCreate`.

## Rationale

This simplifies extension code to be the same as what you'd use with a fully constructed editor instance.

## Changes

De-coupled the event handling to be done on the EventManager instance by making it into an EventEmitter

## Impact
Should be no change

## Testing
Added tests to prove that this works, also added tests for the new onMount & onUnMount events that I exposed as part of this

## Checklist

### DIFF
--- a/docs/content/docs/reference/editor/events.mdx
+++ b/docs/content/docs/reference/editor/events.mdx
@@ -12,7 +12,7 @@ BlockNote provides several event callbacks that allow you to respond to changes 
 
 The editor emits events for:
 
-- **Editor lifecycle** - When the editor is mounted, unmounted, etc.
+- **Editor lifecycle** - When the editor is created, mounted, unmounted, etc.
 - **Content changes** - When blocks are inserted, updated, or deleted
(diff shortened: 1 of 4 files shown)


docs/content/docs/reference/editor/events.mdx
```diff
@@ -12,7 +12,7 @@ BlockNote provides several event callbacks that allow you to respond to changes
 
 The editor emits events for:
 
-- **Editor initialization** - When the editor is ready for use
+- **Editor lifecycle** - When the editor is mounted, unmounted, etc.
```
nperez0111:
```suggestion
- **Editor lifecycle** - When the editor is created, mounted, unmounted, etc.
```

PR #2071, opened 2025-10-06
fix: Toggle and check list item blocks

### LINKED ISSUE #2067: Toggle lists broken v0.39.1 onwards
**Describe the bug**
Toggle lists don't work. You can try at in the example https://www.blocknotejs.org/examples/basic/minimal

**To Reproduce**
Just go to https://www.blocknotejs.org/examples/basic/minimal and try to create a toggle list.

- Node version: v22.14.0
- Package manager: pnpm
- Browser: Chrome

matthewlipski:
# Summary

This PR fixes a variety of issues with the toggle list item block, as well as fixing styling for check list item blocks.

Closes #2067

## Rationale

These two blocks are currently not working as intended. The check list item block is mostly ok and just needs styles tweaked. The toggle list item block has a bunch of issues though, regarding rendering, exporting, etc.

## Changes

- Adjusted check list item `render` and styles.
- Fixed toggle list item blocks not showing the toggle button.
- Fixed toggle list items not being exported to external HTML in the proper list structure.
- Fixed `serializeBlock` incorrectly creating `Block` objects from `PartialBlock` objects to pass into `render` (missing fields).

## Impact

The check list and toggle list item blocks are part of the default blocks, so this fix is fairly critical.

## Testing

- Added toggle list item blocks to unit tests which include all default blocks.
- Added e2e test to screenshot all basic blocks, to prevent styling regressions.

## Checklist


packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
```diff
@@ -28,7 +28,7 @@ export const createToggleWrapper = (
   ignoreMutation?: (mutation: ViewMutationRecord) => boolean;
   destroy?: () => void;
 } => {
-  if (!("isToggleable" in block.props) || !block.props.isToggleable) {
+  if ("isToggleable" in block.props && !block.props.isToggleable) {
```
nperez0111, line 31: if it doesn't have the `isToggleable` property, in block props, it should return just the dom node no?

packages/core/src/blocks/ListItem/CheckListItem/block.ts
```diff
@@ -75,6 +75,8 @@ export const createCheckListItemBlockSpec = createBlockSpec(
       getListItemContent(el, schema, "checkListItem"),
     render(block, editor) {
       const dom = document.createDocumentFragment();
+      const checkboxWrapper = document.createElement("div");
```
nperez0111: Do we need this wrapper element? What is different than before to need this?

packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts
matthewlipski: Actually no, `isToggleable` is used in toggle headings but not toggle list item blocks, as the toggle list items are distinct blocks whereas the toggle headings are integrated into the regular heading blocks. So because toggle list items don't have an `isToggleable` prop, the toggle wrapper should still be rendered when a block has no `isToggleable` prop.

I think we messed smth up here in the default blocks refactor, as this PR makes reverts this line to be the same as in an [older version](https://github.com/TypeCellOS/BlockNote/blob/v0.36.1/packages/core/src/blocks/ToggleWrapper/createToggleWrapper.ts).

packages/core/src/blocks/ListItem/CheckListItem/block.ts
matthewlipski: Since I added the basic blocks e2e test I noticed the check box was off center in Safari. Cleanest fix was to just wrap in a flexbox as trying to fix it with widths, margins, and paddings was difficult across browsers.

tests/src/end-to-end/basics/basicblocks.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    focusOnEditor(page);
+    await page.waitForTimeout(100);
+    await page
+      .locator(`${PARAGRAPH_SELECTOR} > p`, {
+        hasText: "Welcome to this demo!",
+      })
+      .click();
+
+    // Scroll to top of page.
+    await page.mouse.wheel(0, -9999);
```
YousefED: you can use `fullPage: true` (see docs for `screenshot`)

packages/core/src/blocks/ListItem/CheckListItem/block.ts
nperez0111: Can we do that without a wrapper element? I don't believe we should be making changes to html output for CSS reasons. We'd also have to check our parsing behavior, and I'd prefer to keep things simple, than having to remember what this element is for

PR #2089, opened 2025-10-08
fix: Partial block to block conversion in HTML export

matthewlipski:
# Summary

This PR fixes numerous runtime errors that could occur when passing `PartialBlock` arrays to `blocksToFullHTML`/`blocksToHTMLLossy`/`blocksToMarkdownLossy` rather than `Block` arrays. This is because we need full `Block` objects when doing the conversion as this is the type that gets passed to `render`/`toExternalHTML`, and were not doing the conversion properly.

## Rationale

This issue wasn't really noticed as it's rare that a block's `render`/`toExternalHTML` functions will ever access anything aside from the block's props, which were previously correctly getting converted. However, it's a pretty major issue as the object being passed to those functions was basically half missing.

## Changes

Implemented `partialBlockToBlock` function and used it in `serializeBlocksInternalHTML`/`serializeBlocksExternalHTML`.

## Impact

N/A

## Testing

Added a custom block to the unit test editor schema which validates that the block passed to `render`/`toExternalHTML` is of the correct shape. Added unit tests which export this block but specifically omit fields to ensure that they are fixed during the export.

## Checklist

### DIFF
--- a/packages/core/src/util/table.ts
+++ b/packages/core/src/util/table.ts
@@ -25,7 +25,7 @@ export function mapTableCell<
     : isPartialTableCell(content)
       ? {
           type: "tableCell",
-          content: content.content as InlineContent<T, S>[],
+          content: ([] as InlineContent<T, S>[]).concat(content.content as any),
           props: {
             backgroundColor: "default",
             textColor: "default",
@@ -35,7 +35,7 @@ export function mapTableCell<
         }
       : {
           type: "tableCell",
-          content: content as InlineContent<T, S>[],
+          content: ([] as InlineContent<T, S>[]).concat(content as any),
           props: {
             backgroundColor: "default",
             textColor: "default",


YousefED:
Thanks @matthewlipski . Could you give some background on how you ran into this / what was the background / trigger / scenario you're solving?

Asking this because;
- I was revisiting the zod propschemas today, this seems relevant
- maybe we should zoom out and look at the concept of "partial" more holistically

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -175,20 +176,12 @@ function serializeBlock<
...
-  // set default props in case we were passed a partial block
-  const props = block.props || {};
-  for (const [name, spec] of Object.entries(
-    editor.schema.blockSchema[block.type as any].propSchema,
-  )) {
-    if (!(name in props) && spec.default !== undefined) {
-      (props as any)[name] = spec.default;
-    }
-  }
+  const fullBlock = partialBlockToBlock(block, editor);
```
nperez0111: Would recommend `partialBlockToFullBlock`

packages/core/src/api/partialBlockToBlock.ts
```diff
@@ -0,0 +1,324 @@
...
+      ) {
+        const propTypeMatches =
+          typeof partialProps[propName] ===
+          (propSpec.default !== undefined
+            ? typeof propSpec.default
+            : propSpec.type);
+        const propValueIsValid =
+          propSpec.values === undefined
+            ? true
+            : propSpec.values.includes(partialProps[propName]);
```
nperez0111:
```suggestion
        const propValueIsValid =
          Array.isArray(propSpec.values)
            ? propSpec.values.includes(partialProps[propName])
            : true;
```

Prefer testing for the thing you need, not for it's absence

packages/core/src/api/partialBlockToBlock.ts
```diff
@@ -0,0 +1,324 @@
...
+      if (
+        partialProps &&
+        propName in partialProps &&
+        partialProps[propName] !== undefined
+      ) {
+        const propTypeMatches =
+          typeof partialProps[propName] ===
+          (propSpec.default !== undefined
+            ? typeof propSpec.default
+            : propSpec.type);
```
nperez0111:
```suggestion
        const propTypeMatches =
          typeof partialProps[propName] ===
          ('type' in propSpec
            ? propSpec.type
            : typeof propSpec.default);
```

If there is a type, it should be respected

packages/core/src/api/partialBlockToFullBlock.ts
```diff
@@ -0,0 +1,324 @@
...
+        typeof content === "undefined"
+          ? undefined
+          : typeof content === "string"
+            ? [textStringToStyledText(content)]
+            : content,
+    };
+  });
+};
+
+const partialTableContentToTableContent = (
```
nperez0111, line 149: I already had written a utility for this `mapTableCell`, I'd suggest augmenting it with anything that this does better

packages/core/src/api/partialBlockToBlock.ts
```diff
@@ -0,0 +1,324 @@
...
+    return partialInlineContentToInlineContent(
+      partialBlockContent,
+      inlineContentSchema,
+    );
+  }
+
+  return undefined;
+};
+
+export const partialBlockToBlock = <
```
nperez0111:
```suggestion
export const partialBlockToFullBlock = <
```

matthewlipski:
> Thanks @matthewlipski . Could you give some background on how you ran into this / what was the background / trigger / scenario you're solving?
> 
> Asking this because;
> 
> * I was revisiting the zod propschemas today, this seems relevant
> * maybe we should zoom out and look at the concept of "partial" more holistically

I originally ran into this issue recently when working on #2071. The toggle list item uses `block.children` in its `render` function, and after adding the block to our list/default block tests, they were throwing an error. This is because `serializeBlocksInternalHTML`/`serializeBlocksExternalHTML` were only converting the props, and if no children were passed, `block.children` would be undefined while the toggle block's `render` expected an array.

In that PR, I made a quick fix for the children but left the other fields (id, type, and content) unconverted were since they were out of scope. But the whole conversion seemed like a major oversight so I wanted to get this PR out right away. Admittedly though, there are not many use cases I can think of where a block's ID or content should change how it's rendered/exported.

packages/core/src/api/partialBlockToFullBlock.ts
matthewlipski: The only real advantage of `partialTableContentToTableContent` vs `mapTableCell` is that it properly converts partial inline content to full inline content in each cell. This is a nice improvement, but requires an inline content schema to be passed.

Turns out that doing so causes way too many complications downstream, to the point that we'd need to refactor the PDF, DOCX, etc exporters (as these also use `mapTableCell`) as each block mapping now needs an editor to be passed.

So because of the added complexity, I'd rather leave `mapTableCell` as is.

nperez0111: Ok, then can we do the reverse & have some of this simplified by using mapTableCell?

packages/core/src/util/table.ts
```diff
@@ -25,24 +25,21 @@ export function mapTableCell<
     : isPartialTableCell(content)
       ? {
           type: "tableCell",
-          content: ([] as InlineContent<T, S>[]).concat(content.content as any),
```
matthewlipski, line 28: @nperez0111 what's up with the `concat` here? I get that it's to make sure it's an array, but does it matter since the items in the array have the wrong type anyway? Same for line 39

nperez0111: It's just to coerce it to always be an array, since a string is a valid inline content, this just gives it more structure than allowing just anything

matthewlipski:
Yeah I think it makes sense to wait and see if this is really needed, since it's a decent chunk of code, especially if we add more tests. Are there any major features/refactors that we could bundle this investigation with?

nperez0111:
I think we can close this?

matthewlipski:
I think so, @YousefED your changes to where PartialBlocks are used will fully replace this right?

PR #2092, opened 2025-10-10
fix: Emoji picker issues

### LINKED ISSUE #2053: Comments: EmojiPicker clipped in the Thread modal
We can have lot of comments in a thread, so we add a scroller inside the Thread, we can have the input at the bottom easily accessible as well like that.
The problem with that is the EmojiPicker is clipped inside the Thread container, so part of it is hidden and not easily accessible.

Image

An easy solution would be to use `createPortal` around the EmojiPicker, to have the modal floating above the container: https://react.dev/reference/react-dom/createPortal#rendering-a-modal-dialog-with-a-portal

**Bonus**
Blocknote Version: 
[ ] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

matthewlipski:
# Summary

This PR makes a few changes to the emoji picker, mostly related to overflow handling. A few minor styling changes have also been made for the comments & thread examples.

Closes #2053

## Rationale

To deal with longer threads, consumers may set a may height on them and make them scrollable. In this case, the emoji picker gets cut off, leading to bad UX.

## Changes

The Mantine popover component now gets auto-sized based on the available space. Most UI elements that use this component are small and don't really need it, but it's useful for the emoji picker which can overflow more often (especially on the small website example boxes).

The popover for the emoji picker now also gets rendered in a portal which attaches it to the editor element (`.bn-container`). To ensure it remains on top, the z-index of popover components has been increased. This shouldn't have much of an effect outside the emoji picker. The emoji picker in the threads sidebar also still works as expected with these changes.

Finally, the logic for keeping the action toolbar open while picking emojis has been updated to work with the portal.

## Impact

Z-index of popovers has been increased, but this shouldn't have a noticable impact on other elements that use popovers like the create link button of replace file button.

## Testing

N/A

## Checklist

### DIFF
--- a/examples/07-collaboration/05-comments/src/App.tsx
+++ b/examples/07-collaboration/05-comments/src/App.tsx
@@ -4,8 +4,8 @@ import {
   DefaultThreadStoreAuth,
   YjsThreadStore,
 } from "@blocknote/core/comments";
-import { BlockNoteView } from "@blocknote/shadcn";
-import "@blocknote/shadcn/style.css";
+import { BlockNoteView } from "@blocknote/mantine";
+import "@blocknote/mantine/style.css";
 import { useCreateBlockNote } from "@blocknote/react";
 import { YDocProvider, useYDoc, useYjsProvider } from "@y-sweet/react";
 import { useMemo, useState } from "react";
--- a/packages/react/src/components/Comments/Comment.tsx
+++ b/packages/react/src/components/Comments/Comment.tsx
@@ -167,7 +167,7 @@ export const Comment = ({
             onEmojiSelect={(emoji: { native: string }) =>
               onReactionSelect(emoji.native)
             }
-            onOpenChange={(open) => setEmojiPickerOpen(open)}
+            onOpenChange={setEmojiPickerOpen}
           >
             <Components.Generic.Toolbar.Button
               key={"add-reaction"}
@@ -282,7 +282,7 @@ export const Comment = ({
                           onEmojiSelect={(emoji: { native: string }) =>
                             onReactionSelect(emoji.native)
                           }
-                          onOpenChange={(open) => setEmojiPickerOpen(open)}
+                          onOpenChange={setEmojiPickerOpen}
                         >
                           <Components.Generic.Badge.Root
                             className={mergeCSSClasses(


examples/07-collaboration/05-comments/src/App.tsx
```diff
@@ -4,8 +4,8 @@ import {
   DefaultThreadStoreAuth,
   YjsThreadStore,
 } from "@blocknote/core/comments";
-import { BlockNoteView } from "@blocknote/mantine";
-import "@blocknote/mantine/style.css";
+import { BlockNoteView } from "@blocknote/shadcn";
+import "@blocknote/shadcn/style.css";
```
nperez0111: shadcn now?

packages/mantine/src/popover/Popover.tsx
```diff
@@ -16,7 +16,13 @@ export const Popover = (
   assertEmpty(rest);
 
   return (
-    <MantinePopover withinPortal={false} opened={opened} position={position}>
+    <MantinePopover
+      middlewares={{ size: { padding: 20 } }}
+      withinPortal={false}
+      opened={opened}
+      position={position}
+      zIndex={10000}
```
nperez0111, line 24: I prefer more reasonable values for z-index, but I've not seen what this is compared to

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -166,6 +167,7 @@ export const Comment = ({
             onEmojiSelect={(emoji: { native: string }) =>
               onReactionSelect(emoji.native)
             }
+            onOpenChange={(open) => setEmojiPickerOpen(open)}
```
nperez0111:
```suggestion
            onOpenChange={setEmojiPickerOpen}
```
Keeps the function reference the same across re-renders

packages/react/src/components/Comments/Comment.tsx
```diff
@@ -279,6 +282,7 @@ export const Comment = ({
                           onEmojiSelect={(emoji: { native: string }) =>
                             onReactionSelect(emoji.native)
                           }
+                          onOpenChange={(open) => setEmojiPickerOpen(open)}
```
nperez0111:
```suggestion
                          onOpenChange={setEmojiPickerOpen}
```

packages/mantine/src/popover/Popover.tsx
matthewlipski: Each of the UI elements is a multiple of 1000 (1000 for side menu, 3000 for formatting toolbar, etc), so this is continuing that pattern.

packages/mantine/src/popover/Popover.tsx
```diff
@@ -16,7 +16,13 @@ export const Popover = (
   assertEmpty(rest);
 
   return (
-    <MantinePopover withinPortal={false} opened={opened} position={position}>
+    <MantinePopover
+      middlewares={{ size: { padding: 20 } }}
+      withinPortal={false}
```
YousefED, line 21: @matthewlipski only seeing this now, but can't we use `withinPortal` instead of using our own `createPortal`?

matthewlipski: Technically yes, functionally it is the same. However, since we only use a portal for the emoji picker, and neither Ariakit nor ShadCN have popover options to render in a portal, I decided to go with this approach as it's less code and easier to maintain.

PR #2096, opened 2025-10-13
refactor: zod for props

YousefED:
# Summary

This PR drops our custom schema for Block and Inline Content Props and uses Zod instead.

## Rationale

- Less "non-standard" code to maintain
- Easier to explain in docs / to LLMs

## Impact

Consumers using custom blocks would need to migrate

- [ ] TODO: Determine if we want to accept old style props for backwards compatibility

## Testing

Mostly by existing unit tests. TODO:

- [x] build of examples doesn't pass. couldn't figure out the build errors for withMultiColumn and withPageBreak examples. Playing around with `BlockNoteSchema` vs `CustomBlockNoteSchema`. Perhaps @nperez0111 can help here?
- [ ] Manual testing + code review. @matthewlipski can you help with this (after build + tests pass), and specifically the file blocks and anywhere where block type guards are used
- [ ] add tests / example for more complex schemas (e.g.: a prop with a nested object). @matthewlipski can you help?
- [ ] check compatibility with yjs, does it break with https://github.com/yjs/y-prosemirror/issues/116? (cc @nperez0111)
- [ ] heading props / customization (marked as TODO)

## Screenshots/Video

N/A

## Checklist

- [ ] Y: The documentation has been updated to reflect the new feature
- [ ] Y: Review and add comments + followup work
- [ ] All: validate this doesn't make typescript / IDE significantly slower when developing
- [ ] All: Decide on zod as a dep or as a peer dep. Can we use standard schema? 
- [ ] Check remaining comments in https://github.com/TypeCellOS/BlockNote/pull/2111

## Additional Notes

Future work:
- [ ] Migrate style props as well?
- [x] Refactor File Blocks / File Props?
- [x] Remove / refactor "Partial" blocks as much as possible


tests/src/unit/core/clipboard/paste/__snapshots__/text/html/pasteImage.json
```diff
@@ -24,6 +24,7 @@
       "backgroundColor": "default",
       "caption": "",
       "name": "",
+      "previewWidth": undefined,
```
YousefED, line 27: Note, when getting a block you now also get optional props as undefined. I think this is ok?

cc @nperez0111

packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.test.ts
```diff
@@ -50,6 +50,22 @@ describe("Test updateBlock typing", () => {
     } catch (e) {
       // ID doesn't exist, which is fine - this is a compile-time check
     }
+
+    try {
+      // TODO: props.level type should propagate correctly
```
YousefED, line 55: @nperez0111 I didn't get this to work with the way blocks are defined. Can you have a look at this or can we pair up?

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -176,12 +176,19 @@ function serializeBlock<
   const BC_NODE = editor.pmSchema.nodes["blockContainer"];
 
   // set default props in case we were passed a partial block
+  // TODO: should be a nicer way for this / or move to caller
```
YousefED: fyi, I don't think these functions should take PartialBlocks. I want to review the PartialBlock concept later. My thought at this moment is that they should only be on the highest level, to make calling things like `updateBlock` / `insertBlocks` more convenient

packages/core/src/blocks/defaultBlockTypeGuards.ts
```diff
@@ -1,35 +1,20 @@
+import { Selection } from "prosemirror-state";
 import { CellSelection } from "prosemirror-tables";
 import type { BlockNoteEditor } from "../editor/BlockNoteEditor.js";
-import { BlockConfig, PropSchema, PropSpec } from "../schema/index.js";
+import { BlockConfig, PropSchema } from "../schema/index.js";
 import { Block } from "./defaultBlocks.js";
-import { Selection } from "prosemirror-state";
 
+// TODO: matthew review + tests
```
YousefED: @matthewlipski , you recently refactored this file I think. afaik there are no unit tests covering the type guards. Could you add these? I'm not sure if my implementation matches the scenarios you had in mind

You can also add them to the `main` branch implementation and then I can migrate them to validate they still work after this zod refactor

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -156,7 +156,7 @@ export type BlockNoteEditorOptions<
    * @remarks `CommentsOptions`
    */
   comments?: {
-    schema?: BlockNoteSchema<any, any, any>;
+    schema?: CustomBlockNoteSchema<any, any, any>;
```
YousefED: @nperez0111 please double check my changes in this file (not sure I correctly get the difference between CustomBlockNoteSchema and BlockNoteSchema)

packages/server-util/src/context/ServerBlockNoteEditor.ts
```diff
@@ -75,15 +76,19 @@ export class ServerBlockNoteEditor<
...
-    return new ServerBlockNoteEditor(options) as ServerBlockNoteEditor<
-      BSchema,
-      ISchema,
-      SSchema
-    >;
+    Options extends Partial<BlockNoteEditorOptions<any, any, any>> | undefined,
+  >(
+    options?: Options,
+  ): Options extends {
+    schema: CustomBlockNoteSchema<infer BSchema, infer ISchema, infer SSchema>;
```
YousefED, line 84: @nperez0111 another schema change

packages/core/src/schema/blocks/internal.ts
```diff
@@ -34,41 +39,19 @@ export function propsToAttributes(propSchema: PropSchema): Attributes {
...
-          }
-
-          return null;
+        // TBD: this might not be fault proof, but it's also ugly to store prop="&quot;...&quot;" for strings
+        try {
+          const jsonValue = JSON.parse(value);
+          // it was a number / boolean / json object stored as attribute
+          return z.parse(spec, jsonValue);
+        } catch (e) {
+          // it might have been a string directly stored as attribute
```
nperez0111: I'd recommend https://github.com/unjs/destr & maybe this is helpful? https://zod.dev/codecs?id=jsonschema

packages/core/src/schema/blocks/internal.ts
```diff
@@ -20,9 +22,12 @@ import {
 export function propsToAttributes(propSchema: PropSchema): Attributes {
   const tiptapAttributes: Record<string, Attribute> = {};
 
-  Object.entries(propSchema).forEach(([name, spec]) => {
+  Object.entries(propSchema._zod.def.shape).forEach(([name, spec]) => {
+    const def =
+      spec instanceof z.$ZodDefault ? spec._zod.def.defaultValue : undefined;
```
nperez0111: I've seen this in a number of places, would prefer to make a util function for this, since it isn't obvious & dependent on zod's API not our own. I don't love reach into zod for this, but don't see another way to do this unless we provide defaults separately

packages/core/src/blocks/defaultBlockTypeGuards.ts
matthewlipski: Added tests in #2103

tests/src/unit/core/clipboard/paste/__snapshots__/text/html/pasteImage.json
matthewlipski: Seems fine to me, probably better even.

matthewlipski:
>  add tests / example for more complex schemas (e.g.: a prop with a nested object). @matthewlipski can you help?

@YousefED so with this PR are we introducing the ability to have objects as props alongside strings/numbers/booleans? Because I remember this was something we considered when first making the custom blocks API, but skipped it mainly because objects need to be set on HTML attributes with `JSON.stringify` so they can then be parsed properly. IIRC we were concerned about the stringified objects making the DOM super messy, and also issues if you have functions since they can't be stringified. Do we have a solution for this?

PR #2103, opened 2025-10-14
test: unit tests for block type guards

matthewlipski:
# Summary

This PR adds unit tests for `editorHasBlockWithType`.

The code for `editorHasBlockWithType` has also been changed slightly - when checking for a block with a given `propSchema` and `values`, the `values` can be an unordered subset of the ones in the schema. Previously, the 2 arrays had to be identical in order for the type guard to return `true`.

## Rationale

Needed for #2096 and useful in general.

## Changes

See above.

## Impact

N/A

## Testing

See above.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/tests/src/unit/core/typeGuards/runTests.test.ts
+++ b/tests/src/unit/core/typeGuards/runTests.test.ts
@@ -83,4 +83,53 @@ describe("Editor block schema type guard tests", () => {
       }),
     ).toBeFalsy();
   });
+
+  it("customBlockType", () => {
+    expect(editorHasBlockWithType(getEditor(), "simpleImage")).toBeTruthy();
+  });
+
+  it("customBlockWithPropTypes", () => {
+    expect(
+      editorHasBlockWithType(getEditor(), "simpleImage", {
+        name: "string",
+        url: "string",
+      }),
+    ).toBeTruthy();
+  });
+
+  it("customBlockWithPropTypesInvalidType", () => {
+    expect(
+      editorHasBlockWithType(getEditor(), "simpleImage", {
+        name: "string",
+        url: "number",
+      }),
+    ).toBeFalsy();
+  });
+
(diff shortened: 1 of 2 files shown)


tests/src/unit/core/typeGuards/runTests.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import { testSchema } from "../testSchema.js";
+
+// Tests for verifying that type guards which check if an editor's schema
+// contains a block (and its props) are working correctly.
+describe("Editor block schema type guard tests", () => {
+  const getEditor = createTestEditor(testSchema) as () => BlockNoteEditor<
+    any,
+    any,
+    any
+  >;
```
nperez0111, line 14: Can we add a test using a custom block, just to make sure that it can work for that too, even just as a smoke test

matthewlipski:
Also @YousefED we talked on call yesterday about whether to put additional tests in the `tests` directory or with the files that they're actually testing. Should we have the same discussion with @nperez0111 before merging this or is that for another time?

matthewlipski:
Yeah the `!== undefined` additions are to fix bugs, since things like empty strings set as default props would cause issues. Not sure if other change with `values` counts as a fix or not, it just makes it easier e.g. if you want to see if the `textAlignment` prop allows for `"justify"` alignment. You'd previously have to check with:

```ts
editorHasBlockWithType(editor, "paragraph", {
  textAlignment: {
    default: "left",
    values: ["left", "center", "right", "justify"]
  }
})
```

and now it's just:

```ts
editorHasBlockWithType(editor, "paragraph", {
  textAlignment: {
    default: "left",
    values: ["justify"]
  }
})
```

YousefED:
Cool, proofs the value of investing in these tests :)

Op vr 17 okt 2025, 11:26 schreef Matthew Lipski ***@***.***>:

PR #2108, opened 2025-10-15
fix: add more lenient parsing for code blocks, to accept newlines #2105

### LINKED ISSUE #2105: Line breaks lost in code blocks after using tryParseHTMLToBlocks and replaceBlocks
**Describe the bug**
When I try to convert an HTML string containing `<code>` blocks into BlockNote blocks using `tryParseHTMLToBlocks`, and then replace the current blocks using `replaceBlocks`, all the line breaks inside the code blocks are lost and replaced with spaces. This causes multi-line code to appear as a single line, breaking formatting and readability.

Image

**To Reproduce**

[github-vamzb28v (1).zip](https://github.com/user-attachments/files/22923815/github-vamzb28v.1.zip)

**Misc**

- Node version: 22.0
- Package manager: pnpm
- Browser: chrome

nperez0111:
# Summary
I only validated that this did what I expected, I do not think that this is actually complete, we should add tests for this specific behavior.

Closes #2105 

## Rationale

This allows us to more leniently parse existing code blocks from arbitrary HTML in a way which is more user-friendly.

## Checklist

### DIFF
--- a/tests/src/unit/core/formatConversion/parse/__snapshots__/html/codeBlocksMultiLine.json
+++ b/tests/src/unit/core/formatConversion/parse/__snapshots__/html/codeBlocksMultiLine.json
@@ -1,0 +1,19 @@
+[
+  {
+    "children": [],
+    "content": [
+      {
+        "styles": {},
+        "text": "console.log("First Line")
+console.log("Second Line")
+console.log("Third Line")",
+        "type": "text",
+      },
+    ],
+    "id": "1",
+    "props": {
+      "language": "text",
+    },
+    "type": "codeBlock",
+  },
+]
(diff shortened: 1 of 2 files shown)


matthewlipski:
Alright I think a single test case to cover multi-line `<pre><code>` elements should be enough to prevent regressions. Otherwise I think this change looks good.

The one thing I was hesitant about was that blocks generally mark newlines with a line break node, but the parsed code block content is just one big multi-line string. But then I saw that it's because of the `code: true` in the spec, so no issue there.

nperez0111:
I think this is sufficient, thanks @matthewlipski

PR #2110, opened 2025-10-15
fix: Deleting last block in column

### LINKED ISSUE #1323: Cannot convert from Multi Column to single column
When using the multicolumn feature it is straight for ward to create the single columns to multiple column by jsimple drag and drop

But when doing reverse i.e. if we try to convert the two column format to 1, by moving the second column text to bottom it only moves the text

It should also change the 2 column of the previous row to one column but  it doesn't do that

The example is at 
link:: [https://github.com/TypeCellOS/BlockNote/tree/main/examples/01-basic/03-multi-column/](https://github.com/TypeCellOS/BlockNote/tree/main/examples/01-basic/03-multi-column/)
Image::

bug

### LINKED ISSUE #2000: UX hickups - multi-column cursor drop
**Describe the bug**
When drag and droping a line users complain that it oftens create a column on the left. 
And since it's difficult to delete a column (see https://github.com/TypeCellOS/BlockNote/issues/1323) this creates a major UX problem.
Using multi column is a bit of niche feature but the fact that the drop zone that creates a new columns is too big it's impacting a lot of users.
They have to be extra careful where the blue highlight is located
We are actually disabling the cursor drop for column creation for now

**To Reproduce**
- create a few lines
- drag a line bellow another one, your mouse cursor should be on the left hand side
- see that the drop zone is quite large for the left column creation

**Misc**
- Node version:
- Package manager:
- Browser:

matthewlipski:
# Summary

Currently, when there is only one block in a column, removing it via `removeBlocks`/`replaceBlocks` won't collapse the column as it should, leading to confusing UX when deleting the block from the side menu. This behavior is also different to removing the block by hitting backspace at its start.

This PR makes empty `column`s, and `columnLists` with <2 children get collapsed. This happens by running the new  `fixColumnList` function in `removeBlocks`/`replaceBlocks`. The Backspace keyboard handler has also been updated to use `fixColumnList` when removing the last block in a column.

Closes #1323
Fixes #2000

## Rationale

The current behavior of the remove block button in the side menu is unintuitive when used on the only block in a column, and should be improved.

## Changes

- Added `fixColumnList` function.
- Made `removeAndInsertBlocks` and Backspace keyboard handler use `fixColumnList`.

## Impact

Since `removeAndInsertBlocks` can now remove `column`/`columnList` nodes that it's actively traversing in `doc.descendants`, there's an edge case where I'm not sure what will happen.

E.g. consider this structure:

- `columnList`
  - `column`
    - `blockContainer` 1
  - `column`
    - `blockContainer` 2
  - `column`
    - `blockContainer` 3

If we call `removeBlocks` for both `blockContainer` 2 & 3, I'm not sure this will work properly. Worth adding tests for this?

EDIT: Added tests, works ok for both removing and replacing blocks 👍

## Testing

See above.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
@@ -37,6 +37,7 @@ export function removeAndInsertBlocks<
     ),
   );
   const removedBlocks: Block<BSchema, I, S>[] = [];
+  const columnListPositions = new Set<number>();
 
   const idOfFirstBlock =
     typeof blocksToRemove[0] === "string"
@@ -73,34 +74,14 @@ export function removeAndInsertBlocks<
     const oldDocSize = tr.doc.nodeSize;
 
     const $pos = tr.doc.resolve(pos - removedSize);
+
+    if ($pos.node().type.name === "column") {
+      columnListPositions.add($pos.before(-1));
+    } else if ($pos.node().type.name === "columnList") {
+      columnListPositions.add($pos.before());
+    }
+
     if (
-      node.type.name === "column" &&
-      node.attrs.id !== $pos.nodeAfter?.attrs.id
-    ) {
-      // This is a hacky work around to handle removing all columns in a
-      // columnList. This is what happens when removing the last 2 columns:
-      //
-      // 1. The second-to-last `column` is removed.
-      // 2. `fixColumnList` runs, removing the `columnList` and inserting the
-      // contents of the last column in its place.
-      // 3. `removedSize` increases not just by the size of the second-to-last
-      // `column`, but also by the positions removed due to running
-      // `fixColumnList`. Some of these positions are after the contents of the
-      // last `column`, namely just after the `column` and `columnList`.
-      // 3. `tr.doc.descendants` traverses to the last `column`.
-      // 4. `removedSize` now includes positions that were removed after the
-      // last `column`. This causes `pos - removedSize` to point to an
-      // incorrect position, as it expects that the difference in document size
-      // accounted for by `removedSize` comes before the block being removed.
-      // 5. The deletion is offset by 3, because of those removed positions
-      // included in `removedSize` that occur after the last `column`.
-      //
-      // Hence why we have to shift the start of the deletion range back by 3.
-      // The offset for the end of the range is smaller as `node.nodeSize` is
-      // the size of the second `column`. Since it's been removed, we actually
-      // care about the size of its children - a difference of 2 positions.
-      tr.delete(pos - removedSize + 3, pos - removedSize + node.nodeSize + 1);
-    } else if (
       $pos.node().type.name === "blockGroup" &&
       $pos.node($pos.depth - 1).type.name !== "doc" &&
       $pos.node().childCount === 1
@@ -111,12 +92,6 @@ export function removeAndInsertBlocks<
       tr.delete($pos.before(), $pos.after());
     } else {
       tr.delete(pos - removedSize, pos - removedSize + node.nodeSize);
-    }
-
-    if ($pos.node().type.name === "column") {
-      fixColumnList(tr, $pos.before(-1));
-    } else if ($pos.node().type.name === "columnList") {
-      fixColumnList(tr, $pos.before());
     }
 
     const newDocSize = tr.doc.nodeSize;
@@ -134,6 +109,8 @@ export function removeAndInsertBlocks<
         notFoundIds,
     );
   }
+
+  columnListPositions.forEach((pos) => fixColumnList(tr, pos));
 
   // Converts the nodes created from `blocksToInsert` into full `Block`s.
   const insertedBlocks = nodesToInsert.map((node) =>
--- a/packages/core/src/api/blockManipulation/commands/replaceBlocks/util/fixColumnList.ts
+++ b/packages/core/src/api/blockManipulation/commands/replaceBlocks/util/fixColumnList.ts
@@ -4,7 +4,7 @@ import { ReplaceAroundStep } from "prosemirror-transform";
 
 /**
  * Checks if a `column` node is empty, i.e. if it has only a single empty
- * block.
+ * paragraph.
  * @param column The column to check.
  * @returns Whether the column is empty.
  */
@@ -26,7 +26,7 @@ export function isEmptyColumn(column: Node) {
   return (
     column.childCount === 1 &&
     blockContainer.childCount === 1 &&
-    blockContent.type.spec.content === "inline*" &&
+    blockContent.type.name === "paragraph" &&
     blockContent.content.content.length === 0
   );
 }
@@ -63,7 +63,7 @@ export function removeEmptyColumns(tr: Transaction, columnListPos: number) {
     }
 
     if (isEmptyColumn(column)) {
-      tr.delete(columnPos, columnPos + column?.nodeSize);
+      tr.delete(columnPos, columnPos + column.nodeSize);
     }
   }
 }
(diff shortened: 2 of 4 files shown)


packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -1,15 +1,165 @@
...
+ */
+export function moveFirstBlockInColumn(
+  tr: Transaction,
+  blockBeforePos: ResolvedPos,
+): ResolvedPos {
+  const blockInfo = getBlockInfoFromResolvedPos(blockBeforePos);
+  if (!blockInfo.isBlockContainer) {
+    throw new Error(
+      "Invalid blockBeforePos passed to moveFirstBlockInColumn: does not point to blockContainer node.",
+    );
```
nperez0111:
```suggestion
    throw new Error(
      "Invalid blockBeforePos: does not point to blockContainer node.",
      { cause: { blockBeforePos } }
    );
```

We should standardize errors & error messages in the future, but I'm liking this sort of pattern

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -1,15 +1,165 @@
...
+ * block to the start of the next column if it's in the first column.
+ * Otherwise, moves the block to the end of the previous column.
+ * @param tr The transaction to apply changes to.
+ * @param blockBeforePos The position just before the first block in the column.
+ * @returns The position just before the block, after it's moved.
+ */
+export function moveFirstBlockInColumn(
+  tr: Transaction,
+  blockBeforePos: ResolvedPos,
+): ResolvedPos {
```
nperez0111: I read this like 3 times, and couldn't understand what it was doing

nperez0111: It's hard for me to evaluate this logic, I don't completely understand what this logic is trying to achieve. What I was sort of expecting from this was that given a transaction where some operation involving a column was involved, this would "fix up" the columns to make them make sense (i.e. remove columns/columnLists etc.). I'm unsure whether this is feasible, but I prefer something more generic, than trying to infer things while content is being moved around

matthewlipski: I also think a function that applies a transaction to check and clean up a given invalid columnList makes more sense tbh. 

The reason this function is the way it is, is just to maximize how much code can be reused from the existing keyboard handler. Since the logic is fairly complex, I felt like refactoring was gonna put me down a rabbithole and I didn't wanna spend too much time on it. If you think it's worth it though, I can spend a day or 2 refactoring it into a `fixColumnList` function or smth along those lines.

nperez0111: Maybe this is worth doing together. The way that I see this implemented would follow what you outlined in the comment here:

>  There are 3 different cases:
  a) remove entire column list (if no columns would be remaining)
  b) remove just a column (if no blocks inside a column would be remaining)
  c) keep columns (if there are blocks remaining inside a column)
  
So, what I'd do is these 3 passes in reverse order. Something like:

1. Check that a column has non-empty blocks within it
	- If it does, then it is valid, & bail
	- If not, then empty the column of any blocks it may have & continue
2. Check if a column is empty
    - If it is, remove it
    - If not, bail
3. Check that a columnList has only 1 column
    - If it only has 1, take out it's contents, & remove the columnList completely
    - If not, bail

Theoretically, at the end of all of these passes, we should be left with a valid column structure

nperez0111:
Where are we at with this @matthewlipski, am I missing the changes you made here?

matthewlipski:
Nope, #2113 merges into this PR though - do we wanna close this one and merge #2113 straight into main instead?

nperez0111:
That's what it was. Let's merge that into this & review altogether I think

packages/core/src/api/blockManipulation/commands/replaceBlocks/util/fixColumnList.ts
```diff
@@ -0,0 +1,163 @@
...
+  const blockContent = blockContainer.firstChild;
+  if (!blockContent) {
+    throw new Error("Invalid blockContainer: does not have child node.");
+  }
+
+  return (
+    column.childCount === 1 &&
+    blockContainer.childCount === 1 &&
+    blockContent.type.spec.content === "inline*" &&
+    blockContent.content.content.length === 0
```
nperez0111: This would return true for an empty list item, is that what we want?

packages/core/src/api/blockManipulation/commands/replaceBlocks/util/fixColumnList.ts
```diff
@@ -0,0 +1,163 @@
...
+      .resolve($columnListPos.pos + 1)
+      .posAtIndex(columnIndex);
+    const $columnPos = tr.doc.resolve(columnPos);
+    const column = $columnPos.nodeAfter;
+    if (!column || column.type.name !== "column") {
+      throw new Error("Invalid columnPos: does not point to column node.");
+    }
+
+    if (isEmptyColumn(column)) {
+      tr.delete(columnPos, columnPos + column?.nodeSize);
```
nperez0111:
```suggestion
      tr.delete(columnPos, columnPos + column.nodeSize);
```

It should be available right? And if it isn't we should skip the delete rather than trying to delete at all

packages/core/src/api/blockManipulation/commands/replaceBlocks/util/fixColumnList.ts
```diff
@@ -0,0 +1,163 @@
...
+  if (!columnList || columnList.type.name !== "columnList") {
+    throw new Error(
+      "Invalid columnListPos: does not point to columnList node.",
+    );
+  }
+
+  if (columnList.childCount > 2) {
+    // Do nothing if the `columnList` has at least two non-empty `column`s.
+    return;
+  }
```
nperez0111, line 103: The comment doesn't line up with the implementation. 

> has at least two

Which means that it _can_ have 2, therefore it should be:

```
  if (columnList.childCount >= 2) {
```

packages/core/src/api/blockManipulation/commands/replaceBlocks/util/fixColumnList.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  if (columnList.childCount < 2) {
+    throw new Error("Invalid columnList: contains fewer than two children.");
+  }
```
nperez0111, line 112: I don't understand why we are throwing an error here, shouldn't we fix it in this case?

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -70,26 +71,61 @@ export function removeAndInsertBlocks<
...
+      // incorrect position, as it expects that the difference in document size
+      // accounted for by `removedSize` comes before the block being removed.
+      // 5. The deletion is offset by 3, because of those removed positions
+      // included in `removedSize` that occur after the last `column`.
+      //
+      // Hence why we have to shift the start of the deletion range back by 3.
+      // The offset for the end of the range is smaller as `node.nodeSize` is
+      // the size of the second `column`. Since it's been removed, we actually
+      // care about the size of its children - a difference of 2 positions.
+      tr.delete(pos - removedSize + 3, pos - removedSize + node.nodeSize + 1);
```
nperez0111: I'm not certain how well this will all work. Does it really handle remove & inserting blocks properly? What about adding/removing within the column that may be "fix up"? Do we have test cases for this?

To avoid this:

Honestly, what I'd prefer is to do this in two passes:

1. where we do the full table scan, without trying to fix the columns along the way, capture the positions of any columns/columnLists (should just need columnLists right?)
2. do the column list cleanup afterwards for any columnLists found along the way

This avoids us needing to account for any changes in positioning based on the stuff we add/delete along the way. And fixes only after all of the changes have been applied (so we don't need to care where those changes are).

packages/xl-multi-column/src/test/commands/removeBlocks.test.ts
nperez0111, line 1: Let's add a test for remove & insert into the same column:

```
<column> para-1  </column>
```

& remove para-1 & insert para-2, to make sure that it doesn't try to remove the column

matthewlipski:
### IMPORTANT NOTE

There is a known issue with this PR, in that `fixColumnList` will remove all empty columns from a columnList (i.e. ones which only have a single empty paragraph). This is because we want to remove any empty columns that ProseMirror adds when it tries to "fix" columnLists that don't conform to the schema, and we don't differentiate between these and empty columns intentionally created by the user. This can result in poor UX as below:


https://github.com/user-attachments/assets/094ed331-f049-4fa9-9208-06f2ff3f9436

nperez0111 (approved):

PR #2111, opened 2025-10-16
refactor: less use of PartialBlocks and type cleanup

YousefED:
# Summary

Clean types and Partial / full block concepts

## Rationale

https://github.com/TypeCellOS/BlockNote/pull/2089#pullrequestreview-3330760564

## Changes

- Removes a lot of uses of `PartialBlock` internally, where it's not necessary
- Cleans up some of the internal types by removing / not exporting types

## Testing

Should mostly be covered by existing unit tests

- [ ] Manual testing still needed

## Screenshots/Video

-

## Checklist

## Additional Notes

TODO:
- manual testing

Follow up:
- clean more of blockToNode (there's logic there that handles partial content / blocks which is not needed anymore)
- potentially, we could move the definition of Block and Partial lock to Zod as well, and use a transformer to distinguish between input and output types

### DIFF
--- a/packages/core/src/api/exporters/html/externalHTMLExporter.ts
+++ b/packages/core/src/api/exporters/html/externalHTMLExporter.ts
@@ -7,6 +7,7 @@ import {
   InlineContent,
   InlineContentSchema,
   StyleSchema,
+  TableContent,
 } from "../../../schema/index.js";
 import {
   serializeBlocksExternalHTML,
@@ -57,7 +58,7 @@ export const createExternalHTMLExporter = <
     },
 
     exportInlineContent: (
-      inlineContent: InlineContent<I, S>[],
+      inlineContent: InlineContent<I, S>[] | TableContent<I, S>,
       options: { document?: Document },
     ) => {
       const domFragment = serializeInlineContentExternalHTML(
--- a/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
+++ b/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
@@ -4,8 +4,10 @@ import type { BlockNoteEditor } from "../../../../editor/BlockNoteEditor.js";
 import {
   BlockImplementation,
   BlockSchema,
+  InlineContent,
   InlineContentSchema,
   StyleSchema,
+  TableContent,
 } from "../../../../schema/index.js";
 import { UnreachableCaseError } from "../../../../util/typescript.js";
 import {
@@ -34,7 +36,7 @@ export function serializeInlineContentExternalHTML<
   S extends StyleSchema,
 >(
   editor: BlockNoteEditor<BSchema, I, S>,
-  blockContent: Block<BSchema, I, S>["content"],
+  blockContent: InlineContent<I, S>[] | TableContent<I, S>,
   serializer: DOMSerializer,
   options?: { document?: Document },
 ) {
@@ -43,8 +45,6 @@ export function serializeInlineContentExternalHTML<
   // TODO: reuse function from nodeconversions?
   if (!blockContent) {
     throw new Error("blockContent is required");
-  } else if (typeof blockContent === "string") {
-    nodes = inlineContentToNodes([blockContent], editor.pmSchema);
   } else if (Array.isArray(blockContent)) {
     nodes = inlineContentToNodes(blockContent, editor.pmSchema);
   } else if (blockContent.type === "tableContent") {
--- a/packages/core/src/api/clipboard/toClipboard/copyExtension.ts
+++ b/packages/core/src/api/clipboard/toClipboard/copyExtension.ts
@@ -81,7 +81,7 @@ function fragmentToExternalHTML<
 
     // Wrap in table to ensure correct parsing by spreadsheet applications
     externalHTML = `<table>${externalHTMLExporter.exportInlineContent(
-      ic as any,
+      ic,
       {},
     )}</table>`;
   } else if (isWithinBlockContent) {
--- a/packages/core/src/api/exporters/html/util/serializeBlocksInternalHTML.ts
+++ b/packages/core/src/api/exporters/html/util/serializeBlocksInternalHTML.ts
@@ -4,8 +4,10 @@ import { Block } from "../../../../blocks/defaultBlocks.js";
 import type { BlockNoteEditor } from "../../../../editor/BlockNoteEditor.js";
 import {
   BlockSchema,
+  InlineContent,
   InlineContentSchema,
   StyleSchema,
+  TableContent,
 } from "../../../../schema/index.js";
 import { UnreachableCaseError } from "../../../../util/typescript.js";
 import {
@@ -20,7 +22,7 @@ export function serializeInlineContentInternalHTML<
   S extends StyleSchema,
 >(
   editor: BlockNoteEditor<BSchema, I, S>,
-  blockContent: Block<BSchema, I, S>["content"],
+  blockContent: InlineContent<I, S>[] | TableContent<I, S>,
   serializer: DOMSerializer,
   blockType?: string,
   options?: { document?: Document },
@@ -30,8 +32,6 @@ export function serializeInlineContentInternalHTML<
   // TODO: reuse function from nodeconversions?
   if (!blockContent) {
     throw new Error("blockContent is required");
-  } else if (typeof blockContent === "string") {
-    nodes = inlineContentToNodes([blockContent], editor.pmSchema, blockType);
   } else if (Array.isArray(blockContent)) {
     nodes = inlineContentToNodes(blockContent, editor.pmSchema, blockType);
   } else if (blockContent.type === "tableContent") {
@@ -151,7 +151,7 @@ function serializeBlock<
   if (ret.contentDOM && block.content) {
     const ic = serializeInlineContentInternalHTML(
       editor,
-      block.content, // TODO
+      block.content,
       serializer,
       block.type,
       options,


packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
```diff
@@ -6,19 +6,21 @@ import {
...
-import { getPmSchema } from "../../../pmUtil.js";
+import { getBlockNoteSchema, getPmSchema } from "../../../pmUtil.js";
 
 export function insertBlocks<
   BSchema extends BlockSchema,
   I extends InlineContentSchema,
   S extends StyleSchema,
 >(
   tr: Transaction,
+  // TBD: allow PartialBlock here?
```
YousefED: I think we could move it a layer up to the BlockNoteEditor (and only allow partials there, not in the commands. However, for `updateBlock` I don't think that's feasible, so maybe better to keep it here

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -185,7 +183,7 @@ type OccupancyGrid = (RelativeCellIndices & {
  * @returns an {@link OccupancyGrid}
  */
 export function getTableCellOccupancyGrid(
-  block: BlockFromConfigNoChildren<DefaultBlockSchema["table"], any, any>,
```
YousefED, line 188: Stopped exporting `BlockFromConfigNoChildren` so we have more consistency in the codebase

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -177,25 +176,25 @@ function serializeBlock<
...
-  )) {
-    if (
-      !(name in props) &&
-      spec instanceof z.$ZodDefault &&
-      spec._zod.def.defaultValue !== undefined
-    ) {
-      (props as any)[name] = spec._zod.def.defaultValue;
-    }
-  }
+  // const props = block.props || {};
```
YousefED: TODO: this code can now be deleted

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -37,18 +37,19 @@ import type { TableHandlesProsemirrorPlugin } from "../extensions/TableHandles/T
...
-  CustomBlockNoteSchema,
-  InlineContentSchema,
-  InlineContentSpecs,
-  PartialInlineContent,
-  Styles,
-  StyleSchema,
-  StyleSpecs,
+import {
+  partialBlockToFullBlock,
+  type BlockIdentifier,
```
YousefED, line 42: fyi, some of the changes in `type` imports were to fix a circular dependency issue that was breaking the build during development

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1467,7 +1472,7 @@ export class BlockNoteEditor<
    * @returns The blocks, serialized as an HTML string.
    */
   public blocksToHTMLLossy(
-    blocks: PartialBlock<BSchema, ISchema, SSchema>[] = this.document,
+    blocks: Block<BSchema, ISchema, SSchema>[] = this.document,
```
YousefED, line 1475: breaking API change, should be documented in PR / release notes

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -535,10 +546,16 @@ export class TableHandlesView<
     }
 
     // Hide handles if the table block has been removed.
-    this.state.block = this.editor.getBlock(this.state.block.id)!;
```
YousefED, line 538: TODO: In the old version we were potentially setting `this.state.block` to undefined. Now, it will keep the old (unavailable block). can this break anything?

packages/core/src/schema/blocks/types.ts
```diff
@@ -314,35 +289,6 @@ export type SpecificBlock<
...
 };
 
-/** CODE FOR PARTIAL BLOCKS, analogous to above
- *
- * Partial blocks are convenience-wrappers to make it easier to
- *create/update blocks in the editor.
- *
- */
-
-export type PartialTableCell<
```
YousefED, line 324: moved to separate table file

packages/core/src/schema/blocks/types.ts
```diff
@@ -385,23 +331,6 @@ export type PartialBlockNoDefaults<
     children: PartialBlockNoDefaults<BSchema, I, S>[];
   }>;
 
-export type SpecificPartialBlock<
```
YousefED, line 388: unused, removed

packages/core/src/util/table.ts
```diff
@@ -36,6 +37,7 @@ export function mapTableCell<
         }
       : {
           type: "tableCell",
+          // FIXME: content can actually be Partial, we should probably handle this as well
```
YousefED: This comment is to document a finding Matthew earlier found while writing the partialToFull function

.vscode/settings.json
```diff
@@ -1,6 +1,9 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.addMissingImports": "explicit"
```
YousefED, line 5: fyi, this makes sure missing imports are added automatically on save, I kept finding myself to do this manually (or AI-powered)

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
nperez0111: These don't get run anywhere but from the editor instance, so imo it doesn't make too much a difference anyway

packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.ts
```diff
@@ -127,17 +132,19 @@ export function updateBlockTr<
     // currently, we calculate the new node and replace the entire node with the desired new node.
     // for this, we do a nodeToBlock on the existing block to get the children.
     // it would be cleaner to use a ReplaceAroundStep, but this is a bit simpler and it's quite an edge case
-    const existingBlock = nodeToBlock(blockInfo.bnBlock.node, pmSchema);
+    const newFullBlock = partialBlockToFullBlock(schema, block);
```
nperez0111:
```suggestion
    const newBlock = partialBlockToFullBlock(schema, block);
```

There are partial blocks & blocks. Full block gives a third name which is not helpful

packages/core/src/api/blockManipulation/tables/tables.ts
```diff
@@ -656,7 +654,8 @@ function isCellEmpty(
     return true;
   }
   if (isPartialTableCell(cell)) {
-    return isCellEmpty(cell.content);
+    // TODO: what happened here?
+    return isCellEmpty(cell);
```
nperez0111: Is this wrong? It is recursing, what you've written is an infinite loop no?

packages/core/src/api/blockManipulation/tables/tables.ts
nperez0111: While we are cleaning up types, I believe that:

```ts
BlockFromConfig<DefaultBlockSchema["table"], any, any>
```

Should only be:

```ts
BlockFromConfig<DefaultBlockSchema["table"]>
```

Most of the time we don't care about the other parameters, and have to remove their defaults with any. This goes for most instances of B, I, S

packages/core/src/api/exporters/html/externalHTMLExporter.ts
```diff
@@ -62,7 +62,7 @@ export const createExternalHTMLExporter = <
     ) => {
       const domFragment = serializeInlineContentExternalHTML(
         editor,
-        inlineContent as any,
+        inlineContent as unknown as never,
```
nperez0111: ~~ `as never`? Why so? ~~
nvm, I see the todo

packages/core/src/api/nodeConversions/blockToNode.ts
```diff
@@ -322,16 +320,10 @@ function blockOrInlineContentToContentNode(
...
+  block: Block<any, any, any>,
   schema: Schema,
   styleSchema: StyleSchema = getStyleSchema(schema),
 ) {
-  let id = block.id;
-
-  if (id === undefined) {
-    id = UniqueID.options.generateID();
-  }
-
```
nperez0111, line 334: I'm actually ok with this function generating ids if the block doesn't have one. I think this makes the caller's life more difficult to be forced to present an ID

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
```diff
@@ -278,7 +282,14 @@ export class TableHandlesView<
...
     );
 
-    if (editorHasBlockWithType(this.editor, "table")) {
+    if (
+      blockHasType(
+        block,
+        this.editor,
+        "table",
+        defaultBlockSpecs.table.config.propSchema,
+      )
```
nperez0111: Why are we changing this?

Am I right to assume that this would mean that it can now only match the block if it is both a table & conforming to the propSchema? Why does the propSchema matter here? What if someone extends the table to have additional props, wouldn't this break?

packages/core/src/extensions/TableHandles/TableHandlesPlugin.ts
nperez0111: It isn't clear to me why this was modified, but it certainly is a change in behavior that would need to be validated

packages/core/src/schema/partialBlockToFullBlock.ts
```diff
@@ -0,0 +1,258 @@
...
+  } as Block<BSchema, I, S>;
+}
+
+export function partialBlocksToFullBlocks<
+  BSchema extends BlockSchema,
+  I extends InlineContentSchema,
+  S extends StyleSchema,
+>(
+  // TODO: I think this should be a CustomBlockNoteSchema,
+  // but that breaks docxExporter etc
```
nperez0111: Yep, it should be

packages/core/src/api/nodeConversions/blockToNode.ts
YousefED: if there's no `id`, then `block` doesn't conform to `Block`, so that would mean an issue higher up in the call stack

nperez0111: Right, but I mean to say that blockToNode can allow id to not be present, since it is either we generate an id here or at call sites. It will be randomly generated anyway, so I don't think it actually presents a problem

PR #2112, opened 2025-10-17
fix: `BlockTypeSelect` item filtering based on schema

matthewlipski:
# Summary

This PR makes the `BlockTypeSelect` filter its default items based on whether they're allowed in the schema, same as how the slash menu already works.

## Rationale

You can configure things like allowed levels for the default heading block, but the `BlockTypeSelect` won't reflect this change.

## Changes

- Refactored `blockTypeSelectItems` to `getDefaultBlockTypeSelectItems`.
- Fixed typo in default block configuration example.

## Impact

N/A

## Testing

Not sure how feasible testing smth like this is atm. We'd have to set up component testing across all of the examples.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/examples/03-ui-components/03-formatting-toolbar-block-type-items/src/App.tsx
+++ b/examples/03-ui-components/03-formatting-toolbar-block-type-items/src/App.tsx
@@ -4,7 +4,7 @@ import { BlockNoteView } from "@blocknote/mantine";
 import "@blocknote/mantine/style.css";
 import {
   FormattingToolbarController,
-  getDefaultBlockTypeSelectItems,
+  blockTypeSelectItems,
   useCreateBlockNote,
   BlockTypeSelectItem,
   FormattingToolbar,
@@ -60,7 +60,7 @@ export default function App() {
             // Sets the items in the Block Type Select.
             blockTypeSelectItems={[
               // Gets the default Block Type Select items.
-              ...getDefaultBlockTypeSelectItems(editor),
+              ...blockTypeSelectItems(editor.dictionary),
               // Adds an item for the Alert block.
               {
                 name: "Alert",
--- a/examples/03-ui-components/13-custom-ui/src/MUIFormattingToolbar.tsx
+++ b/examples/03-ui-components/13-custom-ui/src/MUIFormattingToolbar.tsx
@@ -1,6 +1,6 @@
 import { Block } from "@blocknote/core";
 import {
-  getDefaultBlockTypeSelectItems,
+  blockTypeSelectItems,
   useBlockNoteEditor,
   useEditorContentOrSelectionChange,
 } from "@blocknote/react";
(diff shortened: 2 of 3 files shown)


matthewlipski:
@nperez0111 sort of related but only tangentially - I noticed that switching to the default block configuration example and then back to any other one doesn't revert the heading prop schema back to having 6 levels. It's stuck having the 3 levels.

I traced the issue back to `BlockNoteViewComponent`, somehow the `editor` passed to it doesn't have the correct heading prop values. This is especially weird because changing the block schema in e.g. the custom alert example and switching to the default schema example correctly updates to only have the default blocks and no alert.

Could you take a look at it?

packages/react/src/components/FormattingToolbar/DefaultSelects/BlockTypeSelect.tsx
```diff
@@ -41,146 +41,123 @@ export type BlockTypeSelectItem = {
...
+  }
+
+  if (editorHasBlockWithType(editor, "heading", { level: "number" })) {
+    (
+      editor.schema.blockSchema.heading.propSchema.level.values || [1, 2, 3]
+    ).forEach((level) => {
+      items.push({
+        name: editor.dictionary.slash_menu[
+          `heading${level === 1 ? "" : "_" + level}` as keyof typeof editor.dictionary.slash_menu
+        ].title,
```
nperez0111: Let's leave a todo here that we need to update this, we are changing this in the zod pr if I remember correctly, to just be normalized to `heading_X` instead of `1` being special cased

packages/react/src/components/FormattingToolbar/DefaultSelects/BlockTypeSelect.tsx
nperez0111, line 1: Looking at this with fresh eyes again, and I'm thinking that the `editorHasBlockWithType` type guards should maybe be gated within the `BlockTypeSelect` component's `filteredItems` instead. Take this for example:
```ts
{
      name: editor.dictionary.slash_menu.quote.title,
      type: "quote",
      icon: RiQuoteText,
      isSelected: (block) => block.type === "quote",
    }
```

Like, can't we know exactly what block this is trying to select based on the `type` & optional `props` object? This would allow us to filter the list down & cut down on the type guards needed.

Maybe this is over complicating it, but having a filter for this sort of thing is sort of useful generically, like in the slash menu, block type select & other places surely

packages/react/src/components/FormattingToolbar/DefaultSelects/BlockTypeSelect.tsx
```diff
@@ -27,160 +27,106 @@ import {
...
   icon: IconType;
-  isSelected: (
-    block: Block<BlockSchema, InlineContentSchema, StyleSchema>,
-  ) => boolean;
 };
 
-export const blockTypeSelectItems = (
-  dict: Dictionary,
+export const getDefaultBlockTypeSelectItems = (
+  editor: BlockNoteEditor<any, any, any>,
```
nperez0111: We can keep this as the same shape that it was before right @matthewlipski, I don't see a reason to make this a breaking change when everything just depends on the dictionary

PR #2113, opened 2025-10-17
feat: Clean up invalid multi-column handling

matthewlipski:
# Summary

This PR replaces the `moveFirstBlockInColumn` function with `fixColumnList`. While the old function reused a lot of code from the existing backspace keyboard handler for the first block in a column, it was unclear and hard to understand. The new function is much simpler, as it just takes a `columnList` node and deletes any empty columns. If all columns are empty, the whole node is deleted.

## Rationale

Makes the code easier to understand and maintain.

## Changes

- Removed `moveFirstBlockInColumn`.
- Added `fixColumnList`.
- Updated multi-column keyboard handler and `removeAndInsertBlocks`.

## Impact

N/A

## Testing

All existing tests for removing/replacing blocks/columns pass. All cases for backspace handling have been manually tested.

## Screenshots/Video

N/A

## Checklist


packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -8,156 +8,162 @@ import type {
...
+ * - Removes all empty `column` nodes. A `column` node is empty if it has only
+ * a single empty block.
+ * - If all but one `column` nodes are empty, replaces the `columnList` with
+ * the content of the non-empty `column`.
+ * - If all `column` nodes are empty, removes the `columnList` entirely.
+ * @param tr The `Transaction` to add the changes to.
+ * @param columnListPos
+ * @returns The position just before the `columnList` node.
+ */
+export function fixColumnList(tr: Transaction, columnListPos: number) {
```
nperez0111: I'd like this function & it's dependents to be in a different file. I think this would be a good candidate for a bit more granular of a unit test to make sure it is doing everything we expect out of it.

I know that you've already got some tests for the higher level API. But I just wanna make sure we know exactly what this is doing at this level, since we will likely use it for more than just the replaceBlocks API

packages/core/src/extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -123,8 +123,35 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
+              tr.delete(
+                blockInfo.bnBlock.beforePos,
+                blockInfo.bnBlock.afterPos,
+              );
+
+              if ($columnPos.index() === 0) {
+                // Fix `columnList` and insert the block before it.
+                fixColumnList(state.tr, columnListPos);
+                tr.insert(columnListPos, fragment);
+                tr.setSelection(TextSelection.create(tr.doc, columnListPos));
```
nperez0111: I've been preferring [`TextSelection.near`, since it will find the closest valid position which is more tolerant](https://prosemirror.net/docs/ref/#state.Selection^near)

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -8,156 +8,162 @@ import type {
...
-      tr.setSelection(TextSelection.between(pos, pos));
-
-      return pos;
-    }
-  } else if (shouldRemoveColumn) {
-    if (isFirstColumn) {
-      // delete column
-      tr.delete(column.bnBlock.beforePos, column.bnBlock.afterPos);
+  const firstColumnEmpty = isEmptyColumn(firstColumn);
+  const lastColumnEmpty = isEmptyColumn(lastColumn);
```
nperez0111: I'd like to see what happens when the first & last column are the same

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -8,156 +8,162 @@ import type {
...
   }
 
-  const column = parentBlockInfo;
-  const columnList = getParentBlockInfo(tr.doc, column.bnBlock.beforePos);
-  if (columnList?.blockNoteType !== "columnList") {
+  for (
+    let columnIndex = columnList.childCount - 1;
+    columnIndex >= 0;
+    columnIndex--
+  ) {
```
nperez0111: Is this not just a columnList.children.forEach?

packages/core/src/extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -24,7 +24,7 @@ export const KeyboardShortcutsExtension = Extension.create<{
   addKeyboardShortcuts() {
     // handleBackspace is partially adapted from https://github.com/ueberdosis/tiptap/blob/ed56337470efb4fd277128ab7ef792b37cfae992/packages/core/src/extensions/keymap.ts
     const handleBackspace = () =>
-      this.editor.commands.first(({ chain, commands }) => [
```
nperez0111, line 27: The tr should be pulled from the individual command, not this `first` command

`commands.command(({ tr }) =>`

packages/core/src/extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -123,8 +123,35 @@ export const KeyboardShortcutsExtension = Extension.create<{
               return false;
             }
 
+            const $blockPos = state.doc.resolve(blockInfo.bnBlock.beforePos);
```
nperez0111: I prefer everything relative to `tr` since `state` can be stale at this point

nperez0111:
@matthewlipski, Unsure what the right thing to do here is, but when dragging a block out of a column, I'd expect the column to collapse unless it had content. Maybe the fixColumns should run as an appendTransaction in a plugin so we don't have to specifically run it everywhere that a column may be touched in some way?

https://github.com/user-attachments/assets/40e18112-cad1-49fa-b947-0b7e09a5de2b

matthewlipski:
Good point, though I would do that in a separate PR though since I want to get it out soon so we can close #1323, and I feel like figuring out how to nicely get the positions of changed tables after each transaction might take some time.

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
matthewlipski: I decided to throw an error in this case. Since if the columnList only has a single non-empty column, this is both an unexpected case that should never happen (PM should automatically add an empty column to make the columnList valid), and it would not be covered by the existing cases.

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
matthewlipski: Eh imo it's way more intuitive in this case to use a for loop since we can go backwards. That way we don't have to worry about having to terminate early. 

The bigger issue though is that `forEach` doesn't account for children being removed - e.g. removing child 0 should mean that child 3 now becomes child 2, but it remains child 3 in `forEach`, and the index ends up being wrong. So overall it's just way easier to use the reverse for loop.

Issue #1016, opened 2024-08-20
Side Menu showing when the cursor is outside the editor


christojeffrey:
**Describe the bug**
the side menu is showing up when the mouse cursor is not inside the editor. it's happening when the mouse is correct vertically, but wrong horizontally. I think it's because the side handle the vertical part of the cursor. 

**To Reproduce**
check the documentation, find a page with an editor inside. hover your mouse to the left of the editor, up to the sidebar for example.
![image](https://github.com/user-attachments/assets/15d05476-36d7-42db-b0e8-18ce33da6026)

**Misc**
- Browser: chrome
- [ ] I'm a [sponsor](https://github.com/sponsors/YousefED) and would appreciate if you could look into this sooner than later 💖

christojeffrey:
looks like the code that handle this is somewhere around here 
https://github.com/TypeCellOS/BlockNote/blob/6d9b8447a106307fb4b53d2032d523aca977b1e1/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts#L344-L361

YousefED:
It's intentional for our internal use-cases, but I agree an option for this would be nice.

What are you building with BlockNote btw? (always curious)

andreyleskov:
have the same issue, we put blocknote inside a note card in a whiteboard

christojeffrey:
> looks like the code that handle this is somewhere around here
> 
> https://github.com/TypeCellOS/BlockNote/blob/6d9b8447a106307fb4b53d2032d523aca977b1e1/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts#L344-L361

I think I can try to do it by limiting the drag handle to only be shown when the cursor is inside the parent div

![image](https://github.com/user-attachments/assets/3bf5cb00-2c18-4fbd-8aaf-a918d157cd31)

yantakus:
I have the same problem. I have a sidebar to the left of the editor. When user hovers over menu items in the sidebar this drag handle appears and it's annoying. Is there any way to fix this?

meko-dev:
+1 for making this behavior optional.
Also, thanks for your hard work on this editor. It’s the only one I’ve been able to use without running into any issues.

nperez0111:
[...]

PR #2117, opened 2025-10-20
feat: add configurable event handling for SideMenu

Aslam97:
This PR demonstrates how to scope the mouse tracking inside the editor by providing an opt-in mechanism to use ProseMirror's `handleDOMEvents` instead of root-level event listeners.

# Summary

Adds a configurable `useHandleDOMEvents` option to the SideMenu plugin, allowing users to choose between two event handling strategies for mouse tracking.

## Rationale

Issue here https://github.com/TypeCellOS/BlockNote/issues/1016

## Checklist

### DIFF
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -139,6 +139,8 @@ export class SideMenuView<
 
   public isDragOrigin = false;
 
+  public useHandleDOMEvents = false;
+
   constructor(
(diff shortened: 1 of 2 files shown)


nperez0111:
@Aslam97 thanks for this, but I don't think this is the direction of where we want to go in exposing this configuration. For one, this should be known upfront, so the useEffect feels like a hack, than a supported solution.

Secondly, I don't know that it is worth it to switch between which API to use just for this. Instead of listening to the root level element this should switch to the editor dom element instead & listen on that, so no need for handleDOMEvents api which does essentially the same thing as an addEventListener would

PR #2119, opened 2025-10-20
fix: Parsing priority for custom inline content and styles

### LINKED ISSUE #2064: Drag and drop with clear the inline content attribute
Describe the bug
After dragging and dropping a block, the inline content attribute dissapeared (reset to a default plain text).

To Reproduce
Simply use the demo on the [custom inline content page](https://www.blocknotejs.org/docs/features/custom-schemas/custom-inline-content)  to reproduce, no other steps or environments necessary.

Misc
Node version: N/A
Package manager: N/A
Browser: Chrome

Image

matthewlipski:
# Summary

This PR fixes an issue where default styles like text and background color were prioritised over custom inline content & styles.

This could be seen in the mentions inline content example, where, copying and pasting a mention would parse it as a background color style instead of a mention.

Closes #2064

## Rationale

N/A

## Changes

- Added `runsBefore` to custom inline content and style implementations.
- Made custom styles and inline content run before default content, including parse rules.
- Removed old unused text and background color marks.
- Made text and background color marks non-consuming, so other parse rules can still run on the same elements.

## Impact

This PR should fix any potential issues that come from custom inline content and styles priority.

## Testing

- Changed Mention in the unit test schema to have `backgroundColor` inline style, to replicate the issue fixed in this PR.
- Added unit test case for copying & copy/pasting Mention.
- Updated snapshot for exporting Mention.

## Screenshots/Video

N/A

## Checklist


packages/core/src/schema/schema.ts
```diff
@@ -84,8 +84,12 @@ export class CustomBlockNoteSchema<
...
-      if (specDef.implementation.runsBefore) {
+    for (const [key, specDef] of Object.entries({
+      ...this.opts.blockSpecs,
+      ...this.opts.inlineContentSpecs,
+      ...this.opts.styleSpecs,
+    })) {
+      if (specDef.implementation?.runsBefore) {
         dag.set(key, new Set(specDef.implementation.runsBefore));
       } else {
         defaultSet.add(key);
```
nperez0111, line 95: Oh, by default everything gets added to the default set, so we actually don't need to set the defaults, we can just leave them as it

packages/core/src/schema/styles/createSpec.ts
```diff
@@ -46,6 +47,7 @@ export function getStyleParseRules<T extends StyleConfig>(
   if (customParseFunction) {
     rules.push({
       tag: "*",
+      consuming: false,
```
nperez0111, line 53: May be worth adding a comment

```suggestion
      // By default, styles can overlap each other, so the rules should not completely consume the element they parse (which can have multiple styles)
      consuming: false,
```

packages/core/src/schema/styles/createSpec.ts
```diff
@@ -107,7 +109,9 @@ export function createStyleSpec<const T extends StyleConfig>(
   });
 
   return createInternalStyleSpec(styleConfig, {
+    ...styleImplementation,
     mark,
+    runsBefore: styleImplementation.runsBefore || ["default"],
```
nperez0111: We can actually remove these now, given that it does the defaults for us

matthewlipski: The `...styleImplementation` too?

nperez0111: Nah, we will keep that.

tests/src/unit/core/formatConversion/parse/__snapshots__/html/googleDocs.json
```diff
@@ -4,6 +4,7 @@
     "content": [
       {
         "styles": {
+          "backgroundColor": "transparent",
```
YousefED, line 7: do we need the "transparent"? or can we get rid of this?

tests/src/unit/core/clipboard/copy/copyTestInstances.ts
```diff
@@ -650,4 +650,30 @@ export const copyTestInstancesHTML: TestInstance<
     },
     executeTest: testCopyHTML,
   },
+  {
+    testCase: {
```
YousefED, line 654: @matthewlipski does this test also test pasting / parsing?

Are we confident the test that was added would have failed pre-fix? (and thus will catch potential regressions correctly?

tests/src/unit/core/formatConversion/parse/__snapshots__/html/googleDocs.json
matthewlipski: For context, this is a result of us now being able to parse multiple styles from a single HTML element, as GDocs puts all text decorations into a single `style` attribute.

Right now, any `style` attribute on a `span` element with `color`/`backgroundColor` fields gets parsed as a `TextColor`/`BackgroundColor` style respectively, regardless of the value. I think it makes sense to filter out values like `transparent` or `unset` though, but this is out of scope for this PR - I think we should make a new issue for it.

tests/src/unit/core/clipboard/copy/copyTestInstances.ts
matthewlipski: Yeah I reverted the other changes while keeping the test and it did indeed fail. It does also test copy/paste (the copy/paste equality tests also use the copy test cases).

It's not obvious but export/parse is also tested. We previously had tests for mentions, the reason they didn't catch this issue though was because the mentions were unstyled. After adding `dom.style.backgroundColor = "red"` to the mention DOM element in `testSchema.ts`, the export/parse equality test failed without the other changes from this PR.

tests/src/unit/core/formatConversion/parse/__snapshots__/html/googleDocs.json
nperez0111: @matthewlipski can you make that issue since you've the most context on it?

PR #2125, opened 2025-10-23
fix: `setSelection` not using `MultipleNodeSelection`

matthewlipski:
# Summary

This PR rewrites `setSelection` to use `MultipleNodeSelection` when spanning multiple blocks. When the start and end blocks are the same, the entire block content is selected, i.e.:

- When a block has inline content, a `TextSelection` spanning it is used.
- When a block has table content, a `CellSelection` spanning all cells is used.
- When a block has no content, a `NodeSelection` spanning the `blockContent` node is used.

## Rationale

Currently, `setSelection` uses a `TextSelection` when spanning multiple nodes. This can cause issues when the start/end blocks do not have content, as it's not possible to find a `TextSelection` that spans them.

## Changes

- Rewrote `setSelection`
- Updated `getSelection`
- Added `fromJSON` to `MultipleNodeSelection`
- Updated `moveBlocks` to work with a `MultipleNodeSelection`.
- Added `"Mod-a"` keyboard handler which calls `setSelection` on the whole document. The default `"Mod-a"` shortcut attempts to create a `TextSelection`, which has the aforementioned issues.

## Impact

Should not be a breaking change as the BlockNote API is unchanged.

## Testing

We already have tests for `getSelection`/`setSelection`. When implementing the changes, I ran them against the tests to ensure they were correct.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/core/src/api/blockManipulation/selections/selection.ts
+++ b/packages/core/src/api/blockManipulation/selections/selection.ts
@@ -1,8 +1,13 @@
-import { TextSelection, type Transaction } from "prosemirror-state";
-import { TableMap } from "prosemirror-tables";
+import {
+  NodeSelection,
+  TextSelection,
+  type Transaction,
+} from "prosemirror-state";
+import { CellSelection } from "prosemirror-tables";
 
 import { Block } from "../../../blocks/defaultBlocks.js";
 import { Selection } from "../../../editor/selectionTypes.js";
+import { MultipleNodeSelection } from "../../../extensions-shared/MultipleNodeSelection.js";
 import {
   BlockIdentifier,
   BlockSchema,
@@ -15,7 +20,7 @@ import {
   prosemirrorSliceToSlicedBlocks,
 } from "../../nodeConversions/nodeToBlock.js";
 import { getNodeById } from "../../nodeUtil.js";
-import { getBlockNoteSchema, getPmSchema } from "../../pmUtil.js";
+import { getPmSchema } from "../../pmUtil.js";
 
 export function getSelection<
   BSchema extends BlockSchema,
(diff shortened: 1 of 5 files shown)


packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -24,16 +29,24 @@ export function getSelection<
...
 
-  const $startBlockBeforePos = tr.doc.resolve(
-    getNearestBlockPos(tr.doc, tr.selection.from).posBeforeNode,
-  );
-  const $endBlockBeforePos = tr.doc.resolve(
-    getNearestBlockPos(tr.doc, tr.selection.to).posBeforeNode,
-  );
+  const $startBlockBeforePos =
+    tr.selection instanceof MultipleNodeSelection
+      ? tr.selection.$anchor
```
nperez0111, line 38: anchor can be different than `from`, depending on direction, so we need to decide whether we are using `$from` or `$anchor`

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -140,84 +153,123 @@ export function setSelection(
...
+        const lastCellAfterPos = lastRowAfterPos - 1;
+
+        tr.setSelection(
+          CellSelection.create(
+            tr.doc,
+            firstCellBeforePos,
+            lastCellAfterPos -
+              tr.doc.resolve(lastCellAfterPos).nodeBefore!.nodeSize,
+          ),
+        );
```
nperez0111, line 209: Should it be a cell selection or a node selection over the table 🤔

nperez0111: I prefer having less special cases for things like tables & columns, so if we can simplify it. Let's do that instead

packages/core/src/api/blockManipulation/selections/selection.ts
```diff
@@ -140,84 +153,123 @@ export function setSelection(
...
+      const firstBlockBeforePos = blockInfo.bnBlock.beforePos + 1;
+      const lastBlockAfterPos = blockInfo.bnBlock.afterPos - 1;
+
+      // Run recursively as the column may only have one block.
+      setSelection(
+        tr,
+        tr.doc.resolve(firstBlockBeforePos).nodeAfter!.attrs.id,
+        tr.doc.resolve(lastBlockAfterPos).nodeBefore!.attrs.id,
+      );
+    }
```
nperez0111, line 247: Instead of always searching for the correct position, can we use helpers to get this so we don't have to write out this logic every time.

In this case specifically though, I think we should be using the getNearestBlock APIs so we don't have to care about the specific strcutures. If we ever change this for columns it will be a hell of a refactor & I'd rather things not be special cased everywhere

nperez0111:
@matthewlipski I'm going to absorb this into the Locations API as part of #2044 instead, so I will close this

PR #2126, opened 2025-10-23
fix: Missing Delete key handlers

matthewlipski:
# Summary

This PR adds full Delete key handling, building on the existing logic. The additional cases are mainly for nested blocks and multi-column blocks.

R.e. multi-column, Backspace key handling has also been slightly changed to be more consistent. Specifically, whenever a block moves into a column after pressing Backspace or Delete, it now always gets added as a new block. Previously, it would sometimes be merged with the current block.

## Rationale

While Delete isn't used as much as Backspace, proper handling for it is still needed.

## Changes

- Added cases for Delete key handling.
- Added a case for Backspace key handling when right after `columnList`.
- Added `getNextBlockInfo` helper.
- Updated `getParentBlockInfo` helper.

## Impact

N/A

## Testing

Added e2e tests for Delete key handling:

- Basic functionality (merging adjacent blocks, deleting selection, etc.)
- Nested blocks (merging child with parent, merging next block with child, etc.)
- Multi-column (moving the first block from the next column into the current one, moving the first block in a column list out of it, etc.)

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
+++ b/packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
@@ -93,6 +93,9 @@ export const KeyboardShortcutsExtension = Extension.create<{
               state.doc,
               blockInfo.bnBlock.beforePos,
             );
+            // If the previous block has no inline content, it can't be merged.
+            // It's instead deleted, which is done later in the chan, so we
+            // return early here.
             if (
               !prevBlockInfo ||
               !prevBlockInfo.isBlockContainer ||


packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -383,50 +377,150 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
+        // Deletes next block if it contains no content and isn't a table,
+        // when the selection is empty and at the end of the block. Moves the
+        // current block into the deleted block's place.
+        () =>
+          commands.command(({ state }) => {
+            const blockInfo = getBlockInfoFromSelection(state);
+
+            if (!blockInfo.isBlockContainer) {
+              // TODO
+              throw new Error(`todo`);
```
nperez0111: todo

matthewlipski: Since this case is for deleting columns/column lists, I'd rather wait until #2110 is merged before fixing this.

packages/core/src/api/blockManipulation/commands/mergeBlocks/mergeBlocks.ts
```diff
@@ -19,8 +19,11 @@ export const getParentBlockInfo = (doc: Node, beforePos: number) => {
 
   // get start pos of parent
   const parentBeforePos = $pos.posAtIndex(
-    $pos.index($pos.depth - 1),
-    $pos.depth - 1,
+    // TODO: This works for blockContainer nodes as we need to traverse 2
+    // nesting levels due to the blockGroup in between. However, that's not the
+    // case for columns.
+    $pos.index($pos.depth - 2),
+    $pos.depth - 2,
```
nperez0111: Yea, let's do this

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -83,6 +89,18 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
+              state.doc,
+              blockInfo.bnBlock.beforePos,
+            );
+            if (
+              !prevBlockInfo ||
+              !prevBlockInfo.isBlockContainer ||
+              prevBlockInfo.blockContent.node.type.spec.content !== "inline*"
+            ) {
+              return false;
+            }
```
nperez0111, line 105: can we add a todo on why we early return here

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -130,31 +184,22 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
                   TextSelection.near(tr.doc.resolve(columnListPos)),
                 );
               } else {
-                // Insert the block at the end of the first column and fix
-                // `columnList`.
-                tr.insert($columnPos.pos - 1, fragment);
+                tr.insert($columnPos.pos - 1, blockInfo.bnBlock.node);
                 tr.setSelection(
-                  TextSelection.near(tr.doc.resolve($columnPos.pos - 1)),
+                  TextSelection.near(tr.doc.resolve($columnPos.pos)),
```
nperez0111, line 204: text selection is after now? Intentional?

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -294,50 +333,397 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
+                    .insertContentAt(blockInfo.bnBlock.afterPos, childBlocks)
+                    .run();
+                }
+
+                return chain()
+                  .deleteRange({
+                    from: nextBlockInfo.bnBlock.beforePos,
+                    to: nextBlockInfo.bnBlock.afterPos,
+                  })
+                  .run();
```
nperez0111: These two can be merged right. insetContentAt accepts null no-ops right

matthewlipski:
> How do you test the delete key on macOS? I'm unsure how to do it

I think it's Fn+Backspace but not 100% sure since I use a keyboard with Windows modifiers

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
matthewlipski: Not really a TODO, we return early intentionally as we can't merge blocks unless they both have inline content. Otherwise, we delete the previous block, which is handled in a different command.

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
matthewlipski: Wdym? Like it's after `fixColumnList` or the selection itself is different?

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
matthewlipski: Also if the previous block is a column list, that also gets handled separately

nperez0111: Sorry not a todo, a comment

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
nperez0111: This changes the position of where the selection is restored to (at least from the previous behavior) is it intentional?

matthewlipski: Ah `TextSelection.near` resolves it to the same position with the -1 or without. I double checked and the behaviour is correct either way.

nperez0111: Alright good enough for me

nperez0111 (approved):
lgtm

Issue #593, opened 2024-02-23
Tailwind example in docs


matthewlipski:
Currently, the "Adding DOM Attributes" page/feature is pretty useless since you can just access the elements using their CSS classes anyway. However, it is somewhat useful for applying Tailwind classes, so the docs page should be changed into one specific for how to use TailwindCSS with BlockNote.

PR #2128, opened 2025-10-24
fix: Firefox invisible text cursor after dropping blocks

### LINKED ISSUE #2036: Drag and drop caret disappears
**Describe the bug**
After dragging and dropping a block the caret for input disappears.

**To Reproduce**
Simply use the demo on the blocknote homepage to reproduce, no other steps or environments necessary.
I was experiencing this in my local project and went looking for examples to see if it was specific to me, but it seems it applies to the examples.

**Misc**

- Node version: N/A
- Package manager: N/A
- Browser: Firefox 142.0

matthewlipski:
# Summary

This PR blurs the editor after blocks are dragged and dropped. This is mainly to fix an issue in Firefox that causes the text cursor to become invisible after drag & drop. However, it's also generally better for UX, as having the editor keep focus makes the formatting toolbar show up. Notion also doesn't show the toolbar after blocks are dropped.

R.e. the Firefox invisible cursor issue, this was a bit of a rabbithole.

First off, this also happens with TipTap's drag handle implementions (vanilla JS, React, and Vue):

[Vanilla JS Drag Handle](https://tiptap.dev/docs/editor/extensions/functionality/drag-handle)

[React Drag Handle](https://tiptap.dev/docs/editor/extensions/functionality/drag-handle-react)

[Vue Drag Handle](https://tiptap.dev/docs/editor/extensions/functionality/drag-handle-vue)

Then I found that TinyMCE also had this issue ages ago, but was since fixed:

[TinyMCE #2823](https://github.com/tinymce/tinymce/issues/2823)

[TinyMCE #1497](https://github.com/tinymce/tinymce/issues/1497)

Unfortunately, despite searching for it, I couldn't find the PR with the fix for these issues.

There's also a ProseMirror bug report that's been open for years regarding what seems to be the same issue:

[ProseMirror #593](https://github.com/ProseMirror/prosemirror/issues/583)

So overall, looks like this is just a bug with Firefox, and the best we can do is work around it. Blurring the editor and refocusing it works. This PR just blurs it (as this more closely matches Notion's UX), but could be changed to refocus it after too.

Closes #2036

## Rationale

The caret disappearing in Firefox is a major annoyance, and users have to click somewhere outside the editor, then back on it, to make it reappear.

## Changes

Made editor DOM element get blurred at the end of `blockDragEnd`.

## Impact

N/A

## Testing

Is there even a way to test for this? Even with e2e test screenshots, the text cursor blinks so they seem like they'd be flaky at best.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -1107,6 +1107,13 @@ export class BlockNoteEditor<
     this.prosemirrorView.focus();
   }
 
+  public blur() {
+    if (this.headless) {
+      return;
+    }
+    this.prosemirrorView.dom.blur();
+  }
+
   public onUploadStart(callback: (blockId?: string) => void) {
     this.onUploadStartCallbacks.push(callback);
 
--- a/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
@@ -739,7 +739,7 @@ export class SideMenuProsemirrorPlugin<
       this.view.isDragOrigin = false;
     }
 
-    this.editor.domElement?.blur();
+    this.editor.blur();
   };
   /**
    * Freezes the side menu. When frozen, the side menu will stay


nperez0111:
Good sleuthing @matthewlipski, I've been annoyed by this for a while myself using firefox daily.

Only question is why we'd blur the domElement, instead of doing `editor.blur()` (editor is `BlockNoteEditor` here). Does it not work for that? I've just not seen this API really used & it could be unclear.

In either case, we should definitely leave ourselves a comment linking to this PR since you've captured a lot of info here

matthewlipski:
So there actually is no `editor.blur` atm, only `editor.focus`. Even under the PM view there is no `blur`. I'll add it to the editor API, but we still have to use `editor.prosemirrorView.dom.blur`.

dlgoodchild:
Nice investigative work @matthewlipski ....looking forward to it, thanks!

YousefED:
@matthewlipski nice work.

> in either case, we should definitely leave ourselves a comment linking to this PR since you've captured a lot of info here

I think that's still a good idea!

PR #2143, opened 2025-10-31
feat: Major Extensions & UI Refactor

nperez0111:
# Problem

Currently, there is not much consistency in how the editor's feature set is extended, both internally and externally. Things like UI extensions (formatting toolbar, side menu, etc) are baked directly into the editor, while the trailing block extension gets added as a TipTap extension, and the placeholder extension is a ProseMirror Plugin.

The BlockNote extension API was somewhat haphazardly created out of the need for a unified API to add features to the editor, such as keyboard shortcuts, input rules, and exposing the ProseMirror plugin API for more advanced use cases. We migrated this in piecemeal, supporting ProseMirror plugins, TipTap extensions & a stripped down BlockNote extension API, but we need more!

# Solution

## BlockNote Extensions API Refactor

We've made big improvements to the BlockNote extension API, making extensions much more scalable. In addition to all previous functionality (like adding keyboard extensions), extensions can now:

- Have a proper life cycle, i.e. initialise themselves, add event listeners, and cleanup.
- Have a state that other parts of the app/UI can read from and write to.
- Expose methods which other parts of the app/UI can call.
- Be added & removed from the editor even after initialisation.
- Be configured with different options.

Pretty much all existing editor features have been converted into BlockNote extensions, and use the same API to be created, as well as added to/removed from the editor. This is a great change to the codebase quality, as the editor is a lot more modular thanks to this change, and there is now a clear way to extend the editor's functionality.

## UI Extension Refactor

Additionally, the UI extensions have been partially rewritten (currently includes the formatting toolbar, link toolbar, and file panel) to take advantage of the new extension functionality. The states they keep have been minimised as much as possible, and many of their responsibilities have been moved to the React layer. This massively cuts down on the overall code and makes them much more understandable.

Speaking of the React layer, alongside refactoring the UI extensions in the editor core, the React components and hooks have also been overhauled. FloatingUI now handles much of what was previously done by the extensions, including:

- Positioning the element in the viewport.
- Updating the element position from user interaction (scrolling, resizing the editor, etc).
- Dismissing the element from user interaction (clicking outside the editor, pressing escape, etc).
- Ensuring that the element transitions in/out correctly.

Alongside that, a bunch of helper components and hooks have been created to streamline the code.

# In-Depth Changes

### Creating Extensions

To create a BlockNote extension, you use the `createExtension` function. It works similarly to the old `createBlockNoteExtension` function, and you can still add an extension key, keyboard shortcuts, input rules, TipTap extensions, and ProseMirror plugins similarly to before. Also like before, you can add your own custom properties to the extension:

```ts
const YourExtension = createExtension({
  key: "your-extension",
  
  keyboardShortcuts: {
    "Mod+Shift+1": ({ editor }) => { ... }
  },
  inputRules: [
    {
      find: new RegExp(`^[-+*]\\s$`),
      replace: ({ match, range, editor }) => { ... }
    }
  ],
  tiptapExtensions: [
    Extension.create({ ... })
  ],
  prosemirrorPlugins: [
    new Plugin({ ... })
  ],
  
  customString: "...",
  customObject: { ... },
  customFunction: (...) => { ... },
  
  ...
});
```

You can also now give your extensions configurable options by passing a function to `createExtension` instead of an object:

```ts
const YourExtension = createExtension(({ editor, options }) => ({
  key: "your-extension",
  
  ...
}));
```

Additionally, there are now three additional properties that you can give your extensions:

`store`: A [TanStack store](https://tanstack.com/store/latest/docs/quick-start) which holds the extension's state. For example, a table of contents extension may keep a state containing a list of heading blocks in the editor, as well as their text content. This state can also be accessed and modified by other extensions.

`mount`: A callback which runs when the editor gets mounted to the DOM. You can use it to setup event listeners and update the state.

`runsBefore`: An array which holds `key`s of other BlockNote extensions that the extension depends on. This ensures that an extension's state will be initialised before other extensions try to access it.

```ts
const YourExtension = createExtension(({editor, options}) => {
  const store = createStore({ ... })
  
  return {
    key: "your-extension",
    
    ...
    
    store,
    mount: ({ dom, root, signal }) => {
      const onMouseDown = () => {
        ...
        store.setState((state) => ({ ... }))
        ...
      };
      
      const onMouseUp = () => {
        ...
        store.setState((state) => ({ ... }))
        ...
      };
      
      const onChange = () => {
        ...
        store.setState((state) => ({ ... }))
        ...
      };
    
      dom.addEventListener("mousedown", onMouseDown);
      root.addEventListener("mouseup", onMouseUp);

      const destroyOnChange = editor.onChange(onChange);
      
      signal.addEventListener("abort", () => {
        dom.removeEventListener("mousedown", onMouseDown);
        root.removeEventListener("mouseup", onMouseUp);
        
        destroyOnChange();
      });
    },
    runsBefore: [ ... ]
  }
});
```

### Adding/Removing Extensions

You can still pass your BlockNote extensions to the `extensions` option in `useCreateBlockNote`:

```ts
  extensions: [YourExtension]
});
```

However, you can now also dynamically add/remove extensions after the editor is initialised using `editor.registerExtension` and `editor.unregisterExtension`:

```ts
editor.registerExtension(YourExtension);

editor.unregisterExtension("your-extension");
```

### Accessing Extensions

You can access extensions and their states using `editor.getExtension`:

```ts
const yourExtension = editor.getExtension(YourExtension);
```

### Editor Hooks

A universal `useEditorState` hook has been added to listen to changes in the editor state. It replaces existing hooks which read parts of the editor state, such as `useActiveStyles` and `useSelectedBlocks`.

```ts
const { ... } = useEditorState({ selector: (editor) => {
  ...
  
  return { ... }
}});
```

The `selector` allows you return specific properties from the editor state, or derive new properties from it. Whenever it's called (i.e. when the editor content of selection changes), it compares the new result with the old one. Only if the new result is different does it trigger a re-render. This works not only when the result is a primitive type (boolean, number, string, etc), but also if it's a plain object, i.e. an object without function properties.

For example, here's how you would redefine the old `useActiveStyles` hook:

```ts
const { bold, italic, ...rest } = useEditorState({
    selector: ({ editor }) => editor.getActiveStyles(),
});
```

This will only trigger a re-render if any of the active styles have actually changed.

The `useEditorState` hook takes several other options:

```ts
const { ... } = useEditorState({
  selector: (editor) => { ... },
  editor: ...,
  equalityFn: (oldResult, newResult) => ...,
  on: ...
});
```

`editor`: The BlockNote editor to use. Only needed if used outside a `.bn-container` element.

`equalityFn`: Function to check whether the new result of calling the `selector` is different to the old result, in order to trigger a re-render. Uses deep equals by default, and only really needed of the returned object contains functions, or other niche cases where deep equals may throw false positives.

`on`: Determines when the `selector` is called. Calls on both editor selection and content (`"all"`) changes by default, but can be changed to only listen for content (`"change"`) or selection (`"selection"`) changes.

`runsBefore`: An array which holds `key`s of other BlockNote extensions that the extension depends on. This ensures that an extension's state will be initialised before other extensions try to access it.

### Extension Hooks

When creating custom blocks or UI elements like menus & toolbars, it's likely you'll want to access editor extensions and their states. To do this, the `useExtension` and `useExtensionState` hooks have been created.

The `useExtension` hook lets you access the properties of an extension in the editor, except:

- `keyboardShortcuts`
- `inputRules`
- `tiptapExtensions`
- `prosemirrorPlugins`
- `runsBefore`

```ts
const yourExtension = useExtension(YourExtension);

const { key, store, mount, customString, customObject, customFunction } = yourExtension;
```

It also has the same `editor` option as `useEditorState`:

```ts
const yourExtension = useExtension(YourExtension, { editor: ... });
```

The `useExtensionState` hook lets you access the state of an extension in the editor:

```ts
const yourExtensionState = useExtensionState(YourExtension);
```

It also has the same `editor` and `selector` options as `useEditorState`:

```ts
const yourExtensionState = useExtensionState(YourExtension, { editor: ..., selector: (state) => { ... }});
```

### Popover Components

The old `useUIElementPositioning` has been removed in favour of components which are easier to use and understand.

The `BlockPopover` component is used to create a popover around a specific block, and is used by elements like the side menu as well as the file panel:

```tsx
<BlockPopover blockId={...} {...floatingUIOptions}>...</BlockPopover>
```

The `PositionPopover` component is used to create a popover around a ProseMirror position range, and is used by elements like the formatting toolbar:

```tsx
<PositionPopover position={{ from: ..., to: ... }} {...floatingUIOptions}>...</PositionPopover>
```

The `GenericPopover` component is used to create a popover around DOM element or FloatingUI virtual element, and is used by both the `BlockPopover` as well as the `PositionPopover`:

```tsx
<GenericPopover reference={...} {...floatingUIOptions}>...</GenericPopover>
```

All 3 components take `floatingUIOptions`, which is an object that combines props from all the FloatingUI hooks used in `GenericPopover`, e.g. `useFloating` and `useDismiss`.

# Breaking Changes

## `@blocknote/core`
- Existing extensions using the `BlockNoteExtension` class are no longer supported. Use the `createExtension` function instead.
- Extension-specific imports have been moved from `@blocknote/core` to `@blocknote/core/extensions`.
- Comment-specific imports have been moved from `@blocknote/core` to `@blocknote/core/commnents`.
- Extensions are no longer listed as properties of `BlockNoteEditor`, e.g. `editor.formattingToolbar`. Use the `getExtension` method instead, e.g. `editor.getExtension(FormattingToolbarExtension)`.
- `editor.openSuggestionMenu` has been removed. Use `editor.getExtension(SuggestionMenu).openSuggestionMenu` instead.
- `editor.getForceSelectionVisible`/`editor.setForceSelectionVisible` have been removed. Use `editor.getExtension(ShowSelectionExtension).store.state.enabled`/`editor.getExtension(ShowSelectionExtension).showSelection` instead.
- `editor.onCreate` has been removed. Use `editor.onMount` instead.
- `insertOrUpdateBlock` has been renamed to `insertOrUpdateBlockForSlashMenu`.
- `editor.updateCollaborationUserInfo` has been removed.

## `@blocknote/react`
- `useEditorContentOrSelectionChange` has been removed. Use `useEditorState` instead.
- `useUIPluginState` has been removed. Use `useExtension` or `useExtensionState` instead.
- `useUIElementPositioning` has been removed. Use `BlockPopover`, `PositionPopover`, or `GenericPopover` instead.
- `useEditorForceUpdate` has been removed.

## `@blocknote/xl-ai`
- `BlockPositioner` has been removed. Use `BlockPopover` instead.

# Testing

While this PR brings many changes to UI elements and their extensions, our UI testing setup is sadly not up to the standard we need to ensure nothing breaks when it gets merged. We could expand our suite of Playwright E2E tests, but decided not to as these have historically slow, flaky, and hard to debug. At some point, we will need to look at implementing a solid component testing setup to prevent UI regressions.

In the meantime though, we have compiled a thorough list of internal test cases for each default UI element, and manually ensured that each of them pass.

The only exception are the table handles, which have very minor regressions.

### DIFF
--- a/packages/react/src/components/Popovers/BlockPopover.tsx
+++ b/packages/react/src/components/Popovers/BlockPopover.tsx
@@ -1,10 +1,9 @@
 import { getNodeById } from "@blocknote/core";
-import { VirtualElement } from "@floating-ui/react";
-import { ReactNode, useMemo, useRef } from "react";
+import { ReactNode, useMemo } from "react";
 
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
 import { FloatingUIOptions } from "./FloatingUIOptions.js";
-import { GenericPopover } from "./GenericPopover.js";
+import { GenericPopover, GenericPopoverReference } from "./GenericPopover.js";
 
 export const BlockPopover = (
   props: FloatingUIOptions & {
@@ -16,50 +15,35 @@ export const BlockPopover = (
 
   const editor = useBlockNoteEditor<any, any, any>();
 
-  // Stores the last created `boundingClientRect` to use in case `blockId` is
-  // not defined, or the block could not be found in the editor.
-  // TODO: Move this logic to the `GenericPopover`.
-  const boundingClientRect = useRef<DOMRect>(new DOMRect());
-  // Uses a virtual element instead of directly using the block's DOM element
-  // to better handle when the block is deleted from the editor. When using the
-  // DOM element directly, this will cause the element to unmount. FloatingUI
-  // will then call `getBoundingClientRect` on a the unmounted DOM element,
-  // resulting in an incorrect `DOMRect` with x, y, height, and width of 0.
-  // With a virtual element, we can instead use the last `DOMRect` from when
-  // the block still existed.
-  const virtualElement = useMemo<VirtualElement>(
+  const reference = useMemo<GenericPopoverReference | undefined>(
     () =>
       editor.transact((tr) => {
         if (!blockId) {
-          return { getBoundingClientRect: () => boundingClientRect.current };
+          return undefined;
         }
 
         // TODO use the location API for this
         const nodePosInfo = getNodeById(blockId, tr.doc);
         if (!nodePosInfo) {
-          return { getBoundingClientRect: () => boundingClientRect.current };
+          return undefined;
         }
 
         const { node } = editor.prosemirrorView.domAtPos(
           nodePosInfo.posBeforeNode + 1,
         );
         if (!(node instanceof Element)) {
-          return { getBoundingClientRect: () => boundingClientRect.current };
+          return undefined;
         }
 
         return {
-          getBoundingClientRect: () => {
-            boundingClientRect.current = node.getBoundingClientRect();
-            return boundingClientRect.current;
-          },
-          contextElement: node,
+          element: node,
         };
       }),
     [editor, blockId],
   );
 
   return (
-    <GenericPopover reference={virtualElement} {...floatingUIOptions}>
+    <GenericPopover reference={reference} {...floatingUIOptions}>
       {blockId !== undefined && children}
     </GenericPopover>
   );
--- a/packages/react/src/components/Popovers/GenericPopover.tsx
+++ b/packages/react/src/components/Popovers/GenericPopover.tsx
@@ -6,16 +6,78 @@ import {
   useMergeRefs,
   useTransitionStatus,
   autoUpdate,
-  ReferenceType,
   useHover,
 } from "@floating-ui/react";
 import { HTMLAttributes, ReactNode, useEffect, useRef } from "react";
 
 import { FloatingUIOptions } from "./FloatingUIOptions.js";
 
+export type GenericPopoverReference =
+  | {
+      // A DOM element to use as the reference element for the popover.
+      element: Element;
+      // To update the popover position, `element.getReferenceBoundingRect`
+      // is called. This flag caches the last result of the call while the
+      // element is mounted to the DOM, so it doesn't update while the
+      // popover is closing and transitioning out. Useful for if the
+      // reference element unmounts, as `element.getReferenceBoundingRect`
+      // would return a `DOMRect` with x, y, width, and height of 0.
+      // Defaults to `true`.
+      cacheMountedBoundingClientRect?: boolean;
+    }
+  | {
+      // When no reference element is provided, this can be provided as an
+      // alternative "virtual" element to position the popover around.
+      getBoundingClientRect: () => DOMRect;
+    }
+  | {
+      element: Element;
+      cacheMountedBoundingClientRect?: boolean;
+      // If both `element` and `getBoundingClientRect` are provided, uses
+      // `getBoundingClientRect` to position the popover, but still treats
+      // `element` as the reference element for all other purposes. When
+      // `cacheMountedBoundingClientRect` is `true` or unspecified, this
+      // function is not called while the reference element is not mounted.
+      getBoundingClientRect: () => DOMRect;
+    };
+
+// Returns a modified version of `getBoundingClientRect`, if
+// `reference.element` is passed and `reference.cacheMountedBoundingClientRect`
+// is `true` or `undefined`. In the modified version, each new result is cached
+// and returned while `reference.element` is connected to the DOM. If it is no
+// longer connected, the cache is no longer updated and the last cached result
+// is used.
+//
+// In all other cases, just returns `reference.getBoundingClientRect`, or
+// `reference.element.getBoundingClientRect` if it's not defined.
+export function getMountedBoundingClientRectCache(
+  reference: GenericPopoverReference,
+) {
+  let lastBoundingClientRect = new DOMRect();
+  const getBoundingClientRect =
+    "getBoundingClientRect" in reference
+      ? () => reference.getBoundingClientRect()
+      : () => reference.element.getBoundingClientRect();
+
+  return () => {
+    if (
+      "element" in reference &&
+      (reference.cacheMountedBoundingClientRect ?? true)
+    ) {
+      if (reference.element.isConnected) {
+        lastBoundingClientRect = getBoundingClientRect();
+      }
+
+      return lastBoundingClientRect;
+    }
+
+    return getBoundingClientRect();
+  };
+}
+
 export const GenericPopover = (
   props: FloatingUIOptions & {
-    reference?: ReferenceType;
+    reference?: GenericPopoverReference;
     children: ReactNode;
   },
 ) => {
@@ -47,20 +109,15 @@ export const GenericPopover = (
 
   useEffect(() => {
     if (props.reference) {
-      if (props.reference instanceof Element) {
-        refs.setReference(props.reference);
-      } else {
-        // FloatingUI's virtual elements can have a `contextElement`, which is
-        // useful for automatically calling `getBoundingClientRect` when said
-        // element changes position. However, it's pretty limited as FloatingUI
-        // doesn't attach listeners to `contextElement` for e.g. `useHover`.
-        // Therefore, we also set the reference element to `contextElement` so
-        // event listeners can be attached.
-        if (props.reference.contextElement) {
-          refs.setReference(props.reference.contextElement);
-        }
-        refs.setPositionReference(props.reference);
+      if ("element" in props.reference) {
+        refs.setReference(props.reference.element);
       }
+
+      refs.setPositionReference({
+        getBoundingClientRect: getMountedBoundingClientRectCache(
+          props.reference,
+        ),
+      });
     }
   }, [props.reference, refs]);
 
--- a/packages/react/src/components/Popovers/PositionPopover.tsx
+++ b/packages/react/src/components/Popovers/PositionPopover.tsx
@@ -1,10 +1,9 @@
-import { ClientRectObject, VirtualElement } from "@floating-ui/react";
 import { posToDOMRect } from "@tiptap/core";
-import { ReactNode, useMemo, useRef } from "react";
+import { ReactNode, useMemo } from "react";
 
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
 import { FloatingUIOptions } from "./FloatingUIOptions.js";
-import { GenericPopover } from "./GenericPopover.js";
+import { GenericPopover, GenericPopoverReference } from "./GenericPopover.js";
 
 export const PositionPopover = (
   props: FloatingUIOptions & {
@@ -16,40 +15,24 @@ export const PositionPopover = (
   const { from, to } = position || {};
 
   const editor = useBlockNoteEditor<any, any, any>();
-  const { headless } = editor;
 
-  // Stores the last created `boundingClientRect` to use in case `position` is
-  // not defined.
-  // TODO: Move this logic to the `GenericPopover`.
-  const boundingClientRect = useRef<ClientRectObject>(new DOMRect());
-  const virtualElement = useMemo<VirtualElement>(
-    () => ({
-      getBoundingClientRect: () => {
-        if (from === undefined || to === undefined) {
-          return boundingClientRect.current;
-        }
+  const reference = useMemo<GenericPopoverReference | undefined>(() => {
+    if (from === undefined || to === undefined) {
+      return undefined;
+    }
 
-        boundingClientRect.current = posToDOMRect(
-          editor.prosemirrorView,
-          from,
-          to ?? from,
-        );
-
-        return boundingClientRect.current;
-      },
-      contextElement:
-        // Use first child as the editor DOM element may itself be scrollable.
-        // For FloatingUI to auto-update the position during scrolling, the
-        // `contextElement` must be a descendant of the scroll container.
-        !headless && editor.domElement?.firstChild instanceof Element
-          ? editor.domElement.firstChild
-          : undefined,
-    }),
-    [editor, headless, from, to],
-  );
+    return {
+      // Use first child as the editor DOM element may itself be scrollable.
+      // For FloatingUI to auto-update the position during scrolling, the
+      // `contextElement` must be a descendant of the scroll container.
+      element: editor.domElement?.firstElementChild || undefined,
+      getBoundingClientRect: () =>
+        posToDOMRect(editor.prosemirrorView, from, to ?? from),
+    };
+  }, [editor, from, to]);
 
   return (
-    <GenericPopover reference={virtualElement} {...floatingUIOptions}>
+    <GenericPopover reference={reference} {...floatingUIOptions}>
       {position !== undefined && children}
     </GenericPopover>
   );
--- a/packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
+++ b/packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
@@ -2,7 +2,7 @@ import { BlockSchema, InlineContentSchema, StyleSchema } from "@blocknote/core";
 import { LinkToolbarExtension } from "@blocknote/core/extensions";
 import { flip, offset, safePolygon } from "@floating-ui/react";
 import { Range } from "@tiptap/core";
-import { FC, useEffect, useMemo, useRef, useState } from "react";
+import { FC, useEffect, useMemo, useState } from "react";
 
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
 import { LinkToolbar } from "./LinkToolbar.js";
@@ -10,7 +10,10 @@ import { LinkToolbarProps } from "./LinkToolbarProps.js";
 import { useEditorState } from "../../hooks/useEditorState.js";
 import { useExtension } from "../../hooks/useExtension.js";
 import { FloatingUIOptions } from "../Popovers/FloatingUIOptions.js";
-import { GenericPopover } from "../Popovers/GenericPopover.js";
+import {
+  GenericPopover,
+  GenericPopoverReference,
+} from "../Popovers/GenericPopover.js";
 
 export const LinkToolbarController = (props: {
   linkToolbar?: FC<LinkToolbarProps>;
@@ -95,10 +98,7 @@ export const LinkToolbarController = (props: {
     };
   }, [frozen, linkToolbar, mouseHoverLink?.url, selectionLink]);
 
-  const link = useMemo(
-    () => selectionLink || mouseHoverLink,
-    [mouseHoverLink, selectionLink],
-  );
+  const link = selectionLink || mouseHoverLink;
 
   const floatingUIOptions = useMemo<FloatingUIOptions>(
     () => ({
@@ -151,17 +151,10 @@ export const LinkToolbarController = (props: {
     ],
   );
 
-  // When a link is deleted, the element representing it in the DOM gets
-  // unmounted. However, the reference still exists and so FloatingUI can still
-  // call `getBoundingClientRect` on it, which will return a `DOMRect` that has
-  // an x, y, height, and width of 0. This is why we instead use a virtual
-  // element for the reference, and use the last obtained `DOMRect` from when
-  // the link element was still mounted to the DOM.
-  // TODO: Logic should be in `GenericPopover`.
-  const mountedBoundingClientRect = useRef(new DOMRect());
-  if (link?.element && editor.prosemirrorView.root.contains(link.element)) {
-    mountedBoundingClientRect.current = link.element.getBoundingClientRect();
-  }
+  const reference = useMemo<GenericPopoverReference | undefined>(
+    () => (link?.element ? { element: link.element } : undefined),
+    [link?.element],
+  );
 
   if (!editor.isEditable) {
     return null;
@@ -170,13 +163,7 @@ export const LinkToolbarController = (props: {
   const Component = props.linkToolbar || LinkToolbar;
 
   return (
-    <GenericPopover
-      reference={{
-        getBoundingClientRect: () => mountedBoundingClientRect.current,
-        contextElement: link?.element,
-      }}
-      {...floatingUIOptions}
-    >
+    <GenericPopover reference={reference} {...floatingUIOptions}>
       {link && (
         <Component
           url={link.url}
--- a/packages/react/src/components/SuggestionMenu/GridSuggestionMenu/GridSuggestionMenuController.tsx
+++ b/packages/react/src/components/SuggestionMenu/GridSuggestionMenu/GridSuggestionMenuController.tsx
@@ -1,6 +1,6 @@
 import { BlockSchema, InlineContentSchema, StyleSchema } from "@blocknote/core";
 import { SuggestionMenu } from "@blocknote/core/extensions";
-import { flip, offset, shift, size, VirtualElement } from "@floating-ui/react";
+import { flip, offset, shift, size } from "@floating-ui/react";
 import { FC, useEffect, useMemo } from "react";
 
 import { useBlockNoteEditor } from "../../../hooks/useBlockNoteEditor.js";
@@ -9,7 +9,10 @@ import {
   useExtensionState,
 } from "../../../hooks/useExtension.js";
 import { FloatingUIOptions } from "../../Popovers/FloatingUIOptions.js";
-import { GenericPopover } from "../../Popovers/GenericPopover.js";
+import {
+  GenericPopover,
+  GenericPopoverReference,
+} from "../../Popovers/GenericPopover.js";
 import { getDefaultReactEmojiPickerItems } from "./getDefaultReactEmojiPickerItems.js";
 import { GridSuggestionMenu } from "./GridSuggestionMenu.js";
 import { GridSuggestionMenuWrapper } from "./GridSuggestionMenuWrapper.js";
@@ -98,11 +101,15 @@ export function GridSuggestionMenuController<
     selector: (state) => state?.referencePos || new DOMRect(),
   });
 
-  const virtualElement = useMemo<VirtualElement>(
+  const reference = useMemo<GenericPopoverReference>(
     () => ({
+      // Use first child as the editor DOM element may itself be scrollable.
+      // For FloatingUI to auto-update the position during scrolling, the
+      // `contextElement` must be a descendant of the scroll container.
+      element: editor.domElement?.firstChild || undefined,
       getBoundingClientRect: () => referencePos,
     }),
-    [referencePos],
+    [editor.domElement?.firstChild, referencePos],
   );
 
   const floatingUIOptions = useMemo<FloatingUIOptions>(
@@ -161,7 +168,7 @@ export function GridSuggestionMenuController<
   }
 
   return (
-    <GenericPopover reference={virtualElement} {...floatingUIOptions}>
+    <GenericPopover reference={reference} {...floatingUIOptions}>
       {triggerCharacter && (
         <GridSuggestionMenuWrapper
           query={state.query}
--- a/packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
+++ b/packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
@@ -5,7 +5,6 @@ import {
 } from "@blocknote/core/extensions";
 import {
   UseFloatingOptions,
-  VirtualElement,
   flip,
   offset,
   shift,
@@ -16,7 +15,10 @@ import { FC, useEffect, useMemo } from "react";
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
 import { useExtension, useExtensionState } from "../../hooks/useExtension.js";
 import { FloatingUIOptions } from "../Popovers/FloatingUIOptions.js";
-import { GenericPopover } from "../Popovers/GenericPopover.js";
+import {
+  GenericPopover,
+  GenericPopoverReference,
+} from "../Popovers/GenericPopover.js";
 import { SuggestionMenu } from "./SuggestionMenu.js";
 import { SuggestionMenuWrapper } from "./SuggestionMenuWrapper.js";
 import { getDefaultReactSlashMenuItems } from "./getDefaultReactSlashMenuItems.js";
@@ -100,11 +102,15 @@ export function SuggestionMenuController<
     selector: (state) => state?.referencePos || new DOMRect(),
   });
 
-  const virtualElement = useMemo<VirtualElement>(
+  const reference = useMemo<GenericPopoverReference>(
     () => ({
+      // Use first child as the editor DOM element may itself be scrollable.
+      // For FloatingUI to auto-update the position during scrolling, the
+      // `contextElement` must be a descendant of the scroll container.
+      element: editor.domElement?.firstChild || undefined,
       getBoundingClientRect: () => referencePos,
     }),
-    [referencePos],
+    [editor.domElement?.firstChild, referencePos],
   );
 
   const floatingUIOptions = useMemo<FloatingUIOptions>(
@@ -164,7 +170,7 @@ export function SuggestionMenuController<
   }
 
   return (
-    <GenericPopover reference={virtualElement} {...floatingUIOptions}>
+    <GenericPopover reference={reference} {...floatingUIOptions}>
       {triggerCharacter && (
         <SuggestionMenuWrapper
           query={state.query}
--- a/packages/react/src/components/TableHandles/TableHandlesController.tsx
+++ b/packages/react/src/components/TableHandles/TableHandlesController.tsx
@@ -9,16 +9,19 @@ import {
 import { TableHandlesExtension } from "@blocknote/core/extensions";
 import { FC, useMemo, useState } from "react";
 
-import { offset, size, VirtualElement } from "@floating-ui/react";
+import { offset, size } from "@floating-ui/react";
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
+import { useExtensionState } from "../../hooks/useExtension.js";
+import {
+  GenericPopover,
+  GenericPopoverReference,
+} from "../Popovers/GenericPopover.js";
 import { ExtendButton } from "./ExtendButton/ExtendButton.js";
 import { ExtendButtonProps } from "./ExtendButton/ExtendButtonProps.js";
 import { TableHandle } from "./TableHandle.js";
 import { TableCellButton } from "./TableCellButton.js";
 import { TableCellButtonProps } from "./TableCellButtonProps.js";
-import { useExtensionState } from "../../hooks/useExtension.js";
 import { TableHandleProps } from "./TableHandleProps.js";
-import { GenericPopover } from "../Popovers/GenericPopover.js";
 
 export const TableHandlesController = <
   I extends InlineContentSchema = DefaultInlineContentSchema,
@@ -41,7 +44,20 @@ export const TableHandlesController = <
 
   const state = useExtensionState(TableHandlesExtension);
 
-  const elements = useMemo(() => {
+  const references = useMemo<
+    | {
+        tableReference: undefined;
+        cellReference: undefined;
+        rowReference: undefined;
+        columnReference: undefined;
+      }
+    | {
+        tableReference: GenericPopoverReference;
+        cellReference: GenericPopoverReference;
+        rowReference: GenericPopoverReference;
+        columnReference: GenericPopoverReference;
+      }
+  >(() => {
     if (
       state === undefined ||
       state.rowIndex === undefined ||
@@ -65,8 +81,10 @@ export const TableHandlesController = <
       tableBeforePos + 1,
     ).node;
     if (!(tableElement instanceof Element)) {
-      return undefined;
+      return {};
     }
+
+    const tableReference = { element: tableElement };
 
     const rowBeforePos = editor.prosemirrorState.doc
       .resolve(tableBeforePos + 1)
@@ -77,15 +95,18 @@ export const TableHandlesController = <
 
     const cellElement = editor.prosemirrorView.domAtPos(cellBeforePos + 1).node;
     if (!(cellElement instanceof Element)) {
-      return undefined;
+      return {};
     }
 
-    const tableBoundingRect = tableElement.getBoundingClientRect();
-    const cellBoundingRect = cellElement.getBoundingClientRect();
+    const cellReference = { element: cellElement };
 
-    const rowVirtualElement: VirtualElement = {
-      getBoundingClientRect: () =>
-        new DOMRect(
+    const rowReference = {
+      element: tableElement,
+      getBoundingClientRect: () => {
+        const tableBoundingRect = tableElement.getBoundingClientRect();
+        const cellBoundingRect = cellElement.getBoundingClientRect();
+
+        return new DOMRect(
           tableBoundingRect.x,
           state.draggingState &&
           state.draggingState.draggedCellOrientation === "row"
@@ -93,12 +114,16 @@ export const TableHandlesController = <
             : cellBoundingRect.y,
           tableBoundingRect.width,
           cellBoundingRect.height,
-        ),
-      contextElement: tableElement,
+        );
+      },
     };
-    const columnVirtualElement: VirtualElement = {
-      getBoundingClientRect: () =>
-        new DOMRect(
+    const columnReference = {
+      element: tableElement,
+      getBoundingClientRect: () => {
+        const tableBoundingRect = tableElement.getBoundingClientRect();
+        const cellBoundingRect = cellElement.getBoundingClientRect();
+
+        return new DOMRect(
           state.draggingState &&
           state.draggingState.draggedCellOrientation === "col"
             ? state.draggingState.mousePos - cellBoundingRect.width / 2
@@ -106,15 +131,15 @@ export const TableHandlesController = <
           tableBoundingRect.y,
           cellBoundingRect.width,
           tableBoundingRect.height,
-        ),
-      contextElement: tableElement,
+        );
+      },
     };
 
     return {
-      tableElement,
-      cellElement,
-      rowVirtualElement,
-      columnVirtualElement,
+      tableReference,
+      cellReference,
+      rowReference,
+      columnReference,
     };
   }, [editor, state]);
 
@@ -129,7 +154,7 @@ export const TableHandlesController = <
   return (
     <>
       <GenericPopover
-        reference={elements?.rowVirtualElement}
+        reference={references?.rowReference}
         useFloatingOptions={{
           open:
             state.show &&
@@ -156,7 +181,7 @@ export const TableHandlesController = <
           )}
       </GenericPopover>
       <GenericPopover
-        reference={elements?.columnVirtualElement}
+        reference={references?.columnReference}
         useFloatingOptions={{
           open:
             state.show &&
@@ -183,7 +208,7 @@ export const TableHandlesController = <
           )}
       </GenericPopover>
       <GenericPopover
-        reference={elements?.cellElement}
+        reference={references?.cellReference}
         useFloatingOptions={{
           open:
             state.show &&
@@ -211,7 +236,7 @@ export const TableHandlesController = <
           )}
       </GenericPopover>
       <GenericPopover
-        reference={elements?.tableElement}
+        reference={references?.tableReference}
         useFloatingOptions={{
           open:
             state.show &&
@@ -246,7 +271,7 @@ export const TableHandlesController = <
           )}
       </GenericPopover>
       <GenericPopover
-        reference={elements?.tableElement}
+        reference={references?.tableReference}
         useFloatingOptions={{
           open:
             state.show &&


packages/core/src/extensions/Collaboration/CursorPlugin.ts
```diff
@@ -1,189 +1,180 @@
...
-    const uicolors = [r / 255, g / 255, b / 255];
-    const c = uicolors.map((col) => {
-      if (col <= 0.03928) {
-        return col / 12.92;
-      }
-      return Math.pow((col + 0.055) / 1.055, 2.4);
-    });
-    const L = 0.2126 * c[0] + 0.7152 * c[1] + 0.0722 * c[2];
-    return L <= 0.179;
+export const CursorPlugin = createExtension((_editor, options) => {
```
nperez0111: TODO, there were some defaults here that were being set that may no longer be working

packages/core/src/extensions/Collaboration/ForkYDoc.ts
```diff
@@ -1,192 +1,164 @@
...
+    | undefined
+  >(undefined, {
+    onUpdate() {
+      store.setState({ isForked: store.state !== undefined });
+    },
+  });
+
+  return {
+    key: "yForkDoc",
+    store,
```
nperez0111: I'm not entirely certain on this pattern, perhaps we can also [allow derived values](https://tanstack.com/store/latest/docs/reference/classes/derived) to be the store?

The point here was to hide the internal state, which should not need to be accessed by anyone other than the plugin.

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -57,306 +61,279 @@ function getUpdatedThreadPositions(doc: Node, markType: string) {
...
-    return "comments";
-  }
-
-  public readonly userStore: UserStore<User>;
-
-  /**
-   * Whether a comment is currently being composed
-   */
-  private pendingComment = false;
+export const CommentsPlugin = createExtension(
```
nperez0111: Ideally, this would be part of a separate package now. But will leave for now

packages/core/src/extensions/FilePanel/FilePanel.ts
```diff
@@ -1,206 +1,79 @@
...
-    return "filePanel";
+import {
+  createExtension,
+  createStore,
+} from "../../editor/BlockNoteExtension.js";
+import { Plugin } from "@tiptap/pm/state";
+
+export const FilePanelPlugin = createExtension((editor) => {
+  const store = createStore({
+    blockId: undefined as string | undefined,
```
nperez0111: I think that the minimal state that a file panel really needs, is the block which it is currently open to. This being a string implies that it is open, and being `undefined` means that it is closed. To get the position, you can read it from the DOM.

I think we can do scroll updates & mouse events from floating UI, if not let's discuss.

packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
```diff
@@ -1,363 +1,113 @@
...
-    this.addProsemirrorPlugin(
+  createExtension,
+  createStore,
+} from "../../editor/BlockNoteExtension.js";
+
+export const FormattingToolbarExtension = createExtension((editor) => {
+  const store = createStore({
+    show: false,
+    referencePos: null as DOMRect | null,
+  });
```
nperez0111: This can likely be simplified further. The only state that this plugin _really_ needs is whether it is being shown or not (for programmatic dismissal). Even the reference position should be derive-able from the current editor state. We will likely want a hook which makes it easy to have a menu at wherever the current selection is, and this should store the majority of the state.

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -587,9 +587,11 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
         if (
           this.options.tabBehavior !== "prefer-indent" &&
-          (this.options.editor.formattingToolbar?.shown ||
-            this.options.editor.linkToolbar?.shown ||
-            this.options.editor.filePanel?.shown)
+          true
+          // TODO what??
+          // (this.options.editor.formattingToolbar?.shown ||
+          //   this.options.editor.linkToolbar?.shown ||
+          //   this.options.editor.filePanel?.shown)
```
nperez0111: Ideally this would not have to be aware of every menu that could be open. I'm not sure if this means that there should be a single "menu" extension which stores the current state of whatever menu is currently active? Or, if these menus should be setting keyboard handlers to capture the Tab key or otherwise not giving this event to the editor.

packages/core/src/extensions/Placeholder/Placeholder.ts
```diff
@@ -1,25 +1,16 @@
...
-    placeholders: Record<
-      string | "default" | "emptyDocument",
-      string | undefined
-    >,
-  ) {
-    super();
-    this.addProsemirrorPlugin(
+export const PlaceholderPlugin = createExtension((editor, options) => {
+  // TODO defaults?
+  const placeholders = options.placeholders;
```
nperez0111: There were some defaults here that we should also apply

packages/core/src/extensions/ShowSelection/ShowSelection.ts
```diff
@@ -10,48 +13,36 @@ const PLUGIN_KEY = new PluginKey(`blocknote-show-selection`);
...
-
-            return DecorationSet.create(doc, [dec]);
-          },
+export const ShowSelectionPlugin = createExtension(
+  (editor: BlockNoteEditor<any, any, any>, _options) => {
+    const store = createStore(
+      { enabled: false },
+      {
+        onUpdate() {
+          editor.transact((tr) => tr.setMeta(PLUGIN_KEY, {}));
```
nperez0111: TODO check this

packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
```diff
@@ -682,83 +685,72 @@ export class SideMenuView<
...
+    key: "sideMenu",
+    store,
+    plugins: [
       new Plugin({
         key: sideMenuPluginKey,
         view: (editorView) => {
-          this.view = new SideMenuView(editor, editorView, (state) => {
-            this.emit("update", state);
+          view = new SideMenuView(editor, editorView, (state) => {
+            store.setState(state);
```
nperez0111: TODO, I did not refactor this extension & maybe should have. It was doing a lot so I didn't take the time for it.

Honestly the gist of the state would be that it knows what the currently hovered block is & whether it is open. There is a lot of handling for drag & drop, which could use some refactoring, but maybe later.

For now, if this integration is enough, then maybe we can stick with it, if not I'll help with the refactor here.

packages/core/src/extensions/SuggestionMenu/SuggestionPlugin.ts
```diff
@@ -163,34 +157,68 @@ const suggestionMenuPluginKey = new PluginKey("SuggestionMenuPlugin");
...
-  I extends InlineContentSchema,
-  S extends StyleSchema,
-> extends BlockNoteExtension {
-  public static key() {
-    return "suggestionMenu";
-  }
-
-  private view: SuggestionMenuView<BSchema, I, S> | undefined;
-  private triggerCharacters: string[] = [];
+export const SuggestionMenuPlugin = createExtension((editor) => {
```
nperez0111: Same as the SideMenuPlugin, did not refactor, but hopefully the shim works.

packages/core/src/extensions/TableHandles/TableHandles.ts
```diff
@@ -144,12 +142,8 @@ function hideElements(selector: string, rootEl: Document | ShadowRoot) {
   }
 }
 
-export class TableHandlesView<
-  I extends InlineContentSchema,
-  S extends StyleSchema,
-> implements PluginView
-{
-  public state?: TableHandlesState<I, S>;
+export class TableHandlesView implements PluginView {
```
nperez0111, line 145: Same as the SideMenuPlugin, shimmed, but unsure if it works

packages/core/src/index.ts
```diff
@@ -10,14 +10,13 @@ export * from "./api/pmUtil.js";
...
 export * from "./editor/selectionTypes.js";
 export * from "./exporter/index.js";
 export * from "./extensions-shared/UiElementPosition.js";
 export * from "./extensions/FilePanel/FilePanelPlugin.js";
 export * from "./extensions/FormattingToolbar/FormattingToolbarPlugin.js";
-export * from "./extensions/LinkToolbar/LinkToolbarPlugin.js";
+export * from "./extensions/LinkToolbar/LinkToolbar.js";
 export * from "./extensions/LinkToolbar/protocols.js";
 export * from "./extensions/SideMenu/SideMenuPlugin.js";
 export * from "./extensions/SuggestionMenu/DefaultGridSuggestionItem.js";
```
nperez0111: Ideally just one export from `extensions`

packages/core/package.json
```diff
@@ -80,7 +80,9 @@
   },
   "dependencies": {
     "@emoji-mart/data": "^1.2.1",
+    "@handlewithcare/prosemirror-inputrules": "0.1.3",
```
nperez0111, line 93: Pretty nice to be able to add this, allows us to undo an input rule

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -453,56 +409,6 @@ export class BlockNoteEditor<
   public readonly inlineContentImplementations: InlineContentSpecs;
   public readonly styleImplementations: StyleSpecs;
 
-  public get formattingToolbar(): FormattingToolbarProsemirrorPlugin {
```
YousefED, line 456: 🥰

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -608,22 +490,22 @@ export class BlockNoteEditor<
...
-          name: "User",
-          color: "#FF0000",
-        },
-        provider: newOptions.collaboration?.provider || null,
-        renderCursor: newOptions.collaboration?.renderCursor,
-        showCursorLabels: newOptions.collaboration?.showCursorLabels,
-        comments: newOptions.comments,
-        resolveUsers: newOptions.resolveUsers,
-      };
+      // const collaborationOptions: CollaborationOptions = {
```
YousefED: is this still a todo or can the code be removed?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1013,21 +723,25 @@ export class BlockNoteEditor<
...
-   * @returns The extension instance
+   * Remove an extension(s) from the editor
    */
-  public extension<T extends BlockNoteExtension>(
-    ext: { new (...args: any[]): T } & typeof BlockNoteExtension,
-    key = ext.key(),
-  ): T {
-    return this._extensionManager.extension(ext, key);
-  }
+  public removeExtension: ExtensionManager["unregisterExtension"] = (
```
YousefED: let's pick the same name (remove / unregister) for both layers?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -1047,7 +761,7 @@ export class BlockNoteEditor<
       return;
     }
 
-    this._tiptapEditor.mount({ mount: element });
+    this._tiptapEditor.mount({ mount: element } as any);
```
YousefED: any idea why this is now necessary? (as any)

packages/core/src/extensions/Collaboration/ForkYDoc.ts
YousefED: This seems a bit hacky indeed. On one hand we're deriving `isForked`, but we're also setting it explicitly. Triggering the following related questions / ideas:

- Can we expose a readonly store (even if only on TS level)? (I've used this trick to do this earlier for zustand: https://github.com/TypeCellOS/BlockNote/blob/916fc8cc644f7fe3ef783424ec15447035c0a8ed/packages/xl-ai/src/AIExtension.ts#L86)
- Can we use a single store with isForked as derived value? The follow up question would then of course be how we can hide the other members. Either as private members or by changing the return type to only expose public members?

packages/core/src/extensions/Collaboration/YUndo.ts
```diff
@@ -1,18 +1,16 @@
...
-import { BlockNoteExtension } from "../../editor/BlockNoteExtension.js";
+import { redoCommand, undoCommand, yUndoPlugin } from "y-prosemirror";
+import { createExtension } from "../../editor/BlockNoteExtension.js";
 
-export class UndoPlugin extends BlockNoteExtension {
-  public static key() {
-    return "yUndoPlugin";
+export const UndoPlugin = createExtension((editor, options) => {
+  if (!options.collaboration) {
+    return;
```
YousefED: shouldn't we throw an error in this case? It seems like a caller-error to create an extension without the right options

packages/core/src/extensions/Collaboration/YSync.ts
```diff
@@ -1,18 +1,15 @@
...
-import type * as Y from "yjs";
-import { BlockNoteExtension } from "../../editor/BlockNoteExtension.js";
+import { createExtension } from "../../editor/BlockNoteExtension.js";
 
-export class SyncPlugin extends BlockNoteExtension {
-  public static key() {
-    return "ySyncPlugin";
+export const SyncPlugin = createExtension((_editor, options) => {
+  const fragment = options?.collaboration?.fragment;
+  if (!fragment) {
```
YousefED: shouldn't we throw an error in this case? It seems like a caller-error to create an extension without the right options

packages/react/src/components/Popovers/GenericPopover.tsx
```diff
@@ -0,0 +1,97 @@
...
+    // block will cause the popover to close as the selection collapses. At the
+    // same time, the toolbar re-renders, showing buttons corresponding to the
+    // block at the new selection.
+    return (
+      <div
+        ref={mergedRefs}
+        style={mergedStyles}
+        {...mergedProps}
+        dangerouslySetInnerHTML={
+          status === "close" && ref.current
```
nperez0111: should always be true no? @matthewlipski

packages/react/src/components/Popovers/GenericPopover.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const { style, ...rest } = props.elementProps || {};
+
+  const mergedStyles = {
+    display: "flex",
+    ...styles,
+    ...floatingStyles,
+    ...style,
+  };
+  const mergedProps = { ...getFloatingProps(), ...rest };
```
nperez0111: why not just put styles in the props, they get spread on anyway

packages/react/src/components/Popovers/BlockPopover.tsx
```diff
@@ -0,0 +1,80 @@
...
+        return editor.transact((tr) => {
+          const blockId = getBlockId(editor);
+          if (!blockId) {
+            return boundingClientRect.current;
+          }
+
+          // TODO use the location API for this
+          const nodePosInfo = getNodeById(blockId, tr.doc);
+          if (!nodePosInfo) {
+            return boundingClientRect.current;
```
nperez0111: This feels wrong, it will use a stale position which could lead to being hard to debug, I'd rather an empty domrect or purposefully positioned offscreen (negative coords).

Maybe a console log if this is something that can actually occur

packages/react/src/components/Popovers/BlockPopover.tsx
```diff
@@ -0,0 +1,80 @@
...
+          return boundingClientRect.current;
+        });
+      },
+      // contextElement:
+      //   // TODO should not need to know the DOM structure here
+      //   editor.prosemirrorView.dom.querySelector(
+      //     `[data-id="${props.blockId}"]`,
+      //   ) || editor.prosemirrorView.dom,
+    }),
+    [getBlockId, editor, placement],
```
nperez0111: This will re-create each time since getBlockId is a function & likely to be declared on render. Is that okay?

nperez0111: If not, throw it in a ref that gets updated each render so we always can pull the latest value but not re-create this object

packages/react/src/components/Popovers/PositionPopover.tsx
```diff
@@ -0,0 +1,62 @@
...
+            range.from,
+            range.to ?? range.from,
+          ),
+        );
+
+        return boundingClientRect.current;
+      },
+      // contextElement: editor.prosemirrorView.dom,
+    }),
+    [editor, getPosition],
```
nperez0111: Same as above

packages/react/src/components/Popovers/PositionPopover.tsx
```diff
@@ -0,0 +1,62 @@
...
+          posToDOMRect(
+            editor.prosemirrorView,
+            range.from,
+            range.to ?? range.from,
+          ),
+        );
+
+        return boundingClientRect.current;
+      },
+      // contextElement: editor.prosemirrorView.dom,
```
nperez0111: Any reason to disable the `contextElement`s? Or just disabled for testing?

I think they help with the scoping of events

packages/react/src/components/FormattingToolbar/FormattingToolbarController.tsx
```diff
@@ -33,107 +35,61 @@ const textAlignmentToPlacement = (
...
+    selector: (state) => state.show,
+  });
+
+  const getPosition = useCallback(
+    (editor: BlockNoteEditor<any, any, any>) => ({
+      from: editor.prosemirrorState.selection.from,
+      to: editor.prosemirrorState.selection.to,
+    }),
+    [],
+  );
```
nperez0111: Ah, yea you had to put this in a callback because you don't want the virtual element being re-created. Better to fix it there, so that consumers can just use an inline function & not have to worry about using useCallback

packages/react/src/components/FormattingToolbar/DefaultButtons/TextAlignButton.tsx
```diff
@@ -128,20 +144,10 @@ export const TextAlignButton = (props: { textAlignment: TextAlignment }) => {
         }
       }
     },
-    [editor, selectedBlocks],
+    [editor, state],
```
nperez0111: Only care about state.blocks here right?

packages/core/src/blocks/File/helpers/render/createAddFileButton.ts
```diff
@@ -36,11 +37,7 @@ export const createAddFileButton = (
   };
   // Opens the file toolbar.
   const addFileButtonClickHandler = () => {
-    editor.transact((tr) =>
-      tr.setMeta(editor.filePanel!.plugins[0], {
-        block: block,
-      }),
-    );
+    editor.getExtension(FilePanelPlugin)?.showMenu(block.id);
```
YousefED: minor; I think better to throw an error (! instead of ?) if the extension is not there

packages/core/src/editor/managers/ExtensionManager/index.ts
```diff
@@ -1,130 +1,585 @@
...
+
+    const registeredExtensions = extensions.map((extension) =>
+      this.addExtension(extension),
+    );
+
+    const pluginsToAdd = new Set<Plugin>();
+    for (const extension of registeredExtensions) {
+      if (extension?.tiptapExtensions) {
+        // eslint-disable-next-line no-console
+        console.warn(
```
YousefED, line 147: it's not clear what "separate the extension" means and why this is necessary

packages/core/src/editor/managers/ExtensionManager/index.ts
```diff
@@ -1,130 +1,585 @@
...
+  private extensionPlugins: Map<Extension, Plugin[]> = new Map();
+
+  constructor(
+    private editor: BlockNoteEditor<any, any, any>,
+    private options: BlockNoteEditorOptions<any, any, any>,
+  ) {
+    /**
+     * When the editor is first mounted, we need to initialize all the extensions
+     */
+    editor.onMount(() => {
```
YousefED, line 59: The fact that you can have an editor that can be unmounted and mounted is not super straightforward to grasp I think. Do we want (need) to mirror this PM concept directly to the BlockNote API, or can we come up with something simpler? When do we need this?

Example; it will be quite confusing to explain that an `init` method can be called multiple times. I'd rather prevent having to explain this by coming up with something simpler - if possible (I haven't thought about solutions so not sure it's feasible)

packages/core/src/editor/managers/ExtensionManager.ts
```diff
@@ -1,130 +1,585 @@
...
   /**
-   * Get all extensions
+   * Register one or more extensions to the editor
    */
-  public getExtensions() {
-    return this.editor.extensions;
+  public registerExtension(
+    extension:
+      | undefined
+      | string
```
YousefED: when is "string" used?

packages/core/src/editor/managers/ExtensionManager/index.ts
```diff
@@ -1,130 +1,585 @@
...
-   * Get the link toolbar plugin
+   * Allows resetting the current prosemirror state's plugins
+   * @param update - A function that takes the current plugins and returns the new plugins
+   * @returns void
    */
-  public get linkToolbar(): LinkToolbarProsemirrorPlugin<any, any, any> {
-    return this.editor.extensions[
-      "linkToolbar"
-    ] as LinkToolbarProsemirrorPlugin<any, any, any>;
+  private updatePlugins(update: (plugins: Plugin[]) => Plugin[]): void {
```
YousefED, line 302: do we need to pass a function here? Or can we just pass the new plugins? (seems a bit cumbersome / difficult to read)

packages/core/src/editor/managers/ExtensionManager.ts
```diff
@@ -1,130 +1,585 @@
...
-   * Check if collaboration is enabled (Yjs or Liveblocks)
+   * Get all the Tiptap extensions BlockNote is configured with by default
    */
-  public get isCollaborationEnabled(): boolean {
-    return (
-      this.hasExtension("ySyncPlugin") ||
-      this.hasExtension("liveblocksExtension")
-    );
-  }
+  private getDefaultTiptapExtensions = () => {
```
YousefED: Let's move this to a separate file, so that the ExtensionManager is really only about the logic around extension management?

(moving would imo more cleanly separate a lot of imports which are not directly related to managing extensions)

packages/core/src/editor/managers/ExtensionManager/index.ts
```diff
@@ -1,130 +1,585 @@
...
+        );
+      }
+
+      this.getProsemirrorPluginsFromExtension(extension).forEach((plugin) => {
+        pluginsToAdd.add(plugin);
+      });
+    }
+
+    // TODO there isn't a great way to do sorting right now. This is something that should be improved in the future.
+    this.updatePlugins((plugins) => [...plugins, ...pluginsToAdd]);
```
YousefED, line 171: this always triggers a state update (reconfigure) when initializing the plugins, right? Is this ok or would we rather find a way to directly pass the initial set of plugins to EditorState.create? Or does this not really matter / infeasible?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -740,105 +551,14 @@ export class BlockNoteEditor<
...
-          return undefined;
-        },
-      ),
-    ].filter((ext): ext is Extension => ext !== undefined);
+    this._eventManager = new EventManager(this as any);
+    this._extensionManager = new ExtensionManager(this, newOptions);
+
+    const tiptapExtensions = this._extensionManager.getTiptapExtensions();
+
+    this.tiptapExtensions = tiptapExtensions;
```
YousefED: do we need to store this as a class member?

packages/core/src/editor/BlockNoteExtension.ts
```diff
@@ -60,17 +77,27 @@ export abstract class BlockNoteExtension<
...
-      editor: BlockNoteEditor<BlockSchema, InlineContentSchema, StyleSchema>;
-    }) => boolean
+    (ctx: { editor: BlockNoteEditor<any, any, any> }) => boolean
   >;
 
-  public tiptapExtensions?: AnyExtension[];
+  /**
+   * Add additional prosemirror plugins to the editor.
+   */
+  readonly plugins?: ReadonlyArray<Plugin>;
```
YousefED: maybe name `prosemirrorPlugins`?

packages/core/src/editor/BlockNoteExtension.ts
```diff
@@ -96,23 +123,83 @@ export type InputRule = {
   }) => undefined | PartialBlockNoDefaults<any, any, any>;
 };
 
+export type ExtensionFactory<
```
YousefED: what's the use of an ExtensionFactory instead of an Extension directly? Let's document here

packages/core/src/extensions/BlockChange/BlockChange.ts
```diff
@@ -3,33 +3,27 @@ import {
   BlocksChanged,
   getBlocksChangedByTransaction,
 } from "../../api/getBlocksChangedByTransaction.js";
-import { BlockNoteExtension } from "../../editor/BlockNoteExtension.js";
+import { createExtension } from "../../editor/BlockNoteExtension.js";
 
 /**
  * This plugin can filter transactions before they are applied to the editor, but with a higher-level API than `filterTransaction` from prosemirror.
```
YousefED, line 9: how does this plugin relate to EventManager.onChange? where is this plugin used?

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -57,306 +61,279 @@ function getUpdatedThreadPositions(doc: Node, markType: string) {
...
+  (editor: BlockNoteEditor<any, any, any>, options) => {
+    const commentsOptions = options?.comments as
```
YousefED: should we remove `BlockNoteEditor.options.comments`, and just let the consumer create this plugin with the correct options?

packages/core/src/extensions/FilePanel/FilePanel.ts
```diff
@@ -1,206 +1,79 @@
...
-            editor,
-            filePanelPluginKey as any,
-            editorView,
-            (state) => {
-              this.emit("update", state);
-            },
-          );
-          return this.view;
-        },
+  // reset the menu when the document changes
```
YousefED: when can this happen / are we sure this is needed?

packages/core/src/extensions/FilePanel/FilePanelPlugin.ts
```diff
@@ -1,206 +1,79 @@
...
+    store,
+    closeMenu,
+    plugins: [
+      // TODO annoying to have to do this here
+      new Plugin({
         props: {
           handleKeyDown: (_view, event: KeyboardEvent) => {
-            if (event.key === "Escape" && this.shown) {
-              this.view?.closeMenu();
+            if (event.key === "Escape" && store.state.blockId) {
```
YousefED: can floating ui take care of this?

packages/core/src/extensions/FilePanel/FilePanelPlugin.ts
```diff
@@ -1,206 +1,79 @@
...
+    init({ dom, root, abortController }) {
+      dom.addEventListener("mousedown", closeMenu, {
+        signal: abortController.signal,
+      });
+      dom.addEventListener("dragstart", closeMenu, {
+        signal: abortController.signal,
+      });
+
+      root.addEventListener(
+        "scroll",
```
YousefED: my guess is that this was previously here to update the position on scroll, but that floatingui can handle this now indeed

packages/core/src/extensions/FormattingToolbar/FormattingToolbar.ts
```diff
@@ -1,363 +1,75 @@
...
               return true;
             }
             return false;
           },
         },
       }),
-    );
-  }
+    ],
+    init({ dom }) {
```
YousefED: same as in FilePanelPlugin

I'm vary curious which of these handlers are actually still needed / can't be done by floating UI.

I'd also suggest that any code remaining in the plugins themselves related to event handling should be closely documented (why it's needed here + how to reproduce the scenario)

packages/core/src/extensions/History/History.ts
```diff
@@ -0,0 +1,16 @@
+import { History } from "@tiptap/extension-history";
+import { createExtension } from "../../editor/BlockNoteExtension.js";
+import { redo, undo } from "@tiptap/pm/history";
+
+export const HistoryExtension = createExtension((_editor, options) => {
+  if (options.collaboration) {
```
YousefED: I'd argue the entire extension shouldn't be added instead of adding this check here (or throw an error if collaboration is enabled)

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
YousefED: this was here for accessibility reasons - but indeed, there needs to be a better / more scalable architecture

packages/core/src/extensions/FormattingToolbar/FormattingToolbarPlugin.ts
```diff
@@ -1,363 +1,75 @@
...
-
-    // Checks if menu should be shown/updated.
-    if (!this.preventShow && (shouldShow || this.preventHide) && !jsdom) {
-      // Unlike other UI elements, we don't prevent the formatting toolbar from
-      // showing when the editor is not editable. This is because some buttons,
-      // e.g. the download file button, should still be accessible. Therefore,
-      // logic for hiding when the editor is non-editable is handled
-      // individually in each button.
-      const newReferencePos = this.getSelectionBoundingBox();
+export const FormattingToolbarExtension = createExtension((editor) => {
```
YousefED: I think the "Plugin" files need to be renamed, right? (to Extension)

packages/core/src/extensions/SuggestionMenu/getDefaultSlashMenuItems.ts
```diff
@@ -257,11 +259,7 @@ export function getDefaultSlashMenuItems<
         });
 
         // Immediately open the file toolbar
-        editor.transact((tr) =>
-          tr.setMeta(editor.filePanel!.plugins[0], {
-            block: insertedBlock,
-          }),
-        );
+        editor.getExtension(FilePanelPlugin)?.showMenu(insertedBlock.id);
```
YousefED: Nice. let's do a search on `tr.setMeta` at end of refactor to see if we missed any occurences throughout the codebase

packages/core/src/extensions/UI_PLUGINS.md
```diff
@@ -0,0 +1,36 @@
+# UI Plugins to refactor
```
YousefED: TODO: remove file before merging

packages/core/src/index.ts
YousefED: Agree. I think to keep our codebase manageable we should migrate (can be step by step) to a system where every directory has an `index` file defining the exports. And then parent directories should only import those.

Open to better patterns of course - but at least good to make steps in this area

packages/react/src/hooks/useExtension.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Use an extension instance
+ */
+export function usePlugin<const T extends ExtensionFactory | Extension>(
+  plugin: T,
+  ctx?: { editor?: BlockNoteEditor },
+): T extends ExtensionFactory ? NonNullable<ReturnType<T>> : T {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const editor = ctx?.editor ?? useBlockNoteEditor();
+
+  const instance = useMemo(() => editor.getExtension(plugin), [editor, plugin]);
```
YousefED: Technically, I think this might not work when a plugin is added later (editor, plugin will be stable so the memo won't trigger). Probably that code path is not used rn

packages/react/src/hooks/usePlugin.ts
```diff
@@ -0,0 +1,58 @@
...
+import { useStore } from "@tanstack/react-store";
+import { useMemo } from "react";
+import { useBlockNoteEditor } from "./useBlockNoteEditor.js";
+
+type Store<T> = ReturnType<typeof createStore<T>>;
+
+/**
+ * Use an extension instance
+ */
+export function usePlugin<const T extends ExtensionFactory | Extension>(
```
YousefED: useExtension?

packages/react/src/components/FilePanel/FilePanelController.tsx
```diff
@@ -1,68 +1,42 @@
...
-  if (!editor.filePanel) {
-    throw new Error(
-      "FileToolbarController can only be used when BlockNote editor schema contains file block",
-    );
-  }
+  const filePanel = usePlugin(FilePanelPlugin);
+  const state = usePluginState(FilePanelPlugin, {
+    selector: (state) => {
+      return {
+        blockId: state.blockId || filePanel.store.prevState.blockId,
```
YousefED: why prevState?

packages/react/src/components/Popovers/BlockPopover.tsx
```diff
@@ -0,0 +1,80 @@
...
+    getBlockId: (editor: BlockNoteEditor<any, any, any>) => string | undefined;
+    placement?: "before" | "after" | "across";
+  },
+) => {
+  const { getBlockId, placement, floatingUIOptions, elementProps, children } =
+    props;
+
+  const editor = useBlockNoteEditor<any, any, any>();
+
+  // Store the current `boundingClientRect` to use in case `getBlockId` returns
```
YousefED: when / why would this happen? Doesn't that indicate a bug?

packages/react/src/hooks/useEditorSelectionBoundingBox.ts
```diff
@@ -1,43 +1,13 @@
...
-  const [boundingBox, setBoundingBox] = useState(() => {
-    if (!enabled) {
-      return undefined;
-    }
-
-    return editor.getSelectionBoundingBox();
+  return useEditorState({
+    editor,
+    selector: ({ editor }) =>
+      enabled ? editor.getSelectionBoundingBox()?.toJSON() : undefined,
```
YousefED: - why `toJSON()`? (a) wasn't here before and (b) better to have this hook return the same thing as the method on the editor?
- (nice to have) maybe an `enabled` boolean on the hook would be nice.

packages/react/src/hooks/useElementHover.ts
```diff
@@ -0,0 +1,433 @@
...
+} from "@floating-ui/react";
+
+function clearTimeoutIfSet(timeoutRef: React.RefObject<number>) {
+  if (timeoutRef.current !== -1) {
+    clearTimeout(timeoutRef.current);
+    timeoutRef.current = -1;
+  }
+}
+
+export interface UseElementHoverProps<T = HTMLElement> {
```
YousefED: is this the file that was mostly copied from floating ui, and that we figured we might be able to simplify / wrap?

packages/react/src/hooks/useEditorState.ts
```diff
@@ -0,0 +1,229 @@
...
+   * @default "all"
+   */
+  on?: "all" | "selection" | "change";
+};
+
+/**
+ * To synchronize the editor instance with the component state,
+ * we need to create a separate instance that is not affected by the component re-renders.
+ */
+class EditorStateManager<
```
YousefED, line 56: I suppose this is mostly copied from tiptap? What were the changes needed?

packages/react/src/hooks/useMenuAt.ts
```diff
@@ -0,0 +1,118 @@
+import { BlockNoteEditor, getNodeById } from "@blocknote/core";
+import { VirtualElement } from "@floating-ui/utils";
+import { posToDOMRect } from "@tiptap/core";
+import { useEditorState } from "./useEditorState.js";
+
+/**
```
YousefED: this file is not used?

pnpm-lock.yaml
```diff
@@ -17603,6 +17643,14 @@ snapshots:
     transitivePeerDependencies:
       - supports-color
 
+  '@handlewithcare/prosemirror-inputrules@0.1.3(prosemirror-model@1.25.3)(prosemirror-state@1.4.3)(prosemirror-view@1.41.3)':
```
YousefED: can we get rid of `prosemirror-inputrule` as dependency?

nperez0111: Tiptap imports it, so we can't yet.
It is very small though, so I'm not too worried about it

packages/react/src/hooks/useElementHover.ts
nperez0111: It was doing more than what floating UI was doing, so I couldn't get rid of it

packages/core/src/extensions/Collaboration/YUndo.ts
nperez0111: Extensions can return undefined if their pre-conditions aren't valid. I think this is useful pattern to have since it does not complicate the caller to have to check each pre-condition for each extension individually. All extensions should act the same.

packages/core/src/blocks/File/helpers/render/createAddFileButton.ts
nperez0111: Disagree

packages/core/src/editor/managers/ExtensionManager/index.ts
nperez0111: > The fact that you can have an editor that can be unmounted and mounted is not super straightforward to grasp I think.

React has the concept of mounting & unmounting components, what is different here?

> Do we want (need) to mirror this PM concept directly to the BlockNote API, or can we come up with something simpler? When do we need this?

An unmounted editor is useful for server-side editing. It does not touch any of the DOM APIs and can still edit documents, even without JSDOM, which is very useful. It also is what is first created by the `useCreateBlockNote` hook, before being attached to a DOM element. So anything interacting with the editor on first render will touch this API. Before, this was worked around, by firing everything in effects so that they always had an editor that was mounted, but that complicates callsites for an implementation detail.

> Example; it will be quite confusing to explain that an init method can be called multiple times.

I can rename it to `mount`, if that is the confusing part (where init is assumed to happen once).

> I'd rather prevent having to explain this by coming up with something simpler - if possible

I don't know what is complicated about this to know how to make it simpler.

packages/core/src/editor/managers/ExtensionManager/index.ts
nperez0111: It allows you to re-order the plugins. Which is important for priority. It is a private method, so readability is not of much concern for me

packages/core/src/editor/managers/ExtensionManager/index.ts
nperez0111: Nope, so this function only ever gets called _after the editor is initialized_, so this state update is actually desirable. This allows us to switch on & off extensions at runtime like the ForkYDoc plugin was doing

packages/core/src/editor/managers/ExtensionManager/index.ts
nperez0111: This is necessary because this can only switch out prosemirror plugins at runtime, it can't switch out Tiptap extensions since that can have more widespread effects (since a Tiptap extension can even add/remove to the schema). So the idea here is to warn: "hey that won't work. If you really want that Tiptap extension, you need it at initialization time & it stays there for the lifetime of the editor instance". Happy to workshop wording & naming here.

packages/react/src/hooks/useEditorState.ts
nperez0111: The only change needed here was the `on` and being able to pass in the BlockNoteEditor instance and typing for that to work properly. Otherwise is what I wrote for Tiptap

packages/react/src/hooks/useEditorSelectionBoundingBox.ts
nperez0111: .toJSON because result of the `selector` needs to be comparable, and Tiptap defines posToDom as having a new `toJSON` function each time. I've just fixed what the editor returns to not have to worry about this.

Having enabled on the `useEditorState` is interesting, but I fear it may encourage somewhat bad patterns, so I'll leave as is for now

packages/core/src/extensions/Collaboration/ForkYDoc.ts
nperez0111: I just made all the members private and flipped the store when I need it to be different. No need to try to make a derived or anything like that right not

packages/core/src/index.ts
nperez0111: Yep, did that. Also exported extensions only as `@blocknote/core/extensions` so really cleaned up the top-level API.

examples/09-ai/01-minimal/src/App.tsx
```diff
@@ -34,14 +35,13 @@ export default function App() {
...
     },
     // Register the AI extension
-    extensions: [
-      createAIExtension({
-        transport: new DefaultChatTransport({
-          // URL to your backend API, see example source in `packages/xl-ai-server/src/routes/regular.ts`
-          api: `${BASE_URL}/regular/streamText`,
-        }),
+    extensions: [AIExtension()],
+    ai: {
```
YousefED: shouldn't this be passed to AIExtension()?

packages/core/src/editor/BlockNoteExtension.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      key: "test",
+      prosemirrorPlugins: [],
+      arbitraryProperty: "arbitraryValue",
+      arbitraryMethod: () => {
+        return "arbitraryValue";
+      },
+    } as const;
+  });
+
+  const factory = extension({ x: 1 })({ editor });
```
YousefED: nitpick; `factory` should be `extensionInstance` or sth?

packages/core/src/editor/BlockNoteExtension.ts
```diff
@@ -123,78 +121,104 @@ type InputRule = {
...
+        ) => ExtensionFactoryInstance<ReturnType<Factory>>
+      : (options: Options) => ExtensionFactoryInstance<ReturnType<Factory>>
+    : () => ExtensionFactoryInstance<ReturnType<Factory>>;
 
-  if (!extension) {
-    return (() => undefined) as any;
+/**
+ * Constructs a BlockNote {@link ExtensionFactory} from a factory function or object
+ */
+export function createExtension<
```
YousefED, line 178: I think it could help (me at least..) to make sense of all overloads if every overload has a tiny 1-line example of what the usage looks like (`// this overload is for createExtension({...})`

packages/core/src/editor/managers/ExtensionManager/extensions.ts
```diff
@@ -0,0 +1,221 @@
...
+    FilePanel(options),
+    FormattingToolbar(options),
+    LinkToolbar(options),
+    NodeSelectionKeyboard(),
+    Placeholder(options),
+    ShowSelection(options),
+    SideMenu(options),
+    SuggestionMenu(options),
+    TrailingNode(),
+  ].filter((a) => a !== undefined) as ExtensionFactoryInstance[];
```
YousefED: do we still need the filter? are there still valid cases where an extension can return undefined?

packages/core/src/editor/managers/ExtensionManager/extensions.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  if (options.collaboration) {
```
YousefED, line 192: sweet

packages/core/src/editor/managers/ExtensionManager/extensions.ts
```diff
@@ -0,0 +1,221 @@
...
+    extensions.push(
+      Comments({ ...options.comments, resolveUsers: editor.resolveUsers! }),
+    );
+  }
+
+  if ("table" in editor.schema.blockSpecs) {
+    extensions.push(TableHandles(options));
+  }
+
+  if (options.animations === false) {
```
YousefED: hmm should this be !== false?

packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
```diff
@@ -2,73 +2,86 @@ import { getMarkRange, posToDOMRect } from "@tiptap/core";
...
-      });
-    },
+      return {
+        range: markRange,
+        mark,
+        get text() {
+          return tr.doc.textBetween(markRange.from, markRange.to);
+        },
+        get position() {
+          // to minimize re-renders, we convert to JSON, which is the same shape anyway
```
YousefED, line 40: I'm curious how converting to json minimizes rerenders?

packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
```diff
@@ -23,15 +23,18 @@ export const ExperimentalMobileFormattingToolbarController = (props: {
     InlineContentSchema,
     StyleSchema
   >();
-  const state = useUIPluginState(
-    editor.formattingToolbar.onUpdate.bind(editor.formattingToolbar),
-  );
+
+  const show = useExtensionState(FormattingToolbarExtension, {
+    editor,
+    selector: (state) => state,
```
YousefED: this selector looks a bit weird?

packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
nperez0111: Tiptap's implementation of `posToDomRect` creates a new instance of `toJSON` everytime that the function is called. In `useEditorState` we rely on deep equal comparisons, which will always fails with this since it is a new function instance.

A DOMRect is really just a struct for x, y, etc. So it being a JSON is almost exactly the same since it has no methods on it.

PR #2151, opened 2025-11-03
test: add test cases for complex attributes

nperez0111:
# Summary

## Checklist

### DIFF
--- a/packages/core/src/blocks/defaultBlockTypeGuards.ts
+++ b/packages/core/src/blocks/defaultBlockTypeGuards.ts
@@ -27,21 +27,19 @@ export function editorHasBlockWithType<
     return true;
   }
 
-  const editorProps: PropSchema =
-    editor.schema.blockSpecs[blockType].config.propSchema;
+  // const editorProps: PropSchema =
+  //   editor.schema.blockSpecs[blockType].config.propSchema;
 
   // make sure every prop in the requested prop appears in the editor schema block props
-  return Object.entries(props._zodSource._zod.def.shape).every(
-    ([key, value]) => {
-      return true;
-      // we do a JSON Stringify check as Zod doesn't expose
-      // equality / assignability checks
-      // return (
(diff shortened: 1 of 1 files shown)


tests/src/unit/core/testSchema.ts
```diff
@@ -33,13 +38,27 @@ const SimpleImage = addNodeAndExtensionsToSpec(
...
+  createBlockConfig(
+    () =>
+      ({
+        type: "customParagraph",
+        propSchema: defaultPropSchema,
+        content: "inline",
+      }) as const,
+  ),
   {
+    parse: (e) => {
```
YousefED, line 51: why did you add a parse function here?

tests/src/unit/react/testSchema.tsx
```diff
@@ -65,6 +65,35 @@ const createContextParagraph = createReactBlockSpec(
   },
 );
 
+// const ComplexAttributeNode = addNodeAndExtensionsToSpec(
```
YousefED, line 68: we can just add this, right? (without `addNodeAndExtensionsToSpec`)

tests/src/unit/core/testSchema.ts
nperez0111: For the round trip parse tests, this also makes sure that custom blocks will always have priority over the default blocks. 

Paragraph would also be able to match this: https://github.com/TypeCellOS/BlockNote/pull/2151/files#diff-f8e7f7017e58c7256b18e4d7b5fe8b4007e0ada0dadadf7306bb232cafbe84fa
So this makes sure that it round-trips back to a custom paragraph

tests/src/unit/react/testSchema.tsx
nperez0111: We can add it, it should not use that `addNodeAndExtensionstoSpec` function. I think I just didn't do it because the type guards weren't passing to get this to work right

PR #2164, opened 2025-11-10
fix(comments): always surface the closest mark to the current position #2073

### LINKED ISSUE #2073: Overlapping comment
Currently you can't comment on a selection that already has a comment.

Be able to comment on a portion or whole selection that already got a comment. 

Google Docs allows this but it's far from optimal
- the highlighting is the same color you can't see the over lapping comment
- the comment icon is hard to find (you don't see it if your screen doesnt show the whole toolbar)

https://github.com/user-attachments/assets/599850e1-8647-42e1-aa07-e41c4fe783cc

**Additional context**

https://github.com/user-attachments/assets/4a3accc6-614a-4997-9ebd-6c02f5b79503

**Bonus**

nperez0111:
# Summary

This changes the logic for selecting a mark to try to surface the mark which is actually closest to the clicked position. Otherwise, it will give preference to marks which span shorter lengths (which are more likely to be hidden when overlapping with other marks).

## Rationale

What is happening here is that a text range can contain several marks overlapping on it (pointing to different thread ids). This attempts to surface the "correct" comment by picking which ever is actually closest to the selected position in absolute distance.

## Changes

Sorts the list of marks by distance, and chooses the closest mark to the current position to select the thread of

## Testing

Manual testing was done for this. I'm not sure how we'd test this event handler right now, and we want to get this in quickly

## Checklist

## Additional Notes

Fixes #2073

### DIFF
--- a/packages/core/src/extensions/Comments/CommentsPlugin.ts
+++ b/packages/core/src/extensions/Comments/CommentsPlugin.ts
@@ -12,6 +12,7 @@ import { BlockNoteEditor } from "../../editor/BlockNoteEditor.js";
 import { BlockNoteExtension } from "../../editor/BlockNoteExtension.js";
 import { CustomBlockNoteSchema } from "../../schema/schema.js";
 import { UserStore } from "./userstore/UserStore.js";
+import { getMarkRange } from "@tiptap/core";
 
 const PLUGIN_KEY = new PluginKey(`blocknote-comments`);
 const SET_SELECTED_THREAD_ID = "SET_SELECTED_THREAD_ID";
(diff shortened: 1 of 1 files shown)


matthewlipski (approved):
Looks good, though am I right in thinking that there's still an issue if 2 comments span the exact same range?

nperez0111:
> Looks good, though am I right in thinking that there's still an issue if 2 comments span the exact same range?

Yep, theoretically, we could cycle between the marks at a range, (i.e. you have to click again to see the "next" mark), but honestly that is an edge case of an edge case that isn't worth it right now

matthewlipski:
Yeah makes sense, probably more sensible to come up with a UI solution to see multiple threads at once at that point anyway (which ofc is way out of scope), since cycling through threads on click would probably be confusing to users.

PR #2165, opened 2025-11-10
fix: Scrolling bug when clicking comment marks

### LINKED ISSUE #2158: Comments: Opening a thread trigger a scroll to the top
**Describe the bug**
When we click on a text with a Thread, it triggers a scroll to the top of the page.

It is still doing it on the BN exemple: https://www.blocknotejs.org/examples/collaboration/comments, but not only, we have the problem on our side as well.

This fix didn't seem to work: https://github.com/TypeCellOS/BlockNote/pull/2043

**To Reproduce**

https://github.com/user-attachments/assets/ea3c5033-c54c-4ac6-876c-11d101e0ad86

https://github.com/user-attachments/assets/59e9bf94-4056-4bcf-8241-f02d1eb43c38

**Misc**

- Node version:
- Package manager:
- Browser:

matthewlipski:
# Summary

This PR fixes a bug which could cause the page to be scrolled to the top when clicking a comment mark. This was because we would attempt to scroll to the floating thread view that pops up after the comment is selected. This would happen before FloatingUI had a chance to position the thread view, so the default position of {0, 0} was used and scrolled to.

This PR removes scrolling to the floating thread view when selecting a comment mark, and instead just scrolls to the mark itself. This makes it reusable for all UI libraries and means we don't have to deal with any FloatingUI weirdness.

Additionally, the `FloatingThreadController` now uses FloatingUI's `autoUpdate` so threads scroll with the editor.

Closes #2158

## Rationale

The scrolling to the top of the page is pretty annoying when trying to view comments.

## Changes

See above.

## Impact

N/A

## Testing

N/A

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/react/src/components/Comments/FloatingThreadController.tsx
+++ b/packages/react/src/components/Comments/FloatingThreadController.tsx
@@ -6,7 +6,13 @@ import {
   InlineContentSchema,
   StyleSchema,
 } from "@blocknote/core";
-import { UseFloatingOptions, flip, offset, shift } from "@floating-ui/react";
+import {
+  UseFloatingOptions,
+  autoUpdate,
+  flip,
+  offset,
+  shift,
+} from "@floating-ui/react";
 import {
   ComponentProps,
   FC,
@@ -55,6 +61,7 @@ export const FloatingThreadController = <
           editor.focus();
         }
       },
+      whileElementsMounted: autoUpdate,
       ...props.floatingOptions,
     });
 


packages/mantine/src/comments/Card.tsx
nperez0111, line 1: Always happy to remove an effect!

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -289,10 +289,11 @@ export class CommentsPlugin extends BlockNoteExtension {
...
       // the editor.
-      (
-        this.editor.prosemirrorView?.domAtPos(selectedThreadPosition.from)
-          .node as Element | undefined
-      )?.scrollIntoView({
+      const node = this.editor.prosemirrorView?.domAtPos(
+        selectedThreadPosition.from,
+      ).node as Element | undefined;
+
+      node?.scrollIntoView({
```
nperez0111: Is this not equivalent?

packages/core/src/extensions/Comments/CommentsPlugin.ts
```diff
@@ -245,7 +245,7 @@ export class CommentsPlugin extends BlockNoteExtension {
             );
 
             const threadId = commentMark?.attrs.threadId as string | undefined;
-            self.selectThread(threadId, false);
+            self.selectThread(threadId);
```
nperez0111, line 248: What is the second param? Why is it removed

examples/07-collaboration/05-comments/src/style.css
nperez0111, line 1: rm

examples/07-collaboration/05-comments/src/App.tsx
nperez0111, line 1: rm

packages/core/src/extensions/Comments/CommentsPlugin.ts
matthewlipski: You could set it to false to prevent scrolling to the mark (`scrollToThread`). We did this before and handled the scrolling in the `Card` component instead. Turns out. this was the only place it was set so I removed it outright.

nperez0111:
Unsure if this was introduced @matthewlipski but when I was testing this I ran into this

https://github.com/user-attachments/assets/e22da0ad-3084-49d5-99d4-3dc197f4afcb

Is this what you meant with your comment? Could we do something quick about this? I think it might just be a race condition of the scroll happening after the floating happens (or vice versa?)

nperez0111:
Fantastic, thanks @matthewlipski

Issue #2170, opened 2025-11-12
A RangeError: No value supplied for attribute previewWidth has occurred


MyungJiwoo:
When using the `/image` command in the default editor setup following the documentation, the image is not added, and the console displays the following error.

This issue occurs in both **v0.41.1, v0.42.0.**

<img width="656" height="662" alt="Image" src="https://github.com/user-attachments/assets/af207f20-4376-4d96-8dc4-49d2f05ee96e" />

[According to the documentation](https://www.blocknotejs.org/docs/features/blocks/embeds#image), `previewWidth: number = 512;` is specified, but in the actual code, [it was declared as `undefined`.](https://github.com/TypeCellOS/BlockNote/blob/main/packages/core/src/blocks/Image/block.ts)

In the current version, applying the following temporary fix allowed the editor to work without any errors.
```tsx
import {
  BlockNoteSchema,
  defaultBlockSpecs,
  createCodeBlockSpec,
} from "@blocknote/core";

const createSchema = () => {
  const blockSpecs = {
    ...defaultBlockSpecs,

    image: {
      ...defaultBlockSpecs.image,
      config: {
        ...defaultBlockSpecs.image.config,
        propSchema: {
          ...defaultBlockSpecs.image.config.propSchema,
          previewWidth: {
            ...defaultBlockSpecs.image.config.propSchema.previewWidth,
            default: 512 as const,
          },
        },
      },
    },

    codeBlock: createCodeBlockSpec({}),
  };

  return BlockNoteSchema.create({ blockSpecs });
};
```
```tsx
const editor = useCreateBlockNote({ schema: createSchema() });
```

AntoLC:
You can see the error here: https://stackblitz.com/edit/github-gj7clar3?file=package.json,index.html

https://github.com/user-attachments/assets/7e72c275-2363-40df-b11e-1656048b5b74

The problem is not only with images.

fraudenwp:
I hit the same RangeError: “No value supplied for attribute start.” The root cause was saved JSON blocks for `numberedListItem` missing `props.start`, but ProseMirror treats `start` as required.

I fixed it by sanitizing the content before passing it to BlockNote:

```ts
const sanitizeContent = (raw: string) => {
  const fallback = [
    {
      type: 'paragraph',
      props: { textColor: 'default', backgroundColor: 'default', textAlignment: 'left' },
      content: [{ type: 'text', text: '', styles: {} }],
    },
  ];

  try {
[...]

PR #2171, opened 2025-11-12
fix: set default previewWidth to 512 to prevent RangeError

MyungJiwoo:
#2170 There was an issue similar to this one.

While it is possible for users to override it manually, I believe it would be more appropriate to update the type definition to match the official documentation.


nperez0111:
Closing this as it is not actually the solution to the problem, the problem is upstream in Tiptap, downgrade if you need a quick fix, they hopefully will fix it soon, or we can set to null

PR #2177, opened 2025-11-14
fix: remove dependency array from comments re-rendering

nperez0111:
# Summary

I think that this was being used to allow the editor to re-render if it saw a new comment.body, but the only scenario that I can even imagine that happening is that the same person is logged in on two different browsers & updates their comment on one and has the comment editor open on the other. In that case they will now see a stale value. If we really cared about this case, then we can change the Comments component to use a React.key that forces the Comment component to re-render based on the new comment body content, but I think it is enough of an edge case that it isn't even worth implementing that.

## Checklist

### DIFF
--- a/packages/react/src/components/Comments/Comments.tsx
+++ b/packages/react/src/components/Comments/Comments.tsx
@@ -23,7 +23,7 @@ export const Comments = ({
   // Maps all comments to elements.
   const comments = thread.comments.map((comment, index) => (
     <Comment
-      key={comment.id}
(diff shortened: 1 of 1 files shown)


YousefED (approved):
I think you're right. Can you make sure with manual testing this doesn't cause stale data outside of the edge-case you mentioned?

nperez0111:
I've just gone with the React.key, to make doubly sure. This should be exactly the same behavior now.

Issue #2176, opened 2025-11-13
Export fails when paste with unknown color


AntoLC:
**Describe the bug**

When we copy a text that has a style attach, and paste it in the doc, the export will then fails (pdf / odt / docx..).

## How to reproduce

- Copy any text from github
- Paste it to a doc
- Export
- The export fails

## Why

A paste color can be in `rgb`:

<img width="567" height="237" alt="Image" src="https://github.com/user-attachments/assets/c70d29ae-ad81-4020-a1b4-edb2bd04ef35" />

## Problem location
https://github.com/TypeCellOS/BlockNote/blob/d192b37eadfae7d241a5d90200ad0c21a0d0a42b/packages/xl-pdf-exporter/src/pdf/pdfExporter.tsx#L299-L301
https://github.com/TypeCellOS/BlockNote/blob/d192b37eadfae7d241a5d90200ad0c21a0d0a42b/packages/xl-pdf-exporter/src/pdf/pdfExporter.tsx#L305-L307

The array is possibly undefined, it should be something like:
`this.options.colors?.[props.backgroundColor as keyof typeof this.options.colors]?.background`

https://github.com/search?q=repo%3ATypeCellOS%2FBlockNote%20props.textColor%20as&type=code

**Misc**

- Node version:
- Package manager:
- Browser:
- [x] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

AntoLC:
Btw, a comment has this color (rgb(237 180 0 / 40%)), so making a comment and trying to export will trigger an error.

PR #2179, opened 2025-11-17
fix: do not error on invalid backgroundColor or textColor #2176

nperez0111:
# Summary
There can be invalid `textColor` or `backgroundColor` in a document. We were assuming that it always was a valid color. This changes all of the places where we read the backgroundColor or textColor to not make that assumption & access a property which does not actually exist.

There is some digging to be done here to understand whether we can prevent these invalid values from getting within the document, but at this point I think having this defensive code is still needed as I'm unsure that we can actually fully validate these string attributes in prosemirror.

Perhaps this is helped with the zod props, so this may not be required soon.

## Checklist

## Additional Notes

#2176


YousefED (approved):
ok for now, can you create a follow up ticket to investigate the root cause? (e.g.: how should we handle color pasting from external html

PR #2222, opened 2025-12-03
chore: Memoize Floating UI options in `TableHandlesController`

matthewlipski:
# Summary

This PR adds memoization to Floating UI options in `TableHandlesController`.

Additionally, a bug has been fixed with the extend rows/columns buttons. If the mouse cursor would move from the table element to the button and started dragging it, the position would update ok. If the mouse cursor would move from outside the table element (i.e. from elsewhere in the table block) and started dragging it, the position would no longer update. This was because in the latter case, the extension state's `rowIndex` and `colIndex` would be undefined, and that incorrectly prevented the table reference element from being updated, which in turn prevented the button positions from updating.

## Rationale

Minor code cleanup & useful bug fix.

## Changes

N/A

## Impact

N/A

## Testing

Went through the UI testing doc to ensure no regressions & updated doc with new test case.

## Screenshots/Video

N/A

## Checklist


matthewlipski:
Not really tbh, just clears up the code a little and makes things more readable.

PR #2223, opened 2025-12-03
fix: shadowdom support and example

YousefED:
# Summary

Support for using BlockNote in shadowdom was broken due to usage of `document.queryselector`

## Rationale

n/a

## Changes

- fix support by using `.closest`
- add example (to playground, not to website), to simplify future testing

## Impact

n/a

## Testing

manual

## Screenshots/Video

n/a

## Checklist

## Additional Notes

n/a


YousefED:
> Any other usage we should be aware of?

I didn't find any other breaking functionality. Checked explicitly for usage of `document.queryselector*`).

In general, we should be careful with any `document.` and `window.` calls (also because of usage in server-side environments of course). (cc @matthewlipski )

Unfortunately I don't think it's easy to enforce this with eslint

matthewlipski:
R.e. e2e failures, wait for #2231 to be merged since that fixes them. Nothing major, the custom block e2e tests just use different URLs which aren't appended with the `hideMenu` query, so the sidebar gets shown in their screenshots. That means adding unrelated examples will cause the screenshot comparison to fail.

PR #2224, opened 2025-12-03
chore: Move side menu height setting to FloatingUI

matthewlipski:
# Summary

This PR makes it so that the height of the side menu is now set in FloatingUI middleware on `SideMenuController`. This is a change from the old method, which uses a more convoluted `useEffect` hook + CSS.

Additionally, it's been changed so that the height is no longer hardcoded, but instead the block content element's height is used. In some cases, the side menu is instead aligned to an element nested deeper inside, like for table blocks.

To streamline everything and make it more consistent, the side menu is now always centred vertically. Before, it would change depending on the block. E.g. it would be centred to the first line of paragraph and heading blocks, but aligned to the top of most file blocks, though not the audio block. This was done to make it more similar to Notion, but this resulted in a bunch of arbitrary exceptions having to be made.

Additionally, two bugs with the file panel and side menu have been fixed. The file panel would show up below any nested blocks, while the side menu was horizontally indented with nested blocks. These were fixed by adding more options to the `BlockPopover` component, detailed in the changes list below.

## Rationale

This continues with the theme of moving responsibilities to FloatingUI where possible, which we started doing as part of the major refactor in #2143. The bugs are also regressions that should be fixed.

## Changes

- Removed `useEffect` hook from `SideMenu` component.
- Removed setting side menu height in CSS.
- Added `size` middleware to `SideMenuController` to set the side menu height.
- Added `includeNestedBlocks` option to `BlockPopover`. When `true`, this works the same as before, using the entire block as the reference element. When `false`, it uses the block's content as the reference element. Defaults to `true`.
- Added `spanEditorWidth` option to `BlockPopover`. When `false`, this works the same as before, where the `boundingClientRect` is the same size & position as the reference element. When `true`, the `x` position and width are used for the `boundingClientRect`, and only the reference element `y` position and height are used.

## Impact

This actually also fixes a previously unknown bug with the file panel thanks to that addition of `includeNestedBlocks`. Because it used the whole block as the reference element, it would show beneath all child blocks.

## Testing

Went through the Notion UI testing doc to ensure no regressions for UI elements using the `BlockPopover` (file panel, side menu & AI menu). Added test cases to all UI elements for nested blocks.

## Screenshots/Video

N/A

## Checklist


packages/react/src/components/Popovers/BlockPopover.tsx
```diff
@@ -28,18 +30,51 @@ export const BlockPopover = (
...
-        if (!(node instanceof Element)) {
+        ).node;
+        if (!(blockElement instanceof Element)) {
           return undefined;
         }
 
-        return {
-          element: node,
-        };
+        const blockContentElement = blockElement.firstElementChild;
```
YousefED, line 40: I'm not sure I like the use of `firstElementChild`. It assumes a specific dom structure to get the dom element - which is not typesafe / futureproof.

Better approach would be to use the methods we have like getblockinfo to get the content block, and then call getdomatpos for it

(example; the current approach would probably break for column blocks which don't have blockContent)

packages/react/src/components/SideMenu/SideMenuController.tsx
```diff
@@ -29,6 +34,72 @@ export const SideMenuController = (props: {
       useFloatingOptions: {
         open: show,
         placement: "left-start",
+        middleware: [
+          size({
+            apply({ elements }) {
+              // TODO: Need to fetch the block from extension, else it's
```
YousefED, line 48: this code is called in `useMemo`, so the apply function only gets recreated when  the deps array changes

packages/react/src/components/SideMenu/SideMenuController.tsx
```diff
@@ -29,6 +34,72 @@ export const SideMenuController = (props: {
...
+              // time it changes?
+              const block =
+                editor.getExtension(SideMenuExtension)?.store.state?.block;
+              if (block === undefined) {
+                return;
+              }
+
+              if (block.type === "heading") {
+                if (!block.props.level || block.props.level === 1) {
+                  elements.floating.style.setProperty("height", "78px");
```
YousefED: we can't base this on pixels here. What if someone changes the font or font-size in css?

YousefED (changes requested):
see comments.

Also it made me think of a separate thing; did we test the block-side menu in multi-column? afaik it wasn't listed under the test cases

nperez0111:
I want to understand this better, so @matthewlipski can we go over this at some point later? There is probably another approach here

packages/react/src/components/SideMenu/SideMenuController.tsx
```diff
@@ -47,56 +55,97 @@ export const SideMenuController = (props: {
...
+                return;
+              }
+
+              const blockContentBoundingClientRect =
+                blockContentElement.getBoundingClientRect();
+
+              // Special case for file blocks with captions. In this case, we
+              // align the side menu with the first sibling of the file caption
+              // element.
+              const fileCaptionParentElement =
```
YousefED, line 78: imo we can't have these special cases hardcoded here. The sidemenu should not be aware of our block types or their HTML structure. Can we come up with a more future-proof API?

Also, is it easy to add e2e tests for this to prevent regressions?

YousefED (changes requested):

packages/react/src/components/SideMenu/SideMenuController.tsx
matthewlipski: I'm not sure there's an alternative solution tbh. Either we make an exception for them, or things end up looking misaligned and wonky. I discussed an idea with @nperez0111 to maybe make it so that blocks that are taller than some height threshold have their side menus aligned to the top left corner of the block instead, but it wouldn't really address the issue for e.g. toggle blocks and audio blocks.

The only solution I think makes sense is if you extract the logic in `apply` to a separate function, and expose that to consumers, who can add their own exceptions. Something like this:

```ts
export type SideMenuException = (
  elements: {
    reference: ReferenceElement;
    floating: FloatingElement;
  }
) => { yOffset: number, height: number };

export function applySideMenuHeight(
  elements: {
    reference: ReferenceElement;
    floating: FloatingElement;
  },
  exceptions?: SideMenuException[]
): void;
```

Ofc naming and exact typing is tbd, but do you think this is worth implementing?

packages/react/src/components/SideMenu/SideMenuController.tsx
matthewlipski: Even when changing the `block` inside `apply` to use `state.block` instead, & adding it to the deps, it still doesn't seem to recreate the function which is why I was confused.

packages/react/src/components/SideMenu/SideMenuController.tsx
YousefED: what about a classname or other child element selector that we can query for? (unless @nperez0111 can come up with something cleaner)?

Imo most important is that this component should imo have no knowledge of different block types and their implementations

nperez0111: I thought the logic would look like:

```ts
const exceptions = [ ".file-caption", ".xyz" ]

let subtractFromHeight = 0;

exceptions.forEach(e=>{
subtractFromHeight += document.querySelector(e)?.getBoundingBox().height ?? 0
})
```

But, it's unclear to me whether @matthewlipski is doing any sort of alignment of buttons or not.

If it were just to subtract things from being counted as the height then either this, or just introducing a classname to put on any element we want to be discounted from the height would work like @YousefED mentioned.

PR #2229, opened 2025-12-04
fix: add preload option to video blocks

### LINKED ISSUE #2203: Gives ability to add props on block Video
We have pages with lot of videos and some of them can be huge. 
We don't want that they download automatically in background.
To do so, it could be nice to have the ability to add `preload="none"` on the video tag.

The problem is to override the block Video seems quite complicated from [packages/core/src/blocks/Video/block.ts](https://github.com/TypeCellOS/BlockNote/blob/9e5ffa8d117c749bd699fba3f97523184ba5fcb1/packages/core/src/blocks/Video/block.ts).

But we could use directly and [createVideoBlockSpec](https://github.com/TypeCellOS/BlockNote/blob/9e5ffa8d117c749bd699fba3f97523184ba5fcb1/packages/core/src/blocks/defaultBlocks.ts#L52), it could be nice to be able to add some property to it, to have the ability to add props on the video tag, example:

```ts
...
video: createVideoBlockSpec({
   preload: 'none', // ("none" | "metadata" | "auto")
}),
```
Then here: https://github.com/TypeCellOS/BlockNote/blob/9e5ffa8d117c749bd699fba3f97523184ba5fcb1/packages/core/src/blocks/Video/block.ts#L92-L95
It would be:
```ts
      video.preload = config.preload;
```

Thank you !

**Bonus**

YousefED:
# Summary

Closes https://github.com/TypeCellOS/BlockNote/issues/2203

## Rationale

With this fix, it's possible to set a `preload` option in the config of your video block in the schema.

I think this is cleaner than setting it on a block prop (as "preload" seems a UI concern, that something that should be stored in the document). The downside is that it's a single configuration that applies to all blocks

## Changes

x

## Impact

x

## Testing

manual

## Screenshots/Video

x

## Checklist

### DIFF
--- a/packages/core/src/blocks/Video/block.ts
+++ b/packages/core/src/blocks/Video/block.ts
@@ -11,6 +11,7 @@ export const FILE_VIDEO_ICON_SVG =
 
 export interface VideoOptions {
   icon?: string;
+  preload?: "none" | "metadata" | "auto";
 }
 
 export type VideoBlockConfig = ReturnType<typeof createVideoBlockConfig>;
@@ -93,6 +94,9 @@ export const createVideoBlockSpec = createBlockSpec(
       video.contentEditable = "false";
       video.draggable = false;
       video.width = block.props.previewWidth;
+      if (config.preload) {
+        video.preload = config.preload;
+      }
       videoWrapper.appendChild(video);
 
       return createResizableFileBlockWrapper(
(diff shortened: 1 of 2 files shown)


nperez0111:
I want to go a little further with this rather than just the preload option, but just any HTML attribute you may want on the element, or even a callback to apply whatever you want to the element. I'll get back to this.

Commented here: https://github.com/TypeCellOS/BlockNote/issues/2203#issuecomment-3631789485

AntoLC (commented):
Works well, thank you !

PR #2230, opened 2025-12-04
fix: html diff error with whitespace

YousefED:
# Summary

To reproduce the issue:
- add a block ending with whitespace, e.g.: "Hello, "
- open the AI menu (/ai)
- ask "continue writing, same line"
- use Sonnet 4.5

Without the code change in this PR, it would throw an "html diff" error because the whitespace would be (invalidly) discarded when parsing the HTML returned from the LLM.

This PR contains a fix, but breaks existing unit tests. Next steps:

- [x] Determine when we do and when we don't want `preserveWhitespace`
- [x] Add additional unit tests if needed

### DIFF
--- a/tests/src/unit/core/formatConversion/parse/__snapshots__/html/whitespace prefix.json
+++ b/tests/src/unit/core/formatConversion/parse/__snapshots__/html/whitespace prefix.json
@@ -1,0 +1,19 @@
+[
+  {
+    "children": [],
+    "content": [
+      {
+        "styles": {},
+        "text": " hello,",
+        "type": "text",
+      },
+    ],
+    "id": "1",
+    "props": {
+      "backgroundColor": "default",
+      "textAlignment": "left",
+      "textColor": "default",
+    },
+    "type": "paragraph",
+  },
+]
--- a/tests/src/unit/core/formatConversion/parse/parseTestInstances.ts
+++ b/tests/src/unit/core/formatConversion/parse/parseTestInstances.ts
@@ -942,6 +942,13 @@ console.log("Third Line")</code></pre>`,
     },
     executeTest: testParseHTML,
   },
+  {
+    testCase: {
+      name: "whitespace prefix",
+      content: `<p> hello,</p>`,
+    },
+    executeTest: testParseHTML,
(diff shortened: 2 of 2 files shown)


nperez0111:
@YousefED, my thinking on this one is, whether to preserve newlines when parsing HTML between nodes is actually dependent on the source of the HTML more than it is based on the schema. If we want to support that, I recommend that we add an option to the `tryParseHTMLToBlocks` which allows us to configure whether we want to preserve whitespace or not.

But, in this case, I think that what we are more concerned with is newline behavior within a node's content. In this case, from what I've seen I believe that preserveWhitespace: true is a good default to have, since we treat blocks as isolated units, which is slightly different from how prosemirror would typically structure this.

To get this to pass tests, I had to modify how a blockquote was being exported to HTML from markdown, in the intermediate step for markdown parsing to blocks. It was doing:

```
<blockquote>
<p>Text</p>
</blockquote>
```
when what we needed was:

```
<blockquote><p>Text</p></blockquote>
```

This, I think, should only be an issue for blockquotes, because most other elements do not have this "wrapper" structure, where the blockquote is wrapping a paragraph. Technically, the most correct solution, would be actually to take the contents of the paragraph & dump them into the blockquote, but it can be parsed correctly without it, as long as it does not have this newline start & endings added.

I've added a commit which implements this. I don't have my AI stuff set up correctly right now, so I didn't do further testing to see if this resolves the issue you were having. But, hopefully, you can understand my thinking & change course yourself if you need to.

YousefED:
@nperez0111 thanks for your thoughts and changes!

Unfortunately, removing the `preserveWhitespace` option in parseHTML breaks the issue this PR was trying to fix.

In https://github.com/TypeCellOS/BlockNote/pull/2230/commits/d38ecde879cc90277fa31effadd8f8f1d96be891, I've:
- added a testcase that's a good reproduction of the AI issue
- re-enabled `preserveWhitespace` option in parseHTML, but unfortunately that breaks other tests

When you look at the testcase, the interesting thing is actually that it's whitespace within a node (before the `</p>` closing tag, not between nodes).

PR #2232, opened 2025-12-04
fix: Suggestion menu positioning

matthewlipski:
# Summary

This PR fixes an issue where if the suggestion menu was displayed above the trigger character (when there was not enough space to display it below), and only showed a few items, there would be a large gap between the menu and trigger character. This was because the suggestion menu had a hard-coded minimum height, which wasn't filled when displaying a small number of items.

Also, the positioning has generally been improved to better take advantage of the available screen space.

## Rationale

This issue has made the suggestion menu feel janky in some situations.

## Changes

- Removed minimum height being set in `SuggestionMenuController` Floating UI options `size` middleware.
- Replaced `flip` middleware with `autoPlacement`. The difference is that `autoPlacement` always displays the menu on the side that has more space, whereas `flip` only changes the side when there is not enough available space for the menu.

## Impact

N/A

## Testing

Updated Notion UI testing doc.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/react/src/components/SuggestionMenu/GridSuggestionMenu/GridSuggestionMenuController.tsx
+++ b/packages/react/src/components/SuggestionMenu/GridSuggestionMenu/GridSuggestionMenuController.tsx
@@ -9,10 +9,7 @@ import {
   useExtensionState,
 } from "../../../hooks/useExtension.js";
 import { FloatingUIOptions } from "../../Popovers/FloatingUIOptions.js";
-import {
-  GenericPopover,
-  GenericPopoverReference,
-} from "../../Popovers/GenericPopover.js";
+import { GenericPopover } from "../../Popovers/GenericPopover.js";
 import { getDefaultReactEmojiPickerItems } from "./getDefaultReactEmojiPickerItems.js";
 import { GridSuggestionMenu } from "./GridSuggestionMenu.js";
 import { GridSuggestionMenuWrapper } from "./GridSuggestionMenuWrapper.js";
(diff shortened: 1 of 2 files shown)


packages/react/src/components/SuggestionMenu/GridSuggestionMenu/GridSuggestionMenuController.tsx
```diff
@@ -126,17 +126,22 @@ export function GridSuggestionMenuController<
...
+            allowedPlacements: ["bottom-start", "top-start"],
+            padding: 10,
           }),
           shift(),
           size({
-            apply({ availableHeight, elements }) {
-              Object.assign(elements.floating.style, {
-                maxHeight: `${availableHeight - 10}px`,
-              });
+            apply(p) {
```
YousefED: I'm not sure about this, maybe @nperez0111 can double check when he's back

nperez0111: Hm, yea I don't like this fix either. Will get back to it to see what we should do here. I think we may just need it to be re-evaluated once the underlying number of items changes

nperez0111: Alright, I see what @matthewlipski was trying to do here. I think that I would prefer not to keep the fix as it feels like more of a hack than needed.

What is happening here is that the autoplacement middleware (which is very similar to the flip middleware) is trying to find the best spot to put the menu, and is unaware on the first render how large the element actually is, because it has not yet requested the items since they can be async.

To fix this, we would need to introduce something which would allow the child to request that the parent re-calculate it's position when the number of items changes. Like a context of our own for popovers or https://floating-ui.com/docs/FloatingTree

I think we can look into doing this at a later time though, because while the UX is not ideal, I'd prefer correctness in the code we introduce.

packages/react/src/components/SuggestionMenu/SuggestionMenuController.tsx
```diff
@@ -127,18 +119,16 @@ export function SuggestionMenuController<
           offset(10),
           // Flips the menu placement to maximize the space available, and prevents
```
YousefED, line 120: is this comment still accurate?

PR #2233, opened 2025-12-04
fix: AI menu not updating position on new line

matthewlipski:
# Summary

There was a regression introduced in #2143 which made the AI menu no longer update its position when it creates a new line.

## Rationale

See above.

## Changes

- Made `AIMenuController` FloatingUI options call `autoUpdate` on each animation frame.

## Impact

N/A

## Testing

Added test case to Notion UI testing doc.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/xl-ai/src/components/AIMenu/AIMenuController.tsx
+++ b/packages/xl-ai/src/components/AIMenu/AIMenuController.tsx
@@ -5,7 +5,7 @@ import {
   useExtension,
   useExtensionState,
 } from "@blocknote/react";
-import { offset, size } from "@floating-ui/react";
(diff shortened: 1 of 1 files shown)


YousefED (approved):
Approved assuming that this was there before as well, correct?

(a nicer solution might be to update it on editor change, but I think this is also ok)

matthewlipski:
Correct, at some point while refactoring the AI menu components, it wasn't necessary so I removed it. But that was I think caused because of unnecessary re-renders every time the document updated (or something along those lines). So after clearing that up, the `animationFrame` option was needed again but not re-added.

Issue #2245, opened 2025-12-09
AIMenuController missing @floating-ui/react positioning


VihangaN:
### Problem
AIMenuController does not use @floating-ui/react for positioning. As a result, the menu does not automatically flip or adjust when near the bottom edge of the viewport, unlike other controllers. This can cause the menu to get cut off or appear too close to the edge, leading to a poor user experience.

### Proposed Solution
Integrate @floating-ui/react in AIMenuController so the menu automatically adjusts its position based on viewport boundaries, matching the behavior of other controllers.

### Workarounds Considered
Manually adjusting menu position via custom CSS or inline styles, which is inconsistent across different screen sizes and viewport changes.

### Additional Context
The issue is most noticeable on smaller windows or when the menu is triggered near the bottom of the screen, causing partial visibility or overlap.

### Screenshot

<img width="1271" height="860" alt="Image" src="https://github.com/user-attachments/assets/f41db301-68d1-4b55-8bf4-89512b8c5a46" />

nperez0111:
Thanks for pointing this out @VihangaN, it looks like we have the flip middleware commented out. I'll get this resolved

VihangaN:
Thanks for fixing this @nperez0111  🙌
When do you think this can go to a production release ? We’re currently having this issue on our web app as well.

PR #2247, opened 2025-12-09
fix(ai): re-enable flipping the AIMenu when there is not enough space #2245

nperez0111:
# Summary

I think we may want to do a couple more things in addition to this.

It makes sense to me that we would move the selection to be the range of blocks that the menu is trying to cover (it is hard to tell what the menu is referencing, and if the selection was visible I think it would make it easier).

There also seems to be an issue of scrolling to the wrong position when the menu is first opened. I think we should only try to scroll if it is actually out of bounds or something

## Checklist

## Additional Notes

Fixes https://github.com/TypeCellOS/BlockNote/issues/2245


matthewlipski (approved):
Seems to work ok

PR #2250, opened 2025-12-09
feat: allow customizing the rendering of video and audio elements #2203

### LINKED ISSUE #2203: Gives ability to add props on block Video

nperez0111:
# Summary
An attempt to supercede #2229

## Checklist

## Additional Notes

Resolves #2203


YousefED (commented):
Makes sense, couple of questions to nail this:

- should we use the name `render`? (it's a term we use in other places as well, so might get confusing? Or good for consistency?)
- Can we add a simple test for this?
- Do we want to implement this at the React layer (React File blocks) as well? Or rather first determine what our strategy is in the react vs native file blocks?

nperez0111:
> Makes sense, couple of questions to nail this:
> 
>     * should we use the name `render`? (it's a term we use in other places as well, so might get confusing? Or good for consistency?)

🤷 I haven't got a better name at the moment. 

>     * Can we add a simple test for this?

Of course.

>     * Do we want to implement this at the React layer (React File blocks) as well? Or rather first determine what our strategy is in the react vs native file blocks?

Now that you mention this, I think this might go a bit deeper, whether this is working towards or against the idea of separating the schema config from it's rendering. Like maybe a user could choose to override the rendering to put whatever they want instead.

PR #2254, opened 2025-12-10
fix: pass dragHandleMenu prop to DragHandleButton

maqen:
# Summary

Making sure to pass `dragHandleMenu` prop to `DragHandleButton` as expected

### DIFF
--- a/packages/react/src/components/SideMenu/SideMenu.tsx
+++ b/packages/react/src/components/SideMenu/SideMenu.tsx
@@ -58,7 +58,7 @@ export const SideMenu = (props: SideMenuProps & { children?: ReactNode }) => {
       {props.children || (
         <>
           <AddBlockButton />
-          <DragHandleButton />
+          <DragHandleButton dragHandleMenu={props.dragHandleMenu} />
         </>
       )}
     </Components.SideMenu.Root>


packages/react/src/components/SideMenu/SideMenu.tsx
```diff
@@ -58,7 +58,7 @@ export const SideMenu = (props: SideMenuProps & { children?: ReactNode }) => {
       {props.children || (
         <>
           <AddBlockButton />
-          <DragHandleButton />
+          <DragHandleButton dragHandleMenu={props.dragHandleMenu} />
```
nperez0111, line 61: @matthewlipski I see you made a comment here about this which may be related?

https://github.com/TypeCellOS/BlockNote/blob/883ab943fe94b37c2c2e33e6d0fd21b2d413b8d5/packages/react/src/components/SideMenu/SideMenu.tsx#L11

matthewlipski: Ah, this comment is just about a minor typing issue. Ideally, if you pass children to the `SideMenu` component, you shouldn't be able to pass `dragHandleMenu` as it won't do anything. But the change in this PR is good, fixes a regression from #2143.

nperez0111: OK, cool thanks for confirming

Issue #2106, opened 2025-10-15
Cannot find node position with custom blocks and shadcn


kegesch:
**Describe the bug**
There is a rendering error once a a block with a custom schema is in the content and certain Shadcn Components are also rendered:
```
Uncaught Error: Cannot find node position
    at $t (internal.ts:100:11)
    at ReactNodeViewRenderer.className (ReactBlockSpec.tsx:265:31)
    at Object.react_stack_bottom_frame (react-dom-client.development.js:25904:20)
    at renderWithHooks (react-dom-client.development.js:7662:22)
    at updateFunctionComponent (react-dom-client.development.js:10166:19)
    at beginWork (react-dom-client.development.js:11778:18)
    at runWithFiberInDEV (react-dom-client.development.js:871:30)
    at performUnitOfWork (react-dom-client.development.js:17641:22)
    at workLoopSync (react-dom-client.development.js:17469:41)
    at renderRootSync (react-dom-client.development.js:17450:11)
```
I initially thought that this might have to be connected with portals, as those component per default use portals and which might confuse tiptap / prosemirror's `getPos` function to resolve the actual elements position after the component that uses portals has rendered. 
But even if I remove the portals it does not work and also only with custom blocks. 

**To Reproduce**
https://stackblitz.com/~/github.com/kegesch/blocknote-shadcn-bug?file=src/App.tsx
- Include certain shadcn components e.g. `DropDown(MenuContent)` and the `BlockNoteView` and include a custom block in the `initialContent`.

**Misc**

- Node version: v22.15.0
- Package manager: pnpm
- Browser: edge
- [ ] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

mehdigmira:
Same here ! Any workaround would be much appreciated

mehdigmira:
Ok instead of setting initial content, setting it in a `setTimeout(() => editor.replaceBlocks(editor.document, savedContent), 0) ` seems t be working

nperez0111:
Hm, this is pretty surprising to me too. For some reason commenting out the shadcn components makes the editor work again. Maybe Prosemirror cannot find the DOM position or something? I didn't see it have an additional render or anything.

@matthewlipski might know more here

kegesch:
@nperez0111 Thank you very much for the fix. It works. But it think something else is now broken, see #2134.

AmayerGogh:
Same here !  
[...]

PR #2256, opened 2025-12-10
Catch failure on getting block position.

sportnak:
# Summary

I was encountering a similar problem to this issue: https://github.com/TypeCellOS/BlockNote/issues/2106 but when I unmounted the BlockNoteView.

The problem seems to be that when the view is unmounted, the react view renderer tries to render the node one more time (not sure why). But the node still exists as does the reference, but it has been disconnected from it's parent (the document tree), causing the `getBlockFromPos` call to fail. So instead catching this failure and early exiting the `render` method prevents this from happening.

A simple way to test this is to create a schema with a block created using `createReactBlockSpec` - then add one of those blocks to the blocknote view. If you put the blocknote view in a toggleable state (i.e. in a popout panel) it will throw the `cannot find node at position` error when you unmount the blocknote view programmatically.

## Checklist

### DIFF
--- a/packages/react/src/schema/ReactBlockSpec.tsx
+++ b/packages/react/src/schema/ReactBlockSpec.tsx
@@ -262,15 +262,24 @@ export function createReactBlockSpec<
                 // only created once, so the block we get in the node view will
                 // be outdated. Therefore, we have to get the block in the
                 // `ReactNodeViewRenderer` instead.
-                const block = getBlockFromPos(
-                  props.getPos,
-                  editor,
-                  props.editor,
-                  blockConfig.type,
-                );
+                let block;
+                try {
+                  block = getBlockFromPos(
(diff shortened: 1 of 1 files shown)


sportnak:
@nperez0111 would love your feedback when you get a chance!

nperez0111:
I don't think this is a valid solution here either, unfortunately. This is treating the symptom rather than the problem.

The problem here is that the React Renderer has to be asynchronous, so it can technically mount & unmount within the time that it takes to render.

Unfortunately, this is only exposed because of React's StrictMode behavior (which will synchronously mount -> unmount -> mount, to check that renders are strict mode safe), which is unfortunately not actually likely to occur in a production scenario.

We will look into our options for this at a later time, I don't have a good solution for this at this point in time

Issue #2244, opened 2025-12-09
Undo / Redo are not working when collaboration is enabled


notABot101010:
**Describe the bug**

Hello and thank you for the work on BlockNote!

It seems that undo / redo (Ctrl + Z / Ctrl + Y) are not working when the collaboration feature with Yjs is enabled.

When enabling collaboration with a minimal YJsProvider, `Ctrl + Z` does nothing. If we comment the collaboration part of the config it's working again.

Also, when trying to implement it by hand with the following code, it's still not working:

```typescript
const undoHandler = useCallback((e: KeyboardEvent) => {
    if ((e.metaKey || e.ctrlKey) && (e.key === 'z' || e.key === 'Z')) {
      console.log('undo'); // displayed
      editor.undo(); // <- NOT WORKING
    }
  }, []);

  useWindowEvent('keydown', undoHandler);
```

**To Reproduce**
Here is a minimal reproduction in a sandbox: https://stackblitz.com/edit/github-qp3au5av?file=src%2FApp.tsx

And the code, in case the sandbox is not working:

**App.jsx**
```jsx
import '@blocknote/core/fonts/inter.css';
import { BlockNoteView } from '@blocknote/mantine';
import '@blocknote/mantine/style.css';
import { useCreateBlockNote } from '@blocknote/react';
import * as Y from 'yjs';

export default function App() {
  const ydoc = new Y.Doc();
  const yLoggingProvider = new YProvider(ydoc);
  const editor = useCreateBlockNote({
    collaboration: {
      provider: yLoggingProvider,
      fragment: ydoc.getXmlFragment('blocks'),
      user: {
        name: 'My Username',
        color: '#ff0000',
      },
      // showCursorLabels: 'always',
    },
  });

  // Renders the editor instance using a React component.
  return <BlockNoteView editor={editor} />;
}

export class YProvider {
  private doc: Y.Doc;
  private name: string;

  constructor(doc: Y.Doc, name = 'YProvider') {
    this.doc = doc;
    this.name = name;

    this.doc.on('update', this.onLocalUpdate);

    console.log(`[${this.name}] provider started`);
  }

  private onLocalUpdate = (update: Uint8Array, origin: any) => {
    Y.applyUpdate(this.doc, update);
  };

  /**
   * Apply a received update from an external source.
   * This bypasses the debounce and applies immediately.
   */
  public receiveUpdate(update: Uint8Array): void {
[...]

PR #2260, opened 2025-12-11
fix: always include the cursor extension #2244

nperez0111:
# Summary

Since extensions can depend on one another, I think that we need for this extension to always be present if in collab mode, and just remove the prosemirror plugin if it should not actually apply to the current editor.

## Checklist

## Additional Notes

#2244

### DIFF
--- a/packages/core/src/editor/managers/ExtensionManager/extensions.ts
+++ b/packages/core/src/editor/managers/ExtensionManager/extensions.ts
@@ -191,11 +191,9 @@ export function getDefaultExtensions(
 
   if (options.collaboration) {
     extensions.push(ForkYDocExtension(options.collaboration));
-    if (options.collaboration.provider?.awareness) {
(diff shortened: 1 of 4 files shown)


YousefED:
@nperez0111 I'm not exactly sure what you mean with "Since extensions can depend on one another" in this context.

My thinking; if the Cursor extension doesn't do anything (i.e.: all it's functionality is disabled), it makes more sense that it's not added at all? But from your description I take this conflicts with something else?

nperez0111:
@YousefED, The `runsBefore` property creates a DAG between the strings. But, if you do not include the cursor extension for example, it cannot know what you meant for it to run before, it defaults to "default" but this means the ordering changes depending on whether the cursor extension is present or not.

This affects the ordering of prosemirror plugins. Really, this would only be an issue for things like the collaboration extensions which have some inter-dependency between them to order the plugins (and sort of should just be a single extension like I made now with the Collaboration extension).

PR #2285, opened 2025-12-12
fix: Make quote blocks isolating

### LINKED ISSUE #2236: Copy paste in Quote block
## Bug Report

**Problematic behavior**
I wanted to create a quote block from a text I'm copy/pasting. 
I tried 2 ways to do it before I found the third way that works...
This caveat applies to all types of blocks.

**Expected behavior/code**
I feel the 2 stories detailed below should work, especially the first one. 

**Steps to Reproduce**

First story:
1. write a paragraph in any text editor on your computer. Copy the text to your pastbin.
2. go to Docs and create a new document
3. Do "/" and choose the "Quote" block You now have your cursor at the beginning of a quote block
4. Paste the content of your pastebin
5. The quote block is gone

Second story:
1. go to Docs and create a new document
2. write a paragraph 
3. Bring your cursor to the beginning of the block which now contains your paragraph. 
4. Do "/" and choose the "Quote" block
5. A quote block is created below your paragrah but your paragraph is still a standard block

The third story that already works:
1. go to Docs and create a new document
2. write a paragraph 
3. select the paragraph
4. the format toolbar appears, choose "quote" in the select box with all formats

**Misc**

- Node version:
- Package manager:
- Browser:

matthewlipski:
# Summary

This PR makes quote blocks isolating. That means when pasting content into an empty quote block, it won't overwrite the block type to that of the clipboard content.

Closes #2236 

## Rationale

There is a scenario where the user wants to paste some text for a quote, so they create a new quote block using the slash menu. They then paste the quoted text into the newly created block, and the type changes to something else. This is not great for UX.

## Changes

- Added `isolating` flag to quote block.

## Impact

N/A

## Testing

We already have tests for pasting in custom blocks which are isolating.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/core/src/blocks/Quote/block.ts
+++ b/packages/core/src/blocks/Quote/block.ts
@@ -23,9 +23,6 @@ export const createQuoteBlockConfig = createBlockConfig(
 export const createQuoteBlockSpec = createBlockSpec(
   createQuoteBlockConfig,
   {
-    meta: {
-      isolating: false,
-    },
     parse(element) {
       if (element.tagName === "BLOCKQUOTE") {
         const { backgroundColor, textColor } = parseDefaultProps(element);
--- a/tests/src/unit/core/clipboard/paste/__snapshots__/text/html/pasteEmptyQuote.json
+++ b/tests/src/unit/core/clipboard/paste/__snapshots__/text/html/pasteEmptyQuote.json
@@ -1,0 +1,18 @@
+[
+  {
+    "children": [],
+    "content": [
+      {
+        "styles": {},
+        "text": "Paragraph",
+        "type": "text",
+      },
+    ],
+    "id": "1",
+    "props": {
+      "backgroundColor": "default",
+      "textColor": "default",
+    },
+    "type": "quote",
+  },
+]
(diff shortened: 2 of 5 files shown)


matthewlipski:
@nperez0111 maybe we want to make all non-paragraph blocks isolating? This seems to be closer to what Notion does.

nperez0111:
@matthewlipski  is isolating what we are looking for here or [defining](https://prosemirror.net/docs/ref/#model.NodeSpec.definingAsContext)?

matthewlipski:
I thought defining makes more sense here too, but blocks are already defining by default. I found that isolating also needs to be `true` for the pasting fix.

YousefED (changes requested):
can we add a test to `pasteTestInstances.ts` or similar?

YousefED:
> @nperez0111 maybe we want to make all non-paragraph blocks isolating? This seems to be closer to what Notion does.

was there any conclusion to this discussion?

nperez0111:
Let's put this on hold for now, what we don't completely understand at this point is why paste operations care about `isolating` when the prosemirror docs says that only `defining` should be relevant. 

We looked into this at one point, but forgot what the conclusions were. For now, this isn't important to resolve, so will just convert this back to draft & revisit again later

PR #2251, opened 2025-12-09
Fix mobile toggle placeholder overlapping issue

MDSAM05:
# Summary

This PR fixes an issue where the placeholder element overlaps the second list item on mobile devices when the first toggle list item is collapsed.

## Rationale

- On screens ≤ 600px, the .ProseMirror-trailingBreak placeholder was positioned incorrectly, causing visual overlap.
- This affected usability when editing toggle lists in mobile view.

## Changes

- Added responsive CSS to reposition the trailing placeholder element.
- Ensured placeholder displays inline and maintains proper spacing on mobile.
- Updated Block.css with a mobile-specific rule:

``` @media (max-width: 600px) {
  .bn-inline-content:has(> .ProseMirror-trailingBreak:only-child)::before {
    position: relative !important;
    top: 0 !important;
    left: 0 !important;
    margin-left: 4px !important;
    display: inline-block !important;
    line-height: inherit !important;
  }
}
```

## Impact

- No breaking changes.
- Desktop behavior is unchanged.
- Fixes overlapping visual bug for mobile users.

## Testing

- Manual tests performed:
- Created a toggle list with multiple items.
- Collapsed the first item.
- Verified no overlapping placeholder text on mobile viewport (≤ 600px).

## Screenshots/Video

 Before
image
After
image

## Checklist

- [ x] Code follows the project's coding standards.
- [ ] Unit tests covering the new feature have been added.
- [ ] All existing tests pass.
- [x ] The documentation has been updated to reflect the new feature

PR #2291, opened 2025-12-15
fix: placeholder when overflowing now wraps

### LINKED ISSUE #2160: Placeholder text overlaps the list element in toggle list on mobile
**Describe the bug**
The placeholder test overlaps the second element in the list

Image

**To Reproduce**
1. On your mobile device: Type / to open the menu

2. Select Toggle list

3. Create a list of at least two elements one after another without collapsing them

4. Collapse the first one

**Expected behavior**
There should not be any overlapping

matthewlipski:
# Summary

This PR fixes some issues with overflowing placeholders when there is not enough space to display them on one line.

## Rationale

This was causing issues on mobile especially, since viewports can get pretty small.

## Changes

- Moved placeholders to `.bn-block-content:after` elements instead of `.bn-inline-content:before`.
- Removed `position: absolute` and other styles from placeholders.
- Made `.bn-inline-content` elements no longer span container width.
- Adjusted placeholder styles.

## Impact

I was wary of making `.bn-inline-content` elements no longer span container width due to #1523 and #1553, as we've had some cursor selection issues related to that before. However, I've done some fairly extensive manual testing across browsers against the preview from #731 (PR that broke the selection referenced in #1553), and everything looks good.

## Testing

N/A

## Screenshots/Video

Before:

https://github.com/user-attachments/assets/821cd351-5150-4333-bb03-20614e22f90c

After:

https://github.com/user-attachments/assets/7fc49f00-ebe0-4a55-95c8-8277617e343e

## Checklist

## Summary by CodeRabbit

* **Style**
  * Refined placeholder styling across the editor for improved visual consistency.
  * Improved placeholder positioning and pseudo-element behavior to reduce overlap and improve cursor offset.
  * Removed legacy inline placeholder rules for a cleaner layout and more reliable placeholder display.
  * Small visual tweaks to ensure placeholders render predictably in block content.


packages/core/src/editor/Block.css
```diff
@@ -542,15 +542,11 @@ NESTED BLOCKS
 }
 
 /* PLACEHOLDERS*/
-.bn-inline-content:has(> .ProseMirror-trailingBreak:only-child):before {
-  /*float: left; */
+/* TODO: should this be here? */
```
nperez0111: I think we can remove the TODO right?

nperez0111 (approved):
Looks about right to me

YousefED:
In my experience, changes to these CSS rules can have unintended consequences and should be tested well across browsers (and probably we need to increase our e2e coverage here).

E.g., when I try this PR and this scenario on Chrome:

image

I can't navigate to the empty list element with arrow up / down anymore

YousefED (changes requested):
see comment

matthewlipski:
I changed the approach - the hacky workaround of hiding the trailing break element has been removed and the placeholder is now rendered in the parent `.bn-block-content` element instead. This should avoid the issues with mouse and keyboard selection and still fixes the same issue as #2251.

packages/core/src/editor/Block.css
```diff
@@ -40,10 +40,6 @@ BASIC STYLES
   overflow: visible;
 }
 
-.bn-inline-content {
-  width: 100%;
-}
```
nperez0111, line 51: Double checked this, the previous issue #1523 seems to still be corrected with this in-place

packages/core/src/editor/Block.css
```diff
@@ -542,15 +538,13 @@ NESTED BLOCKS
...
-.bn-inline-content:has(> .ProseMirror-trailingBreak:only-child):before {
-  /*float: left; */
-  pointer-events: none;
-  height: 0;
-  /* width: 0; */
-  position: absolute;
+.bn-block-content:has(.ProseMirror-trailingBreak:only-child):after {
   font-style: italic;
+  /* Removes text cursor offset. */
+  margin-inline: -2px;
```
nperez0111, line 557: This is to be able to show the text cursor underneath the placeholder.

There may be a way around this, by adding data attributes to make the width of the flex layout different based on whether the placeholder is showing or not.

But, doesn't seem to cause issues, may revisit if it does something weird

nperez0111 (approved):
Seems to look good from my testing, would prefer to have @YousefED look at this though

GuiLeme:
I just found this PR, and it seems to fix the issue on our end as well. Which version will this be included in? Do you have an estimated release timeline?

PR #2304, opened 2025-12-16
feat(ai): expand selections to contain words

YousefED:
# Summary

This adds logic to expand the selection sent to LLMs to span entire words (and adjacent punctuation)

## Rationale

When selecting half-way through a word, AI would often suggest characters that were already there in order to "fix the word". See:

image

## Changes

- Add `expandPMRangeToWords` utility and parameter to BlockNoteEditor methods
- Call this function before sending a document state to LLM
- Add / update related unit tests and snapshots

## Impact

-

## Testing

Manual testing + unit test has been added

## Checklist


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -934,8 +933,8 @@ export class BlockNoteEditor<
    * If the selection starts / ends halfway through a block, the returned block will be
    * only the part of the block that is included in the selection.
    */
-  public getSelectionCutBlocks() {
-    return this._selectionManager.getSelectionCutBlocks();
+  public getSelectionCutBlocks(expandToWords = false) {
```
nperez0111, line 936: I already don't like this API. But, it is even more telling when you need to make a change to core to get access to an API like this.

I have a partial rewrite of this with the locations API refactor, so I'll let this through for now

YousefED: Agree, this entire method (`getSelectionCutBlocks`) was specifically added for AI, would be great if we can deprecate it at some point when we have the right primitives

PR #2307, opened 2025-12-18
Draft: Docs/governance

woutervroege:
# Summary

## Checklist

### DIFF
--- a/.github/RFC_TEMPLATE.md
+++ b/.github/RFC_TEMPLATE.md
@@ -1,0 +1,109 @@
+# RFC Template
+
+- **Status**: Idea | Draft | Accepted | Rejected | Revision Requested
+- **Author(s)**: <name / handle>
+- **Champion**:
+- **Created**:
+- **Last Updated**:
+- **Related**:
+
+---
+
+## 1) Summary (2–5 sentences)
+
+What is being proposed and why? Keep this understandable to non-specialists.
+
+---
+
+## 2) Problem & Motivation
+
+- What problem are we solving?
+- Who is affected?
+
+---
+
+## 3) Goals & Non-Goals
+
+### Goals
+
+What outcomes should this RFC achieve?
+
+- …
+- …
+
+### Non-Goals
+
+What is explicitly out of scope for this RFC?
+
+- …
+- …
+
+---
+
+## 4) Proposed Approach (High-level)
+
+Describe the proposed approach at a level suitable for review:
+
+- Key idea(s)
+- Expected behavior changes (if any)
+- Any public API changes (if any)
+
+---
+
+## 5) Trade-offs & Alternatives
+
+### Trade-offs
+
+What are the downsides or costs of the proposed approach?
+
+### Alternatives
+
+What alternative approaches were considered, and why were they not chosen?
+
+---
+
+## 6) Impact
+
+### Compatibility
+
+Are there breaking changes? Is migration required?
+
+### Performance
+
+Any expected performance impact?
+
+### Security / Privacy
+
+Any security or privacy implications?
+
+### Maintenance
+
+Expected long-term maintenance cost or complexity.
+
+---
+
+## 7) Open Questions
+
+- …
+- …
+
+---
+
+## (Optional) Implementation Notes
+
+Only include if it helps reviewers:
+
+- rollout/migration sketch
+- testing strategy
+- rough milestones
+
+---
+
+## Decision (TSC)
+
+Filled in by the TSC when decided.
+
+- **Decision**: Accepted | Rejected | Revision Requested
+- **Rationale**:
+- **Date**:
+- **Recorded by**:
--- a/BUGS.md
+++ b/BUGS.md
@@ -18,7 +18,7 @@ A bug is behavior that:
 - Regresses previously working functionality
 - Causes crashes, data loss, or security issues
 
-Requests for new features or behavior changes are **not bugs** and should be raised as ideas or proposals instead.
+Requests for new features or behavior changes are NOT bugs and should be raised as ideas or proposals instead.
 
 ---
 
@@ -26,7 +26,7 @@ Requests for new features or behavior changes are **not bugs** and should be r
 
 Anyone may report a bug.
 
-Please open a **GitHub Issue** and include:
+Please open a GitHub Issue and select 'Bug Report'. The template provided will require:
 
 - A clear description of the problem
 - Steps to reproduce (if possible)
@@ -53,7 +53,7 @@ Bug triage is the process of classifying and prioritizing reported bugs.
 
 Maintainers classify bugs based on technical severity.
 
-Severity reflects **technical correctness and safety impact**, not business priority or operational uptime.
+Severity reflects technical correctness and safety impact, not business priority or operational uptime.
 
 Severity levels:
 
@@ -105,7 +105,7 @@ High-severity bugs may be addressed immediately, regardless of roadmap prioritie
 
 ## Prioritization
 
-While maintainers determine technical severity, **Product** may:
+While maintainers determine technical severity, Product may:
 
 - Help prioritize bugs relative to other work
 - Balance bug fixing with feature development
@@ -125,7 +125,7 @@ If fixing a bug:
 
 Then the fix may require an RFC.
 
-See `RFC_PROCESS.md` and **How Work Flows** for guidance.
+See [RFC_PROCESS.md](/RFC_PROCESS.md) and [How Work Flows](/HOW_WORK_FLOWS.md) for guidance.
 
 ---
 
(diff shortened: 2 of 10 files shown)


MEMBERS.md
nperez0111, line 1: There is a thing called CODEOWNERS which may be relevant to this @woutervroege

nperez0111 (approved):
Very well-thought out @woutervroege, I really like this structure. This is a good point to aim towards, I'm interested to hear your thoughts on the specific implementation of this

MEMBERS.md
woutervroege: @nperez0111 Thanks! Where can I find this file? Doesn't seem to be in the repo.

nperez0111: Oh we don't have one, it is just a standard: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

.github/RFC_TEMPLATE.md
```diff
@@ -0,0 +1,109 @@
+# RFC Template
+
+- **Status**: Idea | Draft | Accepted | Rejected | Revision Requested
+- **Author(s)**: <name / handle>
+- **Champion**:
```
YousefED: what's champion?

.github/RFC_TEMPLATE.md
```diff
@@ -0,0 +1,109 @@
...
+- **Created**:
```
YousefED, line 6: created and last updated can be viewed in git, maybe remove this (or at least "last updated")

BUGS.md
```diff
@@ -0,0 +1,151 @@
...
+
+---
+
+## Bug Triage
+
+Bug triage is the process of classifying and prioritizing reported bugs.
+
+### Responsibility
+
+- **Maintainers** are responsible for technical triage and classification
```
YousefED, line 46: I think this could also be the other way around where Product leads technical triage and classification, and maintainers can help with more technical issues and provide input about technical impact etc.

YousefED: (after reading on); For highly technical issues your approach makes sense (maintainers lead classification in p0-p3).

However, I think what we're missing:
- if issues are easy to reproduce and understand, prioritization doesn't require technical input 
- many GH issues are feature-requests
- filed issues are of mixed quality, some might need more input than initially given
- ...

etc. Point is; there's probably quite some work around managing issue lifecycle that's not deeply technical per se

CONTRIBUTING.md
```diff
@@ -1,6 +1,85 @@
...
+
+For larger or higher-impact changes:
+
+- Confirm whether an RFC is required
+- Ensure the change aligns with an accepted RFC, if applicable
+- Coordinate early with maintainers
+
+Pull requests that introduce significant changes without prior discussion may be redirected or closed.
+
+---
```
YousefED, line 80: I think we should separate the technical stuff below to separate guides

GOVERNANCE.md
```diff
@@ -0,0 +1,147 @@
...
+- Sanity-checking alignment with the project mission to prevent long-term drift
+- Acting as stewards of the project’s technical integrity and sustainability
+
+**Privileges**:
+
+- Ability to curate and prioritize problem statements
+- Ability to act as Product Owner (PO) for execution within accepted RFC scope
+- Ability to establish and facilitate regular operational processes (such as bug triage) to support maintainers and contributors
+
+Product management does not make technical decisions, does not override maintainer or TSC authority, and does not bypass the RFC or review process. Product management may facilitate processes such as bug triage, but technical classification and severity decisions remain the responsibility of maintainers. Individuals may hold multiple roles; decision authority depends on the role being exercised.
```
YousefED, line 38: You're introducing the term "Product management" under the section about TSC.

I think that if these are purely clarifications about PM vs. TSC, I think they should be more succint. If they're here to describe the role of PM, it should move to the section below.

GOVERNANCE.md
```diff
@@ -0,0 +1,147 @@
+# Governance
+
+## Overview
+
+BlockNote is an open-source project stewarded by the founding team with the goal of growing into a contributor-led ecosystem over time. Governance is designed to balance long-term technical integrity, transparency, and sustainability while remaining lightweight at the project’s current scale.
```
YousefED, line 5: I'm not sure "growing into a contributor-led ecosystem" is an explicit goal at this moment (or we should further clarify it). Point in case; if there's a profitable organization (like OpenBlocks) supporting it, is it contributor led?

GOVERNANCE.md
```diff
@@ -0,0 +1,147 @@
...
+
+- Aggregating and synthesizing input from users, customers, and contributors
+- Framing problems and maintaining a transparent, indicative roadmap
+- Facilitating the RFC process without exercising decision authority
+- Preparing and facilitating operational processes such as regular bug triage
+
+**Privileges**:
+
+- Ability to curate and prioritize problem statements
+- Ability to act as Product Owner (PO) for execution within accepted RFC scope
```
YousefED, line 102: not everyone might have the same definition of PO. I'm not sure we should clarify a role by using the term of another role so to say. Better to explain what the specific priviliges are that you consider "PO"

GOVERNANCE.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Advisory Input
+
+Enterprise and public-sector users may be invited to provide structured, non-binding feedback through advisory or consultation mechanisms. Advisory input does not grant decision-making authority and does not override governance processes.
+
+---
+
+## Code of Conduct
+
+This project follows a Code of Conduct to ensure a welcoming and respectful community. All participants are expected to uphold these standards.
```
YousefED, line 141: add link

HOW_WORK_FLOWS.md
```diff
@@ -0,0 +1,186 @@
...
+             ↓
+      Product translates into delivery work
+             ↓
+      Engineering implements
+             ↓
+      Maintainers review & merge
+
+```
+
+This pipeline is risk-based: the higher the impact or precedent, the more explicit the decision-making.
```
YousefED, line 32: What does this mean exactly? Does it mean that for smaller things, we don't "need" to follow the above pipeline? If so, good to make explicit

HOW_WORK_FLOWS.md
```diff
@@ -0,0 +1,186 @@
...
+
+---
+
+## Step 1: Idea or Need
+
+Ideas and needs can originate from many places:
+
+- Community contributors
+- Enterprise or public-sector users
```
YousefED, line 41: Maybe better to name this 
- "Sponsored clients"
- or more extensive "Sponsored clients (e.g.: Startups, Enterprise or Public-sector)"

HOW_WORK_FLOWS.md
YousefED: from reading further, I think my assumption is correct (but maybe still good to revisit the phrasing here)

MISSION.md
```diff
@@ -0,0 +1,27 @@
+# Mission
```
YousefED, line 1: will review this doc later

RFC_PROCESS.md
```diff
@@ -0,0 +1,91 @@
...
+- Product
+- Members of the TSC
+
+---
+
+## **RFC Champion**
+
+An RFC is expected to have a **champion** once it is ready for formal consideration.
+
+The champion may be the original author or another volunteer. Their responsibility is to:
```
YousefED, line 39: is it needed to have both "author" and "champion"? seems overkill. What about a single "Owner" role?

TEAM_OPERATIONS.md
```diff
@@ -0,0 +1,214 @@
+# Team Operations
```
YousefED, line 1: I like the clarity / actionability of this doc!

YousefED (commented):
This is great, I think it aligns a lot with how we want to work and more importantly, gives transparency to the community. Left some inline comments, and general thoughts here:

# Size

It's quite the collection of documents. While I understand where they all come from, I have two concerns:

- Maintainability: More docs -> more difficult to maintain.
- Usage / inclusiveness: I'm afraid that the more governance docs we write, less will be read / used. Related; it might also hold people who are not as proficient in English back in contributing / engaging with the project. We could also mitigate this somewhat by keeping the most actionable items in the root of the project, while keeping background processes in a separate dir

! I also feel that at this moment there's quite some duplicate info (I think especially the flows / rfc / roles are described in multiple places)

# Rollout strategy

Our current working process is somewhat unstructured. I think we should consider the trade-off between first adopting these processes (gradually?) internally, and then publishing this, instead of the other way around.

Risks I see with the latter:
- We largely fail to follow the processes described, turning the docs into "fake transparency" which is worse than having no transparency at all
- Or, the other way around; we relentlessly try to follow the processes but find it difficult to admit they don't work for us (yet)

# Company

I think we're not addressing the elephant in the room that people might be curious about; how is the project funded, what's the relationship between the business and open source project, etc. I think this is the no. 1 thing people are curious about that gives them trust in the project or not, and also to judge the "value" / trustworthiness of the other documents. We can also explain a bit more about SLAs for sponsors (prioritized bug reports), etc.


Hope this helps! Curious what you think!

Issue #2313, opened 2025-12-23
Link Menu cannot be closed


jbpenrath:
**Describe the bug**
The link menu cannot be closed. The single way to close the menu is to close the whole toolbar that is not possible for application which are using a static toolbar.

**To Reproduce**
Go to this example : https://www.blocknotejs.org/examples/ui-components/static-formatting-toolbar

1. Click on the link button
2. Try to close the link popover (Click somewhere on the screen, press Esc key)

https://github.com/user-attachments/assets/c2b3a920-f0f5-47e3-b5ab-ecad21cfbf7c

**Misc**

- Browser: Arc 1.126.0 (Chromium 143.0.7499.147)

nperez0111:
Definitely a bug, though given the holidays, we are not likely to get around to this soon.

If you come up with a solution for it, feel free to submit a PR though!

PR #2316, opened 2025-12-23
fix: update CreateLinkButton to be able to toggle popover visibility

jbpenrath:
# Summary

Fix an issue with the `CreateLinkButton` component which cannot be closed. This problem was hidden with non static formatting toolbar as the component was unmount with the toolbar. But with static toolbar, once the link popover was opened, it was not possible to close it.

#2313 

## Rationale

Fix a bug that cause lot of feedbacks/support in the application I'm developing.

## Changes

Make the `CreateLinkButton` Popover fully controlled.

## Testing

Tested during development through the static formatting toolbar example.

## Screenshots/Video

https://github.com/user-attachments/assets/9234d8e5-9ed2-4868-bdce-bdd184105611

## Checklist

- [ ] Unit tests covering the new feature have been added. (I don't see any test suite for the react package)
- [N/A] The documentation has been updated to reflect the new feature

### DIFF
--- a/packages/core/src/blocks/Code/block.ts
+++ b/packages/core/src/blocks/Code/block.ts
@@ -130,14 +130,18 @@ export const createCodeBlockSpec = createBlockSpec(
         select.value =
           block.props.language || options.defaultLanguage || "text";
 
-        const handleLanguageChange = (event: Event) => {
-          const language = (event.target as HTMLSelectElement).value;
+        if (editor.isEditable) {
+          const handleLanguageChange = (event: Event) => {
+            const language = (event.target as HTMLSelectElement).value;
 
-          editor.updateBlock(block.id, { props: { language } });
-        };
-        select.addEventListener("change", handleLanguageChange);
-        removeSelectChangeListener = () =>
-          select.removeEventListener("change", handleLanguageChange);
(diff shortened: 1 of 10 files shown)


packages/react/src/components/FormattingToolbar/DefaultButtons/CreateLinkButton.tsx
```diff
@@ -90,6 +84,11 @@ export const CreateLinkButton = () => {
       };
     },
   });
+  useEffect(() => {
+    if (state?.url === undefined) {
+      setShowPopover(false);
+    }
+  }, [state?.url]);
```
nperez0111: why the change @matthewlipski, won't this close it popover if the url is not set yet?

matthewlipski: Nah the `state` is only responsible for what's in the editor - there's a separate `useState` hook for the popover text input and typing in that doesn't trigger this effect. This only handles the case where the editor's selection/content changes and a link is no longer selected.

I will do a quick 2nd pass on this PR either way though, just to make sure everything still works as expected with both the regular & static formatting toolbars.

PR #2317, opened 2025-12-24
Fix PageDown/PageUp scroll when slash menu is open

MDSAM05:
# Summary

Fix an issue where pressing PageDown / PageUp while the slash command typeahead menu is open causes the entire page to scroll instead of being handled by the suggestion menu.

## Rationale

Restores expected keyboard behavior for the slash command suggestion menu.

## Changes

- Added handleKeyDown handling in the SuggestionMenu ProseMirror plugin
- Prevented default behavior for PageDown / PageUp when the menu is open

## Impact

- No breaking changes
- Improves keyboard navigation UX
- Restores expected behavior for suggestion lists
- Arrow key behavior remains unchanged

## Testing

- Manually tested by opening the slash menu and pressing:
- ArrowUp / ArrowDown → selection moves correctly
- PageDown / PageUp → page no longer scrolls
- Verified menu remains open and focused

## Screenshots/Video

https://github.com/user-attachments/assets/3c29bc33-b5b7-47bb-b73a-94107752473c

## Checklist

### DIFF
--- a/docs/content/docs/features/custom-schemas/custom-blocks.mdx
+++ b/docs/content/docs/features/custom-schemas/custom-blocks.mdx
@@ -121,6 +121,7 @@ type ReactCustomBlockImplementation = {
     block: Block;
     editor: BlockNoteEditor;
     contentRef?: (node: HTMLElement | null) => void;
+    contest: { nestingLevel: number };
   }>;
   parse?: (element: HTMLElement) => PartialBlock["props"] | undefined;
   runsBefore?: string[];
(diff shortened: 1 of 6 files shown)


matthewlipski:
Thanks for this! I did end up moving the page up/down handling to the `useSuggestionMenuKeyboardNavigation` hook though, and made it so they go to the first/last suggestion menu item. I figured this makes sense given we already have arrow key handling there to also navigate the menu.

PR #2329, opened 2026-01-02
feat: Add data-nesting-level to HTML export

YousefED:
# Summary
Adds a `data-nesting-level` attribute to the flattened HTML export of blocks, preserving nesting metadata.
## Rationale
Previously, nested blocks were flattened in the HTML export to keep the output structure simple. However, this meant that indentation information was lost for non-list blocks.
With this change, blocks are still flattened in the export (maintaining simplicity), but we now include a `data-nesting-level` attribute. This metadata allows consumers to apply appropriate styling (e.g., margins) or process the nesting structure without needing a complex nested DOM structure.
## Changes
- Modified `serializeBlocksExternalHTML.ts` to track `nestingLevel` during recursive serialization.
- Added `data-nesting-level` attribute to the output HTML element if `nestingLevel > 0`.
- Updated snapshots to reflect the new attribute.
## Impact
- **HTML Export**: The exported HTML will now include `data-nesting-level` attributes on nested blocks. This is a non-breaking change but affects the output HTML string.
## Testing
- **Automated Tests**:
  - Existing snapshots for HTML export have been updated to include the new attribute.
  - Verified via `pnpm test` (snapshots updated by user).
## Checklist
- [x] Unit tests covering the new feature have been added. (Snapshot tests updated)
- [ ] The documentation has been updated to reflect the new feature (N/A - internal serialization change)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5404,9 +5404,6 @@ importers:
       '@types/react-dom':
         specifier: ^19.2.2
         version: 19.2.2(@types/react@19.2.2)
-      '@vitest/ui':
-        specifier: ^2.1.9
-        version: 2.1.9(vitest@2.1.9)
       eslint:
         specifier: ^8.57.1
         version: 8.57.1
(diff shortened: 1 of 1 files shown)


nperez0111:
I'm increasingly worried about this "simplicity" in the html export. It is more work to maintain and I'm unsure what the benefits are.

Ultimately, HTML is a machine readable format. If someone wanted to parse the full export into a "simple" export, then it really should be up to them, not part of the core package. (I hold a similar position for markdown, that it should be outside of the core & not shipped to users who will not use it)

YousefED:
> I'm increasingly worried about this "simplicity" in the html export. It is more work to maintain and I'm unsure what the benefits are.
> 
> Ultimately, HTML is a machine readable format. If someone wanted to parse the full export into a "simple" export, then it really should be up to them, not part of the core package. (I hold a similar position for markdown, that it should be outside of the core & not shipped to users who will not use it)

Agree, the solution we should work towards is to extract all renderers (whether it's exporters or to the editor) to be independent from the core editor / block definitions

PR #2331, opened 2026-01-04
fix: when you convert a block into checkListItem via inputRule, it should transfer its content into checkListItem content

tmpluto:
Solved problem:
Put the cursor at the start of line and type "[] " to convert the block into checkListItem, but content of the block doesn't transfer into the checkListItem.

https://github.com/user-attachments/assets/daa9e549-0d13-4800-bb6f-836a5e90f4b2


nperez0111:
Thanks for this @tmpluto, good sleuthing on this fix

PR #2335, opened 2026-01-06
fix: LinkToolbar Event Listener leak

### LINKED ISSUE #2327: posAtDOM error when editor view remounts (locale/editable change)
**Describe the bug**

  When the editor view is remounted (e.g., changing `collaboration` or ``locale` or `editable` props), a console error occurs if the mouse cursor is positioned over the editor area:

  Uncaught Error: [tiptap error]: The editor view is not available. Cannot access view['posAtDOM']. The editor may not be mounted yet.
      at Object.getLinkAtElement (TrailingNode-*.js)
      at HTMLDivElement.d (blocknote-react.js)

  The error is thrown from `getLinkAtElement` function which is called on mouse events without checking if the editor view is available.

  **Note:** The editor functionality works correctly - only the console error appears.

  **To Reproduce**

  1. Create a BlockNote editor
  2. Position your mouse cursor over the editor content area
  3. Change the `locale` prop (e.g., 'en' → 'ko') or toggle `editable` (true → false → true)
  4. Observe the console error

  This does NOT happen when changing the `theme` prop (which doesn't trigger view remount).

  **Misc**

  - Node version: 22.x
  - Package manager: pnpm 10.x
  - Browser: Chrome  : 143.0.7499.170
  - BlockNote version: 0.45.0 (also tested on 0.43.0, 0.44.0)
  - Works fine on 0.42.3

  This issue started occurring after v0.43.0 ("Extensions, UI redesign" #2143).

  **Suggested fix:**

  Add a view availability check before accessing `posAtDOM`:

  ```typescript
  // In getLinkAtElement or the mouse event handler
  if (!editor._tiptapEditor?.view) return null;
 ```

- [X] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

YousefED:
# Fix LinkToolbar Event Listener Leak on Remount

## Summary

This PR addresses an issue where event listeners were not being correctly removed in LinkToolbarController during editor remounts, and adds a regression test to verify the fix. 

closes #2327 

## Rationale

When the editor component is unmounted or rapidly re-mounted, the cleanup function in `useEffect` could previously fail to remove the `mouseover` event listener if the underlying `domElement` reference had changed or become undefined. This changes ensures `removeEventListener` is always called on the correct element instance that the listener was originally attached to.

## Changes

- **Fix LinkToolbarController**: captured the element in the `useEffect` scope to ensure `removeEventListener` is called on the correct element during cleanup.
- **Safety Fixes**: applied the same pattern to similar parts of the codebase to prevent consistent issues.
- **New Test**: added unit test to reproduce
- **Discussion Points**: Added comments marked in the code with follow-up questions to discuss with @nperez0111 @matthewlipski 

## Impact

Prevents runtime errors and potential memory leaks when the editor is mounted and unmounted frequently

## Testing

- tested manually
- unit test added

## Screenshots/Video

N/A

## Checklist

## Additional Notes

I've left several comments in the code raising questions about the current architecture for handling these event listeners, which we should discuss in a future refactor.

### DIFF
--- a/packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
+++ b/packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
@@ -67,11 +67,6 @@ export const LinkToolbarExtension = createExtension(({ editor }) => {
 
     getLinkAtElement(element: HTMLElement) {
       return editor.transact(() => {
-        // Q4: posAtDOM can fail if the editor view is not available
-        //     (e.g. if the editor is not mounted)
-        //     a) Unfortunately, TS doesn't give an error about this. Can we make this type safe?
-        //     b) Double check other references of editor.prosemirrorView
-
         const posAtElement = editor.prosemirrorView.posAtDOM(element, 0) + 1;
         return getMarkAtPos(posAtElement, "link");
       });
--- a/packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
+++ b/packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
@@ -100,8 +100,6 @@ export const LinkToolbarController = (props: {
 
     const domElement = editor.domElement;
 
-    // Q 1: why can domElement be available when <LinkToolbarController/> is rendered?
-    // Q 2: this useEffect will not necessarily run when editor.domElement changes
     domElement?.addEventListener("mouseover", mouseCursorCallback);
 
     return () => {
@@ -164,7 +162,7 @@ export const LinkToolbarController = (props: {
     [link?.element],
   );
 
-  // Q3: similar to Q2; are we sure the component rerenders when editor.isEditable changes?
+  // TODO: this should be a hook to be reactive
   if (!editor.isEditable) {
     return null;
   }


packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
```diff
@@ -98,15 +98,18 @@ export const LinkToolbarController = (props: {
     const destroyOnSelectionChangeHandler =
       editor.onSelectionChange(textCursorCallback);
 
-    editor.domElement?.addEventListener("mouseover", mouseCursorCallback);
+    const domElement = editor.domElement;
+
+    // Q 1: why can domElement be available when <LinkToolbarController/> is rendered?
```
nperez0111: Why would `domElement` not be available? It would seem to me that is sort of the default state of things?

packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
```diff
@@ -98,15 +98,18 @@ export const LinkToolbarController = (props: {
...
+    // Q 2: this useEffect will not necessarily run when editor.domElement changes
```
nperez0111: We need to enable to react eslint plugin, it would have caught this.

packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
```diff
@@ -161,6 +164,7 @@ export const LinkToolbarController = (props: {
     [link?.element],
   );
 
+  // Q3: similar to Q2; are we sure the component rerenders when editor.isEditable changes?
```
nperez0111: This should be derived from a `useEditorState`, rather than queried at render time

packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
```diff
@@ -67,6 +67,11 @@ export const LinkToolbarExtension = createExtension(({ editor }) => {
 
     getLinkAtElement(element: HTMLElement) {
       return editor.transact(() => {
+        // Q4: posAtDOM can fail if the editor view is not available
+        //     (e.g. if the editor is not mounted)
+        //     a) Unfortunately, TS doesn't give an error about this. Can we make this type safe?
+        //     b) Double check other references of editor.prosemirrorView
```
nperez0111: posAtDom fails, or does accessing editor.prosemirrorView?

TS could never know if posAtDom fails at runtime
TS could say the editor.prosemirrorView is undefined

anytime editor.prosemirrorView or editor.prosemirrorState is used, it should be considered a failing of our API

packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
YousefED: Whoops my comment was wrong. Question should be why it can be unavailable (i.e. `undefined`).

packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
YousefED: I don't think it's just a matter of eslint; because the component will also not automatically rerender when `domElement` changes; accessing the `domElement` property is non-reactive, so it's "dangerous" to depend on this state from the React layer?

packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
YousefED: accessing `posAtDom` fails fails (because there's no underlying view in the tiptap layer).

> anytime editor.prosemirrorView or editor.prosemirrorState is used, it should be considered a failing of our API

Agree, this seems like the root cause. Will start eng. roadmap and add this

packages/react/src/components/LinkToolbar/LinkToolbarController.tsx
nperez0111: Was this this observed, or a hypothetical?
> because the component will also not automatically rerender when domElement changes;

 My understanding is that this reference should be stable across renders.

PR #2338, opened 2026-01-07
fix: opening suggestion menu in collab mode on empty document

YousefED:
There's a bug that hitting the + button doesn't work in collaboration mode, on an empty document. I tried to address it here but didn't get to the root cause yet.

In this PR, I have:

- improved some of the naming in the suggestion plugin to make it easier for me to untangle the code
- removed some of the offsets that I think shouldn't be there when opening the suggestion menu programmatically

However, this doesn't fix the issue yet. I think there's an issue with using `trackPosition` in this specific case.

We can diagnose further by adding a breakpoint in the if-statement marked by `// Checks if the menu should be hidden.`

### DIFF
--- a/packages/core/src/extensions/SuggestionMenu/SuggestionMenu.ts
+++ b/packages/core/src/extensions/SuggestionMenu/SuggestionMenu.ts
@@ -3,12 +3,12 @@ import { EditorState, Plugin, PluginKey } from "prosemirror-state";
 import { Decoration, DecorationSet, EditorView } from "prosemirror-view";
 
 import { trackPosition } from "../../api/positionMapping.js";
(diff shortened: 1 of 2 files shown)


nperez0111:
On an empty document, y-prosemirror will choose to not initialize the Y.XmlFragment with content until it actually has content to sync (AKA different from a `state.topNodeType.createAndFill()` node).

So, what we would do to fix this is to treat position tracking on an empty Y.XmlFragment (`fragment.length === 0`) as just being normal prosemirror positions rather than trying to find positions in that empty node (which would be wrong).

This is probably a little tricky to implement since a number of things can happen between a position is tracked and it is actually used.

In the next version of y-prosemirror, it will offer a util similar to track position (probably implemented as a position mapper).

PR #2345, opened 2026-01-11
fix: set cursor inside block after input rule

jaehafe:
# Summary

Fix cursor position after creating a block via input rules (e.g., typing ``` + space for code block). Previously, the cursor would move to the next paragraph instead of inside the newly created block.

## Rationale

When users type ``` followed by a space to create a code block, they expect the cursor to be positioned inside the code block so they can immediately start typing code. However, the cursor was incorrectly placed in the paragraph below the code block, requiring users to manually click inside the code block to start typing.

## Changes

- Added TextSelection import to ExtensionManager/index.ts
- After updateBlockTr() is called in the input rule handler, explicitly set the cursor position inside the newly created block using TextSelection.create()

## Testing

  - Manual testing: Type ``` + space in the editor and verify cursor is inside the code block

## Screenshots/Video

**before**
https://github.com/user-attachments/assets/1aeb3742-fbff-428b-8616-a21096f72b47

**after**
https://github.com/user-attachments/assets/7c683737-e823-4a16-b969-b7f237f0b5ef

## Checklist

## Additional Notes

Related issue: The comment in updateBlock.ts:255-258 mentions that selection should be reset after replaceWith, but this was not implemented. This PR addresses that by handling it in the input rule handler.

### DIFF
--- a/packages/core/src/editor/managers/ExtensionManager/index.ts
+++ b/packages/core/src/editor/managers/ExtensionManager/index.ts
@@ -7,7 +7,7 @@ import {
   Extension as TiptapExtension,
 } from "@tiptap/core";
 import { keymap } from "@tiptap/pm/keymap";
-import { Plugin } from "prosemirror-state";
+import { Plugin, TextSelection } from "prosemirror-state";
 import { updateBlockTr } from "../../../api/blockManipulation/commands/updateBlock/updateBlock.js";
 import { getBlockInfoFromTransaction } from "../../../api/getBlockInfoFromPos.js";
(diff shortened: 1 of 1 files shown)


jaehafe:
Hi @nperez0111,  
I was wondering if you had a chance to look at this PR.  
Happy to address any feedback or concerns you might have.

nperez0111:
Hi @jaehafe, I haven't taken a look at this completely yet. My initial reaction is that this likely should live inside of updateBlockTr as an option to move the selection into the block that was updated. There already is some table cell selection stuff that's in there.

So all that to say it will take a bit of untangling to get this right

PR #2348, opened 2026-01-13
fix: Visual differences between live editor and rendered exported HTML

matthewlipski:
# Summary

This PR fixes some visual & DOM inconsistencies between the rendered HTML from a live editor, and the rendered static HTML from exporting blocks.

These differences come mostly from plugins/extensions that use decorations not running when doing an export, as well as a few other fixes due to the HTML being static.

## Rationale

One of the use cases of BlockNote is writing content in the editor, taking advantage of its UX and formatting, then exporting that content to HTML so it can be displayed read-only without requiring an editor instance.

However, this is only useful if the static HTML output closely matches what a user sees when editing content in BlockNote.

## Changes

- Fixed alignment issue with check list items.
- Added `transforms` to `createInternalHTMLSerializer`. These are functions which modify the HTML element created from exporting blocks. See `internalHTMLSerializer.ts` for more info.
- Added example with a live editor to its exported & rendered static HTML side-by-side.
- Added example with the blocks from the [Default Schema Showcase example](https://www.blocknotejs.org/examples/basic/default-blocks) exported to static HTML & rendered (for e2e testing).

## Impact

N/A

## Testing

An e2e test has been added which tests that the Default Schema Showcase example page is visually identical to the new Rendering HTML Converted From Blocks example page.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/examples/05-interoperability/09-blocks-to-html-static-render/src/App.tsx
+++ b/examples/05-interoperability/09-blocks-to-html-static-render/src/App.tsx
@@ -2,7 +2,7 @@ import "@blocknote/core/fonts/inter.css";
 import { BlockNoteView } from "@blocknote/mantine";
 import "@blocknote/mantine/style.css";
 import { useCreateBlockNote, usePrefersColorScheme } from "@blocknote/react";
-import { useCallback, useEffect, useRef } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
 import "./styles.css";
 
@@ -155,6 +155,8 @@ export default function App() {
     ],
   });
 
+  const [html, setHTML] = useState("");
+
   const ref = useRef<HTMLDivElement>(null);
   const systemColorScheme = usePrefersColorScheme();
   const theme =
@@ -162,9 +164,7 @@ export default function App() {
 
   // Function to update the rendered static HTML.
   const updateRenderedHTML = useCallback(async () => {
-    if (ref.current) {
-      ref.current.innerHTML = editor.blocksToFullHTML(editor.document);
-    }
+    setHTML(editor.blocksToFullHTML(editor.document));
   }, []);
   // Updates rendered static HTML with initial editor content.
   useEffect(() => {
@@ -198,7 +198,7 @@ export default function App() {
           >
             <div
               className="ProseMirror bn-editor bn-default-styles"
-              ref={ref}
+              dangerouslySetInnerHTML={{ __html: html }}
             />
           </div>
         </div>
--- a/packages/core/src/api/exporters/html/internalHTMLSerializer.ts
+++ b/packages/core/src/api/exporters/html/internalHTMLSerializer.ts
@@ -24,7 +24,7 @@ const addIndexToNumberedListItems = (element: HTMLElement) => {
         '[data-content-type="numberedListItem"]',
       );
 
-    if (prevNumberedListItem === null || prevNumberedListItem === undefined) {
+    if (!prevNumberedListItem) {
       numberedListItem.setAttribute(
         "data-index",
         numberedListItem.getAttribute("data-start") || "1",
@@ -114,7 +114,7 @@ const addTrailingBreakToEmptyInlineContent = (element: HTMLElement) => {
     ".bn-inline-content:empty",
   );
   emptyInlineContent.forEach((inlineContent) => {
-    // We actually use a `span` instead of a `br` to avoid potential false 
+    // We actually use a `span` instead of a `br` to avoid potential false
     // positives when parsing.
     const trailingBreak = document.createElement("span");
     trailingBreak.className = "ProseMirror-trailingBreak";


packages/core/src/api/exporters/html/internalHTMLSerializer.ts
```diff
@@ -26,13 +142,38 @@ export const createInternalHTMLSerializer = <
...
+  // functionality so that the rendered HTML looks identical to that of a live
+  // editor.
+  const transforms: ((element: HTMLElement) => HTMLElement)[] = [
+    addIndexToNumberedListItems,
+    makeCheckListItemsReadOnly,
+    forceToggleBlocksShow,
+    addTableMinCellWidths,
+    addTableWrappers,
+    addTrailingBreakToEmptyInlineContent,
+  ];
```
nperez0111: This is fantastic @matthewlipski I'm really happy with how this came out

examples/05-interoperability/09-blocks-to-html-static-render/src/App.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  const ref = useRef<HTMLDivElement>(null);
+  const systemColorScheme = usePrefersColorScheme();
+  const theme =
+    systemColorScheme === "no-preference" ? "light" : systemColorScheme;
+
+  // Function to update the rendered static HTML.
+  const updateRenderedHTML = useCallback(async () => {
+    if (ref.current) {
+      ref.current.innerHTML = editor.blocksToFullHTML(editor.document);
```
YousefED: the more standard way would be to use `dangerouslySetInnerHTML` on the `<div>`

examples/05-interoperability/10-static-html-render/src/App.tsx
```diff
@@ -0,0 +1,177 @@
...
+    systemColorScheme === "no-preference" ? "light" : systemColorScheme;
+
+  // Updates rendered static HTML with initial editor content.
+  useEffect(() => {
+    if (ref.current) {
+      ref.current.innerHTML = editor.blocksToFullHTML(editor.document);
+    }
+  }, []);
+
+  // Renders the editor instance using a React component.
```
YousefED: comment is not accurate

packages/core/src/editor/Block.css
```diff
@@ -240,12 +240,16 @@ NESTED BLOCKS
 }
 
 /* Checked */
+.bn-block-content[data-content-type="checkListItem"] > div:has(> input) {
```
YousefED, line 243: why did this and margin-block change?

tests/src/end-to-end/basics/basicblocks.test.ts-snapshots/basicblocks-chromium-linux.png
YousefED, line 1: any idea about why the diff shows a change in top margin?

packages/core/src/editor/Block.css
matthewlipski: Not really related to the other stuff in this PR - vertical alignment was just off for the check list item block.

tests/src/end-to-end/basics/basicblocks.test.ts-snapshots/basicblocks-chromium-linux.png
matthewlipski: Looks to me not that the margin has changed, but that the screenshot is a bit shorter due to the check list item block CSS changes (see comment above), and the screenshots are aligned at their bottom edge. If you change the compare mode to swipe, you can more clearly see that the top edge of the new screenshot is below the top edge of the old one.

nperez0111:
> Architecturally, I don't think it's not ideal that we have the `transforms` living in the serializer. This assumes the serializer has a lot of "knowledge" about the HTML structure outputted from the blocks, whereas actually this knowledge should live side-by-side of where that HTML is created in the first place (currently, where the blocks live).
> 
> However, I think this is probably a symptom of overall architectural issues with the exporters; I don't think it's a concern for this PR (though a general note about this might be nice).

Agreed here, I put @matthewlipski down this path to keep ourselves from the temptation of doing block specific "fixups" during the export. I was imagining that these transforms being a sort of "post-process" step of serializing to HTML, and eventually make that part of the API. But, that is for another time, this is nice in that it encapsulates the change to be completely after the serialization, so I'm quite happy with it as far as these things go.

examples/05-interoperability/09-blocks-to-html-static-render/src/App.tsx
nperez0111: I think this is still not fixed right @matthewlipski

examples/05-interoperability/09-blocks-to-html-static-render/src/App.tsx
```diff
@@ -0,0 +1,208 @@
...
+        props: {
+          url: "https://interactive-examples.mdn.mozilla.net/media/cc0-images/grapefruit-slice-332-332.jpg",
+          caption:
+            "From https://interactive-examples.mdn.mozilla.net/media/cc0-images/grapefruit-slice-332-332.jpg",
+        },
+      },
+      {
+        type: "video",
+        props: {
+          url: "https://interactive-examples.mdn.mozilla.net/media/cc0-videos/flower.webm",
```
nperez0111, line 101: I would bet that if we made this a local video that the vite server served out of the public folder, that we'd have less flaky tests since it wouldn't depend on the network. Same for images

packages/core/src/api/exporters/html/internalHTMLSerializer.ts
```diff
@@ -1,12 +1,131 @@
...
+    '[data-content-type="numberedListItem"]',
+  );
+  numberedListItems.forEach((numberedListItem) => {
+    const prevNumberedListItem = numberedListItem
+      .closest(".bn-block-outer")
+      ?.previousElementSibling?.querySelector(
+        '[data-content-type="numberedListItem"]',
+      );
+
+    if (prevNumberedListItem === null || prevNumberedListItem === undefined) {
```
nperez0111:
```suggestion
    if (prevNumberedListItem) {
```

examples/05-interoperability/09-blocks-to-html-static-render/src/App.tsx
matthewlipski: Apparently not, at least for visual regressions. Even with local files, Chrome's video viewer at least seems to choose to buffer the whole video or not based on how it's feeling at any given moment:

https://drive.google.com/file/d/1r_X3qwSYbCw-Y8TnrDe3sA2ie4u9S9M0/view?usp=sharing

Kind of a long video so it's on Drive

matthewlipski: R.e. images, should be enough for have a timeout until network idle. This is only really an issue with media that has buffering.

nperez0111: Drive link didn't work for me. But, my understanding is that the buffering would be quicker (and more reliable) if it was actually a local file (since it doesn't touch the network)

matthewlipski: Ah my bad, fixed the link - either way the video/audio elements should still be masked as the buffering is inconsistent in the browser video/audio viewer, even with local files.

matthewlipski: Think the masking is ok for this PR - when we get around to re-doing our e2e testing it will definitely be worth trying  e.g. short 1-second base64 videos though. This might also get around the buffering issue.

PR #2357, opened 2026-01-14
fix: `BlockNoteViewEditor` mismatched editable value

matthewlipski:
# Summary

When using `BlockNoteViewEditor` and setting `renderEditor={false}` in `BlockNoteViewComponent`, you currently have to pass the same `editable` prop to both, or it causes issues.

Because it doesn't make sense to need to pass the same `editable` prop to 2 different components in the first place, this PR removes the prop from `BlockNoteViewEditor`. Instead, the editable value is passed in through the `BlockNoteContext` to ensure it's the same between the components.

## Rationale

There is currently an issue with the comments sidebar example caused by this mismatch in editable states, as the prop is passed to `BlockNoteViewComponent` but not `BlockNoteViewEditor`.

This makes the editor not focusable while the user selected has only comment priviledges and `editable` is `false`. That, in turn, prevents the editor from being focusable, meaning the formatting toolbar can't be shown, and so the user can't add comments.

## Changes

See above

## Testing

Technically a breaking change since `BlockNoteViewEditor` no longer supports the `editable` prop? Shouldn't really matter though.

## Screenshots/Video

N/A

## Checklist

### DIFF
--- a/packages/react/src/editor/BlockNoteView.tsx
+++ b/packages/react/src/editor/BlockNoteView.tsx
@@ -268,6 +268,10 @@ export const BlockNoteViewEditor = (props: { children?: ReactNode }) => {
   const mount = useCallback(
     (element: HTMLElement | null) => {
       // Set editable state of the actual editor.
+      // We need to re-mount the editor when changing `isEditable` as TipTap 
+      // removes the `tabIndex="0"` attribute we set (see 
+      // `BlockNoteEditor.ts`). Ideally though, this logic would exist in a 
+      // separate hook.
       editor.isEditable = ctx.editorProps.editable !== false;
       // Since we are not using TipTap's React Components, we need to set up the contentComponent it expects
       // This is a simple replacement for the state management that Tiptap does internally


packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -283,10 +273,10 @@ export const BlockNoteViewEditor = (props: {
   const mount = useCallback(
     (element: HTMLElement | null) => {
       if (
-        props.editable !== undefined &&
-        props.editable !== editor.isEditable
+        ctx.editorProps.editable !== undefined &&
+        ctx.editorProps.editable !== editor.isEditable
       ) {
-        editor.isEditable = props.editable;
+        editor.isEditable = ctx.editorProps.editable;
```
YousefED: I'm wondering if this is needed, don't we already do this on line 145? Can we find edge-cases why we also need to set this here on mount / why we added it originally?

matthewlipski: You're correct, it's redundant to have this logic in both components. Seems like the reason this was done is that the effect on line 145 would only fire after the context value was updated, so `editor.isEditable` may not have the correct value when `mount` is run.

I ended up removing the effect on line 145 and simplifying the code in `mount` instead, to not deal with this weird hook priority thing.

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -282,12 +267,8 @@ export const BlockNoteViewEditor = (props: {
...
   const mount = useCallback(
     (element: HTMLElement | null) => {
-      if (
-        props.editable !== undefined &&
-        props.editable !== editor.isEditable
-      ) {
-        editor.isEditable = props.editable;
-      }
+      // Set editable state of the actual editor.
+      editor.isEditable = ctx.editorProps.editable !== false;
```
nperez0111, line 275: Is it intentional that editable is not "reactive", i.e. you change that context value & it won't switch the editor between editable & not editable modes?

We should check this behavior in an example. Since I think people may have things like:

```ts
function Component(){
const [isEditable, setEditable] = useState(true);

return <><button onClick={()=>setEditable(a=>!a)}>Toggle</button><Editor editable={isEditable} /></>
}
```
And, expect it to react to the button.

packages/react/src/editor/BlockNoteView.tsx
nperez0111: Ah realized I commented on the same sort of behavior https://github.com/TypeCellOS/BlockNote/pull/2357#discussion_r2697384599 . I think we should be using only the effect & not the mount handler. If we need to have this render before layout (as some things will look different in editable vs. not mode), then use a `useLayoutEffect` since it will be done before paint. Though, I could see this triggering a re-render of the editor blocks that are defined as react nodes, so ideally it'd be set on editor creation & the layout effect is purely for reactivity

packages/react/src/editor/BlockNoteView.tsx
matthewlipski: Modified the existing example where this issue was coming up - all good, editor is re-mounted with the toggled editable state and everything works as expected. `ctx.editorProps.editable` is in the dependency array of `mount`, so I think this makes sense.

packages/react/src/editor/BlockNoteView.tsx
matthewlipski: Hmm, even with `useLayoutEffect`, `mount` is still called before the effect, so things still don't work. Also, I don't think we could set it on editor creation since `editable` can only be set on the view, not passed to the editor itself. Maybe worth discussing on call.

PR #2361, opened 2026-01-15
fix(link-toolbar): prevent Enter from submitting during IME composition

hanios123:
# Summary

Prevent the link toolbar form from submitting when the user presses Enter during IME (Input Method Editor) composition.

## Rationale

When users type in languages that require IME input methods (e.g., Japanese, Chinese, Korean), they press Enter to confirm character conversions (e.g., converting hiragana to kanji). Without this fix, pressing Enter during composition would prematurely submit the link form instead of completing the character input, breaking the editing experience for CJK users.

## Changes

- Added a check for event.nativeEvent.isComposing in the handleEnter callback within EditLinkMenuItems.tsx
- The handler now returns early if the keyboard event is part of an ongoing IME composition session

## Impact

- No breaking changes – This only adds a guard condition
- Improves accessibility and usability for users typing in CJK languages
- Existing functionality for non-IME users remains unchanged

## Testing

- Manual testing: Verified with Japanese IME (macOS) that Enter key during kanji conversion no longer submits the form, and Enter after composition completes works as expected
- Tested that normal (non-IME) Enter key behavior still submits the link correctly

## Screenshots/Video

https://github.com/user-attachments/assets/2cd5ebe6-704c-4083-9c8d-211bb2ca914e

## Checklist

### DIFF
--- a/packages/core/src/extensions/NodeSelectionKeyboard/NodeSelectionKeyboard.ts
+++ b/packages/core/src/extensions/NodeSelectionKeyboard/NodeSelectionKeyboard.ts
@@ -40,6 +40,7 @@ export const NodeSelectionKeyboardExtension = createExtension(
                 // Checks if key press is Enter
                 if (
                   event.key === "Enter" &&
+                  !event.isComposing &&
                   !event.shiftKey &&
                   !event.altKey &&
                   !event.ctrlKey &&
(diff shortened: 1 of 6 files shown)


nperez0111:
This is great @hanios123 I would merge this, but I'd probably go around to check whether we have this issue in other places or not first. And I don't have the time to dive into that at this time.

nperez0111:
Thanks @matthewlipski this is exactly what I'd have expected

PR #2397, opened 2026-01-30
feat: update suggestion menu component

MDSAM05:
# Summary

Fixes unexpected page scrolling when navigating the suggestion menu with keyboard arrow keys, especially when wrapping from the last item to the first (and vice versa).

## Rationale

Arrow key navigation caused the page to scroll when wrapping between suggestions, leading to a poor editor experience.
This fix enables smooth circular navigation by keeping focus stable and preventing unwanted scrolling.

## Changes

- Replaced scrollIntoView(true/false) with scrollIntoView({ block: "nearest" }) to prevent page jumps.
- Scrolled the active suggestion into view only when it overflows.
- Enabled smooth circular navigation without focus changes.

## Impact

- Improves keyboard navigation UX by eliminating unexpected page scrolling.
- Keeps suggestion navigation smooth and predictable inside the editor.
- No breaking changes to existing functionality. 

## Testing

- Manually verified arrow key navigation without page scroll.
- Confirmed Enter key selection works as expected.

## Screenshots/Video

https://github.com/user-attachments/assets/b8ce24f0-eaf1-4869-9910-fe9ec210cb84

## Checklist

### DIFF
--- a/packages/shadcn/src/suggestionMenu/SuggestionMenuItem.tsx
+++ b/packages/shadcn/src/suggestionMenu/SuggestionMenuItem.tsx
@@ -26,10 +26,9 @@ export const SuggestionMenuItem = forwardRef<
       itemRef.current,
       document.querySelector(".bn-suggestion-menu, #ai-suggestion-menu")!, // TODO
     );
-    if (overflow === "top") {
-      itemRef.current.scrollIntoView(true);
-    } else if (overflow === "bottom") {
-      itemRef.current.scrollIntoView(false);
+
+    if (overflow !== "none") {
+      itemRef.current.scrollIntoView({ block: "nearest" });
     }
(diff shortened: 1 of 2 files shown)


matthewlipski:
I'm wondering if we're going about the UX of scrolling while the slash menu is open the right way. Notion initially scrolls to the selection/slash menu when opened, but then blocks scrolling the document afterwards.

@nperez0111 do you think it makes any sense to replicate that?

If we do want to deviate from Notion's UX here and allow users to scroll the document while the slash menu is open, we should also make sure that changing the selected item always scrolls it into view. Currently, this happens inconsistently.

For now, I'll merge this PR as it fixes a bug with the scrolling. Depending on what UX solution we decide on, I'll create separate issues later.

PR #2416, opened 2026-02-04
fix: correct return type of getDefaultReactSlashMenuItems to include key

ysds:
# Summary

Fix incorrect return type of `getDefaultReactSlashMenuItems()` to properly include the `key` property.

## Rationale

Items returned by `getDefaultReactSlashMenuItems()` actually contain a `key` property (inherited from `getDefaultSlashMenuItems()`), but the return type `DefaultReactSuggestionItem[]` omits `key`. This causes type errors when users try to access `item.key`:

```typescript
getDefaultReactSlashMenuItems(editor).filter((item) => {
  console.log(item.key); // Error: Property 'key' does not exist
});
```

## Changes

- Updated return type of `getDefaultReactSlashMenuItems()` to include `key` property
- Simplified return type of `getPageBreakReactSlashMenuItems()` by removing unnecessary `Omit`

## Impact

- Users can now access `item.key` without type errors
- No breaking changes to runtime behavior

## Testing

- Type checking passes for all packages

## Screenshots/Video

N/A (type-only change)

## Checklist

### DIFF
--- a/packages/react/src/components/SuggestionMenu/getDefaultReactSlashMenuItems.tsx
+++ b/packages/react/src/components/SuggestionMenu/getDefaultReactSlashMenuItems.tsx
@@ -1,6 +1,7 @@
 import {
   BlockNoteEditor,
   BlockSchema,
+  Dictionary,
   InlineContentSchema,
   StyleSchema,
 } from "@blocknote/core";
@@ -61,7 +62,11 @@ export function getDefaultReactSlashMenuItems<
   BSchema extends BlockSchema,
   I extends InlineContentSchema,
   S extends StyleSchema,
->(editor: BlockNoteEditor<BSchema, I, S>): DefaultReactSuggestionItem[] {
+>(
+  editor: BlockNoteEditor<BSchema, I, S>,
+): (DefaultReactSuggestionItem & {
+  key: keyof Omit<Dictionary["slash_menu"], "page_break">;
+})[] {
   return getDefaultSlashMenuItems(editor).map((item) => {
     const Icon = icons[item.key];
     return {
--- a/packages/react/src/blocks/PageBreak/getPageBreakReactSlashMenuItems.tsx
+++ b/packages/react/src/blocks/PageBreak/getPageBreakReactSlashMenuItems.tsx
@@ -18,7 +18,7 @@ export function getPageBreakReactSlashMenuItems<
   S extends StyleSchema,
 >(
   editor: BlockNoteEditor<BSchema, I, S>,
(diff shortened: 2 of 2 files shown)


packages/react/src/components/SuggestionMenu/getDefaultReactSlashMenuItems.tsx
```diff
@@ -61,7 +62,11 @@ export function getDefaultReactSlashMenuItems<
   BSchema extends BlockSchema,
   I extends InlineContentSchema,
   S extends StyleSchema,
->(editor: BlockNoteEditor<BSchema, I, S>): DefaultReactSuggestionItem[] {
+>(
+  editor: BlockNoteEditor<BSchema, I, S>,
+): (DefaultReactSuggestionItem & {
+  key: keyof Omit<Dictionary["slash_menu"], "page_break">;
```
nperez0111, line 68: I don't think this is correct right? Aren't you marking all of the items as having a page_break key, when that is not actually the case. Shouldn't you change `DefaultReactSuggestionItem` instead?

nperez0111 (changes requested):

packages/react/src/components/SuggestionMenu/getDefaultReactSlashMenuItems.tsx
ysds: Thanks for taking a look. I’ll check it out.

ysds: To clarify, `Omit<Dictionary["slash_menu"], "page_break">` here represents "a union of keys excluding `page_break`", not "marking items as having a page_break key". This pattern follows the same approach used in the [DefaultSuggestionItem.ts](https://github.com/TypeCellOS/BlockNote/blob/8a7f8cb81d354551254599fa40fedb839118b600/packages/core/src/extensions/SuggestionMenu/DefaultSuggestionItem.ts)


I believe the current implementation is correct, but I do agree that modifying `DefaultReactSuggestionItem` could be a better approach.

However, changing `DefaultReactSuggestionItem` requires a deeper understanding of BlockNote's design decisions. The `key` property was removed from `DefaultReactSuggestionItem` in #652, but the PR doesn't explain why. I can only speculate on the reasons:
- Potential conflicts with React's `key` prop
- Custom suggestion items may not need a `key`

Making changes to `DefaultReactSuggestionItem` feels beyond my current understanding of the codebase. This PR at least addresses the immediate type issue.

PR #2447, opened 2026-02-11
feat(i18n): add Persian (fa) localization support

Mrahmani71:
# Summary

Added Persian (fa) localization support to BlockNote


nperez0111:
Thanks for this PR @Mrahmani71 very much appreciate a new locale!

PR #2451, opened 2026-02-13
fix: Backspace/enter behaviour in empty block with children

### LINKED ISSUE #1593: Backspace on empty block before indented block delete both blocks
**Describe the bug**
Backspace on empty block before indented block delete both blocks

**To Reproduce**
1. Press enter multiple times.
2. Indent one block and write something.
3. Go to the block before the indented one and write any char.
4. Press backspace two times two delete the block. It will delete also the indented one.

https://github.com/user-attachments/assets/0736ec6f-9b33-4382-b3dc-89f0f9a7a2ed

**Misc**
- Browser: Chrome

matthewlipski:
# Summary

There are currently 2 issues with backspace/enter handling in empty blocks with children:

- When pressing enter, a new block is created and the selection is moved to it, but the child blocks are not moved to it.
- When pressing backspace, if the previous block is also empty, the child blocks are deleted instead of being un-nested.

Closes #1593

## Rationale

This is a bug affecting core functionality.

## Changes

Added handling for children in relevnt keyboard handlers.

## Impact

N/A

## Testing

Added e2e tests.

## Checklist

### DIFF
--- a/packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
+++ b/packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
@@ -184,6 +184,7 @@ export const KeyboardShortcutsExtension = Extension.create<{
 
               let chainedCommands = chain();
 
+              // Moves the children of the current block to the previous one.
               if (blockInfo.childContainer) {
                 chainedCommands.insertContentAt(
                   blockInfo.bnBlock.afterPos,
@@ -437,6 +438,8 @@ export const KeyboardShortcutsExtension = Extension.create<{
               const newBlockContentPos = newBlockInsertionPos + 2;
 
               if (dispatch) {
+                // Creates a new block with the children of the current block,
+                // if it has any.
                 const newBlock = state.schema.nodes[
                   "blockContainer"
                 ].createAndFill(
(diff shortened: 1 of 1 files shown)


nperez0111:
@matthewlipski it's good that we don't lose content anymore, but in this case, I find the result unexpected, I would expect the item to preserve it's indentation not collapse because I removed the empty line
https://github.com/user-attachments/assets/178cad72-f2ed-4d05-b521-83e8ec32cea5

matthewlipski:
> @matthewlipski it's good that we don't lose content anymore, but in this case, I find the result unexpected, I would expect the item to preserve it's indentation not collapse because I removed the empty line https://github.com/user-attachments/assets/178cad72-f2ed-4d05-b521-83e8ec32cea5

Ah, so I do kind of agree but this is the same behaviour as Notion, so imo it's better to keep it the way it is:

https://github.com/user-attachments/assets/8a45521e-ad87-4d90-b080-ea979f899cd5

nperez0111:
It handles list items differently. So weird decision on their part, imo I would prefer what I'd described, but doesn't have to block this issue if you feel differently

matthewlipski:
@nperez0111 could you show what you mean with list items being handled differently? It seems to work the same for all block types as far as I can tell, but I might be missing something

https://github.com/user-attachments/assets/9dd35779-7b82-4d7f-8146-03392716760c

nperez0111:
https://github.com/user-attachments/assets/1fa376c7-c917-4d89-b6e9-9fd59f30261e

This is what I mean

matthewlipski:
Ok this is indeed a really weird choice from their side. I think for this PR, I'll keep the logic as is. If people end up filing issues that this is unexpected behaviour, changing it to preserve child block nesting should take not much effort. Will also add comments.

nperez0111:
Fine with me @matthewlipski

PR #2503, opened 2026-02-20
fix: no more scrolling to top when opening AI menu

### LINKED ISSUE #2446: Bug AI Menu scroll up
### What’s broken?

When we open the AI menu, the page scroll up.

To stay where the user open the menu.

### Steps to reproduce

https://github.com/user-attachments/assets/869bab2e-8168-4837-8fba-9b5d9d25e112

### BlockNote version

_No response_

### Environment

_No response_

### Additional context

_No response_

### Contribution

### Sponsor

YousefED:
closes #2446 

# Summary

Fix page scrolling to the top when opening the AI menu.

## Rationale

When opening the AI menu, the page's vertical scroll position would jump to the top. This happened because focusing the input element inside the floating popover triggers the browser's native scroll-into-view behavior before the element is properly positioned. By wrapping the popover content in floating-ui's `FloatingFocusManager` (via `focusManagerProps: {}`), focus is managed without causing the scroll jump ([floating-ui docs](https://floating-ui.com/docs/floatingfocusmanager#page-scrolls-to-top-when-opening-the-floating-element)).

## Changes

- Added `focusManagerProps` option to `FloatingUIOptions` and `GenericPopover` to optionally wrap popover content in a `FloatingFocusManager`
- Set `focusManagerProps: {}` in `AIMenuController` to use managed focus instead of native `autoFocus`
- Added Playwright e2e test to catch regressions

## Impact

No breaking changes. Only affects the AI menu's focus behavior when opening.

## Testing

- Playwright test scrolls to the bottom of the page, opens the AI menu, and verifies:
  - `scrollY` hasn't jumped to 0
  - The AI menu input is focused
  - Screenshot snapshot matches

## Screenshots/Video

x

## Checklist

## Additional Notes

N/A


packages/xl-ai/src/components/AIMenu/AIMenuController.tsx
```diff
@@ -26,71 +26,75 @@ export const AIMenuController = (props: {
...
+            zIndex: 100,
+          },
+          ...props.floatingUIOptions?.elementProps,
         },
-        ...props.floatingUIOptions?.elementProps,
-      },
-    }),
+        // we use the focus manager instead of `autoFocus={true}` to prevent "page-scrolls-to-top-when-opening-the-floating-element"
+        // see https://floating-ui.com/docs/floatingfocusmanager#page-scrolls-to-top-when-opening-the-floating-element
+        focusManagerProps: {},
```
nperez0111: This is sort of a smell no? This is equivalent to enabling it, but it doesn't feel right.

Shouldn't we always render the focus manager and just toggle the `disabled` prop?

YousefED: I agree, I improved this in latest commit:
- was a bit cumbersome as now I make sure it's disabled by default on all call sites
- good news is that I was able to fix a bug so that FloatingComposer (for comments) now also autofocuses correctly
- side note; do you know if `TableCellPopover` can be removed? it's not used anywhere afaik

nperez0111: At first I was confused why you needed this, but then it made sense that you needed to pass through these props. I think this is fine for now, and is explicit.

We don't need `TableCellPopover`, I think it stayed from the refactor. I removed it

Issue #2408, opened 2026-02-02
Slash command opens suggestion menu inside tables


expiante:
Using the **`/` command inside a table cell** opens the suggestion menu, but selecting a block inserts it **outside the table**, not inside the cell.

## Expected Behavior

- The “/” command should not open the suggestion menu inside table cells  

## Actual Behavior

- Suggestion menu opens inside a table cell
- Selecting a block inserts it **after the table**
- This behavior is unintuitive and misleading within table contexts

The suggestion menu should either be disabled inside tables.

nperez0111:
This is definitely a bug on our side, we will fix this

PR #2504, opened 2026-02-20
fix: disable slash menu in table content #2408

nperez0111:
# Summary

This disables the slash menu from triggering within table content

## Rationale

Table content cannot have block-level elements inserted within it, so they should be excluded from being considered.

## Impact

I'm unsure how we can filter out _only_ the slash menu from being invoked within table content. It feels to me like we need a callback API too which would filter out the sort of context that we'd be satisfied with triggering the suggestion menu within. 

## Checklist


YousefED:
> I'm unsure how we can filter out only the slash menu from being invoked within table content. It feels to me like we need a callback API too which would filter out the sort of context that we'd be satisfied with triggering the suggestion menu within.

Good catch. I think this our options are:
- refactor `triggerCharacter` to something like `menuOptions: {triggerCharacter: string, filterTransaction: (tr) => boolean}`
- or, zoom out and see if we'd want to refactor this component further (i.e.: make it more inline with the plugin / UI Element refactor from last year)

PR #2505, opened 2026-02-22
fix: correct the code blocks documentation

noahingh:
# Summary

I've made two corrections to the code blocks document.

1) Put the `codeBlock` field under the `blockSpecs` field because it's not a root field.  
2) The `dark` and `light` theme is not supported in [shiki](https://shiki.style/themes). 

https://github.com/TypeCellOS/BlockNote/blob/f150e1e91c73e3b18b8b43336402f580444d7ed6/packages/core/src/schema/schema.ts#L174-L227


nperez0111 (approved):
Thanks @noahingh this is correct!

PR #2550, opened 2026-03-10
refactor: unified drop cursor extension

nperez0111:
# Summary

The goal of this refactor was to add the ability to exclude certain classnames from being considered for drag & drop into the editor. This proved to be difficult with the existing `prosemirror-dropcursor` extension. We have already run into issues with this extension in the past (e.g. with the multi-column dropping), so I decided to just refactor this into a BlockNote extension. Now, the default extension can handle the use cases of both the multi-column extension & the base drop-cursor needs, by exposing options in a similar way to how we already exposed replacing the plugin entirely.

What is cool, is that while the underlying options & interface have changed, this should be a zero code change for consumers to upgrade (assuming they bump both the editor & multi-column at the same time).

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Drag & Drop Exclusion example and interactive demo added.
  * Enhanced drop-cursor with configurable positioning and multi-column edge support.
  * Playground search input to find examples.

* **Documentation**
  * Detailed guide and demo notes for using drag exclusion.

* **Chores**
  * Removed prosemirror-dropcursor from automatic updates and project dependencies.

### DIFF
--- a/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
+++ b/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
@@ -319,7 +319,13 @@ function serializeBlock<
       }
     }
 
-    if (editor.pmSchema.nodes[block.type as any].isInGroup("blockContent")) {
+    if ("childrenDOM" in ret && ret.childrenDOM) {
+      // block specifies where children should go (e.g. toggle blocks
+      // place children inside <details>)
+      ret.childrenDOM.append(childFragment);
+    } else if (
+      editor.pmSchema.nodes[block.type as any].isInGroup("blockContent")
+    ) {
       // default "blockContainer" style blocks are flattened (no "nested block" support) for externalHTML, so append the child fragment to the outer fragment
       fragment.append(childFragment);
     } else {
--- a/packages/core/src/blocks/Heading/block.ts
+++ b/packages/core/src/blocks/Heading/block.ts
@@ -1,11 +1,12 @@
-import { createBlockConfig, createBlockSpec } from "../../schema/index.js";
 import { BlockNoteEditor } from "../../editor/BlockNoteEditor.js";
 import { createExtension } from "../../editor/BlockNoteExtension.js";
+import { createBlockConfig, createBlockSpec } from "../../schema/index.js";
 import {
   addDefaultPropsExternalHTML,
   defaultProps,
   parseDefaultProps,
 } from "../defaultProps.js";
+import { getDetailsContent } from "../getDetailsContent.js";
 import { createToggleWrapper } from "../ToggleWrapper/createToggleWrapper.js";
 
 const HEADING_LEVELS = [1, 2, 3, 4, 5, 6] as const;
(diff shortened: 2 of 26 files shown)


packages/core/src/extensions/DropCursor/DropCursor.ts
```diff
@@ -1,26 +1,263 @@
...
+  getBlockDropRect,
+  getInlineDropRect,
+  getParentOffsets,
+  hasExclusionClassname,
+  type DropCursorPosition,
+} from "./utils.js";
+import type { BlockNoteEditor } from "../../editor/BlockNoteEditor.js";
+import { createExtension } from "../../editor/BlockNoteExtension.js";
+
+export const DRAG_EXCLUSION_CLASSNAME = "bn-drag-exclude";
```
YousefED, line 14: do we want to export this as a default classname? And have a global export that we need to document, etc? Or rather just document `options.exclude` (without default)?

packages/core/src/extensions/DropCursor/DropCursor.ts
```diff
@@ -1,26 +1,263 @@
...
+}
+
+/**
+ * Drop cursor visualization based on prosemirror-dropcursor:
+ * https://github.com/ProseMirror/prosemirror-dropcursor/blob/master/src/dropcursor.ts
+ *
+ * Refactored to use BlockNote extension pattern with mount callback and AbortSignal
+ * for lifecycle management instead of ProseMirror PluginView.
+ */
+export const DropCursorExtension = createExtension<
```
YousefED, line 56: I assume the logic itself is 95% the same as it was right? i.e. we don't need to review this extensively?

(except for the hooks and "exclude"?

examples/03-ui-components/18-drag-n-drop/.bnexample.json
```diff
@@ -0,0 +1,6 @@
+{
+  "playground": true,
+  "docs": true,
+  "author": "must",
```
YousefED: ?

examples/03-ui-components/18-drag-n-drop/.bnexample.json
```diff
@@ -0,0 +1,6 @@
+{
+  "playground": true,
+  "docs": true,
```
YousefED: do you think this should just be a playground example, or also listed on the website? I think it might be a bit specific / overkill for the latter, wdyt?

(maybe we need a general rool for this)

packages/core/src/extensions/DropCursor/DropCursor.ts
nperez0111: I think it's easier to describe that you can just add a classname than having to configure it

packages/core/src/extensions/DropCursor/DropCursor.ts
nperez0111: Yep, is the same, obv want to check behavior but my testing makes me think this is solid

packages/core/src/extensions/DropCursor/DropCursor.ts
nperez0111: export I could lose

PR #2553, opened 2026-03-11
fix(core): slash menu fails in custom blocks after space BLO-1036

### LINKED ISSUE #659: Bug: Cannot add space at the end of a custom block's content

### LINKED ISSUE #2531: Mention element doesn't work as expected within custom blocks
### What’s broken?

The trigger character `@` is not opening the list of users if there's a space before it, it works only next to the text.. For example:
```javascript
some text @ // won't work
```
```javascript
some text@ // works
```

Just when the mention element is implemented, the mention element should always appear in the custom blocks when hitting `@`, regardless of the side conditions.

### Steps to reproduce

_No response_

### BlockNote version

v0.47.0

### Environment

_No response_

### Additional context

_No response_

### Contribution

### Sponsor

- [x] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖 (Representative of Krisp)

YousefED:
# Summary

Fix the slash menu (and other suggestion menus) not opening when the trigger character is typed after a space inside custom blocks with `isolating: true`.

Closes #2531
Closes #659

## Rationale

Tiptap's `NodeViewWrapper` hardcodes `white-space: normal` as an inline style on all React-based custom block wrappers. ProseMirror requires `white-space: pre-wrap` on editable content areas — without it, browsers (especially Chrome) normalize trailing ASCII spaces to NBSP (`\u00a0`) during input.

When a user types `/` after a trailing space inside a custom block, the browser first converts the space to NBSP. ProseMirror's `readDOMChange` → `findDiff` then sees a **replacement** (space→NBSP+char) instead of a pure insertion, calling `handleTextInput(view, from, to, text)` with `from !== to`. The SuggestionMenu plugin's `if (from === to)` guard rejects this, and the menu never opens.

Issue #659 (spaces eaten at end of custom blocks in Firefox) shares the same root cause — `white-space: normal` collapses trailing whitespace. While we could not reproduce #659 on current Firefox versions bundled with Playwright, the fix addresses the underlying CSS problem.

## Changes

- **`packages/core/src/editor/Block.css`**: Added `white-space: pre-wrap` to `.bn-inline-content`, overriding the `normal` inherited from tiptap's NodeViewWrapper.
- **`tests/src/end-to-end/slashmenu/slashmenu-customblock.test.ts`**: New Playwright e2e regression test using the alert-block demo — verifies the slash menu opens both with and without a preceding space.

## Impact

- For **standard blocks** (paragraph, heading, etc.): `white-space` was already `pre-wrap` via inheritance from ProseMirror's default styles. The explicit declaration is a no-op.
- For **custom blocks** wrapped by tiptap's `NodeViewWrapper`: `white-space` changes from `normal` to `pre-wrap`, restoring correct ProseMirror behavior. This is the intended fix.
- No visual changes expected — `pre-wrap` is what ProseMirror already uses everywhere else.

## Testing

- **Playwright e2e**: Verified the test **fails** on Chromium without the fix, and **passes** on both Chromium and Firefox with the fix.
- Manually tested slash menu in paragraphs, headings, and custom alert blocks.
- Existing unit tests for SuggestionMenu continue to pass.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inline whitespace handling to prevent trailing-space normalization from breaking suggestion triggers and restore consistent editor behavior.

* **Tests**
  * Added end-to-end regression tests ensuring the global slash menu opens correctly inside custom alert blocks across typing scenarios, plus related test configuration for the alert-block page.


nperez0111:
I was able to validate that we fixed #659 on firefox with this

PR #2591, opened 2026-03-24
refactor: portal floating UI elements to document.body to prevent overflow clipping BLO-1115

### LINKED ISSUE #2543: Slash menu is clipped inside the editor when editor height is small
### What’s broken?

When the editor height is set to a small value, the slash menu (triggered by typing `/`) gets trapped within the editor's bounding area and is cut off — part of the menu is not visible to the user.

The slash menu should render as a floating overlay that escapes the editor's bounding box and remains fully visible, regardless of the editor's height.

Image

**Steps to reproduce**
1. Set the editor container to a small height (e.g. ~100px)
2. Click inside the editor
3. Type `/` to trigger the slash menu
4. Observe that the slash menu is clipped and cut off within the editor boundary

Optional: You can reproduce this on the official minimal demo without any code changes:
https://www.blocknotejs.org/examples/basic/minimal
→ Open browser DevTools → select the editor element → reduce its height

### Steps to reproduce

_No response_

### BlockNote version

v.0.47.1

### Environment

_No response_

### Additional context

The slash menu appears to be rendered inside the editor's DOM subtree and is constrained by `overflow: hidden` on a parent element. The menu should ideally be portaled to `document.body` or use a positioning strategy (e.g. Floating UI's `strategy: 'fixed'`) to escape parent overflow constraints.

### Contribution

### Sponsor

### LINKED ISSUE #2544: Drag handle menu is clipped when using shadcn with position relative
### What’s broken?

When you render the shadcn editor with 
```
.bn-container {
  position: relative;
}
```
to prevent floating elements escaping their parent, the drag handle menu often renders in an incorrect position and gets clipped by the side of the container

This can be seen in the shadcn example in the documentation
https://www.blocknotejs.org/examples/basic/shadcn

Image

The drag handle menu is not clipped

### Steps to reproduce

1. Go to https://www.blocknotejs.org/examples/basic/shadcn
2. Click on the drag handle button in the example to trigger the menu
3. Observe that the drag handle menu is clipped

### BlockNote version

latest

### Environment

chrome latest, safari latest, mac OS tahoe

### Additional context

_No response_

### Contribution

### Sponsor

### LINKED ISSUE #2558: FormattingToolbar is clipped when editor height is small
### What’s broken?

When the editor container is set to a small height, the FormattingToolbar (shown when text is selected) gets trapped within the editor's bounding area and is cut off — part of the toolbar is not visible to the user.

The FormattingToolbar should render as a floating overlay that escapes the editor's bounding box and remains fully visible, regardless of the editor's height.

Image

Image

### Steps to reproduce

1. Set the editor container to a small height (e.g. ~100px)
2. Click inside the editor and type some text
3. Select the text to trigger the FormattingToolbar
4. Observe that the toolbar is clipped and partially hidden within the editor boundary

Optional: Reproducible on the official minimal demo without any code changes:
https://www.blocknotejs.org/examples/basic/minimal
→ Open browser DevTools → select the editor element → reduce its height → select text

### BlockNote version

0.47.1

### Environment

_No response_

### Additional context

[This appears to be the same root cause as the slash menu clipping issue ](https://github.com/TypeCellOS/BlockNote/issues/2543)

### Contribution

### Sponsor

### LINKED ISSUE #2578: (Comments) GridSuggestionMenuController clipped in parent
### What’s broken?

When we want to add a emoji thanks to the shortcut ":", the modal is clipped inside the parent, the "comment" in this occurence.
Could we fix it as we did for this issue: https://github.com/TypeCellOS/BlockNote/issues/2053.

https://github.com/TypeCellOS/BlockNote/blob/b63b4096daa575f821980ef897fd90f4c76d9e42/packages/react/src/editor/BlockNoteDefaultUI.tsx#L103-L107

Image

### Reproduce

You can reproduce here: https://www.blocknotejs.org/examples/collaboration/comments

To not be clipped in the parent, in a similar fashion as this one, with `createPortal` I guess. https://github.com/TypeCellOS/BlockNote/issues/2053

### Sponsor

YousefED:
# Summary

Floating UI elements (menus, toolbars, emoji picker, etc.) are now portaled to `document.body`, preventing clipping by `overflow: hidden` ancestors.

Closes #2543
Closes #2544
Closes #2558
Closes #2578
Supersedes the approach from #2092

## Rationale

When BlockNote is inside a container with `overflow: hidden` (sidebar, modal, scrollable panel), floating UI gets clipped. [Floating UI docs](https://floating-ui.com/docs/misc#clipping) confirm portaling is the only reliable fix. This is also the default behavior in libraries like Base UI.

### Portal infrastructure
- **`BlockNoteEditor.portalElement`** — lazily-created `div` on the editor instance. Appended to `document.body` (or ShadowRoot) by `mount()`, removed by `unmount()`.
- **`BlockNoteEditor.isWithinEditor(element)`** — new method that checks if an element is inside the editor DOM or its portal container. Used by core extensions to correctly handle portaled elements.
- **`BlockNoteView`** syncs theming classes (`bn-root`, color scheme, user className) onto the portal element via effects.
- **`GenericPopover`** wraps all floating elements with `<FloatingPortal root={portalRoot}>`.

### CSS: `bn-root` vs `bn-container`
- **`bn-root`**: theming (CSS variables, fonts). Applied to both editor and portal containers.
- **`bn-container`**: layout (width, height). Editor container only.
- Theme CSS variables moved from `.bn-container` to `.bn-root`.

### Popover `portalRoot` (for EmojiPicker)
- Added `portalRoot` to generic `Popover.Root` interface.
- **Mantine**: native `withinPortal` + `portalProps`.
- **shadcn**: threads `portalRoot` via React context, uses `createPortal`.
- **Ariakit**: threads `portalRoot` via React context, native `portalElement` prop.

### Core extension fixes
- **SideMenu**: uses `editor.isWithinEditor()` so hovering portaled floating elements doesn't dismiss the side menu.
- **UniqueID**: uses `isWithinEditor` option to detect internal drags from the portaled drag handle, preventing block IDs from being regenerated on drag-drop.

### Z-index
- Portaled elements don't need high z-index values (no stacking context to escape). Hardcoded `z-index: 10000` removed from Mantine popover and ariakit styles when portaling.
- Non-portal fallback preserves existing z-index behavior.
- `--bn-ui-base-z-index` on `.bn-root` remains available for user customization.

### EmojiPicker simplification
- Removed manual `createPortal` — passes `portalRoot` to `Popover.Root` instead.
- Scoped `em-emoji-picker` styles to `.bn-root`, removed stale `z-index: 11000`.

## Impact

- **Breaking (CSS)**: Theme-related properties (fonts, colors, CSS variables) should target `.bn-root` instead of `.bn-container`. `.bn-container` now only wraps the editor itself, while `.bn-root` wraps both the editor and its floating UI (menus, toolbars). Layout related properties (width, height, margin) should still target `.bn-container`.

## Testing

- Manual testing with editor inside `overflow: hidden` containers.
- TypeScript checks pass for all packages.

## Checklist


.claude/worktrees/emoji-fixes
nperez0111, line 1: rm please

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -147,6 +149,22 @@ function BlockNoteViewComponent<
...
+  // Register the portal element on the editor so core extensions (SideMenu,
+  // UniqueID) can identify portaled elements as belonging to this editor.
+  // (through editor.isWithinEditor)
+  useEffect(() => {
+    editor.portalElement = portalRoot ?? undefined;
+    return () => {
+      editor.portalElement = undefined;
+    };
+  }, [portalRoot, editor]);
+
```
nperez0111: Nope, we got to find another way around this. This is a mess.

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -205,6 +224,14 @@ function BlockNoteViewComponent<
         >
           {children}
         </BlockNoteViewContainer>
+        {createPortal(
+          <div
+            ref={setPortalRoot}
+            className={mergeCSSClasses("bn-root", editorColorScheme, className)}
+            data-color-scheme={editorColorScheme}
+          />,
+          document.body,
```
nperez0111: I'd prefer this to not be hard-coded to just add elements into the document.body, we should allow an option to have a selector/callback to pick where to place this. We are a library not an app

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -205,6 +213,18 @@ function BlockNoteViewComponent<
...
+            ref={setPortalRoot}
+            className={mergeCSSClasses(
+              "bn-root",
+              editorColorScheme,
+              className,
+            )}
+            data-color-scheme={editorColorScheme}
+          />,
+          document.body,
+        )}
```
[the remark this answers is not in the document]
nperez0111: Yep, I think this is a fundamental problem in this approach, it being portaled outside of the editor's control does have some serious drawbacks and makes styling more complicated. I think we should re-visit this approach

nperez0111 (changes requested):
Definitely not comfortable with this approach. I agree it is worth solving, but as a lib I don't think we have this luxury

YousefED:
> Definitely not comfortable with this approach. I agree it is worth solving, but as a lib I don't think we have this luxury

Fair, ideally you wouldn't add sth to document.root as a library. Do you think it needs a user-facing API, or would you take a completely different approach?

Note that @ https://floating-ui.com/docs/misc#clipping, they mention a portal is the only reliable solution to fix the clipping problem. It's also similar to what other libraries like base-ui etc do by default, if I'm not mistaken

nperez0111:
> Fair, ideally you wouldn't add sth to document.root as a library. Do you think it needs a user-facing API, or would you take a completely different approach?

Unsure exactly how I'd go about it, I didn't review the existing issues to know where the current limitations are, to know the right problem to solve.

> Note that @ [floating-ui.com/docs/misc#clipping](https://floating-ui.com/docs/misc#clipping), they mention a portal is the only reliable solution to fix the clipping problem. It's also similar to what other libraries like base-ui etc do by default, if I'm not mistaken

Yea, I think we should give the fixed positioning strategy a try, which they also mention:

> Using the 'fixed' strategy will “break” the floating element out of its parent clipping context in the majority of cases. This allows you to keep the floating element’s DOM context in tact. This is not 100% guaranteed though — portalling is the only 100% reliable method.

I think base-ui can get away with it because it is unstyled, whereas I think we will run into some synchronization of styles issues between something within the portal and something within the editor. I'd rather not have to make that distinction if we can get away with it. We would be fighting against the `C`ascade in CSS with this approach and IME, you can win the battle, but not the war with this.

YousefED:
fyi, this is what claude gives when asking for downsides of "fixed" (didn't validate all if this):

The Misc page doesn't go into the downsides directly — let me grab the `computePosition` docs where `strategy` is documented more fully.The Floating UI docs don't give a dedicated "downsides" list, but here's what they state and what follows from it:

**Explicitly from the docs:**

- It's not 100% guaranteed to break out of clipping — portalling is the only fully reliable method.
- `'absolute'` requires the browser to do the least work when updating position, implying `fixed` carries more layout/repaint cost.

**What that means in practice (CSS-level consequences):**

`position: fixed` is broken by any ancestor that creates a new **containing block**, which includes:
- `transform` (any value other than `none`) — the most common gotcha
- `filter` (any value other than `none`)
- `perspective`
- `will-change: transform` / `will-change: filter`
- `contain: layout`, `contain: paint`, or `contain: strict`
- `backdrop-filter` (in some browsers)

So if your floating element lives inside — or gets portalled into — a subtree with any of those CSS properties on an ancestor, `fixed` behaves like `absolute` relative to that ancestor, not the viewport. This is a very common surprise in component libraries and animation-heavy UIs.

**Also relevant for BlockNote specifically:** if you have a scrollable editor container, `fixed` elements stay viewport-anchored while the reference element scrolls, which means `autoUpdate` must run on scroll events to keep coordinates in sync — more overhead than `absolute` which naturally follows the scroll.

---
let's review later 👍

YousefED:
We've decided to continue with this, after confirming that:
- we need to support overflow-constraint layouts (for example, when the editor is in a sidebar)
- strategy: fixed results in janky positioning on scroll

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -714,6 +714,25 @@ export class BlockNoteEditor<
     return this.prosemirrorView?.dom as HTMLDivElement | undefined;
   }
 
+  /**
+   * The portal container element at `document.body` used by floating UI
+   * elements (menus, toolbars) to escape overflow:hidden ancestors.
+   * Set by BlockNoteView; undefined in headless mode.
+   */
+  public portalElement: HTMLElement | undefined;
```
YousefED: @nperez0111 not sure about setting this on the editor. We need it for the UniqueID and SideMenu extensions though. wdyt?

packages/core/src/editor/editor.css
```diff
@@ -20,26 +20,6 @@
   padding: 0;
 }
 
-/*
```
YousefED, line 23: this was unused, lingering code that I removed in this PR

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -147,6 +152,27 @@ function BlockNoteViewComponent<
...
+  // same theming classes as the editor container (bn-root + color scheme), but
+  // not bn-container (which is for layout targeting only).
+  const [portalRoot, setPortalRoot] = useState<HTMLDivElement | null>(null);
+
+  // Ref callback that both updates state (for context consumers) and syncs
+  // the portal element to the editor (for core extensions like SideMenu/UniqueID).
+  const internalPortalRef = useCallback(
+    (node: HTMLDivElement | null) => {
+      editor.portalElement = node ?? undefined;
+      setPortalRoot(node);
```
YousefED: @nperez0111 this will cause an extra render unfortunately, but afaik it's needed because the `blockNoteContext` depends on it.

An alternative would be to not use a ref but just document.createElement ahead-of-time - but that seems less "idiomatic react". wdyt?

packages/shadcn/src/comments/Comment.tsx
```diff
@@ -70,7 +70,7 @@ export const Comment = forwardRef<
     timeString,
     actions,
     edited,
-    emojiPickerOpen, // Unused
+    emojiPickerOpen,
```
YousefED, line 73: this was missing in shadcn and ariakit implementations

packages/ariakit/src/comments/Comment.tsx
```diff
@@ -58,7 +58,7 @@ export const Comment = forwardRef<
     actions,
     children,
     edited,
-    emojiPickerOpen, // Unused
```
YousefED, line 61: fyi, this was not implemented in ariakit / shadcn, so decided to fix

YousefED:
Updated PR to fix:
- shadowroot support
- theming support
- remove as many hardcoded zIndex values as possible, afaik they're not needed anymore

Tested with:
- shadowroot demo
- theming/theming-css-variables-code demo

packages/react/src/editor/BlockNoteView.tsx
YousefED: I changed this to the document.createElement approach, because that also made it easier to support shadow dom

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: Nope, I would not put this on the editor, much less public. Need to find a workaround here

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -714,6 +714,25 @@ export class BlockNoteEditor<
...
+   * Checks whether a DOM element belongs to this editor — either inside the
+   * editor's DOM tree or inside its portal container (used for floating UI
+   * elements like menus and toolbars).
+   */
+  public isWithinEditor = (element: Element): boolean => {
+    return !!(
+      this.domElement?.parentElement?.contains(element) ||
+      this.portalElement?.contains(element)
+    );
+  };
```
nperez0111, line 753: This will have to change if portalElement isn't stored here

packages/core/src/extensions/SideMenu/SideMenu.ts
```diff
@@ -623,11 +620,8 @@ export class SideMenuView<
       // An element is hovered
       event &&
       event.target &&
-      // Element is outside the editor
-      !(
-        editorWrapper === event.target ||
-        editorWrapper.contains(event.target as HTMLElement)
-      )
+      // Element is outside this editor and its portaled UI
+      !this.editor.isWithinEditor(event.target as HTMLElement)
```
nperez0111, line 624: I'm not sure that this is correct. The TODO comment from before made it seem like it was intentionally using the parent element of the editor

packages/core/src/extensions/tiptap-extensions/UniqueID/UniqueID.ts
```diff
@@ -228,14 +232,11 @@ const UniqueID = Extension.create({
...
-                ? void 0
-                : _a.contains(event.target)
-            )
-              ? view.dom.parentElement
-              : null;
+            const editorParent = view.dom.parentElement;
+            const isFromEditor =
+              editorParent?.contains(event.target) ||
+              isWithinEditor?.(event.target);
+            dragSourceElement = isFromEditor ? editorParent : null;
```
nperez0111, line 239: Does dragging from a popover really matter? Why is this check here?

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -147,6 +165,59 @@ function BlockNoteViewComponent<
...
+    [portalRoot],
+  );
+
+  useEffect(() => {
+    if (typeof portalRootRef === "function") {
+      portalRootRef(internalPortalRoot);
+    } else if (portalRootRef) {
+      portalRootRef.current = internalPortalRoot;
+    }
+  }, [portalRootRef, internalPortalRoot]);
```
nperez0111: refs usually have the semantic that it is called when the element is mounted. Why do it like this? Instead of on actual mount?

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -147,6 +165,59 @@ function BlockNoteViewComponent<
...
+
+  // const portalRoot = portalContainer ?? internalPortalEl;
+
+  useEffect(() => {
+    editor.portalElement = internalPortalRoot;
```
nperez0111: This is just asking for trouble. No, this will certainly clobber

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -147,6 +165,59 @@ function BlockNoteViewComponent<
...
+    internalPortalRoot.className = mergeCSSClasses(
+      "bn-root",
+      editorColorScheme,
+      className || "",
+    );
+    internalPortalRoot.setAttribute("data-color-scheme", editorColorScheme);
+  }, [internalPortalRoot, editorColorScheme, className]);
```
nperez0111: why is this done in an effect? Can't you do this in the memo above?

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -147,6 +165,59 @@ function BlockNoteViewComponent<
...
+      const root = element?.getRootNode();
+      const container = root instanceof ShadowRoot ? root : document.body;
+      if (container) {
+        container.appendChild(internalPortalRoot);
+      } else {
+        internalPortalRoot.remove();
+      }
+    },
+    [portalRoot, internalPortalRoot],
+  );
```
nperez0111: It isn't obvious to me why we are doing this. Is this for shadow dom stuff?

nperez0111 (changes requested):
Let's chat

packages/core/src/extensions/tiptap-extensions/UniqueID/UniqueID.ts
YousefED: I think the idea is if you drag from one editor to another, the ids are reset (which might be questionable, but rather keep behavior as is)

packages/core/src/extensions/SideMenu/SideMenu.ts
YousefED: isWithinEditor also checks the parentelement (so it's consistent with the code before)

nperez0111 (approved):
cool that did simplify things

Issue #2122, opened 2025-10-21
iOS Safari: On text selection two overlapping menus


wielinde:
<img width="739" height="1600" alt="Image" src="https://github.com/user-attachments/assets/2cdc58bc-48f6-4a63-a803-c1b292b9e4e6" />

**Describe the bug**
This bug effectively makes it very hard for users to reach all the features of the BlockNote toolbar when on mobile. And users rely on that toolbar even more as they would on a desktop computer. On the desktop they can still use keyboard short for in- and out-denting of lists. But  also marking a text as bold/italic after you have written a text becomes almost impossible.

**To Reproduce**
- on iOS Safari
- open any standard BlockNote editor
- write text
- select a word that is rather at the top of the editor/screen/viewport
- two menu open below (as there is not enough space on top) and overlap. This makes it impossible to reach the BlockNote toolbar under that circumstances

**Misc**

- Node version:
- Package manager:
- Browser:
- [x] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

NarekManukyan:
We are experiencing the same issue on Android.

I maintain a Flutter WebView wrapper for BlockNote, which is already used in production, so this behavior impacts a significant number of mobile users.

On mobile devices, the toolbar is even more critical than on desktop due to the lack of keyboard shortcuts. When the selection is near the top of the viewport, the native selection menu and the BlockNote toolbar overlap, making it impossible to access formatting actions such as bold/italic or list indentation.

Is this issue currently being addressed, or is there a recommended workaround on the BlockNote side?

Flutter wrapper for reference:
https://pub.dev/packages/flutter_blocknote_editor

nperez0111:
@NarekManukyan very cool to see that you've embedded BlockNote in Flutter. Mobile support is definitely on our roadmap for this year, just aren't sure where it slots in yet. So more updates soon on this

judithroth:
I see the same problematic behaviour on Firefox / Android:

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/41755f7b-9dfd-42a7-9b63-b0b5cf596ff0" />

Is mobile support still on the roadmap for this year? Can we (OpenProject) somehow support there?

YousefED:
> Is mobile support still on the roadmap for this year? Can we (OpenProject) somehow support there?

[...]

Issue #938, opened 2024-07-16
Format toolbar goes offscreen on mobile


LeviBensley:
**Describe the bug**
When using the editor on mobile, the formatting toolbar goes partially off the screen

**To Reproduce**
1) Insert basic editor on a page
2) highlight words
3) formatting header appear partially off screen
![screenshot from Safari](https://github.com/user-attachments/assets/71fffaa3-1b0f-40a5-ab2f-466465ff398b)

PR #2617, opened 2026-04-02
fix: flicker-free mobile formatting toolbar via CSS custom properties

Movm:
# Summary

Replaces the React state-driven positioning in `ExperimentalMobileFormattingToolbarController` with a CSS custom property (`--bn-mobile-keyboard-offset`), eliminating the re-render storm that causes visible flickering during keyboard animation.

Tested on iOS Safari (iPhone) — the Visual Viewport API fallback works great, smooth positioning with no flicker.

Relates to #938, #2122, https://github.com/TypeCellOS/BlockNote/discussions/2616

## Rationale

The current implementation calls `setTransform()` → React re-render on every `visualViewport` scroll/resize event. During keyboard animation this fires dozens of times per second, causing visible toolbar flickering. Moving positioning to a CSS custom property lets the browser compositor handle animation via `transition: bottom 0.15s ease-out` — zero React re-renders for positioning.

### `ExperimentalMobileFormattingToolbarController.tsx`

**Two-tier keyboard detection** (progressive enhancement):

1. **VirtualKeyboard API** (Chrome/Edge 94+, Samsung Internet) — sets `overlaysContent = true` and listens to `geometrychange` for exact keyboard bounding rect before animation starts
2. **Visual Viewport API fallback** (Safari iOS 13+, Firefox Android 68+) — computes keyboard height from `clientHeight - vp.height - vp.offsetTop`, with **focus-based prediction** that immediately applies the last-known keyboard height on `focusin` to avoid toolbar jumping after the keyboard animates in

Both tiers write a single CSS custom property `--bn-mobile-keyboard-offset` directly to the wrapper DOM element — no `useState`, no re-renders.

**Additional improvements:**
- `scrollSelectionIntoView()` — auto-scrolls the cursor/selection into view when the keyboard opens, accounting for toolbar height
- `focusout` listener resets offset to 0 when keyboard dismisses
- Proper cleanup of all event listeners

### `styles.css`

New `.bn-mobile-formatting-toolbar` class:
- `bottom: var(--bn-mobile-keyboard-offset, 0px)` — positions above keyboard
- `transition: bottom 0.15s ease-out` — smooth animation
- `touch-action: pan-x` — allows horizontal scrolling on toolbar buttons without vertical scroll interference
- `padding-bottom: env(safe-area-inset-bottom)` — handles notch/home indicator on modern devices

### Browser support

| Browser | API Used | Quality |
|---------|----------|---------|
| Chrome/Edge Android 94+ | VirtualKeyboard | Instant, exact geometry |
| Samsung Internet | VirtualKeyboard | Instant, exact geometry |
| Safari iOS 13+ | Visual Viewport + prediction | Smooth, tested on device ✅ |
| Firefox Android 68+ | Visual Viewport + prediction | Smooth |

### Note on `interactive-widget`

For best results, consumers can add `interactive-widget=resizes-content` to their viewport meta tag. This makes `position: fixed` elements work naturally with the keyboard. However, the implementation works without it — the Visual Viewport fallback handles both cases.

## Summary by CodeRabbit

* **Improvements**
  * Better mobile formatting toolbar positioning with native Virtual Keyboard API support and robust fallback for older browsers
  * Improved keyboard visibility handling to keep the toolbar above on-screen keyboards and preserve selection visibility
* **Style**
  * New toolbar styling for fixed bottom placement, horizontal scrolling, safe-area padding, and smooth bottom-offset transitions

### DIFF
--- a/packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
+++ b/packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
@@ -7,8 +7,6 @@ import { useExtensionState } from "../../hooks/useExtension.js";
 import { FormattingToolbar } from "./FormattingToolbar.js";
 import { FormattingToolbarProps } from "./FormattingToolbarProps.js";
 
-const TOOLBAR_HEIGHT = 44;
-
 /**
  * Flicker-free mobile formatting toolbar controller.
  *
@@ -56,7 +54,8 @@ export const ExperimentalMobileFormattingToolbarController = (props: {
       const rect = sel.getRangeAt(0).getBoundingClientRect();
       const vp = window.visualViewport;
       if (!vp) return;
-      const visibleBottom = vp.offsetTop + vp.height - TOOLBAR_HEIGHT;
+      const toolbarHeight = el.getBoundingClientRect().height || 44;
+      const visibleBottom = vp.offsetTop + vp.height - toolbarHeight;
       if (rect.bottom > visibleBottom) {
         window.scrollBy({
           top: rect.bottom - visibleBottom + 16,
--- a/packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
+++ b/packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
@@ -1,22 +1,31 @@
 import { BlockSchema, InlineContentSchema, StyleSchema } from "@blocknote/core";
 import { FormattingToolbarExtension } from "@blocknote/core/extensions";
-import { FC, CSSProperties, useMemo, useRef, useState, useEffect } from "react";
+import { FC, useRef, useEffect } from "react";
 
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
 import { useExtensionState } from "../../hooks/useExtension.js";
 import { FormattingToolbar } from "./FormattingToolbar.js";
 import { FormattingToolbarProps } from "./FormattingToolbarProps.js";
 
+const TOOLBAR_HEIGHT = 44;
+
 /**
- * Experimental formatting toolbar controller for mobile devices.
- * Uses Visual Viewport API to position the toolbar above the virtual keyboard.
+ * Flicker-free mobile formatting toolbar controller.
  *
- * Currently marked experimental due to the flickering issue with positioning cause by the use of the API (and likely a delay in its updates).
+ * Uses a CSS custom property (`--bn-mobile-keyboard-offset`) instead of React
+ * state to position the toolbar above the virtual keyboard.  This avoids the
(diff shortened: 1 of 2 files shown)


packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
```diff
@@ -1,22 +1,31 @@
...
 import { FormattingToolbarExtension } from "@blocknote/core/extensions";
-import { FC, CSSProperties, useMemo, useRef, useState, useEffect } from "react";
+import { FC, useRef, useEffect } from "react";
 
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
 import { useExtensionState } from "../../hooks/useExtension.js";
 import { FormattingToolbar } from "./FormattingToolbar.js";
 import { FormattingToolbarProps } from "./FormattingToolbarProps.js";
 
+const TOOLBAR_HEIGHT = 44;
```
nperez0111: I'm not sure that it is a great idea to hard code this. I understand reading it once & then using a cached value, but hard coding upfront feels wrong to me

packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
```diff
@@ -28,60 +37,121 @@ export const ExperimentalMobileFormattingToolbarController = (props: {
...
+          top: rect.bottom - visibleBottom + 16,
+          behavior: "smooth",
+        });
+      } else if (rect.top < vp.offsetTop) {
+        window.scrollBy({
+          top: rect.top - vp.offsetTop - 16,
+          behavior: "smooth",
+        });
+      }
+    };
```
nperez0111, line 70: Unclear to me why we would need to be scrolling for this?

nperez0111:
Thanks for looking into this, @Movm I agree that we likely can move a lot of this computation into CSS & avoid re-renders. This seems like a good start for that

packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
Movm: Good call — replaced with `el.getBoundingClientRect().height || 44` so it measures the actual rendered toolbar at call time. Works correctly with custom `formattingToolbar` components too.

packages/react/src/components/FormattingToolbar/ExperimentalMobileFormattingToolbarController.tsx
Movm: When the keyboard opens, the visible area shrinks. If the cursor is near the bottom, it ends up hidden behind the keyboard + toolbar. This scrolls it back into view so users can see what they're typing. That said — happy to split this into a separate PR if you'd prefer to keep this one focused on the flicker fix.

PR #2619, opened 2026-04-02
fix: `useEditorDOMElement` hook

### LINKED ISSUE #2546: Link: no menu appears on hover
### What’s broken?

On link hover, edit menu is not appearing

Link edit menu appears on link hover (for all links, including 1 char links)

### Steps to reproduce

1. Go to https://www.blocknotejs.org/docs/react/components/hyperlink-toolbar
2. Hover over "Home Page" link in the example: link toolbar with "Edit link" and other menu options does not appear on hover.
(Note: this happens 90% of time. Sometimes, very rarely, link hovers are available immediately. So seems for some reason it works sometimes)
3. If I click any link at least once, hovers start appearing 
(Note: this is not connected to focus being inside editor. Even if focus was somewhere on other text within editor before, links are not appearing until at least one link is clicked)

One more use case: Create link containing 1 char. Hover never appears at all

### BlockNote version

Latest

### Environment

Chrome

### Additional context

_No response_

### Contribution

### Sponsor

matthewlipski:
# Summary

This PR adds a `useEditorDOMElement` hook which fetches the editor DOM element. It's necessary for certain cases where components rely on `editor.domElement` to be defined to attach event listeners. However, when it becomes defined, there is nothing which explicitly triggers a re-render, so the hook is needed to force that re-render.

Internally, `useEditorDOMElement` uses `useEditorState`, which has been given a `"mount"` event to specifically listen for when the view becomes defined/undefined.

Closes #2546

## Rationale

See above.

## Changes

- Added `"mount"` event to `useEditorState`.
- Made `"all"` event in `useEditorState` also listen for TipTap `"create"`/`"mount"`/`"unmount"` events.
- Added `useEditorDOMElement` hook.
- Replaced direct references to `editor.domElement` with the hook where necessary.

## Impact

N/A

## Testing

I wasn't able to reproduce the issue in #2546 locally, and have not seen other cases of this elsewhere. Therefore, no tests have been added.

## Screenshots/Video

N/A

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added useEditorDOMElement() hook to expose the editor DOM element to React components.
  * Updated popovers, toolbars, suggestion menus, and keyboard handlers to use the new hook for event wiring and positioning.
  * Editor headless status now follows the editor mount/unmount lifecycle.
  * onMount() and onUnmount() now return unsubscribe handles for easier listener cleanup.


packages/react/src/editor/BlockNoteView.tsx
nperez0111, line 1: I'd really prefer not to do this if at all possible. This causes jank for everyone else. I'd prefer something at the read-side (e.g. making the reading of the domElement reactive, or making the component that needs this element on render to remount). But remounting the whole editor is way too much

nperez0111 (changes requested):
I'd prefer a different solution here, this is too much. Especially for an issue we can't reliably reproduce

nperez0111 (approved):

PR #2592, opened 2026-03-24
fix: add link on commented text removes the comment highlight (2573)

miadnguyen:
# Summary
This pull request fixes an issue where adding or editing a link on commented text removes the comment highlight and comment association. After this change, a link can be added or edited on commented text while preserving both the link underline and the existing comment behavior.

## Rationale
The issue was caused by `LinkToolbarExtension.editLink()` (from packages/core/src/extensions/LinkToolbar/LinkToolbar.ts) replacing the selected text with `insertText(...)` and then reapplying only the `link` mark. As a result, existing marks on that text, including the `comment` mark, were lost. Preserving existing non-link marks makes link editing behave correctly and keeps comment functionality intact.

## Changes
- Updated `LinkToolbarExtension.editLink()` to preserve existing marks on the selected text before replacing it
- Reapplied existing non-link marks after `insertText(...)`
- Reapplied the updated `link` mark after restoring the preserved marks

## Impact
This change affects the link editing flow for text that already contains other marks, especially comments. It should improve behavior by preserving comment highlighting and comment accessibility when links are added or edited. The intended behavior for normal link editing remains the same.

## Testing
- Manually verified: added a link on commented text preserves the comment highlight
- Manually verified: editing a link on bold text preserves the bold mark  
- Added Playwright e2e test in `tests/src/end-to-end/linktoolbar/linktoolbar.test.ts` 
  verifying that bold marks are preserved when editing a link
- Existing tests pass (pre-existing snapshot failures are unrelated platform differences)

## Screenshots/Video
Below is an image showing commented text retaining the comment, highlight, and link after the fix.
image

## Checklist

- [x] Code follows the project's coding standards.
- [x] Unit tests covering the new feature have been added.
- [x] All existing tests pass.
- [ ] The documentation has been updated to reflect the new feature

## Additional Notes
Fixes #2573 

## Summary by CodeRabbit

* **Bug Fixes**
  * Link editing now preserves other text formatting (bold, italic, etc.) when updating a link and ensures editor focus happens after changes are applied.

* **Tests**
  * Added an end-to-end test that verifies link insertion, editing, and preservation of text formatting (e.g., bold) after updating a link.

PR #2620, opened 2026-04-02
fix: Inserting link removes comment & add comment button click buggy

### LINKED ISSUE #2257: Floating comment button stops working for non-editors
**Describe the bug**
The comment button in the formatting toolbar stop appearing when the BlockNoteView `editable` state becomes false.

**To Reproduce**
You can reproduce this in the latest stackblitz from your demo: https://stackblitz.com/github/TypeCellOS/BlockNote/tree/main/examples/07-collaboration/06-comments-with-sidebar?file=src%2FApp.tsx

* The user starts as an editor. 
* Highlight text and see the floating bar.
* Switch to a commenter user
* Highlight text, and you will  not see any floating buttons (what I expect: still see the comment button)

Now, edit App.tsx and change 

```
const [activeUser, setActiveUser] = useState<MyUserType>(HARDCODED_USERS[0]);
```

to

```
const [activeUser, setActiveUser] = useState<MyUserType>(HARDCODED_USERS[2]);
```

So that it starts with a commenter user.  

`Reload the page`

After reloading, the initial commenter user DOES see the add comment button when highlighting text. If you switch to an editor user and back to a commenter, the add comment button does not appear again.

**Expected Behavior**

The add comment button is always visible to all users even if the editable state is false.

**Misc**

This may be affected by `reactStrictMode` -- when I set it to `false` in my config this issue springs up more often. Setting it to true fixes the issue in some of my own cases.

### LINKED ISSUE #2573: Add link on commented text removes the comment highlight
### What’s broken?

Add link on commented text removes the comment highlight

Highlight should be kept and the link (text undeerlined) should appear

### Steps to reproduce

https://github.com/user-attachments/assets/28931ef5-2652-49b6-a1a7-923c4839ce9b

### BlockNote version

_No response_

### Environment

https://www.blocknotejs.org/demo#et2zn

### Additional context

_No response_

### Contribution

### Sponsor

matthewlipski:
# Summary

## Link on top of comment fix

When adding a link, if the end aligns with the end of a comment, the comment will be removed across the content spanned by the link.

This is because adding a link calls `tr.insertText`, which replaces the content spanned by the link. This is a problem especially with comments as they are not `inclusive` in their TipTap mark spec. Therefore, when content that they cover gets replaced, they get cleared instead of preserved.

This PR makes it so that `tr.insertText` only gets called if the text has actually changed, which doesn't happen in most cases. If it does happen, there are a bunch of edge cases that have to be considered, e.g.:

Given the following content, where {} marks the span of a link and [] marks the span of a comment:

```
1{23[456}78]9
```

If the link's text is changed to something completely different, like "qwe", where should the comment start?

Therefore, this PR only covers the case where the text remains the same.

Closes #2573

## Add comment button fix

This PR also adds a fix to the add comment button in the formatting toolbar, as I was getting frustrated with it while working on the other issue. Clicking the button doesn't open the thread creation popup the first time, only the second time. This is because calling `showSelection` updating decorations will trigger a selection change event if the editor isn't focused, causing the popup to instantly close.

Fortunately, this is solved by focusing the editor DOM element before `showSelection` is called.

Closes #2257 

## Rationale

It's annoying for users to accidentally clear comments when trying to add links.

## Changes

- Made `tr.insertText` get called conditionally in `createLink`.
- Made editor DOM element receive focus before `showSelection` is called, when the add comment button is pressed.

## Impact

N/A

## Testing

- Added e2e test

## Screenshots/Video

N/A

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Avoid unnecessary link text replacement when the visible text is unchanged.

* **New Features**
  * Added a new "Comments Testing" example project with playground config and demo app.
  * Added a dedicated example entry in the playground.

* **Refactor**
  * Comments composer updated to support flexible schema types and improved editor interactions.

* **Tests**
  * Added Playwright end-to-end tests for comments and linking behavior.

* **Documentation**
  * Added README and example assets for the comments testing scenario.


nperez0111 (approved):
I'm okay with this as a stop-gap solution for right now, only because it is so simple. But, I would want more out of this later. But, that would come at the cost of higher complexity in comments. Let's just fix the user-facing issue for now

packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
```diff
@@ -88,7 +88,12 @@ export const LinkToolbarExtension = createExtension(({ editor }) => {
         if (!range) {
           return;
         }
-        tr.insertText(text, range.from, range.to);
+
```
nperez0111, line 91: should actually probably use style manager.createLink here no?

nperez0111:
@matthewlipski you may want to review this implementation too, to see which is worth keeping: https://github.com/TypeCellOS/BlockNote/pull/2592 test case might be good to take from there

packages/core/src/extensions/LinkToolbar/LinkToolbar.ts
matthewlipski: Not exactly, since `editLink` takes an optional position whereas `createLink` always uses the selection start position to insert the link. This is necessary for hovered links, as the editor selection isn't related to the link position in this case.

We could ofc just update `createLink` to take an optional position/range, but because it's part of the editor API, I don't think we want to have anything related to ProseMirror positions there. Ideally this would be solved when we finish our own locations API.

matthewlipski:
> @matthewlipski you may want to review this implementation too, to see which is worth keeping: #2592 test case might be good to take from there

Yea I originally had the same idea, i.e. just copying existing marks to the new text, like in #2592. Due to the edge cases involved when replacing text though (see PR description), I decided to omit that. Basically, since we can't know the user intent either way, I opted for the less complex solution.

Will add the test case though!

PR #2623, opened 2026-04-03
feat: simplify links by inlining it to BlockNote

nperez0111:
- **refactor: remove dep on @tiptap/extension-link and linkify**
- **refactor: simplify inlined Link extension by removing unused tiptap options**

# Summary

The goal of this change is that `@tiptap/extension-link` has proved to be somewhat problematic in the past for a couple of reasons:
 - It depends on linkify, which is a fairly heavy dep since it ships a whole parser & state machine for link detection. Which has awkward initialization semantics
 - The link extension is very configurable in Tiptap since there are many different use cases, which we do not expose at the BlockNote level

## Rationale

Given these, I felt it was appropriate to just inline the extension into BlockNote & write a simpler regex based matcher for link detection. It will not be quite as full featured as linkify was, but it is an order of magnitude simpler & lighter.

## Changes

 - Inlines `@tiptap/extension-link` into BlockNote core. Simplfies interface since we don't need all the configuration options it had
 - Removes dependency on `linkify` since it was a heavyweight dep, and somewhat inappropriate for this use-case anyway

## Impact
 - This can possibly lead to more things being linked that weren't before, or something not being linked that would have before. This is mitigated by test cases that were taken based on the existing linkify behaviors.
 - Adds some complexity & maintenance that is our responsibility now, at the trade off of more control over the package to make changes

## Testing

 - Added tests to understand the current linkfiy behavior, and built an algo to mimic the same behaviors that were captured there.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Robust in-editor autolinking while typing, improved URL/email detection/tokenization, paste-to-link, and click-to-open link behavior
  * New Link mark with href/target/rel handling and safer URI validation
  * Editor link API: get link at position, edit link in-place, delete link

* **Tests**
  * Comprehensive unit and integration tests covering detection, tokenization, autolink, paste handling, and editor flows

* **Chores**
  * Removed external link extension dependency and updated Dependabot configuration


matthewlipski:
There seems to be a regression with this r.e. custom inline content with styled text. On the initial render, everything looks fine, but the text within the custom inline content element gets cleared immediately after.

https://github.com/user-attachments/assets/3f0f5d17-3b23-4eb9-8ef3-c3502f5f3bc9

docs/content/docs/features/blocks/inline-content.mdx
```diff
@@ -79,6 +79,55 @@ type Link = {
 };
 ```
 
+### Customizing Links
```
YousefED, line 82: can we also make sure this is reflected correctly in blocknote initialization options?

Looking at it, that part is actually currently broken in docs: https://www.blocknotejs.org/docs/reference/editor/overview#options

(maybe check for other AutoTypeTables that are broken as well)

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -1,4 +1,4 @@
-import { Mark, Node, Schema, Slice } from "@tiptap/pm/model";
+import { Node, Schema, Slice } from "@tiptap/pm/model";
```
YousefED: @nperez0111 I'm missing some context as to why this file was changed. Should not be directly related to pulling in the tiptap link extension, right?

packages/core/src/editor/managers/ExtensionManager/extensions.ts
```diff
@@ -80,22 +73,9 @@ export function getDefaultTiptapExtensions(
...
-          delete (attrs as Record<string, unknown>).title;
-          return attrs;
-        },
-      })
-      .configure({
-      defaultProtocol: DEFAULT_LINK_PROTOCOL,
-      // only call this once if we have multiple editors installed. Or fix https://github.com/ueberdosis/tiptap/issues/5450
-      protocols: LINKIFY_INITIALIZED ? [] : VALID_LINK_PROTOCOLS,
+    Link.configure({
+      HTMLAttributes: options.links?.HTMLAttributes ?? {},
```
YousefED, line 77: What's the new default? No attributes?

If we change from the previous defaults (https://github.com/ueberdosis/tiptap/blob/a4b939127821aac3a2139a467c5b257dd897da4e/packages/extension-link/src/link.ts#L237), I think it's a breaking change. Rather keep previous defaults?

packages/core/src/extensions/tiptap-extensions/Link/helpers/clickHandler.ts
```diff
@@ -0,0 +1,74 @@
...
+            link = null;
+          }
+        }
+
+        if (!link) {
+          return false;
+        }
+
+        if (options.onClick) {
+          const result = options.onClick(event);
```
YousefED: should we pass `link` to the handler?

packages/core/src/api/nodeConversions/nodeToBlock.ts
nperez0111: reverted, will do another pass later

packages/core/src/editor/managers/ExtensionManager/extensions.ts
matthewlipski: The additional `target="_blank"` and `rel="noopener noreferrer nofollow"` attributes are set on links automatically by ProseMirror, unless you override them by passing alternative values in `HTMLAttributes`. So the defaults are the same.

packages/core/src/extensions/tiptap-extensions/Link/helpers/clickHandler.ts
matthewlipski: I don't think the `Link` object is super useful tbh, but adding the BlockNote editor as a second param to the handler makes sense. Since that exposes all the functions that relate to the link, i.e. `getSelectedLinkUrl`, `editLink`, and `deleteLink`.

matthewlipski (approved):

PR #2628, opened 2026-04-04
fix: Drag & drop of blocks without inline content opens formatting toolbar (BLO-1116)

### LINKED ISSUE #2603: Move an image doesn't work with drag and drop
### What’s broken?

When I try to move a block containing an image using drag and drop I get 2 errors
- the toolbar opens (not my intention, I didn't click on it)
- the image is not dropped to where it was highlighted (in light blue). It stays in its location
Currently the only way I can move images is by 
- using the keyboard shortcut to move  a block ctrl + maj + up / down
- or ctrl + x and the ctrl + v

I would expect the image toolbar to open only when hovering over it when on desktop and taping on it on mobile.
it should move to the highlighed location in light blue.

### Steps to reproduce

https://github.com/user-attachments/assets/8742433b-34ba-4400-9462-9816c469f0a2

### BlockNote version

_No response_

### Environment

firefox 148 en NixOS 64bit

### Additional context

tested on https://www.blocknotejs.org/demo#rlgys

### Contribution

### Sponsor

miadnguyen:
# Summary
Fixes #2603. When dragging an image block using the side menu drag handle, the formatting toolbar incorrectly opens mid-drag.

## Rationale
The formatting toolbar should only appear when the user clicks on the image or during text selection, not during a drag-and-drop operation. The unexpected toolbar appearance is disruptive.

## Changes
- Added a `preventShowWhileDragging` flag in `FormattingToolbar.ts` that mirrors the existing `preventShowWhileMouseDown` pattern
- Registered `dragstart` and `dragend` listeners on `editor.prosemirrorView.root` to set and clear this flag for the full duration of any drag operation
- Removed existing check to not show if block with inline content is selected.
- Added a Playwright end-to-end test verifying the toolbar does not appear while dragging an image block
- Moved one existing test from `images` to `dragdrop`

## Impact
The toolbar is now suppressed during any block drag operation. Normal toolbar behavior is unaffected.

## Testing
- Manually verified fix on Chrome and Firefox
- Added Playwright test in `tests/src/end-to-end/images/images.test.ts`: "Formatting toolbar should not appear when dragging image block"
- Existing delete image test continues to pass

## Screenshots/Video
https://github.com/user-attachments/assets/db9f713b-6a82-4e44-b5d6-0b802cbe14a0

## Checklist

## Additional Notes
When the drag handle is used, ProseMirror sets a `NodeSelection` on the image node before the drag begins. The existing `preventShowWhileMouseDown` flag did not cover HTML drag events (`dragstart`/`dragend`), only pointer events, so the toolbar was not suppressed during drags initiated from the side menu.

## Summary by CodeRabbit

* **Bug Fixes**
  * Formatting toolbar now stays reliably hidden during drag interactions (including drag-handle drags) and remains hidden after pointer cancellation to avoid flicker or unwanted appearance.

* **Tests**
  * End-to-end tests updated: added coverage verifying image-block drag behavior and that the formatting toolbar is not shown during/after dragging; one older drag test was replaced/streamlined.


matthewlipski:
This is great, thanks! With your fix, I've also removed the check in `shouldShow` we used before to prevent the formatting toolbar from showing up on drop for blocks with inline content, since your solution works for all blocks.

nperez0111:
Thanks @miadnguyen, we really appreciate your contribution. Should be out sometime early next week

PR #2630, opened 2026-04-07
fix: hide side menu on scroll instead of overflow hacks

YousefED:
# Summary

Hide the side menu when the user scrolls, preventing it from overflowing outside the editor's scroll container (e.g. in docs demos). This reverts the CSS workarounds introduced in #2043 in favor of a proper floating-ui–based solution.

## Rationale

PR #2043 added `overflow: none` and `position: relative` hacks to the docs demo containers to prevent the side menu from visually escaping the demo box. This approach was fragile and didn't address the root cause. The proper fix is to hide the side menu on scroll — the same UX pattern used for keyboard input.

We prefer to handle this on the floating-ui layer instead of adding additional event listeners in the SideMenu extension, as we're trying to refactor more logic out of our custom extensions and into standard floating-ui patterns.

## Changes

- **`docs/app/styles.css`** — Removed `.demo { overflow: none }` and `.demo .bn-container { position: relative }` from #2043
- **`GenericPopover.tsx`** — Added `mergeWhileElementsMounted` helper that composes the default `autoUpdate` with any additional `whileElementsMounted` handler provided via props
- **`SideMenuController.tsx`** — Added a `whileElementsMounted` handler using `autoUpdate` with `ancestorScroll: true` to detect scroll and call `hideMenuIfNotFrozen()`
- **`SideMenu.ts`** — Added `hideMenuIfNotFrozen()` extension method that hides the menu only when it's not frozen (i.e. drag handle submenu is not open)

## Impact

- Side menu now hides on scroll in any context (not just docs demos)
- No impact on drag handle submenu — `hideMenuIfNotFrozen` respects the frozen state
- `GenericPopover`'s new merge behavior is backwards-compatible: when no `whileElementsMounted` is passed, behavior is identical to before

## Testing

- Manual testing: verified side menu hides on scroll in docs demos and playground
- Verified drag handle submenu stays open during scroll (frozen state respected)

## Checklist

## Additional Notes

Reverts CSS changes from https://github.com/TypeCellOS/BlockNote/pull/2043.

## Summary by CodeRabbit

* **New Features**
  * Side menu now automatically hides when not frozen during interactions.
  * Table handles (and related controls) now auto-hide when not frozen during interactions.

* **Refactor**
  * Improved Floating UI lifecycle integration for more stable popovers and handle positioning.

* **Style**
  * Simplified demo container styling (removed certain overflow/position rules), which may affect demo layout/stacking.

### DIFF
--- a/docs/app/styles.css
+++ b/docs/app/styles.css
@@ -53,14 +53,6 @@ body {
   box-shadow: unset !important;
 }
 
-.demo {
-  overflow: none;
-}
-
-.demo .bn-container {
-  position: relative;
-}
-
 .demo .bn-container:not(.bn-comment-editor),
 .demo .bn-editor {
   height: 100%;
--- a/packages/core/src/extensions/SideMenu/SideMenu.ts
+++ b/packages/core/src/extensions/SideMenu/SideMenu.ts
@@ -784,5 +784,17 @@ export const SideMenuExtension = createExtension(({ editor }) => {
       view!.state!.show = false;
       view!.emitUpdate(view!.state!);
     },
+
+    /**
+     * Hides the side menu unless it is currently frozen (e.g. the drag
+     * handle menu is open). Used to dismiss the menu on scroll without
+     * interfering with open submenus.
+     */
+    hideMenuIfNotFrozen() {
+      if (!view!.menuFrozen) {
+        view!.state!.show = false;
+        view!.emitUpdate(view!.state!);
+      }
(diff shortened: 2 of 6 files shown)


nperez0111 (approved):
using floating ui for the scroll dismiss seems like a bit of an abuse, but it works

nperez0111:
table handles would likely want the same fix applied, but it is super minor

PR #2659, opened 2026-04-18
feat: add emoji picker option

### LINKED ISSUE #808: Integrate emoji picker

thegreatalxx:
Adds an emoji picker option to the slash menu. Fixes TypeCellOS/BlockNote#808.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for emoji suggestion menu item functionality, verifying that selecting an emoji item correctly triggers the suggestion menu with appropriate configuration parameters.

### DIFF
--- a/packages/core/src/extensions/SuggestionMenu/SuggestionMenu.test.ts
+++ b/packages/core/src/extensions/SuggestionMenu/SuggestionMenu.test.ts
@@ -1,6 +1,7 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 
 import { BlockNoteEditor } from "../../editor/BlockNoteEditor.js";
+import { getDefaultSlashMenuItems } from "./getDefaultSlashMenuItems.js";
 import { SuggestionMenu } from "./SuggestionMenu.js";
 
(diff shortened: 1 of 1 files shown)


nperez0111 (commented):
This is just adding a test case, what is this meant to do?

PR #2661, opened 2026-04-20
fix(markdown): avoid autolink angle brackets when copying plain URL l…

Ovgodd:
## Purpose

Fix broken links when copying a URL from a BlockNote document and pasting it
into another input (link toolbar).

When copying content from the editor, BlockNote serializes links to
`text/plain` using Markdown. Plain URL links (where text = URL) were
serialized as Markdown autolinks (`<https://example.com>`), causing the angle
brackets to be included wherever the text was pasted, resulting in invalid
hrefs.

## Proposal

* [x] Add a custom `link` handler in `remarkStringify` inside
  `cleanHTMLToMarkdown` to output plain URLs instead of Markdown autolink
  format when the link text equals the URL
* [x] When the link has a distinct label, the standard Markdown format
  `[label](url)` is preserved

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Markdown export link handling to avoid autolink-style rendering when link text equals the URL.
  * Descriptive links continue to export as standard Markdown ([text](url)) for readability.
  * URLs are now escaped when needed to ensure links render correctly in parentheses and other contexts.

### DIFF
--- a/packages/core/src/api/exporters/markdown/markdownExporter.ts
+++ b/packages/core/src/api/exporters/markdown/markdownExporter.ts
@@ -33,7 +33,22 @@ export function cleanHTMLToMarkdown(cleanHTMLString: string) {
     .use(deps.rehypeRemark.default)
     .use(deps.remarkGfm.default)
     .use(deps.remarkStringify.default, {
-      handlers: { text: (node) => node.value },
+      handlers: {
+        text: (node) => node.value,
+        // Prevent autolink format <url> output plain URL instead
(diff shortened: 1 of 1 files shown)


nperez0111:
Thanks @Ovgodd we've incorporated your suggestion with https://github.com/TypeCellOS/BlockNote/pull/2624

Ovgodd:
> Thanks @Ovgodd we've incorporated your suggestion with #2624

nice, my pleasure.

PR #2666, opened 2026-04-21
feat: Link customization options (BLO-913)

### LINKED ISSUE #1539: Click on links are not intercepted
**Describe the bug**
When clicking on a link, the link is directly followed, making it easy to move out of the editor accidentally.

**To Reproduce**
* Write a text and insert a link
* Click on the link

*Expected*: a toolbar is displayed, with actions edit link, remove link, follow link
*Actual*: the user is navigating to the linked page

Note: while it's difficult to add link on mobile, the issue described above does not reproduce.

matthewlipski:
# Summary

This PR adds a new editor option:

```
links: Partial<{
  HTMLAttributes: Record<string, any>;
  onClick?: (event: MouseEvent) => void;
}>
```

These do basically what they say - `HTMLAttributes` adds HTML attributes to rendered link elements and `onClick` replaces the default click behaviour (which opens the link in a new tab).

Closes #1539 

## Rationale

Some users are finding it annoying that links open a new tab on click when they're just trying to move the selection. 

HTML attributes allow for slight customization for link rendering. It's the best we can do atm, but really more of a stopgap solution as consumers should ideally be able to override the default link rendering with whatever they want.

## Changes

See above.

## Impact

N/A

## Testing

N/A (example needed?)

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

### DIFF
--- a/packages/core/src/extensions/tiptap-extensions/Link/helpers/clickHandler.ts
+++ b/packages/core/src/extensions/tiptap-extensions/Link/helpers/clickHandler.ts
@@ -6,7 +6,7 @@ import { Plugin, PluginKey } from "@tiptap/pm/state";
 type ClickHandlerOptions = {
   type: MarkType;
   editor: Editor;
-  onClick?: (event: MouseEvent) => void;
+  onClick?: (event: MouseEvent) => boolean | void;
 };
 
 export function clickHandler(options: ClickHandlerOptions): Plugin {
@@ -48,8 +48,8 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
         }
 
         if (options.onClick) {
-          options.onClick(event);
-          return true;
+          const result = options.onClick(event);
+          return result ?? true;
         }
 
         const attrs = getAttributes(view.state, options.type.name);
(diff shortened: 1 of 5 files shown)


packages/core/src/extensions/tiptap-extensions/Link/helpers/clickHandler.ts
```diff
@@ -46,6 +47,11 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
           return false;
         }
 
+        if (options.onClick) {
+          options.onClick(event);
+          return true;
```
nperez0111: Should we let the consumer decide the return value? i.e. undefined we interpret as true, but false we return false

nperez0111 (approved):
This looks right overall though!

packages/core/src/extensions/tiptap-extensions/Link/helpers/clickHandler.ts
matthewlipski: Yeah I think that makes sense

PR #2677, opened 2026-04-25
fix(clipboard): use ProseMirror selection state for Shadow DOM compatibility

wielinde:
## Problem

BlockNote embeds ProseMirror inside editors that may be mounted in a Shadow DOM (e.g. OpenProject uses `attachShadow({ mode: 'open' })` to isolate the editor from its Angular host). In this setup, `window.getSelection()` returns `null` or a collapsed selection even when text is visually selected — this affects Firefox (all versions), Safari ≤16.3, and Chromium edge cases.

`checkIfSelectionInNonEditableBlock` used `window.getSelection()` as its primary empty-selection guard. Because `getSelection()` misfires in Shadow DOM, the guard always returned `true`, causing both the `copy` and `cut` handlers to bail out early without writing to the clipboard. The browser's default copy then ran, which uses ProseMirror's `DOMSerializer` without BlockNote's semantic wrappers — losing list formatting, headings, and bold/italic on paste into external apps.

## Fix

Use `view.state.selection.empty` as the primary guard. ProseMirror's internal selection state is updated via its own reconciliation loop and is always accurate regardless of DOM mode.

The non-editable-block DOM walk (which does need `window.getSelection()`) is kept as a secondary guard, but only runs when `getSelection()` actually returns a non-collapsed selection — so it's a no-op in Shadow DOM environments where `getSelection()` is unreliable.

## Test

Verified in OpenProject (Shadow DOM setup) with a bullet list: the clipboard now contains all three expected types — `blocknote/html`, `text/html` (with `<ul>/<li>`), and `text/plain` (markdown `* item`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard handling for copy and cut operations in non-editable sections and shadow DOM scenarios, ensuring more reliable selection detection across different browser conditions.

### DIFF
--- a/packages/core/src/api/clipboard/toClipboard/copyExtension.ts
+++ b/packages/core/src/api/clipboard/toClipboard/copyExtension.ts
@@ -145,11 +145,13 @@ export function selectedFragmentToHTML<
   return { clipboardHTML, externalHTML, markdown };
 }
 
-const checkIfSelectionInNonEditableBlock = () => {
-  // Let browser handle event if selection is empty (nothing
-  // happens).
-  const selection = window.getSelection();
-  if (!selection || selection.isCollapsed) {
+const checkIfSelectionInNonEditableBlock = (view: EditorView) => {
+  // Use ProseMirror's internal selection state to check for empty selection.
+  // window.getSelection() returns null or a collapsed selection inside Shadow
+  // DOM (Firefox, Safari, and Chromium edge cases), causing this guard to
+  // misfire and silently skip clipboard writes. view.state.selection is always
+  // accurate regardless of DOM mode.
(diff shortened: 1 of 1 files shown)


matthewlipski:
Hi @wielinde, this looks good! Could you just approve [this PR](https://github.com/opf/BlockNote/pull/2)? It just adds a regression test - I don't have permissions to edit your branch directly so this should be fine.

nperez0111:
@matthewlipski, let's just merge this PR & you can fix up in a follow up.

For next time, it would be nice to give edit rights to the PR @wielinde

wielinde:
Thank you 🙏 That is amazing!

...and yes, note taken for next time!

PR #2679, opened 2026-04-27
fix: align side menu drag handle with table top (#2604)

### LINKED ISSUE #2604: Drag handle is not aligned on tables
### What’s broken?

The drag handle is slightly mis positionned, it floats a little bit higher that the table.

Image

The drag handle should be horizontally aligned with the table.

### Steps to reproduce

1. hover on table

### BlockNote version

_No response_

### Environment

Firefox 146 on NixOS

### Additional context

Tested here https://www.blocknotejs.org/demo#xpbok

### Contribution

### Sponsor

betsyschultz:
# PR Draft — fix: align side menu drag handle with table top (#2604)

> Open against `TypeCellOS/BlockNote:main` from `fix/2604-drag-handle-table-alignment`.
> Title: **fix: align side menu drag handle with table top (#2604)**

---

## What

Closes #2604. The side menu drag handle (and the `+` add button next to it) now align with the visible top of a table block instead of floating ~8px above it.

## Why it was broken

`SideMenuController` uses Floating UI with `placement: "left-start"`, which anchors the popover's top edge to the reference element's top edge. The reference is the block's outer wrapper (`bn-block-content`).

For text blocks (paragraph, heading, list, code) this looks correct because the inline content sits flush at the top of the wrapper.

For table blocks the DOM is:

```
.bn-block-content[data-content-type="table"]   ← Floating UI reference
  .tableWrapper                                ← padding-top: 9px (--bn-table-handle-size)
    .tableWrapper-inner
      <table>                                  ← what the user actually sees
```

`.tableWrapper`'s top padding is intentional — it reserves vertical space for the in-block row handles (the small drag dots that appear above each row on hover). That padding pushes the visible `<table>` element 9px below the wrapper top, but the side menu had no awareness of it.

Measured before fix (initial empty editor, fresh table): drag handle top = `12px`, table element top = `20px` → 8px gap.

## The fix

Add a Floating UI `offset` middleware that, when the reference element contains a direct `.tableWrapper` child, shifts the popover down by that wrapper's computed `padding-top`. The middleware is a no-op for any reference without a `.tableWrapper` child, so non-table blocks are unaffected.

The padding value is read from `getComputedStyle` rather than hardcoded so that the fix tracks the `--bn-table-handle-size` CSS variable automatically.

After fix: drag handle top = `21px`, table element top = `20px` → 1px gap (matches paragraph alignment, where the handle SVG is intentionally 1px below the inline content top).

## Files changed

- `packages/react/src/components/SideMenu/SideMenuController.tsx` (+22 / −0)
- `tests/src/end-to-end/draghandle/draghandle.test.ts` (+24 / −1)

## Test plan

- [x] Added regression test in `draghandle.test.ts` — asserts `Math.abs(handleY - tableY) <= 5`. Fails on `main` (8px gap), passes with the fix (1px gap).
- [x] `pnpm exec tsc --noEmit` clean for `@blocknote/react`
- [x] `nx lint @blocknote/react` clean (`--max-warnings 0`)
- [x] `nx build @blocknote/react` succeeds
- [x] Manual repro: paragraph hover → handle still aligned (no regression). Table hover → handle now aligned with table top.

### What was NOT run locally

The full E2E Playwright suite (`tests/`) was not run locally — relies on CI for the cross-browser snapshot suite. The added regression test follows the same conventions as the other tests in `draghandle.test.ts` (uses `executeSlashCommand`, `moveMouseOverElement`, `boundingBox()`).

### Edge cases considered

| Case | Status |
|------|--------|
| Paragraph / heading / list (no regression) | Verified — middleware is no-op without `.tableWrapper` |
| Table as the only block | Verified visually |
| Table after a paragraph | Same code path, same result |
| Multi-column block containing a table | `:scope > .tableWrapper` (direct child only) prevents an outer block's reference from picking up a nested table's wrapper |
| Future block types adding their own widget padding | Pattern is reusable — any block whose direct child is `.tableWrapper` (or by extension a similarly named wrapper) gets correct alignment automatically |

## Screenshots

After fix (drag handle aligned with top of table):

after-fix

(Run the playground at `pnpm dev` → `/basic/minimal` → `/table` Enter, then hover the table to reproduce.)

## Notes for reviewer

- The middleware uses a CSS-attribute-driven approach rather than a per-block-type switch, so it stays correct if `--bn-table-handle-size` is themed.
- `tableWrapperOffset` is exported from `@blocknote/react` so consumers who pass their own `middleware: [...]` via `floatingUIOptions.useFloatingOptions` can re-compose it: `middleware: [offset(tableWrapperOffset), ...other]`. Without this, customising `middleware` would silently overwrite the fix (this matches the existing spread pattern in `FloatingComposerController` etc., so the footgun isn't new — but the helper makes it recoverable).
- The fix assumes `placement: "left-start"` (the default for the side menu). Consumers who change placement keep the offset, but `crossAxis` may map to a different visual axis depending on placement; documented in the helper's JSDoc.
- If you'd rather solve this at the layout layer (e.g., move `tableWrapper`'s widget-spacing from padding to absolute-positioned widgets), that's a larger architectural change — let me know and I can spin a separate PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Side menu now aligns correctly with the visible top of table content, accounting for table padding so the menu appears adjacent to rows.
  * Improved drag-handle positioning on table elements; handles now match the table’s visible top within a tight tolerance.
* **Tests**
  * Added an end-to-end regression test to verify drag-handle positioning for tables.

### DIFF
--- a/packages/react/src/components/SideMenu/SideMenuController.tsx
+++ b/packages/react/src/components/SideMenu/SideMenuController.tsx
@@ -101,8 +101,11 @@ export const SideMenuController = (props: {
         open: show,
         placement: "left-start",
         whileElementsMounted,
-        middleware: [offset(tableWrapperOffset)],
         ...props.floatingUIOptions?.useFloatingOptions,
+        middleware: [
+          offset(tableWrapperOffset),
+          ...(props.floatingUIOptions?.useFloatingOptions?.middleware ?? []),
+        ],
       },
       useDismissProps: {
         enabled: false,
--- a/packages/react/src/components/SideMenu/SideMenuController.tsx
+++ b/packages/react/src/components/SideMenu/SideMenuController.tsx
@@ -23,10 +23,21 @@ import { SideMenuProps } from "./SideMenuProps.js";
  * that reserves space for row/column handles) — without this, the side menu
  * floats above the visible top of the table.
  *
- * Assumes `placement: "left-start"`. Other placements map `crossAxis` to a
- * different visual axis.
+ * Only active for left/right placements (where `crossAxis` maps to vertical).
+ * Returns 0 for top/bottom placements to avoid horizontal misalignment if the
+ * export is reused outside its intended placement context.
+ *
+ * Note on composition: this is added unconditionally by `SideMenuController`.
+ * Consumer-supplied middleware is appended after, so any additional `offset(...)`
+ * from a consumer stacks with this one (Floating UI's offset middlewares are
+ * additive). For non-table blocks `tableWrapperOffset` returns 0, so the stack
+ * is a no-op there; for table blocks the compensation always applies.
  */
 export const tableWrapperOffset: OffsetOptions = (state) => {
+  const side = state.placement.split("-")[0];
+  if (side !== "left" && side !== "right") {
+    return 0;
+  }
   const { reference } = state.elements;
   const refEl =
     reference instanceof Element ? reference : reference.contextElement;
--- a/packages/react/src/components/SideMenu/SideMenuController.tsx
+++ b/packages/react/src/components/SideMenu/SideMenuController.tsx
@@ -1,5 +1,10 @@
 import { SideMenuExtension } from "@blocknote/core/extensions";
-import { autoUpdate, ReferenceElement } from "@floating-ui/react";
+import {
+  autoUpdate,
+  offset,
+  OffsetOptions,
+  ReferenceElement,
+} from "@floating-ui/react";
 import { FC, useCallback, useMemo } from "react";
 
 import { useBlockNoteEditor } from "../../hooks/useBlockNoteEditor.js";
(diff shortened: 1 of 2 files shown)


betsyschultz:
Thanks @coderabbitai — addressed two of the three nitpicks in 84b7cc9:

**1. Middleware override (SideMenuController.tsx:104)** — fixed. The default `offset(tableWrapperOffset)` is now composed with consumer-provided middleware instead of being silently overwritten:

```tsx
useFloatingOptions: {
  ...
  ...props.floatingUIOptions?.useFloatingOptions,
  middleware: [
    offset(tableWrapperOffset),
    ...(props.floatingUIOptions?.useFloatingOptions?.middleware ?? []),
  ],
}
```

Worth noting the same latent bug exists in `FloatingThreadController`, `FloatingComposerController`, `FilePanelController`, `LinkToolbarController`, and `FormattingToolbarController` — happy to open a follow-up PR if maintainers want it fixed across the board, or leave it scoped to this fix.

**2. Test tolerance (draghandle.test.ts:60)** — tightened from 5px → 2px. The fix subtracts the exact `tableWrapper` padding value, so the post-fix delta is sub-pixel by construction; 2px is comfortable headroom for cross-browser rounding without inviting flake.

**3. Non-null assertion on `boundingBox()`** — skipped. The inline `(await el.boundingBox())!` pattern is the established idiom across this test suite (`colors.test.ts`, `images.test.ts`, `copypaste.test.ts`, `indentation.test.ts`, `ai-selection.test.ts`, `keyboardhandlers.test.ts`, `tests/src/utils/mouse.ts`, `tests/src/utils/draghandle.ts`). Adding an explicit `expect().not.toBeNull()` only here breaks pattern consistency without functional benefit — the `!` already encodes the assertion. Happy to revisit if you'd prefer the broader test suite move to explicit checks.

betsyschultz:
Round 2 — addressed one, declined the other with reasoning. Pushed in a3625bd.

**Placement guard (lines 29–45)** — added. The export now no-ops for non-left/right placements where `crossAxis` doesn't map to vertical. Three-line guard, defensive insurance for consumers who reuse `tableWrapperOffset` outside its intended context. Good catch.

**Offset stacking (lines 105–108)** — declined; documented the contract instead.

Walking through the alternative: detecting `middleware.name === "offset"` and skipping the default if a consumer provides their own offset would silently evaporate the table compensation for that consumer — which is exactly the bug fixed in round 1. Different shape, same broken-tables outcome.

The current contract is intentional and now spelled out in the JSDoc:

- `tableWrapperOffset` returns 0 for non-table blocks → no-op stack on those.
- For table blocks, the structural compensation always applies. A consumer's additional `offset(N)` stacks with it (Floating UI's offset middlewares are additive), giving them `tableY + N` — i.e. their offset applied on top of the structural fix, which is the correct behavior for a fix that's about visible alignment, not styling preference.

Detecting and stripping the default would mean "consumer offset wins, tables break." Keeping the stack means "consumer offset composes with the structural fix, tables stay aligned." The latter is the right default for a regression fix.

Open to maintainer pushback on this one — happy to switch to detection if the project would rather treat consumer-supplied offset as a full override.

packages/react/src/components/SideMenu/SideMenuController.tsx
```diff
@@ -9,6 +14,47 @@ import { FloatingUIOptions } from "../Popovers/FloatingUIOptions.js";
...
+  }
+  const { reference } = state.elements;
+  const refEl =
+    reference instanceof Element ? reference : reference.contextElement;
+  // The side menu's reference is the block's outer `.bn-block` element.
+  // For tables, the `.tableWrapper` lives one level deeper inside
+  // `.bn-block-content`. Match that exact path so unrelated descendants
+  // (e.g. a nested table inside a multi-column block) don't trigger.
+  const wrapper = refEl?.querySelector(
+    ":scope > .bn-block-content > .tableWrapper",
```
nperez0111, line 49: We've tried to address this before with https://github.com/TypeCellOS/BlockNote/pull/2224 and unfortunately this also suffers from the same problem. We don't want for the sidemenu to have to become aware of each block's quirks and have to adjust for that. This referencing the tableWrapper is exactly the sort of thing we'd like to avoid

nperez0111 (changes requested):
Sorry but this won't meet our standards for a solution here

PR #2684, opened 2026-04-28
fix: Color naming & CSS (BLO-946)

### LINKED ISSUE #2005: Applying default color sometimes doesn't work
**Describe the bug**
Applying default color sometimes doesn't work

**To Reproduce**

https://github.com/user-attachments/assets/b8f4d587-cc8e-4738-8807-2d63f478b6f6

**Misc**
- Node version:
- Package manager:
- Browser: Firefox

matthewlipski:
# Summary

This PR fixes 2 issues:

1. The "Default" option for block text/background color can be misleading if a parent block already sets the color, and the current block/selection inherits from it. As the "Default" option will set the color to the inherited color, rather than the actual default.
2. CSS for block colors is not quite right, causing nested blocks to use the fallback colors rather than those from the editor theme.

Closes #2005 

## Rationale

These issues cause confusion for users.

## Changes

- Changed "Default" naming to "Auto" for colors in the dictionary for all locales.
- Fixed CSS for inherted colors in nested blocks.

## Impact

N/A

## Testing

N/A

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Style**
  * Updated color highlighting so styles apply to blocks when their inner content declares text/background colors, improving consistent rendering of colored blocks.
* **Localization**
  * Changed the color picker’s "default" label to "Auto" across 26+ languages.
* **Examples**
  * Updated sample markup to move color attributes from outer wrappers to inner block content to match the new highlighting rules.


nperez0111 (approved):
Looks good to me

packages/core/src/editor/Block.css
```diff
@@ -561,110 +561,92 @@ NESTED BLOCKS
 
 /* TEXT COLORS */
 [data-style-type="textColor"][data-value="gray"],
-[data-text-color="gray"],
```
nperez0111, line 564: @matthewlipski this seems to have caused this regression: https://github.com/TypeCellOS/BlockNote/issues/2760 we should fix this, I'm unsure if we should be taking the same approach as you did in https://github.com/TypeCellOS/BlockNote/pull/2762 when I think that this line is actually the right fix? Why was this removed again?

PR #2688, opened 2026-04-29
chore: Regression test for #2677

matthewlipski:
This PR adds an e2e test for TypeCellOS#2677 by copying content in a non-editable block and pasting it in the editor. A playground example is also added for this test.

## Summary by CodeRabbit

* **New Features**
  * Added a new "Non-Editable Block" example demonstrating how to create custom blocks with non-editable content using custom schemas. The example is now available in the playground.

* **Tests**
  * Added end-to-end tests to validate copy/paste functionality with non-editable blocks.


nperez0111 (approved):

YousefED:
@matthewlipski while this PR validates whether the original code (non-editable area copy / paste) still works, it does not test the shadow-dom scenario, right? 

More specifically, if I'm not mistaken (haven't tested this), with the code change from https://github.com/TypeCellOS/BlockNote/pull/2677, non-editable block copy _within_ a shadow dom would not work, right? This would mean BlockNote behaves differently in a shadow dom vs regular use. Is there a way to streamline this? e.g.: not rely on window.getSelection at all?


cc @nperez0111

PR #2697, opened 2026-04-30
feat: add container block API for nested blocks

nperez0111:
Adds first-class **container blocks**: custom blocks whose body is other blocks — a Notion-style callout wrapping a paragraph and a code block, a toggle with a title and a body, a multi-column layout, or a table. Declared with a `children` config on the block spec and exposed at runtime as `block.children`, the same field used for indented blocks, now with schema-enforced control over counts, allowed types, and rendering.

The smallest container is one line on any `createBlockSpec` / `createReactBlockSpec` config:

```ts
{
  type: "callout",
  propSchema: {},
  content: "none",
  children: { allow: "any" },
}
```

Child blocks mount into the element the render passes `contentRef` (React) / `contentDOM` (vanilla) to — the same placement mechanism as inline content. The block's JSON shape is unchanged: children live in the standard `children` array.

## API surface

- **`children: { allow, min, max, default, whenEmptied, boundary }`** on block configs.
  - `allow`: `"any"` | `"blocks"` | `"containers"` | an array of container block types. Wildcards never include `containerOnly` types; the array form is exact.
  - `min` / `max`: child counts, compiled into the editor schema.
  - `default`: insertion template used when a container is created without children, and the seed for refills.
  - `whenEmptied`: `"refill"` (top the container back up from `default`) or `"unwrap"` (replace the container with its survivors — how an emptied column list dissolves).
  - `boundary`: `"open"` | `"isolated"` | `"sealed"` — what crosses the container's edge (caret, editing gestures, text selections). `"sealed"` makes a compartment like a table cell with **no hand-written keyboard handlers**: Backspace/Delete/Enter/arrows never implicitly move content across the edge, while the block manipulation API always crosses seals as an intentional act.
- **`placement: "containerOnly"`** restricts a block to containers that name it in their `allow` array (e.g. `column` inside `columnList`).
- **Containers with their own content**: `children` combines with `content: "inline"` / `"plain"` (a toggle's title plus its body), rendered as `[data-content-type]` and `[data-children-of]` regions placed by the same single `contentRef`.
- **`insertBlocks` gains `"start"` / `"end"` placements** to insert into a container's children — addressing containers that currently have no child to point at.
- **Up-front schema validation**: impossible `allow`s, unknown or regular block types in `allow` arrays, invalid `default`s, orphaned `containerOnly` blocks, and container cycles all fail at schema creation with a message naming the block.
- **`meta.draggable`** lets structural blocks opt out of the side-menu drag handle, falling through to the nearest draggable ancestor.

## Interop

- Containers round-trip through internal HTML via `data-node-type`, with non-default props as `data-*` attributes.
- Foreign HTML is recognized via `parse` / `parseContent` on the block implementation; `runsBefore` orders container parse rules against each other (containers register in a priority band below regular blocks).
- External HTML stays semantic via `toExternalHTML` returning a `childrenDOM` (how a toggle exports as `<details>`).
- Markdown flattens containers; exporters (docx/PDF/ODT/email) require an explicit mapping and throw a clear error otherwise.

## Reimplementations & examples

- **`@blocknote/xl-multi-column`** (`columnList` / `column`) is reimplemented on this API, dropping its bespoke node implementations.
- **Container-block example**: a callout holding arbitrary blocks, plus the string-prop pattern for non-document fields.
- **Container-table example**: a stress test of the API — a working table built from four container specs (`table`, `tableRow`, `tableCell`, `tableHeader`) using only public APIs: sealed cell boundaries, `containerOnly` placement, Tab/Shift-Tab cell navigation via `createExtension` keyboard shortcuts, and row/column operations through `insertBlocks` / `removeBlocks` / `updateBlock`. Advanced table features (cell merging, column resizing, table handles) are intentionally out of scope; the point is proving the primitive carries the structural weight.

## Docs

New [Container Blocks](docs/content/docs/features/custom-schemas/container-blocks.mdx) page covering declaration, the two regions, boundaries, defaults/refilling, restricting children, validation, parsing, and interop; `manipulating-content` reference updated for the new insert placements.


nperez0111:
Superseded by the stacked split: #2997 (core container API) → #2998 (multi-column migration) → #2999 (docs & examples).

PR #2698, opened 2026-04-30
feat: Code mark input rule edge case (BLO-938)

### LINKED ISSUE #1970: Bug with inline code using backticks
**Describe the bug**

Image

When creating inline code with backticks the conversion work fine when doing : 
1. backtick 
2. writing code 
3. backtick
4. Conversion to inline code highlight.

But if you don't respect this order, the backticks are not removed and the first letter of your code is dropped, for example with
1. backtick
2. backtick
3. move back between the two with arrow left
4. write code
5. move out of code block with arrow right
6. conversion to inline code with space

**To Reproduce**
Here is a video showing the issue on the blocknotejs playground : https://www.loom.com/share/52096cb8f6194725b9e61c6a412439ca

**Misc**

- Node version:
- Package manager:
- Browser:

matthewlipski:
# Summary

The code mark/style has an input rule which converts text wrapped in backticks into code. However, it misses an edge case where the user inputs 2 backticks, then enters text between them and adds a space after. This PR adds an additional input rule for this case.

Closes #1970

## Rationale

While minor, this is a nice improvement to UX.

## Changes

- Extended `Code` TipTap mark with additional input rule.

## Impact

N/A

## Testing

N/A

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inline code handling so backtick-wrapped text followed immediately by a space is reliably recognized and rendered as code, preserving the trailing space correctly for smoother typing and display.

### DIFF
--- a/packages/core/src/blocks/defaultBlocks.ts
+++ b/packages/core/src/blocks/defaultBlocks.ts
@@ -1,3 +1,4 @@
+import { InputRule } from "@tiptap/core";
 import Bold from "@tiptap/extension-bold";
 import Code from "@tiptap/extension-code";
 import Italic from "@tiptap/extension-italic";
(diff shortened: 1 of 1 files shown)


packages/core/src/blocks/defaultBlocks.ts
```diff
@@ -136,7 +137,32 @@ export const defaultStyleSpecs = {
...
-  code: createStyleSpecFromTipTapMark(Code, "boolean"),
+  code: createStyleSpecFromTipTapMark(
+    Code.extend({
+      // Extends the Code mark with an extra input rule that fires when a space is
+      // typed after the closing backtick. The default rule only fires when typing
+      // the closing backtick itself, so it misses the case where the user opens
+      // both backticks first, then writes content between them.
+      addInputRules() {
+        return [
+          ...(this.parent?.() ?? []),
```
nperez0111: Do we need the default rules if this regex is enough?

nperez0111 (approved):
Works from what I can tell

packages/core/src/blocks/defaultBlocks.ts
matthewlipski: Debatable - without the original regex you always need to add a space after the closing backtick. It's a very minor thing, and I don't really have any opinion on it. Wdyt?

nperez0111: I'd prefer to be in full-control of the code mark, so let's move their regex and logic in here. In the future, we'll remove Tiptap, so it is better to have more in our control

Issue #2693, opened 2026-04-30
Plain text paste does not preserve line breaks — multi-line text is flattened into a single paragraph


yannishin:
### What’s broken?

When pasting **plain text** containing line breaks (`\r\n` / `\n`) into the BlockNote editor, the line breaks are not preserved. All lines are flattened into a single paragraph, separated only by spaces.

This is reproducible on the **official demo site** (`https://www.blocknotejs.org/demo`) without any custom configuration.

### What did you expect to happen?

Each line in the pasted plain text should map to a separate paragraph block (or at minimum a soft line break). Plain text line breaks carry meaning and should not be collapsed into spaces.

### Steps to reproduce

1. Open Windows Notepad (or any plain-text source — Notepad++, terminal output, etc.) and type the following on **separate lines**:
```
   line1
   line2
   line3
   line4
```
2. Select all and copy (`Ctrl+C`)
3. Open `https://www.blocknotejs.org/demo`
4. Click into the editor and paste (`Ctrl+V`)

**Actual result**: A single paragraph — `line1 line2 line3 line4`

**Expected result**: Four separate paragraph blocks, one per line.

<img width="995" height="320" alt="Image" src="https://github.com/user-attachments/assets/23274078-3eb0-4404-b505-c071b092d19c" />

<img width="1094" height="356" alt="Image" src="https://github.com/user-attachments/assets/a108fae1-5936-4477-8bcd-6aca38500d2f" />

### Additional context

- Source: Windows Notepad / Notepad++ (CRLF), also reproduces from other plain-text sources
- Target: `https://www.blocknotejs.org/demo`
- macOS Chrome and Windows Chrome both reproduce
- Default configuration, no custom paste handlers

### Contribution

- [ ] I'd be interested in contributing a fix for this issue

### Sponsor

- [x] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

YousefED:
Thanks, we'll look into this

PR #2702, opened 2026-05-03
fix: Plain text paste does not preserve line breaks

Gxrvish:
# Summary

This PR aims to solve an issue described [here](https://github.com/TypeCellOS/BlockNote/issues/2693)

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,5 +85,8 @@
     "docs",
     "shared",
     "tests"
-  ]
+  ],
+  "dependencies": {
+    "remark-breaks": "^4.0.0"
+  }
 }
--- a/packages/core/src/api/parsers/markdown/parseMarkdown.ts
+++ b/packages/core/src/api/parsers/markdown/parseMarkdown.ts
@@ -1,6 +1,7 @@
 import { Schema } from "prosemirror-model";
 import remarkGfm from "remark-gfm";
 import remarkParse from "remark-parse";
+import remarkBreaks from "remark-breaks";
 import remarkRehype, {
   defaultHandlers as remarkRehypeDefaultHandlers,
 } from "remark-rehype";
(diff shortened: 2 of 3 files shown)


Gxrvish:
@YousefED few remarks?
first time solving bug in BlockNote, Sorry if i missed some internal guidelines :slightly_smiling_face: 
Thanks!

package.json
```diff
@@ -85,5 +85,8 @@
     "docs",
     "shared",
     "tests"
-  ]
+  ],
+  "dependencies": {
+    "remark-breaks": "^4.0.0"
+  }
```
nperez0111, line 91: This is installed to the wrong package.json, it should be installed to the one in packages/core/package.json, this is a monorepo

packages/core/src/api/parsers/markdown/parseMarkdown.ts
```diff
@@ -1,6 +1,7 @@
 import { Schema } from "prosemirror-model";
 import remarkGfm from "remark-gfm";
 import remarkParse from "remark-parse";
+import remarkBreaks from "remark-breaks";
```
nperez0111, line 4: Honestly, would prefer not to introduce a whole package just for this. And would like to have tests added that prove it does not break existing behavior

nperez0111 (changes requested):

packages/core/src/api/parsers/markdown/parseMarkdown.ts
Gxrvish: > Honestly, would prefer not to introduce a whole package just for this

For this operation adding a package is worth, Because doing it from scratch without the help of package might break some edge cases which may introduce unusual behaviour.

> would like to have tests added that prove it does not break existing behavior

Will add it soon :slightly_smiling_face: 
Thanks for the review :blush: !!!

nperez0111:
Thanks for your efforts we've shipped a custom markdown parser which fixes this: https://github.com/TypeCellOS/BlockNote/pull/2624

gppa-odoo:
@nperez0111 Thanks!
Will continue to ship good PRs, Had great learning.

PR #2703, opened 2026-05-04
fix: prevent table row drag from moving an extra adjacent row

### LINKED ISSUE #2691: Table row drag moves two rows instead of one due to duplicate dropHandler   invocation
## Bug Description

  Dragging a row in a table sometimes moves **two rows** (the dragged row plus
  an adjacent one) instead of one. The duplicate move happens because the
  `dropHandler` registered by `TableHandlesView` is invoked twice for a single
  user drop:

  1. First, on the synthetic `drop` event re-dispatched by the SideMenu plugin.
  2. Then, on the original `drop` event as it bubbles to `pmView.root`.

  Between the two invocations, ProseMirror commits the first transaction,
  `view.update()` refreshes `TableHandlesView.state.block`, and the second
  invocation operates on the updated block — moving an additional row.

  ## Affected version

  `@blocknote/core@0.46.2`

  ## Root Cause Analysis

  `SideMenuView` registers a `drop` listener in capture phase and re-dispatches
  a synthetic `drop` event whenever it considers the drop point outside the
  editor bounds:

  ```ts
  // packages/core/src/extensions/SideMenu/SideMenuPlugin.ts
  this.pmView.root.addEventListener("drop", this.onDrop, true);
  // ...
  onDrop = (event) => {
    if (event.synthetic) return;
    const t = this.getDragEventContext(event);
    // ...
    if (!t.isDropWithinEditorBounds && t.isDropPoint) {
      this.dispatchSyntheticEvent(event);
    }
    // ...
  };

  dispatchSyntheticEvent(e) {
    const t = new Event(e.type, e);
    // ...
    t.synthetic = true;
    this.pmView.dom.dispatchEvent(t);
  }
  ```

  For a table-row drop, `isDropPoint` is true and `isDropWithinEditorBounds` is
  sometimes false, so the synthetic dispatch fires. The synthetic event bubbles
  to `pmView.root`, where `TableHandlesView.dropHandler` is registered (bubble
  phase). Then the original event continues bubbling and triggers `dropHandler`
  again.

  `TableHandlesView.dropHandler` does **not** check `event.synthetic` and does
  **not** clear `state.draggingState` after handling, so both invocations:
  - pass the `state.draggingState !== undefined` guard,
  - read `state.block` (refreshed by `view.update()` between calls),
  - call `moveRow` again with the new row at index `originalIndex`.

  Result: a second row is moved.

  ## Confirmation Logs

  Instrumented `dropHandler` showed two invocations on a single user drop:

  ```
  [BN dropHandler ENTRY] instanceId=1 isSynthetic=true  eventTarget=DIV
  eventPhase=3 defaultPrevented=false
  [BN row-drop] before  { originalIndex: 1, targetRowIndex: 8, snapshotRows: 9,
  blockId: ... }
  [BN row-drop] after moveRow { newRowsCount: 9 }
  [BN dropHandler ENTRY] instanceId=1 isSynthetic=undefined eventTarget=DIV
  eventPhase=3 defaultPrevented=true
  [BN row-drop] before  { originalIndex: 1, targetRowIndex: 8, snapshotRows: 9,
  blockId: ... }
  [BN row-drop] after moveRow { newRowsCount: 9 }
  ```

  Both calls share `instanceId=1` (one `TableHandlesView`), confirming this is a
   single instance receiving two events — the synthetic one (from SideMenu) and
  the real one.

  ## Suggested Fix

  **A. Skip synthetic events in `TableHandles.dropHandler`** (mirrors what
  `SideMenuView`'s own handlers already do):

  We've shipped (A) as a `yarn patch` workaround in our app and verified the bug
   is gone.

  **I'd be happy to open a PR with this fix — would that be okay?** Let me know
  if you'd prefer approach A or something different, and I'll send it along.

  Dragging a single table row should move only that one row to the drop target.

  For example, in a table with rows R1–R10:
  - When the user grabs the row handle of R2 and drops it after R10,
  - The expected order is: R1, R3, R4, R5, R6, R7, R8, R9, R10, R2
  - Only R2 should be repositioned. All other rows should keep their original
  order.

  Instead, the dragged row and the row immediately after it (e.g., R2 and R3)
  both get moved to the drop target, producing: R1, R4, R5, R6, R7, R8, R9, R10,
   R2, R3.

### Steps to reproduce

  1. Open the BlockNote demo with a `table` block (no merged cells required).
  2. Insert at least 5 rows.
  3. Drag the second row using the row handle and drop it after the last row.
  4. Expected: only the dragged row is moved.
  5. Actual: the dragged row **and** the row originally below it are both moved
  to the bottom.

### BlockNote version

v0.46.2

### Environment

Chrome  146.0.7680.178 (arm64) , macOS 26.3.1 , React 18

### Additional context

_No response_

### Contribution

### Sponsor

LimChaeJune:
# Summary

Fixes a bug where dragging a single table row could move an extra adjacent row when the drop lands outside `.bn-block-group` (e.g. into the side menu gutter). Closes #2691.

## Rationale

`SideMenuView.dispatchSyntheticEvent` re-dispatches `drop` events as `synthetic` so ProseMirror can handle drops that fall outside the editor bounds. Both `SideMenuView` and `TableHandlesView` listen for `drop` on `pmView.root`. `SideMenuView`'s handlers already guard against the synthetic  event, but `TableHandlesView.dropHandler` does not — so a single user gesture invokes `dropHandler` twice (once for the synthetic, once for the original). Between the two calls, ProseMirror commits the row move and refreshes `state.block`, causing the second call to drag the now-shifted neighbor along with the original target.

The fix clears `state.draggingState` once `dropHandler` has the values it needs, making the handler idempotent. This avoids leaking knowledge of `SideMenuView`'s synthetic-event convention into `TableHandlesView` and protects against any future path that re-enters `dropHandler` for the same gesture.

## Changes

- `packages/core/src/extensions/TableHandles/TableHandles.ts`: After destructuring `draggingState` from `this.state`, clear `this.state.draggingState = undefined`. The re-dispatched drop now short-circuits at the existing `draggingState === undefined` early return.
- `tests/src/end-to-end/tables/tables.test.ts`: Added a Playwright regression test that drags row 2 of a 5-row table onto the side-menu gutter (`blockGroup.x - 50`) and asserts the resulting row order.

## Impact

- **TableHandles row/column drag**: Behavior unchanged for normal in-bounds drops. Out-of-bounds drops (side gutter, below table padding) now move only the dragged row instead of also dragging an adjacent one.
- **`dragEnd`** still sets `draggingState = undefined`, which is now a no-op  after `dropHandler` ran but stays correct for drag cancellations that never reach `dropHandler`.

## Testing

- **Regression test**: New Playwright test drags row 2 of a 5-row table onto the side-menu gutter and asserts the row order is `["R1", "R3", "R4", "R5",  "R2"]`. Fails without the fix with `["R1", "R4", "R5", "R2", "R3"]` (the exact bug pattern from #2691). Passes on chromium and webkit; firefox is skipped following the existing dragdrop test convention.
- **Manual**: Verified the original repro in the BlockNote demo no longer triggers after the fix.

## Screenshots/Video

**Before**
duplicateRows

**After**
화면 기록 2026-05-04 오후 3 43 22

## Checklist

## Additional Notes

the bug is intermittent in real use because it only triggers when the drop coordinate falls outside `.bn-block-group`. The regression test pins the drop to the side gutter to make this deterministic.

## Bug Fixes
- Fixed table row drag-and-drop behavior to prevent unintended row movements during drag operations, ensuring only the selected row is repositioned as intended.
- Added regression test coverage to verify correct row drag behavior and confirm that only the intended row moves during drag operations.

### DIFF
--- a/packages/core/src/extensions/TableHandles/TableHandles.ts
+++ b/packages/core/src/extensions/TableHandles/TableHandles.ts
@@ -466,6 +466,8 @@ export class TableHandlesView implements PluginView {
     event.preventDefault();
 
     const { draggingState, colIndex, rowIndex } = this.state;
+    // Clear so a re-dispatched drop short-circuits above (issue #2691).
+    this.state.draggingState = undefined;
 
     const columnWidths = this.state.block.content.columnWidths;
 
--- a/tests/src/end-to-end/tables/tables.test.ts
+++ b/tests/src/end-to-end/tables/tables.test.ts
@@ -1,5 +1,6 @@
+import { expect } from "@playwright/test";
 import { test } from "../../setup/setupScript.js";
-import { BASE_URL } from "../../utils/const.js";
+import { BASE_URL, TABLE_SELECTOR } from "../../utils/const.js";
 import { compareDocToSnapshot, focusOnEditor } from "../../utils/editor.js";
 import { executeSlashCommand } from "../../utils/slashmenu.js";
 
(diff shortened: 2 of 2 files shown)


nperez0111 (approved):
Thanks for this @LimChaeJune  this fixes the issue well!

PR #2709, opened 2026-05-05
fix: backslash newlines when copying from a code block

nperez0111:
# Summary

Copying content from inside a code block produced a clipboard payload littered with backslashes — every newline in `text/plain` was prefixed with `\` (markdown's hard-break syntax), and `text/html` had `<br>` separators inside the code instead of literal newlines.

Reproduction: select code in a code block → paste into any plain-text app:
```
{\
abc: '34\
\
}
```

This PR fixes the root cause so both `text/html` and `text/plain` come out clean.

## Rationale

Tracing the clipboard pipeline showed two compounding issues:

1. **Inline-only export path was used for code-block selections.** When the selection lives inside a single block, copyExtension takes a shortcut that strips the block wrapper — producing inline HTML for cleaner pastes. For paragraphs that's right; for code blocks it's wrong, because the `<pre><code>` wrapper carries semantic meaning. Without it, the markdown converter has no idea this is code and treats every line break as a paragraph hard-break (`\` + newline).

2. **`serializeInlineContentExternalHTML` re-introduced `<br>` for code content.** It calls `inlineContentToNodes` without a `blockType`, which makes it split every `\n` into a `hardBreak` PM node. Those serialize as `<br>` — even when wrapped in `<pre><code>`. The internal-HTML serializer already passed `blockType` through; the external one didn't.

## Changes

- `copyExtension.ts` — guard the inline-only optimization with a `parent.type.spec.code` check. Code-block selections fall through to the existing block-export path, which invokes the code block's own `toExternalHTML` (proper `<pre><code class="language-X" data-language="X">`).
- `serializeBlocksExternalHTML.ts` — plumb `blockType` through `serializeInlineContentExternalHTML` (via `options.blockType`) so `inlineContentToNodes` keeps `\n` as literal text for code-content blocks. `serializeBlock` passes `block.type`.
- `tests/.../exportTestExecutors.ts` — `prettify(..., { ignore: ["code"] })` so code-block snapshots show real newlines instead of having them collapsed by the prettifier.
- `tests/.../codeBlockMarkdown.test.ts` — three regression tests: full code-block copy, partial selection within a code block, and a paragraph-copy guard to prove non-code paths are untouched.
- Snapshots: 3 code-block HTML snapshots regenerated to show literal newlines (no more `<br />` inside `<pre><code>`).

## Impact

After the fix, for the user's repro:

- `text/html`: `<pre data-language="javascript"><code class="language-javascript" data-language="javascript">{
  abc: '34

}</code></pre>` — proper code semantics, real newlines.
- `text/plain`: clean fenced markdown block, no backslashes.

No behavior change for non-code blocks. The inline-only optimization for paragraphs, headings, etc. is preserved exactly.

## Testing

- Added 3 regression tests in `tests/src/unit/core/clipboard/copy/codeBlockMarkdown.test.ts` (full code-block copy, partial selection, paragraph guard).
- Updated 3 code-block HTML snapshots (now show literal newlines instead of `<br />`).
- All 430 core unit tests pass.
- All 49 clipboard tests (copy, paste, copyPaste, copyPasteEquality) pass — confirming no regressions in paste behavior.
- The pre-existing failures in `src/unit/nextjs/serverUtil` and `src/unit/react/BlockNoteView*Remount*` are unrelated (fail on `main` too).

## Checklist

- [ ] The documentation has been updated to reflect the new feature — N/A, this is a bug fix with no API change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of code block content when copying or exporting selections within code blocks, ensuring raw text formatting is properly preserved instead of converting through intermediate HTML.
  * Enhanced block type information tracking during export serialization for more accurate content handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TypeCellOS/BlockNote/pull/2709)


matthewlipski (approved):
Looks good, agree that copying within code blocks should preserve the block HTML, rather than just the inline HTML. Though I'm seeing different behaviour on main to what you describe. Given this code block:

```javascript
const a = 1;
const b = 2;
```

This is the plain text I see get pasted:

```
const a = 1; const b = 2;
```

Whereas if I understand correctly, you were seeing this:

```
const a = 1;\n
const b = 2;
```

Any idea what's up with that?

nperez0111:
@matthewlipski, it depends on where you paste the content after.

But, on main, this is what is on my clipboard:
image

On this branch, this is on my clipboard:
image

nperez0111:
^ just realizing the code fence will mess this up

nperez0111:
OK, here is the output now (for a text selection within the code block):
image

And here is for a node selection on the code block:
image

So, this looks corrected to me now

PR #2710, opened 2026-05-05
Custom schema example demos: alert title input + macro block

nperez0111:
## Summary

Two new patterns for working with custom blocks, both as runnable examples under `examples/06-custom-schema/`.

### 1. Alert block — multiple "fields" via props (`01-alert-block`)
By rendering an `<input>` inside the block's React view and storing its value in a block prop, a single block can expose multiple editable fields. The alert block now has both:

- a flat, click-to-edit **title input** (stored in the new `title` prop), styled to look like static document text until focused
- the existing **rich-text body** (the BlockNote inline content)

This is a clean way to fake "structured" blocks without splitting the content across multiple blocks or stepping outside the BlockNote document model.

Example here: https://blocknote-git-feat-example-demos-typecell.vercel.app/custom-schema/alert-block

### 2. Macro block — arbitrary HTML injection from outside the document (`09-macro-block`)
Built with the vanilla `createBlockSpec` API (no React). Each macro block stores an `id` prop that's used to look up arbitrary content in a **global registry**, which is injected as the "before" and "after" decoration around the block's editable inline content.

Each registry slot accepts either:
- an **HTML string** → injected via `innerHTML` (read-only templated decoration)
- a live **HTMLElement** → injected via `appendChild` (live, interactive DOM — e.g. an `<input>` or `<button>`)

Combined with pattern #1, this gives you read-only templated wrappers (citations, server-driven labels, etc.) and lightweight interactive controls (inputs, buttons) sitting next to BlockNote's editable inline content — without that decoration ever being part of the document JSON.

Example here: https://blocknote-git-feat-example-demos-typecell.vercel.app/custom-schema/macro-block

### DIFF
--- a/examples/06-custom-schema/09-macro-block/src/Macro.ts
+++ b/examples/06-custom-schema/09-macro-block/src/Macro.ts
@@ -1,0 +1,63 @@
+import { createBlockSpec, defaultProps } from "@blocknote/core";
+
+import { macroRegistry } from "./macroRegistry.js";
+
+// The Macro block — built with the vanilla `createBlockSpec` API.
+//
+// It carries a single `id` prop that is used to look up the HTML that should
+// be rendered in the "before" and "after" slots around the block's editable
+// inline content. This can actually be in any structure, it is in your control
+export const macroBlock = createBlockSpec(
+  {
+    type: "macro",
+    propSchema: {
+      textAlignment: defaultProps.textAlignment,
+      textColor: defaultProps.textColor,
+      id: {
+        default: "" as const,
+      },
+    },
+    content: "inline",
+  },
+  {
+    render: (block) => {
+      const wrapper = document.createElement("div");
+      wrapper.className = "macro-block";
+
+      // We pull from the "registry" what this block should insert before and after
+      const definition = macroRegistry[block.props.id];
+
+      const before = document.createElement("div");
+      before.className = "macro-slot macro-slot-before";
+      before.contentEditable = "false";
+      const beforeContent = definition?.before ?? "";
+      if (typeof beforeContent === "string") {
+        before.innerHTML = beforeContent;
+      } else {
+        before.appendChild(beforeContent);
+      }
+
+      const content = document.createElement("div");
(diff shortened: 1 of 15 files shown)


examples/06-custom-schema/09-macro-block/src/Macro.ts
```diff
@@ -0,0 +1,63 @@
...
+// It carries a single `id` prop that is used to look up the HTML that should
+// be rendered in the "before" and "after" slots around the block's editable
+// inline content. This can actually be in any structure, it is in your control
+export const macroBlock = createBlockSpec(
+  {
+    type: "macro",
+    propSchema: {
+      textAlignment: defaultProps.textAlignment,
+      textColor: defaultProps.textColor,
+      id: {
```
YousefED, line 16: nitpick, would change this to `macroId` instead of `id` to make sure it doesn't conflict with block ids (and to make the code clearer)

examples/06-custom-schema/09-macro-block/src/Macro.ts
```diff
@@ -0,0 +1,63 @@
+import { createBlockSpec, defaultProps } from "@blocknote/core";
+
+import { macroRegistry } from "./macroRegistry";
+
+// The Macro block — built with the vanilla `createBlockSpec` API.
+//
+// It carries a single `id` prop that is used to look up the HTML that should
+// be rendered in the "before" and "after" slots around the block's editable
+// inline content. This can actually be in any structure, it is in your control
+export const macroBlock = createBlockSpec(
```
YousefED, line 10: An alternative solution could be to create a custom block for every macro at runtime, right? Pros / cons of that approach?

nperez0111: This supports either way, and given that XWiki throws away the BlockNote representation and only stores their format, the only difference would be code organization: having 1 very dynamic block, or N dynamic blocks.

Given that their macros are already dynamically created at runtime, I don't think this should make too much of a difference, so I'd leave it to them

PR #2713, opened 2026-05-06
fix: Pasting plain text from VSCode (BLO-366)

matthewlipski:
# Summary

When VSCode content is on the clipboard, it stores content in the `plain/text` MIME type. It also contains a bunch of information about the source, like the language of the code being pasted, which depends on the source file type.

BlockNote always attempts to parse VSCode content as a code block. If this fails for whatever reason, e.g. code blocks are not in the schema, a language wasn't found, etc, it won't paste anything.

This PR makes it so that if `handleVSCodePaste` is unable to parse the `text/plain` content as a code block, we attempt to parse it the usual way.

## Rationale

While minor, it's not that uncommon for people to have plain text files in VSCode, or files without an ending. Copy/pasting from these into BlockNote is totally broken right now, so this should be fixed.

## Changes

- If `handleVSCodePaste` returns false, `format` is set to `text/plain` and the content is parsed normally.
- Made `handleVSCodePaste` for all branches where the content cannot be parsed as a code block.

## Impact

N/A

## Testing

Our unit testing setup doesn't support adding custom content types to the clipboard the way that VSCode does, so we can't really replicate VSCode clipboard content for testing.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard paste handling for VS Code formatted content. Code blocks now paste with proper formatting only when a code block node is available; otherwise, content falls back to plain text.

### DIFF
--- a/packages/core/src/api/clipboard/fromClipboard/handleVSCodePaste.ts
+++ b/packages/core/src/api/clipboard/fromClipboard/handleVSCodePaste.ts
@@ -1,9 +1,6 @@
 import { EditorView } from "prosemirror-view";
 
-export async function handleVSCodePaste(
-  event: ClipboardEvent,
-  view: EditorView,
-) {
+export function handleVSCodePaste(event: ClipboardEvent, view: EditorView) {
   const { schema } = view.state;
 
   if (!event.clipboardData) {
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):
Good catch @matthewlipski

PR #2722, opened 2026-05-07
fix(core): preserve list item type when pasting into empty list items

### LINKED ISSUE #2330: pasteHandler doesn't work correctly for list items when their content is empty prior to pasting text.
Problem:
1. Create bullet/numbered/checklist item
2. Try to paste text into its content
3. List item gets replaced with paragraph (This happens when list item content is empty).

The following code solves the problem, but I don't if it is ideal:
```js
pasteHandler: ({ event, editor, defaultPasteHandler }) => {
	const currentBlock = editor.getTextCursorPosition().block;
	const blockSpec = editor.schema.blockSchema[currentBlock.type];
	
	const hasInlineContent = blockSpec?.content === "inline";
	const isEmpty = !currentBlock.content || 
		(Array.isArray(currentBlock.content) && currentBlock.content.length === 0);
	
	if (hasInlineContent && isEmpty) {
		const plainText = event.clipboardData?.getData("text/plain");
		if (plainText) {
			editor.pasteText(plainText);
			return true;
		}
	}
	
	return defaultPasteHandler();
},
```

https://github.com/user-attachments/assets/47196d1c-db14-48c2-8bc1-fbc4cb2bbf10

nperez0111:
# Summary

Fixes #2330. Pasting plain text (or a paragraph) into an empty bullet/numbered/check list item used to replace the list item with a paragraph; it now keeps its type and absorbs the pasted inline content. Also fixes bare `<li>a</li><li>b</li>` HTML pastes that previously produced a list item followed by a paragraph instead of two list items.

## Rationale

BlockNote's `editor.pasteHTML` round-trips through the BlockNote HTML serializer, producing a closed slice (`blockGroup > blockContainer > paragraph`) that ProseMirror inserts as a new block — walking up the document and replacing the surrounding container. That's surprising when the user is just pasting text into an empty list item. Separately, the `BulletListItem` parse rule requires a `<ul>`/`<ol>` parent, so orphan `<li>` HTML fell back to paragraph parsing.

## Changes

- `packages/core/src/editor/transformPasted.ts` — `retypeLeadingParagraphForEmptyTarget` retypes the slice's leading paragraph to match an empty, non-paragraph, inline-content target block (list items, headings, custom inline-content blocks). Subsequent blocks are kept as-is and become siblings. Non-paragraph leading blocks (heading, list item) keep the existing replace behavior.
- `packages/core/src/api/parsers/html/util/nestedLists.ts` — new `wrapOrphanListItems` step wraps consecutive `<li>` siblings with no `<ul>`/`<ol>` ancestor in a fresh `<ul>` before list-lifting runs.

## Impact

Scoped: only fires when the target is an empty inline-content block whose type isn't `paragraph` and the slice's leading block is a `paragraph`. Pasting into non-empty blocks, into paragraphs, or pasting non-paragraph leading blocks (heading, list item, table) is unchanged. The orphan-`<li>` wrap only affects HTML where `<li>` has no list ancestor.

## Testing

- New `packages/core/src/editor/transformPasted.test.ts` (15 tests) drives the actual paste path via `editor.pasteHTML` for: paragraph into empty bullet/numbered/check list items, paragraphs with marks, multi-paragraph paste, heading paste, heading + paragraph paste, list item paste, nested list paste, bare `<li>` paste, two list items into empty/non-empty list items, and regression cases for non-empty list items and paragraphs.
- New snapshot tests in `packages/core/src/api/parsers/html/util/nestedLists.test.ts` cover orphan `<li>` wrapping and confirm `<li>`s already inside `<ul>` are left alone.
- All existing core (441) and tests-package (844) tests pass.

## Checklist

## Release Notes

* **Bug Fixes**
  * Improved pasting behavior for list items and other content into empty list items, with better inline content preservation and merging.
  * Enhanced parsing of standalone list items pasted without a containing list wrapper.

* **Tests**
  * Added comprehensive test coverage for paste transformation across various content types and scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TypeCellOS/BlockNote/pull/2722)

### DIFF
--- a/packages/core/src/api/parsers/html/util/__snapshots__/nestedLists.test.ts.snap
+++ b/packages/core/src/api/parsers/html/util/__snapshots__/nestedLists.test.ts.snap
@@ -1,9 +1,5 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`Lift nested lists > Does not merge <li> nodes separated by non-whitespace text 1`] = `"<ul><li>a</li></ul>text<ul><li>b</li></ul>"`;
-
-exports[`Lift nested lists > Leaves <li>s already inside a <ul> alone 1`] = `"<ul><li>existing</li></ul><ul><li>orphan</li></ul>"`;
-
 exports[`Lift nested lists > Lifts multiple bullet lists 1`] = `
 "<ul>
     <div><li>
@@ -146,11 +142,3 @@ exports[`Lift nested lists > Lifts nested numbered lists 1`] = `
     </li>
  </ol>"
 `;
-
-exports[`Lift nested lists > Wraps a single bare <li> in a <ul> 1`] = `"<ul><li>only</li></ul>"`;
-
-exports[`Lift nested lists > Wraps bare <li>s mixed with other top-level content 1`] = `"<p>before</p><ul><li>x</li><li>y</li></ul><p>after</p>"`;
-
-exports[`Lift nested lists > Wraps consecutive bare <li> elements in a <ul> 1`] = `"<ul><li>a</li><li>b</li></ul>"`;
-
-exports[`Lift nested lists > Wraps nested orphan <li>s as inner lists 1`] = `"<ul><li>outer</li><li>inner</li></ul>"`;
--- a/packages/core/src/api/parsers/html/util/__snapshots__/nestedLists.test.ts.snap
+++ b/packages/core/src/api/parsers/html/util/__snapshots__/nestedLists.test.ts.snap
@@ -1,5 +1,7 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`Lift nested lists > Leaves <li>s already inside a <ul> alone 1`] = `"<ul><li>existing</li></ul><ul><li>orphan</li></ul>"`;
+
 exports[`Lift nested lists > Lifts multiple bullet lists 1`] = `
 "<ul>
     <div><li>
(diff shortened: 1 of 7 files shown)


matthewlipski:
My main question regarding this is why not set the `isolating` attribute to `true` for list items instead? This is something we do by default in `createBlockSpec` and explicitly disable for default blocks, and has the effect of retaining the existing block type regardless of what content is pasted into it. I assume it's because:

1. When pasting a non-paragraph (heading, quote, etc) into an empty list item, we actually do want to override the list item type to the type of the pasted content. We only don't want to do that for paragraphs because they are the "vanilla" block.
2. That doesn't address the other issue fixed in this PR, i.e. handling pasting of consecutive bar `<li>` elements.

Could you comment on this?

nperez0111:
> My main question regarding this is why not set the `isolating` attribute to `true` for list items instead? This is something we do by default in `createBlockSpec` and explicitly disable for default blocks, and has the effect of retaining the existing block type regardless of what content is pasted into it. I assume it's because:
> 
>     1. When pasting a non-paragraph (heading, quote, etc) into an empty list item, we actually do want to override the list item type to the type of the pasted content. We only don't want to do that for paragraphs because they are the "vanilla" block.
> 
>     2. That doesn't address the other issue fixed in this PR, i.e. handling pasting of consecutive bar `<li>` elements.
> 
> 
> Could you comment on this?

Good question. Your hunch is right that we kind of diverge from prosemirror semantics (because our HTML structure is different) so a `<p>` being pasted into an `<li>` wouldn't just work right away. As for 2, I found that this behave unexpectedly when pasting something that was just `<li>A</li><li>B</li>`as opposed to being wrapped in a `<ul>` or `<ol>`. They weren't being matched as a list item even though they were.

Now that I mention it though, maybe I can just have the bullet list item parse match if it is an li without a parent? I'll give that a try

nperez0111:
Yep, that worked better I think!

PR #2728, opened 2026-05-08
feat(react): Added select component for text align feature

GuiLeme:
# Summary

Added a selector component (Just like we have for the block-type selector) for the text align feature. 

## Rationale

I am thinking of 2 good reasons for this addition:
- With this component, hopefully, customized toolbars can use it in order to fit in less space than with the 3 buttons (besides also adding the justify, so, a bit of improvement there as well).
- Outside of that, we can take advantage of the user-experience with other major text editor softwares that already use this approach.

Besides those, there is one other reason.
- Currently in [BigBlueButton](https://github.com/bigbluebutton/bigbluebutton), we are integrating BlockNote as our new shared-notes, and we are currently making some changes to smooth the transition between Etherpad to BlockNote, one of those changes includes fix the toolbar on the top of the document, instead of appearing only when selecting the text. This ends up taking a lot of space so what we did was exactly this - compile these 3 alignment buttons into 1 selector (and we also removed the text of it just to save some space, but that's a minor detail of ours). See screenshot ahead 

image

With that in mind, it would be much simpler if we had this component for us to use it out-of-the-box.

## Changes

- Added the `TextAlignSelect.tsx` component;
- Added 1 new i18n key to be the label of the new selection tool.  

## Impact

As this will not be part of the default toolbar, I'd say no impact on the other components (it will just be used by developers to create their own customized toolbar).

## Testing

I didn't create any automated tests, but if you guys wanted to have something like that, I can dig further.

As for manual testing, I just changed one of the playground examples and tried to break it some how.

In my case, I used `examples/03-ui-components/02-formatting-toolbar-buttons/src/App.tsx` I added it in place of the three align buttons:

```diff
-- a/examples/03-ui-components/02-formatting-toolbar-buttons/src/App.tsx
+++ b/examples/03-ui-components/02-formatting-toolbar-buttons/src/App.tsx
@@ -11,7 +11,7 @@ import {
   FormattingToolbar,
   FormattingToolbarController,
   NestBlockButton,
-  TextAlignButton,
+  TextAlignSelect,
   UnnestBlockButton,
   useCreateBlockNote,
 } from "@blocknote/react";
@@ -38,9 +38,7 @@ const CustomFormattingToolbar = () => (
     {/* Extra button to toggle code styles */}
     <BasicTextStyleButton key={"codeStyleButton"} basicTextStyle={"code"} />
 
-    <TextAlignButton textAlignment={"left"} key={"textAlignLeftButton"} />
-    <TextAlignButton textAlignment={"center"} key={"textAlignCenterButton"} />
-    <TextAlignButton textAlignment={"right"} key={"textAlignRightButton"} />
+    <TextAlignSelect key={"textAlignSelect"} />

```

After that I just had to build the lib and test it locally.

## Screenshots/Video

image

## Checklist

~~I'll definitely dig into all that for the following commits.~~ All of the above is covered now.

## Additional Notes

I'd just like to mention that the Logic I use for the align text in the selector is exactly the same as for the button. So that's why I say it follows the project coding standards. 

PR I did the same thing for BBB is [here](https://github.com/bigbluebutton/bigbluebutton/pull/25023)

## Summary by CodeRabbit

* **New Features**
  * Added a text alignment dropdown to the formatting toolbar (Left, Center, Right, Justify) and exported it for use.

* **Localization**
  * Added human-readable labels for alignment options across many locales.

* **Documentation**
  * Added docs demonstrating how to add and use the text alignment dropdown.

* **Tests**
  * Added unit tests verifying rendering and reactive behavior of the alignment control.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TypeCellOS/BlockNote/pull/2728)

### DIFF
--- a/docs/content/docs/react/components/formatting-toolbar.mdx
+++ b/docs/content/docs/react/components/formatting-toolbar.mdx
@@ -38,3 +38,29 @@ The first element in the default Formatting Toolbar is the Block Type Select, an
 <Example name="ui-components/formatting-toolbar-block-type-items" />
 
 Here, we use the `FormattingToolbar` component but keep the default buttons (we don't pass any children). Instead, we pass our customized Block Type Select items using the `blockTypeSelectItems` prop.
+
+## Adding a Text Align Select
+
+You can replace the default text alignment buttons with a `TextAlignSelect` dropdown, which groups all four alignment options (Left, Center, Right, Justify) into a single compact select.
+
+To use it, pass `<TextAlignSelect>` as a child of `FormattingToolbar` in a custom toolbar:
+
+```tsx
+import {
+  BlockTypeSelect,
+  FormattingToolbar,
+  FormattingToolbarController,
+  TextAlignSelect,
+} from "@blocknote/react";
+
+const CustomFormattingToolbar = () => (
+  <FormattingToolbar>
+    <BlockTypeSelect key="blockTypeSelect" />
+    {/* ... other toolbar items ... */}
+    <TextAlignSelect key="textAlignSelect" />
+    {/* ... other toolbar items ... */}
+  </FormattingToolbar>
+);
+```
+
+The select is only rendered when the cursor or selection is on a block that supports the `textAlignment` prop (e.g. paragraphs and headings). For table blocks, it applies the chosen alignment to the individual cells covered by the current cell selection.
--- a/packages/core/src/i18n/locales/ar.ts
+++ b/packages/core/src/i18n/locales/ar.ts
@@ -310,15 +310,19 @@ export const ar: Dictionary = {
     },
     align_left: {
       tooltip: "محاذاة النص إلى اليسار",
+      label: "يسار",
     },
     align_center: {
       tooltip: "محاذاة النص في المنتصف",
+      label: "وسط",
     },
     align_right: {
       tooltip: "محاذاة النص إلى اليمين",
+      label: "يمين",
     },
     align_justify: {
       tooltip: "ضبط النص",
+      label: "ضبط",
     },
     table_cell_merge: {
       tooltip: "جمع الخلايا",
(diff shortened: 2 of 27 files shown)


gustavotrott:
I think the floating toolbar would also benefit from this new component!
It would reduce the toolbar size (make it feel less cluttered) and align it with an already established pattern.
Combining the 3 existing buttons into a single dropdown feels like a significant UX refinement.

Of course, that would be for a future PR. I'm just sharing feedback as someone who has been using BlockNote every day

nperez0111:
I'll let @matthewlipski lead here, but I'm generally for this change. I probably wouldn't even have "Left" in the toolbar just the icon perhaps with the existing tooltip

GuiLeme:
Hello, @nperez0111.

Thanks for the feedback. I agree with removing the text from the selected Item, but the problem is:

* With the current toolbar `Select` component's API the text displayed inline in the toolbar is tightly bound to the selected item;

* Besides, the `Select` component doesn't display a tooltip when hovering the selected Item;

So for that, the best approach (we thought of) would be to also change the `Select` component itself adding a flag like `labelDisplayMode?: 'inline' | 'tooltip'` (default: `inline`) and this would control how to display the selected item text (displayed directly in the select component - current behavior - or only show the icon with no text and having a tooltip for the selected item text).

If that's implemented, we could use that flag in the `TextAlignSelect` so that no text would appear in the component in the toolbar (as suggested by you), and we could also remove the new translations added and use the previous `tooltip` attribute of the respective dictionary key instead, since it is not going to be displayed inline anymore.

What do you say?

nperez0111:
I understand that the `Select` API as it is now may not be sufficient to represent this. But we'd change it to support this

I'm proposing that the change here should be:
 - collapse these 3 options into 1 dropdown, where the button icon of the dropdown is the current state. (Otherwise you really aren't saving horizontal space)
 - The options can describe left / right / justify (either inline or as a tooltip, unsure on UX)


I'll let @matthewlipski lead here though

PR #2729, opened 2026-05-08
feat(react): configurable portal targets for floating UI

### LINKED ISSUE #2692: Table handles overlap editor content area in v0.48.0 — likely caused by floating UI portaling to document.body (#2591)
### What’s broken?

After upgrading to **v0.48.0**, the table handles (row drag handle `⋮⋮` on the left, column drag handle `⋮⋮⋮` on top) are mispositioned and overlap the editor's content area:

- The row drag handle (`⋮⋮`) intrudes into the editor's text area, sitting on top of the previous block
- The column drag handle (`⋮⋮⋮`) appears above the table in a way that overlaps surrounding blocks
- The horizontal scrollbar of the table renders inside the editor container instead of at the bottom of the table
- The AI/floating-action button appears overlapped on top of the table

This is reproducible on the **official demo site itself** (`https://www.blocknotejs.org/demo`) without any custom styling, which rules out userland CSS migration as the cause.

Table handles should be positioned at the edges of the table without overlapping other blocks or floating UI, as they did in v0.47.x.

### Steps to reproduce

1. Open `https://www.blocknotejs.org/demo`
2. Click on any cell inside the default table block
3. Observe that the row/column drag handles overlap surrounding blocks and floating UI

Image

Image

### BlockNote version

v0.48.0, v4.48.1, v0.49.0

### Environment

_No response_

### Suspected root cause

I believe this regression is caused by **#2591** ("portal floating UI elements to document.body to prevent overflow clipping") in v0.48.0.

Hypothesis:

- Before v0.48.0, table handles were rendered inside `.bn-container`, sharing the same coordinate space as the table cells. Position calculations worked relative to the editor.
- In v0.48.0, floating UI elements — including table handles — are now portaled to `document.body`. This changes the reference frame to the viewport.
- The table handle positioning logic appears to still compute offsets relative to the editor's internal coordinate system, or fails to account for the new portal target's scroll/transform context. As a result, handles render at the wrong absolute position and overlap editor content.

If table handles were not intentionally included in the portaling change, scoping the portal behavior to only menus/toolbars (and keeping table handles inside `.bn-container` as before) could be a quick fix. Otherwise the position calculation needs to be updated for the portaled context.

The related breaking change in v0.48.0 also splits `.bn-container` (editor only) from `.bn-root` (editor + floating UI), which seems consistent with this portaling change.

### Related question: opt-out for portaling floating UI to `document.body`?

Separately from the bug above, is there (or could there be) an option to render **all floating UI** inside the editor container instead of `document.body`?

Portaling to `document.body` can cause integration issues in host apps (z-index conflicts with host modals/drawers, embedded environments, etc.), so an opt-out would be valuable for users who previously relied on the in-container behavior.

### Contribution

### Sponsor

nperez0111:
# Summary

Adds a `portalElements` prop on `BlockNoteView` plus a per-Controller `portalElement` override for configuring where each floating UI element (slash menu, formatting toolbar, side menu, table handles, etc) portals to. Default mounting moves from `document.body` to the editor's `bn-container`.

## Rationale

Resolves #2692. Previously every floating UI element was force-portaled to `document.body`, which caused z-index conflicts when the editor was embedded in app shells with fixed headers, sidebars, or modals, and let elements like table handles render outside the editor's visual boundary. Per-element configuration lets integrators scope each piece appropriately for their layout — small editors can keep large dropdowns at `body` while keeping table handles pinned inside the editor.

## Changes

- `editor.mount(element, options?)` accepts an optional `portalTarget`. When omitted, defaults to `element.parentElement` (i.e. `bn-container` in normal React usage), falling back to shadow root / `document.body`.
- New `portalElements?: PortalElementsMap` prop on `BlockNoteView`. Keys mirror `BlockNoteDefaultUIProps` (`formattingToolbar`, `linkToolbar`, `slashMenu`, `emojiPicker`, `sideMenu`, `filePanel`, `tableHandles`, `comments`) plus `default` for the global fallback. Values can be `HTMLElement`, a CSS selector string, or `null` for explicit `document.body`.
- Each Controller (`FormattingToolbarController`, `LinkToolbarController`, `SuggestionMenuController`, `GridSuggestionMenuController`, `TableHandlesController`, `SideMenuController`, `FilePanelController`, `FloatingComposerController`, `FloatingThreadController`) accepts an optional `portalElement` prop, threaded through `GenericPopover` / `PositionPopover` / `BlockPopover`.
- `BlockNoteDefaultUI` reads `portalElements` and threads each resolved value into the matching auto-mounted Controller.
- New side-by-side example at `examples/03-ui-components/20-portal-elements/` demonstrating clipped vs escaping behaviour.

## Impact

The default-mount-target change is a behaviour change: floating UI now lands inside `bn-container` instead of `document.body`. Consumers that relied on body-level portaling (e.g. for z-index against modals, or CSS rules keyed on `body > .bn-root`) need to opt back in with `<BlockNoteView portalElements={{ default: document.body }} />`. The Controller prop additions are purely additive.

## Testing

Manually tested in the playground via the new example: verified default mode keeps each floating popover inside `.bn-container`, `portalElements.default = document.body` makes them escape, and per-element overrides (e.g. `tableHandles` to a specific element while everything else goes to body) work as expected. `tsc --noEmit` passes across `@blocknote/{react,mantine,ariakit,shadcn}`.

## Screenshots/Video

Side-by-side example in `examples/03-ui-components/20-portal-elements/` shows the clipped vs escaping behaviour visually.

## Checklist

- [x] All existing tests pass (type-check).
- [x] The documentation has been updated (new example + JSDoc on the new prop).

## Additional Notes

Changing `portalElements.default` at runtime requires remounting the editor (the example uses a `key` to demonstrate this), since `editor.mount()` consults the target once. Per-element keys update reactively because they flow through React props on the Controllers, not through `mount()`.

## Summary by CodeRabbit

* **New Features**
  * Configurable portal mounting for floating UI (per-feature and default), allowing callers to control where toolbars, menus, panels, and comment/thread UIs render.
  * Components now accept an optional portal override prop to target custom DOM nodes.

* **Examples**
  * Added a new interactive example demonstrating portal target configuration and escaping floating UI from clipped containers.

* **Documentation**
  * Docs updated with usage guidance and examples for portal configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TypeCellOS/BlockNote/pull/2729)

### DIFF
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -731,15 +731,27 @@ export class BlockNoteEditor<
   /**
    * Mount the editor to a DOM element.
    *
+   * @param element The DOM element to mount the editor's contenteditable into.
+   * @param options.portalTarget Where to mount `editor.portalElement` — the
+   *   container that floating UI (toolbars, menus, etc) portals into. When
+   *   omitted, defaults to `element.parentElement` (which is the editor's
+   *   `bn-container` in typical React usage), or to `document.body` /
+   *   the surrounding shadow root when no parent is available.
+   *
    * @warning Not needed to call manually when using React, use BlockNoteView to take care of mounting
    */
-  public mount = (element: HTMLElement) => {
+  public mount = (
+    element: HTMLElement,
+    options?: { portalTarget?: HTMLElement | null },
+  ) => {
     const root = element.getRootNode();
-    if (typeof ShadowRoot !== "undefined" && root instanceof ShadowRoot) {
-      root.appendChild(this.portalElement);
-    } else {
-      document.body.appendChild(this.portalElement);
-    }
+    const isInShadowRoot =
+      typeof ShadowRoot !== "undefined" && root instanceof ShadowRoot;
+    const target =
+      options?.portalTarget ??
+      element.parentElement ??
+      (isInShadowRoot ? (root as ShadowRoot) : document.body);
+    target.appendChild(this.portalElement);
     this._tiptapEditor.mount({ mount: element });
   };
 
--- a/docs/content/docs/react/components/index.mdx
+++ b/docs/content/docs/react/components/index.mdx
@@ -14,3 +14,23 @@ BlockNote includes a number of UI Components (like menus and toolbars) that can 
   {/* - [Image Toolbar](/docs/react/components/image-toolbar) */}
 
 <CardTable path="/react/components" />
+
+## Configuring Portal Targets
+
+By default, all floating UI elements (toolbars, menus, table handles, etc.) portal into the editor's `bn-container` so they stay scoped to the editor. If your layout needs them to escape — e.g. an `overflow: hidden` ancestor that would clip large dropdowns, or a host modal with its own stacking context — pass a `portalElements` prop to `BlockNoteView`:
+
+```tsx
+<BlockNoteView
+  editor={editor}
+  portalElements={{
+    // Global default for any element not listed below.
+    default: document.body,
+    // Per-element overrides. Values can be HTMLElement, a CSS selector, or null (= document.body).
+    tableHandles: ".bn-container",
+  }}
+/>
+```
+
+Keys mirror the default UI flags (`formattingToolbar`, `linkToolbar`, `slashMenu`, `emojiPicker`, `sideMenu`, `filePanel`, `tableHandles`, `comments`). Manually-mounted Controllers also accept a `portalElement` prop that takes precedence over the map. See the [Portal Targets example](/examples/ui-components/portal-elements).
+
+Note: changing `portalElements.default` after mount requires remounting the editor (`editor.mount()` consults it once); per-element keys update reactively.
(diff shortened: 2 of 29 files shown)


packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -731,15 +731,27 @@ export class BlockNoteEditor<
...
+   *   omitted, defaults to `element.parentElement` (which is the editor's
+   *   `bn-container` in typical React usage), or to `document.body` /
+   *   the surrounding shadow root when no parent is available.
+   *
    * @warning Not needed to call manually when using React, use BlockNoteView to take care of mounting
    */
-  public mount = (element: HTMLElement) => {
+  public mount = (
+    element: HTMLElement,
+    options?: { portalTarget?: HTMLElement | null },
```
matthewlipski, line 745: Just reiterating coderabbit comment, but `| null` is probably not necessary when `portalTarget` is already optional.

matthewlipski (approved):
Nice! Configuration looks pretty much exactly as I would expect

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: I hate APIs that force you to have the value completely empty. If I want to null it, let me null it

PR #2733, opened 2026-05-11
feat: Trailing block extension rewrite (BLO-843)

matthewlipski:
# Summary

Currently, the trailing block (from the `trailingBlock` editor option/`TrailingNodeExtension`) is part of the document. If the document does not end in an empty paragraph, one is added to serve as the trailing block.

This PR makes it so that the trailing block no longer exists in the document. Instead, it exists as a decoration, simply showing a transparent element below the editor with the same dimensions as an empty paragraph block. Clicking the element will remove it, and add in its place an actual block that exists in the document. If the editor is not editable, the trailing block is not shown.

Closes https://github.com/TypeCellOS/BlockNote/pull/2724

## Rationale

The trailing block is strictly a UX feature - it's useless when the editor is not editable, or when its content is exported to HTML and rendered statically. Therefore, it's common for consumers to want to hide it when the editor content is non-editable. This currently isn't possible without hacking around it, as the trailing block is a permanent part of the document.

This implementation is also more semantically correct as the trailing block is just a UX improvement, and therefore shouldn't be represented in the editor state.

## Changes

- Rewrote `TrailingNode.ts`.
- Added CSS for trailing node in `Block.css`.
- Updated examples to no longer include hard-coded trailing block in the initial content.

## Impact

This change comes with some changes to the trailing block behaviour:

- The side menu no longer appears next to the trailing block as it's not an actual block. Only once the user clicks on it and turns it into an actual block can they use the side menu with it.
- Pressing arrow down in the last block of the document no longer moves the text cursor to the trailing block, as again, it's not actually part of the document. However, it would be easy to add this to the `TrailingNodeExtension`, as the same logic already exists for the `mousedown` handler.

## Testing

All unit & e2e tests have been updated to no longer rely on trailing blocks being added to the document.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A


tests/package.json
```diff
@@ -5,9 +5,9 @@
   "scripts": {
     "build": "tsc",
     "lint": "eslint src --max-warnings 0",
-    "playwright": "playwright test",
+    "playwright": "playwright test --ui",
     "test": "vitest --run",
-    "test:updateSnaps": "docker run --rm -e RUN_IN_DOCKER=true --network host -v $(pwd)/..:/work/ -w /work/tests -it mcr.microsoft.com/playwright:v1.51.1-noble npx playwright test -u",
+    "test:updateSnaps": "docker run --rm -e RUN_IN_DOCKER=true --network host -v $(pwd)/..:/work/ -w /work/tests -it mcr.microsoft.com/playwright:v1.51.1-noble npx playwright test images -u",
```
nperez0111: what does this do?

tests/playwright.config.ts
```diff
@@ -14,7 +14,7 @@ import { devices } from "@playwright/test";
 const config: PlaywrightTestConfig = {
   testDir: "./src/end-to-end",
   /* Maximum time one test can run for. */
-  timeout: 30 * 1000,
+  timeout: 8 * 1000,
```
nperez0111: should we change either of these?

packages/core/src/extensions/TrailingNode/TrailingNode.ts
```diff
@@ -1,76 +1,126 @@
...
+            const node = blockContainerType.create(
+              undefined,
+              paragraphType.create(),
+            );
+            const insertPos = view.state.doc.content.size - 1;
+            const tr = view.state.tr.insert(insertPos, node);
+            tr.setSelection(TextSelection.create(tr.doc, insertPos + 2));
+            tr.scrollIntoView();
+            view.dispatch(tr);
+            view.focus();
```
nperez0111: why not blocknote apis for this?

packages/core/src/extensions/TrailingNode/TrailingNode.ts
```diff
@@ -1,76 +1,126 @@
...
+      if (wasShowing === shouldShow) {
+        return mapped;
+      }
+      if (wasShowing) {
+        return mapped.remove(existing);
+      }
+      return mapped.add(tr.doc, [
+        createTrailingWidget(tr.doc.content.size - 1),
+      ]);
+    }
```
nperez0111, line 91: this is nice

tests/playwright.config.ts
matthewlipski: Not intentional, will revert

nperez0111 (approved):
One last thought, how does this work with columns?

PR #2690, opened 2026-04-29
feat: additional heading top padding (BLO-1008)

matthewlipski:
# Summary

This PR adds extra padding to the tops of heading blocks, as this makes it more clear that the heading relates to the content below it, not above it.

Closes #2414

## Rationale

See above.

## Changes

- Added 15px padding to tops of headings.
- Updated side menu styles to still be centered on headings.

## Impact

N/A

## Testing

Updated e2e screenshots.

## Screenshots/Video

N/A

## Checklist

- [X] Code follows the project's coding standards.
- [X] Unit tests covering the new feature have been added.
- [X] All existing tests pass.
- [X] The documentation has been updated to reflect the new feature

## Additional Notes

N/A

## Summary by CodeRabbit

* **Style**
  * Enhanced visual consistency and usability by refining heading block spacing with adjusted padding and improving menu sizing across all heading levels, maintaining design harmony throughout the editor interface.

* **Tests**
  * Updated end-to-end tests with simplified navigation approaches to improve test reliability and reduce maintenance complexity for future updates.

PR #2600, opened 2026-03-30
fix(core): optimize plugin traversals for large documents BLO-1111

YousefED:
# Summary

Fixes typing/echo lag in documents with many blocks (~50k chars) by optimizing several plugins that were performing O(n) full-document traversals on every transaction.

Closes https://github.com/TypeCellOS/BlockNote/issues/2595

## Rationale

Several plugins (`PreviousBlockType`, `NumberedListIndexingPlugin`) traversed the entire document on every keystroke. At ~500+ blocks, this caused noticeable input lag. The fix scopes traversals to only the changed ranges.

## Changes

- **PreviousBlockType plugin**: Replaced `findChildren` (full doc) with `findChildrenInRange` scoped to `transaction.changedRange()`. Maps new-doc range back to old-doc coordinates for accurate comparison.
- **NumberedListItem IndexingPlugin**: 
  - Uses `DecorationSet.map()` + incremental updates instead of rebuilding all decorations
  - Early exit via `completedGroups` pattern — once a decoration matches past the changed range in a blockGroup, skip remaining siblings
  - Converted recursive `calculateListItemIndex` to iterative to prevent stack overflow at 10k+ blocks
  - Narrowed `DecorationSet.find()` range to the numberedListItem node only (not the full blockContainer subtree) to avoid false matches from nested lists

## Impact

- Typing at 10k blocks: ~30-50ms/keystroke (dominated by ProseMirror's `DecorationSet.map()` which is inherently O(n))
- No behavioral changes — all existing functionality preserved
- UniqueID `getChangedRanges` → `changedRange()` change was evaluated and **not included** (validated to have zero measurable impact: ~3.5μs difference)

## Testing

- **IndexingPlugin.test.ts** (16 tests): basic numbering, structural changes (insert/delete/type-change), nested lists, typing preservation, decoration specs, selection-only transactions
- **PreviousBlockType.test.ts** (2 tests): scoped traversal verification, attribute change detection
- **performance.test.ts** (3 tests): ratio benchmarks at 10k blocks for headings and numbered lists (begin + end of doc)

## Checklist

- [x] Code follows the project's coding standards.
- [x] Unit tests covering the new feature have been added.
- [x] All existing tests pass.
- [ ] The documentation has been updated to reflect the new feature

## Summary by CodeRabbit

* **Chores**
  * Updated ProseMirror Transform dependency to 1.11.0 across packages.

* **Performance**
  * Numbered-list indexing and decoration updates are more efficient and avoid unnecessary recomputation.
[...]

PR #2734, opened 2026-05-11
fix: numbered list item decorations missed on initial render

matthewlipski:
# Summary

This PR fixes a bug introduced in #2600, where list item indices would not be visible on the initial render. They would only show for a given numbered list item block, after the editor content before it was changed. This is because the decoration updates only occur when a transaction has a `changedRange`, which is not the case when the editor is first initialized.

To fix this, if `changedRange` is `null`, the extension scans the whole document to apply decorations, rather than returning early.

## Rationale

This is a bug.

## Changes

See above.

## Impact

N/A

## Testing

An e2e test for this (`"Check static rendering visually matches live editor"` in `static.test.ts`) is already present and shows the issue in the screenshot output. This e2e test wasn't updated in #2600, which caused the issue, and was then updated in #2690 without the mistake being caught. The screenshot has been updated in this PR.

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed numbered list items not displaying correct indexing on initial load.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TypeCellOS/BlockNote/pull/2734)


nperez0111 (approved):
Good catch!

packages/core/src/blocks/ListItem/NumberedListItem/IndexingPlugin.ts
```diff
@@ -135,10 +135,9 @@ function getDecorations(
   // Find the start of the first change to limit traversal scope.
   // We only need to check from the change point forward, since earlier
   // blocks are unaffected and their mapped decorations remain correct.
-  const range = tr.changedRange();
-  if (!range) {
-    return { decorations: nextDecorationSet };
-  }
+  // On init (no steps), changedRange() returns null — fall back to a
+  // full scan so initial content gets indexed.
+  const range = tr.changedRange() ?? { from: 0, to: tr.doc.nodeSize - 2 };
```
YousefED, line 140: @matthewlipski besides init, wouldn't this now also fire on selection changes? e.g. if there have been no updates at all?

PR #2741, opened 2026-05-13
feat: decouple yjs from blocknote/core

### LINKED ISSUE #2759: Initial doc creation menu + not working
### What’s broken?

When we create a totally new doc, we have the id "initialBlockId", it seems to have a problem with it, the menu flickers, it opens and closes instantly (not obvious on the video because it is too fast).
As soon as we add some content, it works as expected.

https://github.com/user-attachments/assets/c9e4c4df-4fb8-4613-b9da-5b42677b5f07

I expect to have the menu to open normally.

### Steps to reproduce

@YousefED I remember we got this issue a few month ago, it got solved then came back. I don't find the issue anymore though.

### BlockNote version

_No response_

### Environment

_No response_

### Additional context

_No response_

### Contribution

### Sponsor

nperez0111:
# Summary

In preparation of the upcoming Y.js 14. Which will introduce breaking changes on the API, type, package, and possibly document level.
I'm doing what we probably should have done from the start & de-coupling BlockNote from Y.js completely.
This means that when importing from `@blocknote/core`, we will no longer depend on anything in `yjs` or `y-prosemirror`.
Before this, certain functionality would import `yjs` or `y-prosemirror` at runtime even if the editor didn't actually need it's methods.

Now, to make your editor collaborative, you essentially just need to wrap your existing editor options with `withCollaboration`, for the editor to become collaborative.
This will automatically configure your editor to be collaborative, installing all of the necessary extensions to make the feature work.

## Rationale

I'm fairly certain that we are going to need to support both Y.js 13 & Y.js 14 for some period of time. Since it will require document migrations at the discretion of the integrator.
So, it is better to get ahead of this & completely de-couple the editor from `yjs` & `y-prosemirror`.

I'm hoping for this to land pretty soon, before we are done with the other `y/prosemirror` changes. Since it will reduce drift in the PR as well as reduce bundle size for existing users

## Changes

This PR removes the tight coupling between Yjs and `@blocknote/core` by moving all Yjs-related code into a dedicated `@blocknote/core/yjs` subpath export.

# Breaking Changes

**`collaboration` is no longer a direct editor option.** Instead, use the new `withCollaboration()` helper to wrap your editor options:

```ts
// Before
  collaboration: { provider, fragment, user },
});

// After
import { withCollaboration } from "@blocknote/core/yjs";

const editor = useCreateBlockNote(
  withCollaboration({
    collaboration: { provider, fragment, user },
  }),
);
```

**Yjs-related imports have moved:**

| Export | Old path | New path |
|---|---|---|
| `YjsThreadStore` | `@blocknote/core/comments` | `@blocknote/core/yjs` |
| `RESTYjsThreadStore` | `@blocknote/core/comments` | `@blocknote/core/yjs` |
| `YjsThreadStoreBase` | `@blocknote/core/comments` | `@blocknote/core/yjs` |
| `ForkYDocExtension` | `@blocknote/core/extensions` | `@blocknote/core/yjs` |
| `YSyncExtension`, `YUndoExtension`, `YCursorExtension`, `SchemaMigration` | `@blocknote/core/extensions` | `@blocknote/core/yjs` |

Non-Yjs exports (`CommentsExtension`, `DefaultThreadStoreAuth`, `TiptapThreadStore`, etc.) remain in `@blocknote/core/comments`.

## Impact

Now, `@blocknote/core` does not depend on `yjs` or `y-prosemirror` unless using the new `@blocknote/core/yjs`. Reducing bundle size & separating concerns better.

## Additional Notes

We should follow this up with some changes to the `@blocknote/xl-ai` & `@blocknote/server-editor` packages to remove dependency on the `yjs` package

Closes #2759

## Summary by CodeRabbit

* **New Features**
  * Added `withCollaboration` helper to simplify real-time editor setup.
  * Improved position/cursor tracking using shared position mapping.

* **Breaking Changes / Bug Fixes**
  * Collaboration is now configured via `withCollaboration` (the direct collaboration option in the editor setup was removed).
  * Comments thread creation now uses a simplified `{ head, anchor }` selection shape.
  * Updated Yjs thread store import paths for comments/thread storage.

* **Documentation**
  * Updated collaboration and comments guides/examples to match the new configuration pattern.

* **Chores**
  * Made Yjs packages optional peer dependencies.


packages/core/src/comments/threadstore/ThreadStore.ts
```diff
@@ -23,14 +23,8 @@ export abstract class ThreadStore {
   abstract addThreadToDocument?(options: {
     threadId: string;
     selection: {
-      prosemirror: {
-        head: number;
-        anchor: number;
```
nperez0111, line 28: this was changed to not depend on yjs positions in the API. There are still ways to get it, this was just exposed in the API for some reason

packages/core/src/yjs/extensions/YCursorPlugin.ts
nperez0111: we should do awareness.off to cleanup this

packages/core/src/yjs/extensions/index.ts
```diff
@@ -44,12 +47,45 @@ export const CollaborationExtension = createExtension(
...
+    ...options,
+    extensions: [
+      ...(options.extensions ?? []),
+      CollaborationExtension(options.collaboration),
+    ],
+    // We disable the default prosemirror history plugin, since it's not compatible with yjs
+    disableExtensions: ["history", ...(options.disableExtensions ?? [])],
+    // We don't want the default initial content, since it will generate a random id for the initial block on each client,
+    // leading to conflicts when syncing happens afterwards.
+    initialContent: [{ type: "paragraph", id: "initialBlockId" }],
```
YousefED, line 89: should we keep the warning for when the consumer has supplied initialContent? (the warning that was removed from `BlockNoteEditor`)

packages/core/src/yjs/comments/RESTYjsThreadStore.ts
```diff
@@ -47,14 +47,8 @@ export class RESTYjsThreadStore extends YjsThreadStoreBase {
...
-      prosemirror: {
-        head: number;
-        anchor: number;
-      };
-      yjs: {
-        head: any;
-        anchor: any;
-      };
+      head: number;
+      anchor: number;
```
YousefED, line 51: The REST call below will need the yjs positions to create the comment server-side right?

See the POC demo [here](https://github.com/TypeCellOS/BlockNote-demo-nextjs-hocuspocus/blob/b1a7a949e9d6714c4bc2e42699106f9ed65e2d58/hocuspocus-server/src/threads.ts#L92)

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -501,17 +494,6 @@ export class BlockNoteEditor<
 
     const tiptapExtensions = this._extensionManager.getTiptapExtensions();
 
-    const collaborationEnabled =
-      this._extensionManager.hasExtension("ySync") ||
-      this._extensionManager.hasExtension("liveblocksExtension");
```
YousefED, line 506: I think we should check if the new setup if still compatible with LiveBlocks and update their examples / blocknote package

YousefED (commented):
Super nice how this is extracted with `withCollaboration`!

packages/core/src/yjs/comments/RESTYjsThreadStore.ts
nperez0111: Yea, this is what I mentioned, if you want the positions you'll have to grab them yourself.
Docs doesn't use this if you were ever wondering: https://github.com/suitenumerique/docs/blob/51364e3d3ceac5fb94b3b6a8fac66bdf9e98533f/src/frontend/apps/impress/src/features/docs/doc-comments/api/DocsThreadStore.tsx#L28

packages/core/src/yjs/extensions/index.ts
nperez0111: It'll just be ignored, which I think is fine 🤷

packages/core/src/editor/BlockNoteEditor.ts
nperez0111: confirmed, fine for liveblocks, they inject a tiptap extension for this separately

packages/core/src/yjs/extensions/index.ts
YousefED: Let's see, won't be the first time users ask "why is `initialContent` not working"

packages/core/src/yjs/comments/RESTYjsThreadStore.ts
YousefED: Ah, you mean, get the yjs positions in `RESTYjsThreadStore` for example?

PR #2799, opened 2026-05-28
draft: suggestion / diffing tests

YousefED:
This draft PR adds visual testing for a large number of diffing scenarios outlined [here](https://docs.blocknotejs.mosacloud.eu/docs/e7d5f65d-c73a-46bf-a17e-6917eacd62e7/)

The goal is two-fold:
a) Serve as an overview of how far we are with implementing the UX / handling edge-cases
b) prevent regressions going forward

Todo:
[ ] integrate with CI
[ ] I think after the latest update to y-prosemirror, more tests are failing than before (try re-running the tests now)


nperez0111:
^ added the latest changes which update to latest @y/prosemirror & moves modification CSS to core for modification rendering

nperez0111:
OK, so it looks like you were just not running with `pnpm run build` & using something which did not have the modification css that I had added. I made vitest pull the css from source so this should not happen. I also regenerated all of the screenshots now that they can have more CSS to them.

Before this, I updated to the latest y/prosemirror so we can check with the new y/prosemirror replacement algo. This fixed a couple of broken tests so I re-enabled them. Screenshots & tests should be accurate now.

Issue #2801, opened 2026-05-30
Cursor goes out of view when pressing enter on the last line of a document


Wqrld:
### What’s broken?

when pressing enter on the last line of a document, i expect the view to scroll with it. Instead the cursor goes out of frame until you start typing characters.

Reproducible on https://www.blocknotejs.org/examples/basic/minimal and LaSuite Docs

Interestingly, i can not reproduce on Tiptap's demo pages.

### What did you expect to happen?

The page autoscrolls to keep up, ideally with a configurable margin.

### Steps to reproduce

 1. hold enter until at the end of the document
 2. on the last line, type a word
 3. press enter
 4. cursor goes out of view
 5. start typing
 6. cursor returns to correct position

### Environment

Firefox, LaSuite docs

### Additional context

I have the following workaround, but this should really be done in blocknote itself

```diff
diff --git a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
index 3f5dbef9..992578e4 100644
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
@@ -225,6 +225,25 @@ export const BlockNoteEditor = ({ doc, provider }: BlockNoteEditorProps) => {
       },
       uploadFile,
       schema: blockNoteSchema,
+      _tiptapOptions: {
+        editorProps: {
+          scrollMargin: { top: 0, right: 0, bottom: 72, left: 0 },
+        },
+        // BlockNote's splitBlock command (Enter on non-empty blocks) never calls
+        // tr.scrollIntoView(), so ProseMirror's scrollMargin doesn't fire for it.
+        // This ensures every transaction that moves the cursor triggers a scroll,
+        // letting ProseMirror's own scrollRectIntoView handle the margin uniformly.
+        onTransaction: ({ editor: tiptapEditor, transaction }) => {
+          if (
+            (transaction.docChanged || transaction.selectionSet) &&
+            !transaction.scrolledIntoView
+          ) {
+            tiptapEditor.view.dispatch(
+              tiptapEditor.view.state.tr.scrollIntoView(),
+            );
+          }
+        },
+      },
     },
     [
       aiExtension,
```
### Videos

workaround:
[...]

PR #2802, opened 2026-05-30
fix(core): call `scrollIntoView()` on `Enter`

Wqrld:
# Summary

Updates the enter handler to scrollintoview both globally and for listItems.

Part of https://github.com/TypeCellOS/BlockNote/issues/2801

## Rationale

Without this handler, the cursor goes out of frame when pressing enter on the last line in the window

## Changes

adds a scrollintoview call for the normal and listitem enter handler

## Impact

None expected.

## Testing

Tested in the playground environment. Does not overscroll the margins yet as in the demo video.

## Screenshots/Video

See referenced issue

## Checklist

## Additional Notes

There should be support for configuring the margin aswell (see original issue). I can try to add that here or in a separate PR

Disclaimer: Code was AI assisted, but tested by a human and PR text is human.

## Summary by CodeRabbit

* **New Features**
  * Allow passing editor configuration props (excluding internally managed attributes and paste handling) when creating the editor, enabling safer customization of editor behavior.

* **Bug Fixes**
  * Ensure pressing Enter to split blocks or empty list items scrolls the newly created content into view automatically.

### DIFF
--- a/packages/core/src/blocks/utils/listItemEnterHandler.ts
+++ b/packages/core/src/blocks/utils/listItemEnterHandler.ts
@@ -34,10 +34,9 @@ export const handleEnter = (
   } else if (blockContent.node.childCount > 0) {
     return editor.transact((tr) => {
       tr.deleteSelection();
-      const result = splitBlockTr(tr, tr.selection.from, true);
       tr.scrollIntoView();
-      return result;
-    });
+      return splitBlockTr(tr, tr.selection.from, true);
+});
   }
 
   return false;
--- a/packages/core/src/blocks/utils/listItemEnterHandler.ts
+++ b/packages/core/src/blocks/utils/listItemEnterHandler.ts
@@ -34,7 +34,9 @@ export const handleEnter = (
   } else if (blockContent.node.childCount > 0) {
     return editor.transact((tr) => {
       tr.deleteSelection();
-      return splitBlockTr(tr, tr.selection.from, true);
+      const result = splitBlockTr(tr, tr.selection.from, true);
+      tr.scrollIntoView();
+      return result;
     });
   }
 
--- a/packages/core/src/editor/BlockNoteEditor.ts
+++ b/packages/core/src/editor/BlockNoteEditor.ts
@@ -114,20 +114,6 @@ export interface BlockNoteEditorOptions<
   domAttributes?: Partial<BlockNoteDOMAttributes>;
 
   /**
-   * Additional ProseMirror editor props to pass to the underlying editor view.
-   * Use this to configure options such as `scrollMargin`, `scrollThreshold`,
-   * `handleDOMEvents`, and other
-   * [ProseMirror EditorProps](https://prosemirror.net/docs/ref/#view.EditorProps).
-   *
-   * Note: `attributes` (use {@link domAttributes} instead) and `transformPasted`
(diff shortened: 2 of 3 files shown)


Wqrld:
Would you be willing to promote editorprops to something that does not need accessing the internal _tiptapOptions  so the margins and maybe other editorprops can be set nicely? 

(Ref workaround diff in the original issue)

YousefED (commented):
Thanks for this.

PR seems good, but I'd rather not expose the generic "editorProps", but rather design our own API surface / properties for the scroll configuration.

wdyt @nperez0111 ?

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -120,6 +120,20 @@ export interface BlockNoteEditorOptions<
...
+   * `handleDOMEvents`, and other
+   * [ProseMirror EditorProps](https://prosemirror.net/docs/ref/#view.EditorProps).
+   *
+   * Note: `attributes` (use {@link domAttributes} instead) and `transformPasted`
+   * are managed by BlockNote and cannot be overridden here.
+   */
+  editorProps?: Omit<
+    NonNullable<EditorOptions["editorProps"]>,
+    "attributes" | "transformPasted"
+  >;
```
nperez0111: I'm not sure that I want to expose this @Wqrld, can you explain what you would be using this for? If it is just for scrollMargin, I think we could set that default within BlockNote instead of having to expose this whole config on our API surface for it.

packages/core/src/blocks/utils/listItemEnterHandler.ts
```diff
@@ -34,7 +34,9 @@ export const handleEnter = (
   } else if (blockContent.node.childCount > 0) {
     return editor.transact((tr) => {
       tr.deleteSelection();
-      return splitBlockTr(tr, tr.selection.from, true);
+      const result = splitBlockTr(tr, tr.selection.from, true);
+      tr.scrollIntoView();
+      return result;
     });
```
nperez0111:
```suggestion
      tr.scrollIntoView();
      return splitBlockTr(tr, tr.selection.from, true);
});
```

The scrollIntoView does not have to be done after

nperez0111 (changes requested):

nperez0111:
Thanks for the contribution, I agree with the goal here, just some minor details to work through

packages/core/src/editor/BlockNoteEditor.ts
Wqrld: It is indeed just for scrollmargin and scrollthreshold. Any way that exposes those in some way or another works for me.

nperez0111: What values are you using for those? I don't think I would expose that low-level of a concept. I would just pick a value that makes sense for BlockNote already

Wqrld: I was playing around with 72 for scrollmargin and 0 for scrollThreshold. What feels right depends a bit on the font and editor size so i prefer to make it configurable (with good defaults)

Wqrld:
Hi @nperez0111 Can you let me know where you'd want to place these settings? 

If it's easier for you to just steal my scrollintoview and make it into a commit with the configurability feel fee to do that aswell. This is just a thing that has been nagging me for weeks ;)

PR #2821, opened 2026-06-04
fix(table): guard stale block after changes for table handles

satoren:
## Summary

Fixes a `TypeError: Cannot read properties of undefined (reading 'id')` in `TableHandlesView.mouseMoveHandler` when hovering a table wrapper after the underlying table block was removed or became invalid (e.g. collaborative edits / Yjs DOM replacement).

## Root cause

1. `update()` refreshes `this.state.block` via `getBlock(id)`. When the block is gone, `block` becomes `undefined`, but the PluginView kept the `state` object (only `show` was set to `false`).
2. On a later `mousemove` over the table wrapper, this expression ran before new state was assigned:

   ```ts
   this.state?.block.id !== tableBlock.id
   ```

   Because optional chaining stops at `block`, `(this.state?.block).id` throws when `block` is `undefined`.

## Fix

- Use `this.state?.block?.id` in the wrapper hover branch.
- When the table block is invalid in `update()`, clear internal PluginView state (`this.state`, `tableId`, `tableElement`) instead of leaving a stale object.
- Allow `emitUpdate` to pass `undefined` to the extension store (matching existing `state.block ? … : undefined` logic).

## Reproduction (consumer apps)

1. Open an editor with collaboration enabled and a table in the document.
2. Hover the table so table handles state is initialized.
3. Remove or replace the table block remotely (or disconnect the table DOM).
4. Move the mouse to the table wrapper area (below/right of the table).

## Reported in

- ovice-ui Sentry: [OVICE-UI-15SZ](https://ovice.sentry.io/issues/7518310990/) (`@blocknote/core` 0.51, staging, `mouseMoveHandler`)

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes during table handle initialization by safely handling uninitialized state.
  * Resolved hover-related crashes with more robust table identification logic.
  * Ensured full cleanup of internal table tracking when a table becomes invalid or disconnected.
  * Allowed the table-handle state to be cleared to undefined to improve stability during view/store updates.

### DIFF
--- a/packages/core/src/extensions/TableHandles/TableHandles.ts
+++ b/packages/core/src/extensions/TableHandles/TableHandles.ts
@@ -163,13 +163,9 @@ export class TableHandlesView implements PluginView {
       any
     >,
     private readonly pmView: EditorView,
-    emitUpdate: (state: TableHandlesState) => void,
+    emitUpdate: (state: TableHandlesState | undefined) => void,
   ) {
     this.emitUpdate = () => {
-      if (!this.state) {
-        throw new Error("Attempting to update uninitialized image toolbar");
-      }
-
       emitUpdate(this.state);
     };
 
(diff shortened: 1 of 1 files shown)


nperez0111:
Thanks for this contribution @satoren, you can use the pkg-pr-new version above if you need this fix, otherwise. We are working on some larger updates and probably will not make a release for a couple of weeks

PR #2830, opened 2026-06-04
fix: React custom blocks cause infinite rerender on mobile (BLO-1212)

### LINKED ISSUE #2804: Alert Block freezes editor when selected from Slash Menu on mobile
### What’s broken?

When using the Slash Menu (/) on mobile, selecting "Alert" block causes the editor to freeze.

In most cases, the editor remains stuck on the Slash Menu screen and becomes
completely unresponsive — no text input is accepted. Occasionally, after an extended wait,
the page crashes with error code 5.

Other block types in the Slash Menu work without issue — the problem is isolated to Alert Block only.

Our Alert Block implementation follows the official BlockNote example exactly:
https://www.blocknotejs.org/examples/custom-schema/alert-block-full-ux

Selecting Alert Block from the Slash Menu on mobile should insert an Alert block
into the editor without freezing, consistent with the behavior of all other block
types and consistent with desktop behavior.

### Steps to reproduce

1. Open https://www.blocknotejs.org/examples/custom-schema/alert-block-full-ux
2. Open Chrome DevTools → Toggle device toolbar (mobile emulation) — e.g. iPhone 12 Pro
3. Focus the editor and type "/"
4. Select "Alert" from the Slash Menu
5. Observe: editor freezes, Slash Menu remains visible, no text input accepted
   (occasionally: page crashes with error code 5 after extended wait)

Reproduced on the official playground with Chrome DevTools mobile emulation.

### BlockNote version

v0.51.3 (also reproduced on v0.47.3 — not a regression, appears to be a long-standing bug)

### Environment

_No response_

### Additional context

Slash Menu open — editor frozen, no input accepted (most common symptom)

Image

Page crash with error code 5 (occasional symptom after extended freeze)

Image

 
1. Open https://www.blocknotejs.org/examples/custom-schema/alert-block-full-ux
2. Open Chrome DevTools → Toggle device toolbar (mobile emulation)
3. Focus the editor and type "/"
4. Select "Alert" from the Slash Menu
5. Observe: editor freezes, Slash Menu remains visible, no text input accepted
   (occasionally: page crashes with error code 5 after extended wait)

Reproduced on the official playground with Chrome DevTools mobile emulation.

### Contribution

### Sponsor

matthewlipski:
# Summary

This PR fixes an issue with React custom blocks sometimes causing infinite re-render loops on mobile. This can be seen most easily with the [Alert Block with Full UX](https://www.blocknotejs.org/examples/custom-schema/alert-block-full-ux) example, where attempting to insert an alert block via the slash menu will result in an infinite re-render loop and crash. This can be replicated via the mobile emulator in Chrome dev tools.

The reason this happens is that the way ProseMirror detects when a block's editable content has changed is different on desktop vs mobile. On desktop, it's detected using key events within the editable content wrapper element. On mobile though, key inputs don't use actually use typical key events, but insert characters using IME composition (see [here](https://discuss.prosemirror.net/t/contenteditable-on-android-is-the-absolute-worst/3810) and [here](https://github.com/facebook/react/issues/6176)). Therefore, ProseMirror instead detects changes in the editable content based on DOM mutations. Yet for some reason, re-renders are triggered by DOM mutations anywhere in the block, not just within its editable content.

And so the problem is that React's own rendering also causes DOM mutations, leading to ProseMirror triggering a re-render, causing React to render the block again, and starting an infinite loop. Therefore, we need to explicitly tell ProseMirror to ignore mutations outside the editable content to avoid this from happening.

Closes #2804

## Rationale

This is a severe issue on mobile.

## Changes

- Added `ignoreMutation` to `createReactBlockSpec` to ignore DOM mutations outside of block's editable content.

## Impact

N/A

## Testing

N/A (we don't have testing infrastructure on mobile)

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Performance**
  * Reduced unnecessary editor re-renders triggered by React-driven DOM changes outside editable content areas, improving responsiveness and smoothing block editing. Selection changes remain unaffected.
* **Chores**
  * No changes to public APIs or exported types; internal behavior adjusted only.

### DIFF
--- a/packages/react/src/schema/ReactBlockSpec.tsx
+++ b/packages/react/src/schema/ReactBlockSpec.tsx
@@ -315,6 +315,39 @@ export function createReactBlockSpec<
               },
               {
                 className: "bn-react-node-view-renderer",
+                // Any change to a block's content causes ProseMirror to
+                // re-render the block. On desktop, this change is detected
+                // using key presses. On mobile, it's instead detected using
+                // DOM mutations as touchscreen keyboards don't use typical key
+                // events. However, React's own rendering causes DOM mutations,
+                // triggering ProseMirror to re-render the block even if the
+                // mutation is outside the block's editable content. Therefore,
+                // we need to explicitly ignore mutations outside the block's
+                // editable content.
+                ignoreMutation: ({ mutation }) => {
+                  if (mutation.type === "selection") {
+                    return false;
+                  }
+
+                  const target =
+                    mutation.target.nodeType === Node.ELEMENT_NODE
+                      ? (mutation.target as HTMLElement)
+                      : mutation.target.parentElement;
+                  const content = target?.closest("[data-node-view-content]");
+
+                  // Ignore mutations outside a block's editable content.
+                  if (!content) {
+                    return true;
+                  }
+
+                  // Also ignore mutations for the editable content wrapper
+                  // element.
+                  if (mutation.target === content) {
+                    return true;
+                  }
+
+                  return false;
+                },
               },
             )(this.props!) as ReturnType<BlockImplementation["render"]>;
           } else {
--- a/packages/react/src/schema/ReactBlockSpec.tsx
+++ b/packages/react/src/schema/ReactBlockSpec.tsx
@@ -340,9 +340,14 @@ export function createReactBlockSpec<
                     return true;
                   }
 
-                  // Also ignore mutations for the editable content wrapper
-                  // element.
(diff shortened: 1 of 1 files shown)


nperez0111 (commented):
Let's double check with Tiptap's new React Renderer, just to make sure we aren't fixing something that is fixed.

matthewlipski:
Closing in favour of #2912

PR #2832, opened 2026-06-04
fix: Inline code clears comments (BLO-1207)

### LINKED ISSUE #2803: Tables are hard to un-nest
### What’s broken?

When you are filling out a table, you usually tab through while filling fields. When you press tab in the last block, the table gets indented instead. However, this is very hard to undo.

Image

Tables should probably not be indentable, or it should be able to easily unindent tables. The first sounds preferable to me, but might impact other people's workflows?

~~shift-tab seems to be the way to unindent elsewhere, but in tables it is used for going back cells.~~ Behavior seems to differ depending on where i look. playground does let me shift-tab on the first cell (but you can only indent from the last cell). Anyways the shift-tab shortcut is not very intuitive so i would rather not require it for unindenting tables.

### Steps to reproduce

1. Create table
2. go to last cell and press tab
3. Table gets indented
4. you cannot unindent by pressing backspace ~~(or shift-tab) ~~

### BlockNote version

_No response_

### Environment

_No response_

### Additional context

_No response_

### Contribution

### Sponsor

matthewlipski:
# Summary

This PR fixes an issues where comments cannot overlap inline code. This is because the inline code mark has `exclude: "_"` in its spec, meaning no other marks can overlap it. This makes sense for things like colors and bold/italic/etc, but not comments. Therefore, an exception has been added.

Closes TypeCellOS/BlockNote#2803 #3084

## Rationale

This is a bug.

## Changes

* Extended `Code` mark with `onBeforeCreate` to remove the comment mark from the `exclude` array.

## Impact

N/A

## Testing

Added e2e test.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue preventing inline code formatting from being applied to previously-commented text. Comment formatting is now preserved alongside code formatting, allowing both styles to coexist on the same text selection.


packages/core/src/blocks/defaultBlocks.ts
```diff
@@ -139,6 +140,23 @@ export const defaultStyleSpecs = {
...
+        const codeType = this.type as MarkType & {
+          excluded?: readonly MarkType[];
+        };
+
+        if (commentType && codeType.excluded?.includes(commentType)) {
+          codeType.excluded = codeType.excluded.filter(
+            (markType) => markType !== commentType,
+          );
+        }
+      },
```
nperez0111, line 159: I don't love this solution, can the `excluded` be a function in tiptap? And we can read the editor to see the other marks that exist in the schema. It being a runtime hack isn't great, especially when we can know all of the marks that the editor has included in it

nperez0111 (approved):
This looks good now, I couldn't find a simpler solution than to flatten the extensions straight from the options like you did

packages/core/src/blocks/defaultBlocks.ts
```diff
@@ -139,6 +139,21 @@ export const defaultStyleSpecs = {
   strike: createStyleSpecFromTipTapMark(Strike, "boolean"),
   code: createStyleSpecFromTipTapMark(
     Code.extend({
+      excludes() {
```
YousefED, line 143: I think we have `NON_FORMATTING_MARK_GROUP` for this right? cc @nperez0111 ?

nperez0111: Ah yea, true. we would need a function similar to nonFormattingMarks which sort of does the same thing based on group membership

PR #2842, opened 2026-06-08
feat: `moveBlocks` behaviour when leaving empty columns (BLO-1109)

### LINKED ISSUE #2594: Multi-columns errors when moving blocks
### What’s broken?

When moving blocks with ctrl+shift, we get some errors sometimes with Multi-columns.

https://github.com/user-attachments/assets/912635a8-a9af-4662-9523-c24d5e55bc61

I suppose we should stay on the left column when we move them up.
No errors should occurs.

### Sponsor

matthewlipski:
# Summary

This PR fixes an edge case when using `moveBlocks` with columns. `moveBlocks` first removes the target block(s), then inserts it/them above/below the previous/next block. In the removal step, `fixColumns` is called for all affected column lists, collapsing them or individual columns when they're empty.

This causes an error  in the following case:

```
<column>
  <paragraph></paragraph>
  <paragraph>Paragraph</paragraph>
</column>
```

When moving the non-empty paragraph block up, it's first removed. This leaves the column with a single empty paragraph, which `fixColumns` reads as an empty column, and so collapses it. Upon re-insertion, the reference block no longer exists and an error is thrown. This also happens when the order of the 2 blocks is flipped and the non-empty one is moved down.

After reviewing this, we have decided to change the behaviour of `moveBlocks` to not call `fixColumns`. If a column is left empty after a move, it's filled with an empty paragraph.

Closes #2594

## Rationale

This is a bug, but also made us realize that we want to change the behaviour of `moveBlocks` to not remove empty columns altogether.

## Changes

See above.

## Impact

N/A

## Testing

Added unit tests.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of block movement operations, ensuring moved blocks are removed and re-inserted reliably while maintaining surrounding layout structure.
  * Fixed edge cases for moving blocks within multi-column layouts, including scenarios involving empty sibling paragraphs.

* **Tests**
  * Added snapshot coverage for moving blocks up/down past empty siblings inside multi-column structures to prevent regressions.

### DIFF
--- a/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
@@ -164,7 +164,12 @@ export function moveBlocks(
     // When the non-empty block is moved up, the column is seen as empty and
     // collapsed in the removal step, so the following insertion fails.
     removeAndInsertBlocks(tr, blocks, [], { fixColumns: false });
-    insertBlocks(tr, flattenColumns(blocks), referenceBlock, placement);
+    insertBlocks<any, any, any>(
+      tr,
+      flattenColumns(blocks),
+      referenceBlock,
+      placement,
+    );
   });
 }
 
--- a/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
@@ -13,7 +13,6 @@ import { getNearestBlockPos } from "../../../getBlockInfoFromPos.js";
 import { getNodeById } from "../../../nodeUtil.js";
 import { insertBlocks } from "../insertBlocks/insertBlocks.js";
 import { removeAndInsertBlocks } from "../replaceBlocks/replaceBlocks.js";
-import { fixColumnList } from "../replaceBlocks/util/fixColumnList.js";
 
 type BlockSelectionData = (
   | {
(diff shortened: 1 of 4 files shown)


YousefED:
@matthewlipski I pushed 2f1af8f on top of your fix — could you review? It changes the approach based on a policy decision:

**Moves no longer collapse emptied columns at all.** Instead of deferring `fixColumnList` until after re-insertion, `moveBlocks` now simply skips it: a move is a rearrangement rather than a deletion, so it shouldn't restructure the document as a side effect. This makes moves reversible (moving a block back restores the original layout), and matches what dragging a block out of a column already does (native PM drops never ran `fixColumnList` either). If a move empties a column, the column stays, with an empty paragraph — the user can see it and delete it deliberately. Explicit deletions (`removeBlocks`/`replaceBlocks`) still clean up as before.

Follow-on simplifications:
- `removeAndInsertBlocks` keeps your `fixColumns` opt-out but no longer collects/returns `affectedColumnLists` — nothing consumes it anymore.
- The columnList reference anchoring workaround at the top of `moveBlocks` is removed: it guarded against `fixColumnList` dissolving the reference columnList mid-move, which can no longer happen. The `moveBlocksDown > Selection across columns` test covers exactly that case (reference = the columnList itself) and passes through the plain path.
- 4 snapshots re-recorded — all in multi-block-selection scenarios that empty out a source column. They now show the columnList surviving with an empty column instead of dissolving into loose paragraphs. Your two "empty sibling" regression tests pass unchanged.

Since this is now a behavior change rather than just an error fix, the PR title/description should probably be updated to say so (e.g. "moveBlocks no longer collapses emptied columns") — that's worth surfacing in the changelog.

matthewlipski:
> @matthewlipski I pushed [2f1af8f](https://github.com/TypeCellOS/BlockNote/commit/2f1af8fbfc94e80b24837a28f970d3fc362f8c1e) on top of your fix — could you review? It changes the approach based on a policy decision:
> 
> **Moves no longer collapse emptied columns at all.** Instead of deferring `fixColumnList` until after re-insertion, `moveBlocks` now simply skips it: a move is a rearrangement rather than a deletion, so it shouldn't restructure the document as a side effect. This makes moves reversible (moving a block back restores the original layout), and matches what dragging a block out of a column already does (native PM drops never ran `fixColumnList` either). If a move empties a column, the column stays, with an empty paragraph — the user can see it and delete it deliberately. Explicit deletions (`removeBlocks`/`replaceBlocks`) still clean up as before.
> 
> Follow-on simplifications:
> 
> * `removeAndInsertBlocks` keeps your `fixColumns` opt-out but no longer collects/returns `affectedColumnLists` — nothing consumes it anymore.
> * The columnList reference anchoring workaround at the top of `moveBlocks` is removed: it guarded against `fixColumnList` dissolving the reference columnList mid-move, which can no longer happen. The `moveBlocksDown > Selection across columns` test covers exactly that case (reference = the columnList itself) and passes through the plain path.
> * 4 snapshots re-recorded — all in multi-block-selection scenarios that empty out a source column. They now show the columnList surviving with an empty column instead of dissolving into loose paragraphs. Your two "empty sibling" regression tests pass unchanged.
> 
> Since this is now a behavior change rather than just an error fix, the PR title/description should probably be updated to say so (e.g. "moveBlocks no longer collapses emptied columns") — that's worth surfacing in the changelog.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

I think it's a good change, don't see issues with it. The one caveat I have is that drag & drop is also a rearrangement rather than a deletion, yet drag & drop does remove empty columns. I think for logical consistency, maybe these should have the same behaviour. But I'd rather do that in a separate PR if anything.

YousefED:
> I think it's a good change, don't see issues with it. The one caveat I have is that drag & drop is also a rearrangement rather than a deletion, yet drag & drop does remove empty columns. I think for logical consistency, maybe these should have the same behaviour. But I'd rather do that in a separate PR if anything.

Good point. @virgile-dev do you have an opinion on this? (It would make dragging the last block out of a column not remove that column as it is now (but keep an empty block around)

PR #2857, opened 2026-06-16
feat: math block

matthewlipski:
# Summary

This PR adds a math block which renders LaTeX, math inline content which does the same, and a diagram block which renders Mermaid diagrams. These are under the `xl-math-block` and `xl-diagram-block` packages respectively.

To facilitate the new blocks, some changes to the core editor and syntax highlighting have been made.

### Plain Text Content

Blocks and inline content can now hold plain text instead of rich text. Internally, this is represented as an array of `StyledText` objects, but with an empty `styles` field:

```ts
type PlainContent = {
  type: "text";
  text: string;
  styles: {};
}[];
```

All of the new blocks/inline content use this, and the code block has also been updated to use plain content as well.

### Syntax Highlighting

Rather than syntax highlighting being bundled with the code block, it's now a separate configurable extension. Configuration works mostly the same, and if multiple instances of the extension are passed, they are deduplicated automatically. This change was necessary as the new blocks/inline content also need syntax highlighting, so it no longer makes sense to couple that to the code block.

### Code Block

The code block has received a small refactor in addition to the already mentioned changes, splitting it into smaller, more composable pieces.

## Rationale

There has been demand for a math block for a while, and to a lesser extent a diagram block too. This PR implements both as well as making more foundational changes to do so properly.

## Changes

- Updated docs.
- Added math/diagram blocks/inline content to existing interoperability examples.
- Added examples for adding math/diagram blocks/inline content to the editor.
- Refactored core code block.
- Added `InlineContentBoundaryEditExtension` for better keyboard handling around empty inline content.
- Added `SourceBlockWithPreviewExtension`/`SourceInlineContentWithPreviewExtension`. These work in tandem with the `SourceBlockWithPreview`/`SourceInlineContentWithPreview` components, which show a preview of source code in the block's text content. Clicking/hitting enter on the preview opens a popup which contains the source code. The extensions handle mostly keyboard navigation, and are added to the editor by default. In order to work properly, blocks using the components must set the `meta.hasPreview` flag.
- Added `SyntaxHighlightingExtension`, which is used for syntax highlighting and automatically deduplicates other instances of itself. Blocks/inline content can declare which language they need to be highlighted in using `meta.highlight` in their config.
- Added support for `"plain"` content for both blocks and inline content.
- Added math/diagram blocks/inline content, alongside slash menu and block type select items to use them in the editor.
- Added support for math/diagram exporters. For math, the blocks/inline content are converted to an editable equation in the target format, and fallback to an image of the equation if the target format doesn't support that. Diagram blocks are always converted to images.
- Adjusted styling for selected blocks.

## Impact

N/A

## Testing

- Added unit tests for `SyntaxHighlighting` extension & math block.
- Added math/diagram test cases for math/diagram import/export.

## Screenshots/Video

TODO

## Checklist

## Additional Notes

N/A

## Release Notes

* **New Features**
  * Added a math block that renders LaTeX formulas as MathML with an editable source editor popup
  * Added code block preview functionality with interactive source code popup
  * Syntax highlighting configuration moved to the editor level for simplified setup

* **Documentation**
  * Updated code block documentation to clarify syntax highlighting configuration

* **Internationalization**
  * Added "Add source code" button text translations across 24+ languages


docs/content/docs/features/blocks/code-blocks.mdx
```diff
@@ -44,15 +43,38 @@ type CodeBlockOptions = {
 
 `supportedLanguages:` The syntax highlighting languages supported by the code block, which is an empty array by default.
 
-`createHighlighter:` The [Shiki highliter](https://shiki.style/guide/load-theme) to use for syntax highlighting.
+**Syntax Highlighting**
 
-BlockNote also provides a generic set of options for syntax highlighting in the `@blocknote/code-block` package, which support a wide range of languages:
+Syntax highlighting is handled by a separate editor extension, configured at the editor level via the `syntaxHighlighting` option (not on the code block itself), so it can highlight any block's content:
```
YousefED: Curious if we can find a better way to do this. This will make it difficult to create easily installable block-plugins later I think

packages/core/src/blocks/Code/helpers/extensions/CodeKeyboardShortcutsExtension.ts
```diff
@@ -0,0 +1,120 @@
...
+        Tab: ({ editor }) => {
+          if (options.indentLineWithTab === false) {
+            return false;
+          }
+
+          return editor.transact((tr) => {
+            const { block } = editor.getTextCursorPosition();
+            if (block.type === blockType) {
+              // TODO should probably only tab when at a line start or already tabbed in
+              tr.insertText("  ");
```
YousefED, line 36: spaces or \t? or configurable?

packages/core/src/blocks/Code/helpers/render/createPreviewWithSourcePopup.ts
```diff
@@ -0,0 +1,170 @@
+import {
```
YousefED: do we prefer react / vanilla for these components? (cc @nperez0111 )

YousefED: first thought; vanilla makes it pretty complex to customize them in a block-view style manner right?

packages/core/src/blocks/Code/CodeBlockOptions.ts
```diff
@@ -0,0 +1,81 @@
...
+ * A `renderPreview` function is only responsible for the preview itself. It has
+ * no opinion on when, where, or how the preview is displayed - that's up to the
+ * code block's `render` function.
+ */
+export type CodeBlockPreview = (
+  block: BlockFromConfig<any, any, any>,
+  editor: BlockNoteEditor<any>,
+) => {
+  dom: HTMLElement;
+  ignoreMutation?: (mutation: ViewMutationRecord) => boolean;
```
YousefED: if a user wants to implement their own preview, would they need to implement `ignoreMutation`? It's a very prosemirror-style (magic 🤯) API. Can we figure an API design that abstracts this out? (maybe same for `dom` and `destroy` actually)

YousefED (commented):
Exciting!

Looking at the code from a high level (no user testing or deep dive into the functions), I have the following high-level questions:

- Should we use React or Vanilla for the components? (especially thinking about customizability)
- Curious if the current code design (with the keyboard handlers) also scales to interfaces where the editor is not in a pop-up (e.g.: Notion / TypeCell style editors that can collapse)

YousefED (commented):
Here's some feedback after a super early round of user-testing:

- [X] pressing up after “s” doesn’t go to the first popup:

Image


- [X] esc should close popup

- [X] formattingtoolbar appears when highlighting text.
- [x] I’m allowed to “bold” things. I think we should see if we can make sure at a lower level code blocks cannot allow formatted text. This probably also caused the issue with the exporters recently (remember?)
- [X] preview opens when dragging block
- [X] keyboard is buggy when two math blocks sit above/below combined with the “gapcursor”


- [x] this is a bit misaligned:

Image

This might also be nice for testing:
- [x] slashmenu support in demos
- [ ] collaboration example

docs/content/docs/features/blocks/code-blocks.mdx
nperez0111: I don't think syntaxHighlighting should be a top-level option like this. But as a single extension, maybe yea.
What's annoying is that it wants to be a singleton, but multiple places need to configure it. I'll think on it.

matthewlipski:
> Exciting!
> 
> Looking at the code from a high level (no user testing or deep dive into the functions), I have the following high-level questions:
> 
> * Should we use React or Vanilla for the components? (especially thinking about customizability)
> * Curious if the current code design (with the keyboard handlers) also scales to interfaces where the editor is not in a pop-up (e.g.: Notion / TypeCell style editors that can collapse)

- No strong preference for this, though admittedly React would be cleaner and these days I don't know if anyone is using BlockNote outside React.
- Maybe worth extending the custom block API with handling when the text cursor moves in to/out of the block, as imo these are the biggest pain points in my experience with writing the keyboard handling for the math block. Would have to be a bit limited though without exposing ProseMirror APIs.

packages/core/src/blocks/Code/CodeBlockOptions.ts
matthewlipski: Added this initially as I though it might be needed, but in practice we really just need the `dom` here. `ignoreMutation`/`update`/`destroy` are needed in `createSourceBlockWithPreview`, but not for rendering just the previews.

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
```diff
@@ -0,0 +1,408 @@
...
+  );
+
+  return {
+    dom: previewWithSourcePopup,
+    contentDOM: sourceBlock.contentDOM,
+    ignoreMutation: (mutation: ViewMutationRecord) =>
+      // Ignore mutations outside of the inline content container. Used mainly to ignore DOM
+      // changes caused preview updates.
+      !sourceBlock.contentDOM.parentElement ||
+      !sourceBlock.contentDOM.parentElement.contains(mutation.target),
```
nperez0111: I'll need to look into this later. But, I think we should make this the default no? Why would we ever want a mutation observer on this element?

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
```diff
@@ -0,0 +1,408 @@
...
+    editor,
+    previewWithSourcePopup,
+    isSourcePopupOpen,
+    setSourcePopupOpen,
+  );
+
+  const sourcePopupPositioner = positionSourcePopup(
+    previewContainer,
+    sourceBlockPopup,
+  );
```
nperez0111: I'm unsure whether this should be here (on rendering) or part of the block's extensions. I think the reason that it is here is for the `isSourcePopupOpen` and `setSourcePopupOpen`? But we can do that by getting the block's id in the dom and getting/setting the attr there.

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
```diff
@@ -0,0 +1,408 @@
...
+              blockContent?.getBoundingClientRect().width ??
+              rects.reference.width;
+            elements.floating.style.width = `${width}px`;
+          },
+        }),
+      ],
+    });
+    sourcePopup.style.left = `${x}px`;
+    sourcePopup.style.top = `${y}px`;
+  });
```
nperez0111: I kind of question this, I get that it is nice to have the slip behavior, but other than that I feel like it is sort of overkill? I think we can simplify the rendering without this and just have it always render to the bottom.

nperez0111:
Alright @matthewlipski, so we decided on the React topic that it is still undecided for now, but we would like to write an example using React for the math block and then see from there whether we would want it to be the default implementation or not.

I think we can move some of the logic that is currently in the render path right into the extension, but we can see how that plays out.

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
matthewlipski: Good point, I can't really think of any good reason not to make it a default

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
matthewlipski: The main reason it's here is because the keyboard handling and rendering is very tightly coupled, so I preferred to not have a separate extension since it would always and only be used when rendering `createSourceBlockWithPreview`. Do you think it makes more sense to create an extension instead for event/selection handling?

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
matthewlipski: Yeah I think that's valid. I will play around with it, if we keep the popup absolutely positioned then scroll handling and sizing might be a bit annoying which is the one thing I'm a bit wary of. Or do you think better to give it a go using CSS anchors?

nperez0111: What is needed for scroll handling? If it is absolute (relative to the container) then it would just scroll no?

Sizing is just always full-width right?

nperez0111: CSS anchors we can leave for another time, I think the browser support wasn't there yet

nperez0111: Actually browser support is there, there are just some niche things that are not yet supported but speced, but still no need to migrate to it, just an FYI https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/anchor-name

packages/math-block/src/helpers/parse/parseMathML.ts
```diff
@@ -0,0 +1,33 @@
+import { MathMLToLaTeX } from "mathml-to-latex";
```
nperez0111: Do we have an idea on how well this round trips? For example going: source latex -> html -> latex
I imagine is lossy. I wonder if katex has some test cases we can try this out on.

nperez0111: 187kb for this. IDK maybe this isn't worth bundling by default and can be a separate export

tests/src/unit/core/formatConversion/export/__snapshots__/html/complex/document.html
```diff
@@ -25,6 +25,6 @@ <h2 data-level="2">Section 1</h2>
 </ul>
 <hr />
 <blockquote>A notable quote</blockquote>
-<pre data-language="javascript">
+<pre>
```
nperez0111, line 28: intentional?

packages/code-block/package.json
```diff
@@ -52,12 +52,14 @@
     "@shikijs/core": "^4",
     "@shikijs/engine-javascript": "^4",
     "@shikijs/langs-precompiled": "^4",
-    "@shikijs/themes": "^4"
+    "@shikijs/themes": "^4",
+    "katex": "^0.16.11"
```
nperez0111: code block doesn't need katex right

packages/core/src/blocks/Code/helpers/extensions/CodeKeyboardShortcutsExtension.ts
nperez0111: was already there. I don't think we need to get into tabs v.s. spaces debate for code blocks

packages/core/src/extensions/SourceBlockWithPreview/SourceBlockWithPreview.ts
```diff
@@ -0,0 +1,140 @@
...
+          const { block } = editor.getTextCursorPosition();
+
+          store.setState((state) => ({
+            selected: blockHasPreview(block) ? block.id : undefined,
+            popupOpen:
+              state.popupOpen && state.popupOpen !== block.id
+                ? undefined
+                : state.popupOpen,
+          }));
+        });
```
nperez0111: should set the state conditionally, otherwise will churn subscribers for no-op updates

packages/math-block/src/helpers/render/MathPreview.tsx
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  // The source is hidden, so highlight the whole block while the cursor is in
+  // it. Mirrors the vanilla `createSourceBlockWithPreview` store sync.
+  const rootRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    rootRef.current
+      ?.closest(".bn-block-content")
+      ?.classList.toggle("ProseMirror-selectednode", selected);
+  }, [selected]);
```
nperez0111: We don't have to put it on the parent block-content element right? I'm sure that class would look fine on the wrapper element here. This is just for visual indication so it is not such a big deal and would help us skip this useEffect

packages/math-block/src/helpers/parse/parseMathML.ts
matthewlipski: For round trips we rely on [annotation elements](https://developer.mozilla.org/en-US/docs/Web/MathML/Reference/Element/annotation) which you can see we check for manually below. It's not guaranteed that external MathML will have these, but the katex MathML export includes them so the round trip is guaranteed to be lossless. `mathml-to-latex` only handles external MathML parsing, which may not include the annotations.

tests/src/unit/core/formatConversion/export/__snapshots__/html/complex/document.html
matthewlipski: Yep, the tests set supported languages which did not include `text` but didn't specify a default language, so that remained `text`. That means the default language wasn't supported and it caused problems with the tests, so I fixed that.

packages/core/src/extensions/SourceBlockWithPreview/SourceBlockWithPreview.ts
matthewlipski: Doesn't TanStack Store do a deep equals when setting the state? I thought that when updating the state, it will only trigger subscribers when the state has actually changed.

nperez0111: I'm pretty sure that it does a shallow equals, but could be wrong

examples/06-custom-schema/09-math-block/src/App.tsx
```diff
@@ -0,0 +1,128 @@
...
+      },
+      {
+        type: "math",
+        content: "\\int_0^\\infty e^{-x^2} dx = \\frac{\\sqrt{\\pi}}{2}",
+      },
+      {
+        type: "paragraph",
+        content: [
+          "Equations can also be inline, like ",
+          { type: "inlineMath", content: "e^{i\\pi} + 1 = 0" },
```
YousefED: wdyt, `inlineMath` or `math_inline`?

examples/06-custom-schema/09-math-block/src/App.tsx
```diff
@@ -0,0 +1,128 @@
...
+  return (
+    <BlockNoteView editor={editor} slashMenu={false}>
+      {/* Replaces the default Slash Menu. */}
+      <SuggestionMenuController
+        triggerCharacter={"/"}
+        getItems={async (query) => {
+          // Gets all default slash menu items.
+          const defaultItems = getDefaultReactSlashMenuItems(editor);
+          // Finds index of last item in "Advanced" group.
+          const lastAdvancedIndex = defaultItems.findLastIndex(
```
YousefED: don't we have a merge function for this?

packages/core/src/blocks/Code/helpers/render/createCodeBlock.ts
```diff
@@ -0,0 +1,98 @@
+import type { BlockNoteEditor } from "../../../../editor/BlockNoteEditor.js";
+import type { BlockFromConfig } from "../../../../schema/index.js";
+
+// Select dropdown to change the block's language. Assumes `block` has a `language` prop.
```
YousefED, line 4: this file is similar to the old version I suppose?

packages/core/src/blocks/index.ts
```diff
@@ -16,6 +16,12 @@ export * from "./Table/block.js";
 export * from "./Video/block.js";
 
 export { EMPTY_CELL_HEIGHT, EMPTY_CELL_WIDTH } from "./Table/TableExtension.js";
+export * from "./Code/helpers/extensions/CodeKeyboardShortcutsExtension.js";
```
YousefED: let's create index files? (we had a pattern for this in block-view, right?)

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
```diff
@@ -0,0 +1,229 @@
+import type { Node as ProsemirrorNode } from "prosemirror-model";
```
YousefED: I suppose this file would drop if we go react-only, right?

packages/math-block/src/helpers/getMathPlainTextContent.ts
```diff
@@ -0,0 +1,16 @@
+// Converts rich text content in math blocks/inline content to plain text.
+// Should be removed once we add plain text support for blocks/inline content
+export const getMathPlainTextContent = (content: unknown): string => {
```
YousefED: this can drop if / when we merge the plain text PR, right?

packages/math-block/src/block/react/render/MathBlockPreviewWithPopup.tsx
```diff
@@ -0,0 +1,104 @@
...
+      className={
+        "bn-preview-with-source-popup" +
+        (selected ? " ProseMirror-selectednode" : "")
+      }
+      data-open={popupOpen ? "true" : "false"}
+    >
+      <div
+        className="bn-preview-container"
+        contentEditable={false}
+        onMouseDown={handlePreviewMouseDown}
```
YousefED: tested on mobile? should be mousedown or pointerdown?

packages/math-block/src/block/react/render/MathBlockPreviewWithPopup.tsx
```diff
@@ -0,0 +1,104 @@
...
+          <pre>
+            <code ref={contentRef} />
+          </pre>
+          <div
+            className="bn-code-block-source-popup-ok-button-wrapper"
+            contentEditable={false}
+          >
+            <button
+              className="bn-code-block-source-popup-ok-button"
+              onMouseDown={handleOkButtonMouseDown}
```
YousefED: should be onClick?

packages/math-block/src/inlineContent/react/render/MathInlinePreviewWithPopup.tsx
```diff
@@ -0,0 +1,127 @@
...
+      className={
+        "bn-preview-with-source-popup" +
+        (selected ? " ProseMirror-selectednode" : "")
+      }
+      data-open={selected ? "true" : "false"}
+    >
+      <span
+        className="bn-preview-container"
+        contentEditable={false}
+        onMouseDown={handlePreviewMouseDown}
```
YousefED: same as before, double check which events we should use

packages/math-block/src/inlineContent/react/render/MathInlinePreviewWithPopup.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const selected = useExtensionState(SourceInlineContentWithPreviewExtension, {
+    editor,
+    selector: (state) => state.selected === pos,
+  });
+
+  const { mathMLString, error } = useLatexToMathMLString(source, true);
+
+  // Opens the popup when clicking the preview.
+  const handlePreviewMouseDown = (event: MouseEvent) => {
+    if (!editor.isEditable || !pos) {
```
YousefED: double check, should we use pos here, or call getPos()? could it be pos has drifted?

packages/core/src/extensions/SourceInlineContentWithPreview/SourceInlineContentWithPreview.ts
```diff
@@ -0,0 +1,138 @@
...
+  InputRule,
+  inputRules as inputRulesPlugin,
+} from "@handlewithcare/prosemirror-inputrules";
+import { Selection } from "prosemirror-state";
+
+/**
+ * Inline-content counterpart of {@link SourceBlockWithPreviewExtension}. Drives
+ * the source popup for inline content with a preview.
+ *
+ * Unlike the block version, the popup isn't toggled with a separate state flag:
```
YousefED, line 16: why do we choose to make this different from the block version? (I assume there's a UX trade-off?)

packages/math-block/src/inlineContent/SourceInlineContentWithPreviewExtension.ts
```diff
@@ -0,0 +1,138 @@
...
+      keyboardShortcuts: {
+        Enter: moveSelectionOut("after"),
+        Escape: moveSelectionOut("after"),
+        ArrowUp: moveSelectionOut("before"),
+        ArrowDown: moveSelectionOut("after"),
+      },
+      // Cannot use `inputRules` field as it only allows for converting matched content to blocks.
+      prosemirrorPlugins: [
+        inputRulesPlugin({
+          rules: [/\$([^$]+)\$$/, /\\\((.+?)\\\)$/].map(
```
YousefED: SourceInlineContentWithPreviewExtension is generic, but this inputrules is specific to math block if I see it correctly

packages/core/src/extensions/SourceInlineContentWithPreview/SourceInlineContentWithPreview.ts
```diff
@@ -0,0 +1,138 @@
...
+          });
+        });
+        signal.addEventListener("abort", unsubscribeSelectionChange);
+
+        // Sets `visibility: hidden` on the popup for a single frame when pressing up/down arrow
+        // keys. The popup is normally hidden through `opacity: 0`, which means it's still visible
+        // to the browser for navigation. Therefore, the up/down arrows can sometimes move the
+        // selection into the popup from unexpected positions, such as on the same line. Setting
+        // `visibility: hidden` makes the browser ignore it when determining the new selection.
+        const handleVerticalArrow = (event: KeyboardEvent) => {
```
YousefED, line 115: this seems hacky. @nperez0111 can you confirm this is the best way to handle this?

YousefED (commented):
Looks great! Still need to think about the API (react / vanilla), but here's a first sweep of comments

examples/06-custom-schema/09-math-block/src/App.tsx
matthewlipski: Should stay `inlineMath`, we already have `codeBlock` which sets the naming convention.

packages/core/src/blocks/Code/helpers/render/createCodeBlock.ts
matthewlipski: Yeah it's functionally the same, the select has just been separated out into its own component

packages/core/src/extensions/SourceInlineContentWithPreview/SourceInlineContentWithPreview.ts
matthewlipski: Yeah exactly, the left/right arrow keys can navigate in/out of inline math as if it's regular text this way

examples/06-custom-schema/09-math-block/src/App.tsx
nperez0111: It should be `math` & `mathBlock` if you wanted consistency (e.g. `code` & `codeBlock`)

YousefED:
Note on `patches/katex@0.16.47.patch`: prod builds (incl. the Vercel preview) rendered every formula as "Error" because rolldown/oxc mangles lone-surrogate escapes when re-emitting string literals (`"\uD800"` → `�d800`), corrupting KaTeX's lexer token regex — `\sqrt` lexed as `\s` ("Undefined control sequence"). Dev builds don't rewrite the strings, so it only showed in builds.

The patch rewrites the affected regex strings to ASCII-only escapes (`"[\\uD800-\\uDBFF]"`), which is semantically identical for `new RegExp()` and immune to the bug. Temporary: oxc has this scheduled as a Vite 8 blocker for Q3 2026 (oxc-project/oxc#15524, oxc-project/oxc#16886) — the patch can be dropped once vite-plus ships a fixed oxc.

packages/math-block/src/helpers/parse/parseMathML.ts
matthewlipski: @nperez0111 can I resolve this issue or do you wanna go over it together?

nperez0111: I really don't think that this is worth having for a fallback, if we round trip without it, I don't think we need to introduce it to every bundle just for the possibility of parsing it.

I've made an issue to remove a large part of their code (that we don't need since we run blocknote in the browser): https://github.com/asnunes/mathml-to-latex/issues/82 

But, honestly, I'm don't think it is worth depending on this for such an edge case. So, let's remove it

nperez0111: Done with 7bf6c7ecb2b53368a1df3b530dd9fbd35e2b9de7

packages/core/src/blocks/Code/helpers/render/createSourceBlockWithPreview.ts
matthewlipski: Fixed in #2912

PR #2861, opened 2026-06-17
feat(comments): confirm before discarding an unsaved comment (blo-1197)

YousefED:
# Summary

Show a confirmation prompt before discarding an unsaved comment in the new-comment composer, so users don't lose their text by accidentally clicking away.

## Rationale

As raised in discussion [#2742](https://github.com/TypeCellOS/BlockNote/discussions/2742): when writing a new comment, clicking outside the composer (or pressing Escape) immediately discards it, silently losing any text the user typed. Full draft support was considered too complex; the agreed approach was to warn the user before discarding when text has been entered.

## Changes

- When the new-comment composer is dismissed (click-outside / Escape) and the editor is **not empty**, a `window.confirm()` prompt asks the user before discarding. Cancelling keeps the composer open with the text intact; confirming discards as before. An empty composer closes silently, unchanged.
- Refactored ownership of the composer editor: it's now created and owned by `FloatingComposerController` (which owns the open/dismiss lifecycle) and passed down to `FloatingComposer`. This lets the dismiss handler check `isEmpty` directly. A fresh editor is created per pending comment, so it always starts empty.
- Added a translatable `comments.discard_pending_comment` string to all 24 locales.

## Impact

- Scope is limited to the new-comment composer (the case in #2742). Reply and edit-comment composers are unchanged.
- `FloatingComposer` is a public export; its props change from `{}` to a required `newCommentEditor`. In practice it can only be used via the controller (it depends on it for the editor and positioning), so real-world impact is minimal.

## Testing

- `tsc -b` passes for `@blocknote/core` and `@blocknote/react`.
- `eslint` passes on the changed files.
- Manually: typing a comment then clicking outside / pressing Escape prompts before discarding; an empty composer closes without a prompt.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a confirmation prompt when users attempt to discard a pending comment, preventing accidental loss of drafts. The confirmation is now available in 23+ languages including English, Spanish, French, German, Chinese, Japanese, Korean, and more.


YousefED:
@virgile-dev See a demo here: https://blocknote-git-comments-discard-pending-confirmation-typecell.vercel.app/collaboration/comments-with-sidebar

Two UX questions:
- We currently use the browser-native "confirm dialog". Would this suffice, or prefer a dedicated confirmation dialog?
- Do you think the confirmation dialog should be only enabled when composing a new comment, or also when replying / editing? (I think the former)

Internal (BlockNote): decide if we want to make this behavior configurable

virgile-dev:
Hey @YousefED thanks for taking at stab at this.
We can work with a browser alert for starters no problem. It would be consistent with the current behaviour in Docs when a user tries to leave and some content is not yet saved.

Regarding your second point I'd argue we need this for replies as well.
Indeed, we use both the comment modal and the side panel unlike you demo link. 
So if you reply but don't hit send when using the comment modal then your reply is lost.

And then for editing, I would just to do it to keep the behaviour consistent

YousefED:
I updated the example to:
- also confirm replies, see: https://blocknote-git-comments-discard-pending-confirmation-typecell.vercel.app/collaboration/comments
- replies in Thread sidebar are excluded, because the replies there are actually saved (when you go back to editing, the comment is still there)
- Whether or not to show the confirm dialog is configurable

> And then for editing, I would just to do it to keep the behaviour consistent

This turned out to be significantly more work unfortunately, so if it's ok, I kept this out of scope

virgile-dev:
That's fine for a a first pass ! You edit your comments much less often than you create a new one. Do you want to me to create another issue for the edit part ?

nperez0111 (approved):
If we have issues with this, we can follow this up. But this behavior is strictly better than what we have now, so I'll merge this now

PR #2868, opened 2026-06-23
feat(core): support "plain" block content (BLO-335)

YousefED:
# Summary

Adds a **"plain"** block content type — for blocks that hold an unstyled text **string** rather than rich inline content — and migrates the code block to use it. At the ProseMirror level a plain block is `text*`; in the block model its `content` is a `string`. Plain blocks disallow **formatting** marks but allow the **non-formatting** marks (comments and suggestions/diffs).

## Rationale

Some blocks (code blocks being the canonical example) should hold plain text only: no bold/italic/links, no inline-content array to reason about. Before this, the only options were `"inline"` (rich content) or `"none"` (no content). `"plain"` fills that gap with a simpler, string-typed content shape — while still supporting comments and tracked-changes on the block's text.

## Changes

- **Type system**: propagate `"plain"` through `BlockFromConfig` / `PartialBlock`, `createBlockSpec`, and the schema helpers so a block/inline-content spec declared with `content: "plain"` infers `string`.
- **Runtime**: node spec emits `text*`; `nodeToBlock` returns the text content as a string; `createSpec` parse/render handle plain; `schemaToJSONSchema` maps `plain → { type: "string" }` (xl-ai).
- **Code block**: migrated to `content: "plain"`.
- **Marks on plain blocks**: a shared `"annotation"` mark group is added to the comment mark and the suggestion marks (`insertion`/`deletion`/`modification`); plain blocks allow that group, so formatting marks are excluded but comments and suggestions/diffs are not. The always-present suggestion marks keep the group non-empty, so the reference is valid even when comments aren't configured (the comment mark is conditional).
- **Exporters**: docx / odt / pdf / email default schemas now read code-block content as a string.
- **Yjs backwards compatibility**: a **pre-sync** migration (`stripDisallowedMarks`). Legacy code blocks whose text carries *formatting* marks are otherwise deleted by y-prosemirror during reconstruction — and the deletion propagates to all peers. The migration strips only the disallowed (formatting) marks from the `Y.XmlFragment` before y-prosemirror touches it (keeping comments/suggestions), covering both documents present before mount and content that syncs in afterwards, and tears down once migration is done.
- **Tooling**: `tests/` is now type-checked/linted by `vp lint`; added a Commands section to `CONTRIBUTING.md`.

## Impact

- Existing `"inline"` / `"none"` / `"table"` blocks are unaffected (additive to the content union).
- **Breaking-ish for stored code blocks**: a code block previously held inline content (could be formatted); it now holds a plain string. Formatting is dropped at the block-model level. The Yjs migration ensures old collaborative documents load gracefully (keeping comments/suggestions) instead of losing the block.
- Limitations (documented in code): the Yjs pre-sync migration is one-shot (like the existing post-sync migration) — disallowed content arriving from an old-version peer after the first sync isn't re-migrated. Plain blocks reference the `"annotation"` group, which assumes the always-present suggestion marks keep it non-empty.

## Testing

- `contentTypePropagation.test.ts`: type-level assertions that `"plain"` (and the baseline kinds) propagate to the right `content` type for blocks and inline content. Verified it fails to compile if propagation regresses.
- `markGroups.test.ts`: a code block allows `insertion`/`deletion`/`modification` (and `comment` when comments are enabled) but not `bold`.
- `stripDisallowedMarks.test.ts`: rule unit tests + the two real collaborative flows (formatted code block present before mount, and synced in after mount) survive with content intact and the Yjs fragment uncorrupted; the migration keeps comment + suggestion marks while stripping formatting; the observer is removed once migration completes.
- Updated stale fixtures/snapshots (core code-block + yjs tests, `tests/` parse/export-equality JSON, xl-ai schema snapshot) to the new string shape.
- Full `@blocknote/core` suite green; exporter, xl-ai, and xl-multi-column suites green.

## Screenshots/Video

N/A — no visual changes.

## Checklist

## Additional Notes

Plain blocks deliberately disallow **formatting** marks (so code blocks can't be bolded/linked) but allow the **non-formatting** marks — comments and suggestions/diffs — via a shared `"annotation"` mark group. This keeps the "plain = unstyled string" semantics at the block-model level (those marks are `blocknoteIgnore` and don't appear in the block's content) while letting collaboration features work on code.

## Summary by CodeRabbit

* **New Features**
  * Code blocks now use plain text content, improving editing, pasting, and export behavior.
  * Plain-content blocks can now preserve text while stripping formatting, with allowed annotation-style marks still supported.

* **Bug Fixes**
  * Fixed compatibility issues when loading older documents that contained formatting in code blocks.
  * Improved collaboration handling so documents with disallowed marks no longer lose affected blocks.
  * Updated schema and conversions to keep code blocks stable across import/export and round trips.

* **Documentation**
  * Clarified local development and test commands in the contributing guide.


YousefED:
Note that this PR takes the approach of mapping `content: "plain"` in the spec to a block with type `content: string`;

This means instantiated blocks can now hold a "string" as their content.

An alternative would have been to map `content: "plain"` to `content: [{text: string, styles: {}]` (i.e.: an unstyled piece of text).

The advantage of the latter is that the type system of instantiated blocks is unaffected, while the former (what's implemented here) is more strictly typed.

cc @nperez0111

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -153,9 +163,22 @@ export function addNodeAndExtensionsToSpec<
       name: blockConfig.type,
       content: (blockConfig.content === "inline"
         ? "inline*"
-        : blockConfig.content === "none"
-          ? ""
-          : blockConfig.content) as TContent extends "inline" ? "inline*" : "",
+        : blockConfig.content === "plain"
+          ? "text*"
```
matthewlipski, line 188: Just noting down that it feels a bit weird that you can technically have multiple separate `text` nodes in the content, but doubt that this would realistically cause any issues.

packages/core/src/schema/markGroups.ts
```diff
@@ -0,0 +1,27 @@
...
+export const NON_FORMATTING_MARK_GROUP = "annotation";
+
+/**
+ * The marks that belong to {@link NON_FORMATTING_MARK_GROUP}, by name.
+ *
+ * Yjs schema migrations run before the ProseMirror schema is available, so they
+ * can't read group membership from the schema and use this list instead. Keep
+ * it in sync with the marks that declare `group: NON_FORMATTING_MARK_GROUP`.
+ */
+export const NON_FORMATTING_MARK_NAMES: readonly string[] = [
```
matthewlipski: Not a fan of having to keep information about which marks are non-formatting in multiple places and having to sync them. Imo would make more sense to just read `NON_FORMATTING_MARK_NAMES` in `createSpec`, and set the node `content` based on that. Then we can get rid of  `NON_FORMATTING_MARK_GROUP`.

packages/core/src/yjs/extensions/FixUpSchema.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    stripDisallowedMarks(fragment, editor.schema.blockSchema);
+
+    // Text preserved, formatting gone.
+    expect(fragment.toJSON()).toBe(
+      '<blockgroup><blockcontainer id="code-0"><codeblock language="javascript">const x = 1;</codeblock></blockcontainer></blockgroup>',
+    );
+  });
+
+  it("keeps comment and suggestion marks while stripping formatting", () => {
+    const doc = new Y.Doc();
```
matthewlipski, line 283: Minor, but any way to make this more readable?

matthewlipski (commented):
Looks good, main question is r.e. `NON_FORMATTING_MARK_GROUP` above. Will also need to remember to update docs.

packages/core/src/schema/blocks/createSpec.ts
YousefED: I think this is prosemirror standard syntax

packages/core/src/schema/markGroups.ts
YousefED: fixed

packages/core/src/blocks/Code/block.ts
```diff
@@ -62,7 +62,7 @@ export const createCodeBlockConfig = createBlockConfig(
           default: defaultLanguage,
         },
       },
-      content: "inline",
+      content: "plain",
```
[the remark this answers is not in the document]
nperez0111: ^ might be valid @YousefED

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -89,6 +92,13 @@ export function getParseRules<
                 }
               }
 
+              if (config.content === "plain") {
+                // Plain blocks hold unstyled text only, so we parse the
+                // element's text content directly into a single text node.
+                const text = (node as HTMLElement).textContent ?? "";
+                return text ? Fragment.from(schema.text(text)) : Fragment.empty;
```
nperez0111: I don't think this is accurate, it would drop comments no? I think we need to parse similar to inline & just use a DOMParser & let it choose what it understands or not

packages/core/src/schema/blocks/createSpec.ts
nperez0111: text* is because you can have marks that split up the text. Comments would do this.

packages/core/src/yjs/extensions/schemaMigration/migrationRules/stripDisallowedMarks.ts
```diff
@@ -0,0 +1,98 @@
...
+  const markKeys = new Set<string>();
+  for (const op of text.toDelta() as {
+    attributes?: Record<string, unknown>;
+  }[]) {
+    if (op.attributes) {
+      for (const key of Object.keys(op.attributes)) {
+        markKeys.add(key);
+      }
+    }
+  }
```
nperez0111: Hm, there are better ways to do this. I'm unsure if this is technically correct in all cases or if we should be iterating the text elements instead

packages/core/src/yjs/extensions/schemaMigration/migrationRules/stripDisallowedMarks.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const keysToStrip = [...markKeys].filter((key) => {
+    // y-prosemirror encodes "overlapping" marks (those that don't exclude
+    // themselves, e.g. comments) as `${markName}--${hash}` so multiple
+    // instances can coexist on a range. Resolve the base mark name before
+    // looking it up in the schema.
+    const markName = key.split("--")[0];
```
nperez0111: We can do this the same way as y-prosemirror, this isn't the most robust

packages/core/src/yjs/extensions/schemaMigration/SchemaMigration.ts
```diff
@@ -1,28 +1,85 @@
...
+    };
+    // `unobserveDeep` is a no-op if the observer isn't (or is no longer)
+    // registered, so this is safe to call more than once.
+    const stopPreSyncObserver = () => fragment.unobserveDeep(preSyncObserver);
+
+    // Migrate content that streams in from a provider after mount. Registered
+    // here — in the extension factory, before the editor mounts — so it runs
+    // before y-prosemirror's own observer and can clean the fragment first.
+    // Removed once migration is done (see `appendTransaction`).
+    fragment.observeDeep(preSyncObserver);
```
nperez0111: Oof, I'm unsure about this. This is just asking for a race condition. This isn't a very robust migration

nperez0111 (commented):
I'm a little wary of making `"plain"` only output text instead of just an inline content, but I guess I might just be over thinking it. If `"plain"` is expected to _ONLY_ contain plaintext then it is fine, but things like comments can never be represented in the JSON format then (which is intentional now, but unsure if we have cases in the future for this)

YousefED:
> I'm a little wary of making "plain" only output text instead of just an inline content, but I guess I might just be over thinking it. If "plain" is expected to ONLY contain plaintext then it is fine, but things like comments can never be represented in the JSON format then (which is intentional now, but unsure if we have cases in the future for this)

The current principle indeed is that we separate metadata (e.g. annotations like comments, suggestions) from the document representation. Of course, currently we don't expose any of this metadata yet so we don't know if this principle holds up. I'd say for now it's overthinking a bit though, because it's likely the existing array format might also not hold up for this and we need sth completely different.

Not saying I'm extremely attached to the pure-string format. I'd be ok with parsing `plain` blocks to InlineContent without styles`[{text: "content", styles{}]` (it's less "pure, strictly typed TS", but it does reduce some complexity) if you feel more comfortable with this. (I'd lean to keeping as-is in this PR)

**migration**
Given you're deeper into Yjs,  could you resolve the migration path or think of a better solution here? Maybe the cleanest would be to have a custom "repair" function in y-prosemirror when it finds it can't instantiate the text node? (instead of silently dropping it)?

nperez0111:
> The current principle indeed is that we separate metadata (e.g. annotations like comments, suggestions) from the document representation. Of course, currently we don't expose any of this metadata yet so we don't know if this principle holds up. I'd say for now it's overthinking a bit though, because it's likely the existing array format might also not hold up for this and we need sth completely different.
> 
> Not saying I'm extremely attached to the pure-string format. I'd be ok with parsing `plain` blocks to InlineContent without styles`[{text: "content", styles{}]` (it's less "pure, strictly typed TS", but it does reduce some complexity) if you feel more comfortable with this. (I'd lean to keeping as-is in this PR)

I'll let you decide on what to do here. I'll just say some things to consider:
 - BlockNote JSON compatibility: input and output
 - I think it is arguably easier for everything to be an array of inline content than to introduce a branch on strings
   - I typically lean toward the "liberal in what you parse, strict in what you output" rule

> **migration** Given you're deeper into Yjs, could you resolve the migration path or think of a better solution here? Maybe the cleanest would be to have a custom "repair" function in y-prosemirror when it finds it can't instantiate the text node? (instead of silently dropping it)?

I've added a commit with de7c391ffaf3ef6ac0436b00932699b36efc60c3 which I think should resolve this. Leaves the Y.Doc with the marks in-place, it, silently, will not materialize the mark on plain text content, which is arguably what y-prosemirror should have been doing in this scenario anyway.

packages/core/src/yjs/extensions/FixUpSchema.ts
```diff
@@ -0,0 +1,88 @@
...
+      content?: Fragment | Node | readonly Node[],
+      marks?: readonly Mark[],
+    ) {
+      const nodeType = typeof type === "string" ? this.nodes[type] : type;
+      // y-prosemirror always passes `content` as a plain array of nodes; only
+      // touch text children that carry marks.
+      if (
+        nodeType &&
+        isPlainContentNodeType(this, nodeType) &&
+        Array.isArray(content)
```
YousefED, line 73: @nperez0111 prefer doing it this way, or calling the original function with a try / catch and then try to "recover"?

packages/core/src/api/nodeConversions/blockToNode.ts
```diff
@@ -52,11 +52,23 @@ function styledTextToNodes<T extends StyleSchema>(
     }
   }
 
+  // Backwards compat: old BlockNote JSON may carry formatting marks (e.g. bold)
+  // on a block whose content type is now "plain". Those marks aren't allowed on
+  // the node, and would make `createChecked` throw when the block is assembled.
+  // Drop them here (for plain blocks only) — comment/suggestion (annotation)
+  // marks are allowed and kept by `allowedMarks`.
+  const allowedMarks =
```
YousefED, line 60: @nperez0111 we now have this backwards compatibility workaround and the pm function override. Do we need both?

YousefED (commented):
@nperez0111 I updated the code to return an inline content array, see last commit. Some minor questions posted as comments

packages/core/src/yjs/extensions/FixUpSchema.ts
nperez0111: I would prefer this way, there are many ways to fail and I would prefer to only do this in situations we know will be problematic.

packages/core/src/api/nodeConversions/blockToNode.ts
nperez0111: This one I will remove

nperez0111: Yep, we do need both. They are invoked differently, and this is the cleanest way to do it on the pm-only side

PR #2880, opened 2026-07-01
Enhance serialization with optional chaining and checks

Bredo:
# Summary
Added optional chaining to safely check for class presence and attribute setting. Improved handling of nesting level attribute assignment.

## Rationale
When serializing blocks to external HTML (via `blocksToHTMLLossy`), the serializer evaluates `(ret.dom as HTMLElement).classList.contains("bn-block-content")`. However, certain block `toExternalHTML` or render implementations might return node types that aren't strictly HTMLElements (such as DocumentFragment or Text nodes). These nodes do not have a `.classList` property, which causes the serializer to crash with `TypeError: Cannot read properties of undefined (reading 'contains')`.

I ran in to this specifically using @twentyhq/twenty with numbered lists `<ol>`

Additionally, if these node types are evaluated in the else branch and nestingLevel > 0, the exporter tries to call `.setAttribute("data-nesting-level", ...)` on them, causing a secondary crash (TypeError: ...setAttribute is not a function).

This PR guards both of these DOM operations to ensure the HTML exporter fails gracefully and doesn't crash when encountering non-element DOM nodes.

## Changes
- Added optional chaining to safely check if `.classList` exists before calling `.contains("bn-block-content")`.
- Guarded `data-nesting-level` assignments by checking if the `.setAttribute` method exists via `typeof === 'function'` before invoking it.

## Impact
This PR increases the stability of the HTML exporters. It prevents hard crashes during `blocksToHTMLLossy` and `blocksToFullHTML` when the editor encounters complex fragment or text structures. Existing functionalities will remain completely unaffected as this merely guards operations against undefined errors.

## Testing
Manual testing was performed by executing `blocksToHTMLLossy` on a document containing block structures that resolve to nodes lacking `.classList` properties. Previously, this would crash the frontend application; with this fix, the serialization completes successfully.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTML export resilience by handling missing DOM features more safely.
  * Prevented errors when certain element properties are unavailable during block serialization.
  * Made nesting-level attribute updates more defensive, reducing the chance of export failures.

### DIFF
--- a/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
+++ b/packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
@@ -227,7 +227,8 @@ function serializeBlock<
 
   const elementFragment = doc.createDocumentFragment();
 
-  if ((ret.dom as HTMLElement).classList.contains("bn-block-content")) {
+  // Needs optional chaining for handling <ol> lists export
+  if ((ret.dom as HTMLElement).classList?.contains("bn-block-content")) {
     const blockContentDataAttributes = [
       ...attrs,
       ...Array.from((ret.dom as HTMLElement).attributes),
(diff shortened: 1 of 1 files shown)


nperez0111:
@Bredo with this fix wouldn't your element not get the styling. Does this actually fix your issue

Bredo:
@nperez0111 I'm actively converting the the styling myself right now, this is a bugfix for a specific built-in exception rather than a deeper fix.

nperez0111:
Sorry, I think this fix is not valid then @Bredo, it doesn't fix the actual problem so there is no point in keeping it in BlockNote.

PR #2883, opened 2026-07-03
feat: reusable React API for source blocks with previews

YousefED:
TLDR: Separates some of the react components and introduces reusable `<PreviewWithSourcePopup>` component. (Is not a decision yet on what to expose or vanilla vs. react)

---
# Summary

Builds on #2857: extracts the math block's React popup UI into a reusable API in `@blocknote/react` for any block or inline content that renders a preview of its source (math today; mermaid, graphviz, etc. tomorrow), and fixes a few bugs found along the way.

## Rationale

In #2857 the React math block and inline math each hand-rolled the full preview/popup DOM, handlers, and store wiring. That made "build your own preview block" (or "customize the math block") mean copying ~100+ lines of internals. This PR makes the consumer story: **config + preview + optionally a popup hook** — the math renders themselves are now ~25-line compositions and serve as the copy-me template.

## Changes

**`@blocknote/react` — new `blocks/SourceWithPreview/`**

- `PreviewWithSourcePopup` — presentational shell (block + inline via an `inline` prop). The `error` prop accepts arbitrary elements, so consumers can render actions (e.g. a "fix with AI" button) alongside the error message.
- `block/SourceBlockWithPreview` & `inlineContent/SourceInlineContentWithPreview` — wired wrappers; callers only supply `source`, `preview`, and optional `error`.
- `block/useSourceBlockPreviewPopup` & `inlineContent/useSourceInlineContentPreviewPopup` — the public popup-state API. Both return the shared `SourcePreviewPopup` shape (`{ isOpen, isSelected, open, close }`), so consumers never touch `useExtension` or raw extension stores. Split per kind deliberately: the block popup is an explicit toggle while the inline popup is selection-driven (`isOpen === isSelected`), and a unified hook meant type-erasing casts and interleaved branches.
- `AddSourceButton` moved here from `@blocknote/math-block` (it was generic).

**`@blocknote/core`**

- `SourceInlineContentWithPreviewExtension` moved from math-block to core, next to its block counterpart, after extracting the math-specific input rules. The two extensions intentionally stay separate — the block/inline UX models differ enough that merging them only hid the differences behind indirection.

**`@blocknote/math-block`**

- New `MathInlineInputRulesExtension` (`$...$` and `\(...\)`), mirroring `MathBlockInputRulesExtension`.
- Block & inline React renders are now thin compositions over the new wrappers.

**Bug fixes**

- Exported/serialized inline math rendered with `data-open="true"` and `ProseMirror-selectednode`: outside the editor `getPos()` returns `undefined`, which matched the store's initial `undefined` state. The hooks now guard against this; the `inlineMath/basic` snapshot reflects the corrected output.
- Inline popup mouse handlers captured `getPos()` once at render time; positions shift without node views re-rendering, so actions could target stale positions. Position is now read fresh at action time.
- Regenerated two stale schema snapshots (`blocks.json`, `inlinecontent.json`) that were already failing on the base branch.

## Impact

No behavior changes to the math block or code block beyond the export fix. Spec-facing APIs are unchanged; the math package no longer exports `AddSourceButton` or the inline preview extension (now in core).

## Testing

- Existing suites pass: math-block (29), react unit (107, includes the inline popup keyboard/click behaviors driven through the new wrappers & hooks), core export (500).
- Manually verified in the playground: popup open/close via click/Enter/Escape/OK, in-place preview updates while typing, KaTeX error display, and both inline input rules.

## Additional Notes

- The `#2857` TODO about copy/paste while a block popup is closed remains open; the suggested direction is content-layer protection (a ProseMirror plugin via `handleTextInput`/`handlePaste`) instead of the current capture-phase keydown suppression, which also misses drop/IME/autocorrect.
- Known follow-up: `getExtension(factory)` returns only the last-registered instance per factory, so block + inline (or code block + math) preview extensions can't be looked up by factory — the new hooks avoid this, but a keyed lookup (e.g. `getExtension(factory, key)`) is the eventual fix.


packages/react/src/blocks/SourceWithPreview/PreviewWithSourcePopup.tsx
```diff
@@ -0,0 +1,90 @@
...
+   * Error from rendering the preview, shown below the source in the popup.
+   * Accepts arbitrary elements, so actions (e.g. a button that fixes the
+   * source) can be rendered alongside the error message.
+   */
+  error?: ReactNode;
+  /**
+   * Ref for the element holding the editable source content.
+   */
+  contentRef: (node: HTMLElement | null) => void;
+  onPreviewMouseDown?: MouseEventHandler;
```
matthewlipski: I'm not a fan of passing individual event handlers like this. I would rather pass a `setSelected` and `setPopupOpen` then have `PreviewWithSourcePopup` create the mouse down handlers. Alternatively, if we really want to make `PreviewWithSourcePopup` as "dumb" as possible, I would still rather make it possible to pass all props rather than picking out just the mouse down handlers (`previewContainerProps: HTMLAttributes<HTMLElement>; okButtonProps: HTMLAttributes<HTMLButtonElement>;`).

packages/react/src/blocks/SourceWithPreview/PreviewWithSourcePopup.tsx
```diff
@@ -0,0 +1,90 @@
...
+export type PreviewWithSourcePopupProps = {
+  /**
+   * Renders with `span` wrappers so the shell can be used for inline content.
+   * Defaults to `false`, which renders with `div` wrappers for blocks.
+   */
+  inline?: boolean;
+  /**
+   * Whether the source popup is shown.
+   */
+  popupOpen: boolean;
```
matthewlipski: We should standardize the naming of `popupOpen` & `selected` to `isOpen` & `isSelected`.

packages/react/src/blocks/SourceWithPreview/inlineContent/SourceInlineContentWithPreview.tsx
```diff
@@ -0,0 +1,91 @@
+import { BlockNoteEditor } from "@blocknote/core";
+import { MouseEvent, ReactNode } from "react";
+
+import { AddSourceButton } from "../AddSourceButton.js";
+import { PreviewWithSourcePopup } from "../PreviewWithSourcePopup.js";
+import { useSourceInlineContentPreviewPopup } from "./useSourceInlineContentPreviewPopup.js";
+
+export type SourceInlineContentWithPreviewProps = {
+  editor: BlockNoteEditor<any, any, any>;
```
matthewlipski: Can be picked out from `InlineContentImplementation` instead of manually re-typed. Same for the block equivalent.

packages/math-block/src/inlineContent/react/createReactMathInlineContentSpec.tsx
```diff
@@ -23,7 +24,8 @@ export const createReactInlineMathSpec = () =>
     [
       SourceInlineContentWithPreviewExtension({
         key: INLINE_MATH_PREVIEW_KEY,
-        inlineContentType: "inlineMath",
+        inlineContentType: mathInlineContentConfig.type,
```
matthewlipski, line 27: If we're changing this here we should probably do the same for the math block

matthewlipski (commented):
Looks good, few comments

packages/react/src/blocks/SourceWithPreview/PreviewWithSourcePopup.tsx
YousefED: went for option 2 to keep this a presentation-only ("dumb") component

PR #2885, opened 2026-07-06
fix: multi-column trailing blocks, column hover borders & drop cursor left edge BLO-1226

### LINKED ISSUE #2534: Multi-block drag to column drop only moves the first block
### What’s broken?

When selecting multiple blocks and dragging them into an adjacent column, only the first of the selected blocks is placed into the column and the remaining blocks are seemingly left behind.

I'd expect all blocks in the selection to be included in the slice.content() and not just the first child. 

Currently if I have blockA, blockB and blockC and I select blockA and blockB together; then drag that selection to the left or right edge of blockC and release, I'd expect all selected blocks to appear in the new column.

### Steps to reproduce

https://www.blocknotejs.org/examples/basic/multi-column

Highlight all blocks in the far right column in this example then drag the selected blocks to another column. Only the first block is relocated to the other column. The others are left behind.

### BlockNote version

0.37.0

### Environment

Chrome 145.0.7632.117

### Additional context

Looking at `MultiColumnDropCursorPlugin.ts` the `handleDrop` extracts only the first node from the drag slice:

const draggedBlock = nodeToBlock(
  slice.content.child(0),
  editor.pmSchema,
);

Could we not iterate over all slice.content() children and create an array of selected blocks to be passed through the code?

### Contribution

### Sponsor

### LINKED ISSUE #2820: Multi-column bug. Add a block on right column creates a gap in right column
### What’s broken?

Add a block on right column creates a gap in right column

There shouldn't be gap

### Steps to reproduce

https://github.com/user-attachments/assets/9b130498-7bab-4d33-8d05-47aab4c0f3ba

### BlockNote version

Tested on https://www.blocknotejs.org/examples/basic/multi-column

### Environment

NixOS

### Additional context

_No response_

### Contribution

### Sponsor

### LINKED ISSUE #2943: Error when dragging blocks from @blocknote/xl-multi-column
### What’s broken?

The block disappears, and an error appears in the console. The video reproduces the bug in the demo at https://www.blocknotejs.org/examples/basic/multi-column.

https://github.com/user-attachments/assets/34e3a498-e641-4365-a566-5543282b040d

The block must take its place

### Steps to reproduce

Watch the playback in the attached video

### BlockNote version

"@blocknote/core": "0.47.3", "@blocknote/mantine": "0.47.3", "@blocknote/react": "0.47.3", "@blocknote/xl-multi-column": "0.47.3"

### Environment

150.0.7871.187

### Additional context

_No response_

### Contribution

### Sponsor

YousefED:
Fixes #2820. Fixes #2534. Fix https://github.com/TypeCellOS/BlockNote/issues/2943

Five multi-column UX fixes, plus the side menu positioning change they required.

## 1. Trailing block support for columns (#2820)

Clicking the empty space below a short column's last block used to do nothing useful, and clicking the strip below the whole multi-column area added a block *outside* the columns — which looked like it was added to a column until another column grew, leaving a "gap" (see the issue video).

The `TrailingNode` extension now renders its clickable trailing-block widget not just at the end of the document, but also at the end of every column, filling the column's leftover height (columns stretch to their tallest sibling, so short columns have dead space). Clicking it appends a paragraph *inside that column* and focuses it. As with the doc-level widget, it's a decoration: hidden in read-only mode, absent from exported/collaborative content, and skipped when the column already ends with an empty paragraph. Schemas without columns skip the doc traversal entirely.

## 2. Column hover borders & resize affordance

It was hard to tell where columns begin and end (part of the confusion behind #2820), and the resize border now has two improvements:

- **bugfix: the hover border did not always show.** The mousemove handler suppressed it whenever the side menu was *shown* — which is true almost any time the pointer is over a larger block. Only the mousedown path (which skips the check) worked, so resize-by-drag functioned but had no hover affordance. The check now uses a new `menuFrozen` getter on `SideMenuExtension` (i.e. drag-handle menu open), which is what the original check appears to have intended.
- **The side menu blocked the boundary next to headings.** The menu element was stretched via per-block-type CSS heights (108px next to an H1) to keep its buttons centered on the first line — and that invisible box swallowed mouse events over the column gutter. The menu now keeps its natural 30px height and is vertically centered via a floating-ui `offset({ crossAxis })` in `SideMenuController` instead. Button positions are pixel-identical; the `data-block-type`/`data-level`/`data-url` attributes that only existed for those CSS rules are removed.

**New behavior (UX improvement)**: hovering anywhere over a column list fades in light `#eee` separators between all its columns (showing each column's full extent, including its clickable empty space). Nearing a boundary instantly switches that boundary to the familiar `#ccc` resize border with a `col-resize` cursor. Both are driven by plugin state (`hover-column-list` / `hover-column` / `resize`) and class-based decorations (`bn-column-list-hovered`, `bn-column-resize-border`), so they survive the pointer crossing the side menu.

This last fix should make it easier for users to understand where columns are in the document

## 3. Drop cursor left-edge detection

`multiColumnDropCursor` treated a drop as a "left edge" (column-creating) drop when the pointer was within a margin *inside* the block. It now only does so when the pointer is actually left of the block.

This caused users to create column-based layouts accidentally (while the intention was to re-order blocks vertically)

## 4. Dragging a column's last block to create a column on the other side threw

In a 2-column layout, dragging the only block of a column to the other side of the column list (to create a new column there) threw `Block with ID ... not found` and did nothing. The drop handler removed the dragged block from the document *before* updating the column list - emptying its column dissolved the (then single-column) column list that the update was targeting. Blocks that are already in the column list are now moved purely via the children update instead of `removeBlocks`.

## 5. Multi-block drag to a column edge only moved the first block (#2534)

The drop handler only converted the first node of the drag slice, so dragging a multi-block selection to a column edge moved one block and left the rest behind. Worse: multi-block selections of list items produce drag fragments where the blocks are nested inside a `blockGroup` node, which made the handler throw (`Node should be a bnBlock, but is instead: blockGroup`) - the drop then did nothing at all. The slice is now converted with core's `fragmentToBlocks` (the same helper the drag/clipboard serialization already uses, now exported from `@blocknote/core`), and all dragged blocks are placed into the new column together.

## How to test

All in the **Multi-Column Blocks** playground example (`/basic/multi-column`):

**Trailing blocks (the #2820 repro):**
- Click the blank space directly below "This paragraph is in a column!" (left column) and type — the text becomes a new block *inside* the left column.

**Hover borders / resize:**
1. Hover anywhere over the columns — light separators fade in between all columns; move away and they disappear.
2. Move near a column boundary — it instantly darkens to the resize border and the cursor becomes `col-resize`; dragging resizes as before (border stays visible during the drag).
3. Hover an H1 next to a column boundary — the drag handle and the boundary remain reachable (previously the invisible 108px-tall menu blocked them). Side menu buttons render at the same position as before for all block types.

**Drop cursor:**
1. Drag a block over another block, pointer within the block's left portion — the drop indicator no longer suggests a side drop.
2. Drag it to the left of the block's left edge — the vertical (column-creating) drop indicator shows.

**Drag & drop into columns:**
1. Make a 2-column layout where one column has a single block. Drag that block via its drag handle to the far edge of the other column - a new column is created on that side (previously: console error, nothing happened).
2. Select multiple blocks (e.g. the example's bullet list items) and drag the selection via the drag handle to a column's edge - all selected blocks move into the new column (previously only the first block moved, or nothing happened with a console error).

Covered by two new e2e tests (`multicolumn.test.tsx`: clicking a column's trailing block, and hover borders persisting over the side menu) plus a new doc snapshot; existing e2e selectors were updated to target the doc-level trailing block specifically. The new `multicolumnDrop.test.tsx` covers the two drag fixes with snapshot tests, both verified to fail without the fixes. Full e2e suite, unit tests, and `vp check` pass.

## Summary by CodeRabbit

* **New Features**
  * Added multi-block drag-and-drop support for multi-column layouts.
  * Trailing insertion widgets now appear at eligible container ends, including columns.
  * Added access to side-menu frozen state and expanded node-conversion exports.

* **Bug Fixes**
  * Improved multi-column hover, resize, and drop-edge behavior.
  * Side-menu positioning now better adapts to selected block content.
  * Copy/paste targets the correct trailing insertion point.

* **UI/Style**
  * Simplified side-menu height styling.

* **Tests**
  * Added end-to-end coverage for multi-column trailing insertion and drag-and-drop.


tests/src/end-to-end/multicolumn/multicolumn.test.tsx
```diff
@@ -48,10 +61,34 @@ describe("Check Multi-Column Behaviour", () => {
...
+
+    // Hovering a block in a column shows the borders between the columns.
+    await moveMouseOverElement(page.getByText("So is this heading!").element());
+    await waitForSelector(".bn-column-list-hovered");
+
+    // The borders must stay visible when the mouse moves onto the side menu,
+    // which is rendered over the boundary between the first two columns.
+    await moveMouseOverElement(await waitForSelector(".bn-side-menu"));
+    expect(document.querySelector(".bn-column-list-hovered")).not.toBeNull();
+  });
```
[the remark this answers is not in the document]
YousefED: The assertion is deterministic rather than racy: mouse moves whose target is inside `.bn-side-menu` are ignored by the column resize plugin's mousemove handler (that's the behavior under test), so no state update can occur between `moveMouseOverElement` resolving and the assertion — the class provably still exists at that point. A retrying wait on presence would pass instantly without strengthening the test, and there's no "wait for it to stay present" primitive that wouldn't just be a sleep. The two nitpicks are addressed in 23539d5d3.

packages/core/src/extensions/TrailingNode/TrailingNode.ts
```diff
@@ -13,15 +13,10 @@ import {
...
-function shouldShowTrailingWidget(doc: PMNode, isEditable: boolean): boolean {
-  if (!isEditable) {
-    return false;
-  }
-
-  const rootGroup = doc.lastChild;
-  const lastBlock = rootGroup?.lastChild;
+// Skip the widget when the container already ends with an empty paragraph
+// block (since the user can just type into it).
+function containerNeedsTrailingWidget(container: PMNode): boolean {
```
matthewlipski, line 18: `isEditable` check not needed anymore?

packages/react/src/components/SideMenu/SideMenu.tsx
```diff
@@ -20,41 +17,8 @@ import { SideMenuProps } from "./SideMenuProps.js";
...
   const Components = useComponentsContext()!;
 
-  const editor = useBlockNoteEditor<any, any, any>();
-
-  const block = useExtensionState(SideMenuExtension, {
-    editor,
-    selector: (state) => state?.block,
-  });
-
-  const dataAttributes = useMemo(() => {
```
matthewlipski, line 30: We originally went with this approach to make it easier for people to adjust the side menu height for custom blocks they add, since it's just CSS. If we want to go back to FloatingUI for this, we should be certain that we're ok with this being more difficult for consumers to do.

packages/react/src/components/SideMenu/SideMenuController.tsx
```diff
@@ -9,6 +10,52 @@ import { FloatingUIOptions } from "../Popovers/FloatingUIOptions.js";
...
 
+// Returns the vertical offset of the side menu for the given block. Blocks
+// whose first line is taller than the side menu need the menu shifted down to
+// stay vertically centered on that line. This is done with a position offset
+// instead of stretching the menu element to the first line's height, as the
+// taller (invisible) element would block mouse interactions with content
+// rendered next to the block, e.g. the column resize borders. The values are
+// ported from the per-block-type CSS height rules that used to stretch the
+// menu: (first line height - 30px menu height) / 2.
+function getBlockOffset(
```
matthewlipski, line 21: Related to previous comment - we should at least make sure that the side menu offset is configurable in some way for custom blocks.

packages/core/src/editor/Block.css
```diff
@@ -723,6 +723,46 @@ NESTED BLOCKS
...
+
+/* Shows separators between a column list's columns while the mouse is over
+it. This clarifies where each column (and its clickable empty space) ends.
+The class is managed by the multi-column package's column resize plugin. The
+transition delay stops the separators from flashing in when the mouse just
+passes through the column list. */
+.bn-editor[contenteditable="true"]
+  .bn-column-list-hovered
+  > .bn-block-column:not(:last-child) {
+  box-shadow: 4px 0 0 #eee;
```
matthewlipski, line 755: Need to use different color for dark mode, also for hover

matthewlipski (commented):
Overall good changes but left a few comments, and also I don't think I'm totally happy with the UX still - having the resize handle on top of the side menu looks quite wonky to me. I think it's worth experimenting with the column padding such that we can display the resize bar and side menu side-by-side, see below.

Image

packages/core/src/extensions/TrailingNode/TrailingNode.ts
YousefED: moved down (to call site)

packages/react/src/components/SideMenu/SideMenuController.tsx
YousefED: we could allow a prop to customize this in SideMenuController. I'd also be ok with skipping this for now until we actually see someone needing this

packages/react/src/components/SideMenu/SideMenu.tsx
YousefED: answered below

packages/core/src/editor/Block.css
YousefED: Thanks, could you add this?

YousefED:
> Overall good changes but left a few comments, and also I don't think I'm totally happy with the UX still - having the resize handle on top of the side menu looks quite wonky to me. I think it's worth experimenting with the column padding such that we can display the resize bar and side menu side-by-side, see below.

Agree, but don't see a great fix for this yet. It would require quite some padding right in order for them not to overlap? This then comes at the cost of space for content

cc @robin-lecomte do you have ideas on how to resolve the overlapping?

image

nperez0111:
> image

In notion, if you hover the directly over the side menu, the column resize will hide. That fulfills my user expectation. I would hide the resize handle whether the target is not just the column container. Also hovering over one column should only show a single resize handle, not two if you are in a three column layout.

virgile-dev:
Congrats @YousefED this is much much better! It solved all the hickups I had identified on my end.

> In notion, if you hover the directly over the side menu, the column resize will hide. I would hide the resize handle whether the target is not just the column container. 

I studied this a bit closer. In Notion you actually display both if your mouse is on the resize column and get near a block side menu. 
The resize column will disappear if you actually hover on one of the 2 buttons of the side-menu. I think it's a good compromize and good behaviour to follow.
image

> Also hovering over one column should only show a single resize handle, not two if you are in a three column layout.

Agreed.

YousefED:
fixed some small ux items based on the feedback:

positioning / overlapping:

image

re. side-menu hovering:

I've opted to show the "light" version when hovering those menu items. Making them hide completely makes it easy to get into a flickering user experience. We can also test / decide on this live at some later point if needed

>> Also hovering over one column should only show a single resize handle, not two if you are in a three column layout.

> Agreed.

I'm keeping this as-is for now; the idea is the borders show the "layout" of all columns, so you get an idea of the "skeleton" of the area. Not showing all edges / borders would not reach that goal imo

virgile-dev:
LGTM this PR improves things by a lot

PR #2888, opened 2026-07-08
[Fix] Show heading options in BlockTypeSelect when allowToggleHeadings is false

### LINKED ISSUE #2845: BlockTypeSelect shows no heading options when createHeadingBlockSpec({ allowToggleHeadings: false })
### What’s broken?

When the heading block spec is configured with allowToggleHeadings: false, the default BlockTypeSelect does not show any heading options.
The schema itself supports normal headings, and headings still work elsewhere, but the formatting toolbar’s block type dropdown does not expose H1-H6.

BlockTypeSelect should still show normal heading options H1-H6. Only toggle-heading options should be removed when allowToggleHeadings: false

### Steps to reproduce

1. Create a schema with toggle headings disabled: `const schema = BlockNoteSchema.create({
  blockSpecs: {
    ...defaultBlockSpecs,
    heading: createHeadingBlockSpec({
      allowToggleHeadings: false,
    }),
  },
});`
2. Render the default block type select: `<FormattingToolbarController
  formattingToolbar={() => (
    <FormattingToolbar>
      <BlockTypeSelect />
    </FormattingToolbar>
  )}
/>`

### BlockNote version

 0.50.0

### Environment

_No response_

### Additional context

_No response_

### Contribution

### Sponsor

hiro-nikaitou:
## Summary
When `createHeadingBlockSpec({ allowToggleHeadings: false })` is used, the heading `propSchema` excludes the `isToggleable` prop. However, `blockTypeSelectItems` always included `isToggleable: false` in non-toggleable heading items. This caused `editorHasBlockWithType` to reject all heading options since `isToggleable` is not present in the schema.

## Related Issue
Closes #2845

## Root Cause
`editorHasBlockWithType` checks each prop against the block spec's `propSchema`. When `isToggleable` is excluded from the schema (`allowToggleHeadings: false`), the function returns `false` for any item that includes `isToggleable` in its props — even `isToggleable: false`.

## Fix
Removed `isToggleable: false` from the 6 non-toggleable heading items (Heading 1–6). 

- When `allowToggleHeadings: true`: the default for `isToggleable` is already `false`, so behavior is identical.
- When `allowToggleHeadings: false`: items without `isToggleable` now pass the schema check, and heading options appear correctly.
- Toggleable heading items (`isToggleable: true`) are unchanged and still respect the config.

## Type of Change
- [x] Bug fix (non-breaking change)
- [ ] New feature

## Checklist
- [ ] I have read the contributing guidelines.
- [x] My code follows the existing style.
- [ ] CI is green (passing all checks).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block type matching in the formatting toolbar by validating only the relevant properties for the selected block type.
  * Fixed cases where heading entries could be inconsistently recognized or mismatched between available heading options.

### DIFF
--- a/packages/react/src/components/FormattingToolbar/DefaultSelects/BlockTypeSelect.tsx
+++ b/packages/react/src/components/FormattingToolbar/DefaultSelects/BlockTypeSelect.tsx
@@ -47,37 +47,37 @@ export const blockTypeSelectItems = (
   {
     name: dict.slash_menu.heading.title,
     type: "heading",
-    props: { level: 1, isToggleable: false },
+    props: { level: 1 },
     icon: RiH1,
   },
   {
     name: dict.slash_menu.heading_2.title,
     type: "heading",
-    props: { level: 2, isToggleable: false },
+    props: { level: 2 },
     icon: RiH2,
   },
   {
     name: dict.slash_menu.heading_3.title,
     type: "heading",
-    props: { level: 3, isToggleable: false },
+    props: { level: 3 },
     icon: RiH3,
   },
   {
     name: dict.slash_menu.heading_4.title,
     type: "heading",
-    props: { level: 4, isToggleable: false },
+    props: { level: 4 },
     icon: RiH4,
   },
   {
     name: dict.slash_menu.heading_5.title,
     type: "heading",
-    props: { level: 5, isToggleable: false },
+    props: { level: 5 },
     icon: RiH5,
   },
   {
     name: dict.slash_menu.heading_6.title,
     type: "heading",
-    props: { level: 6, isToggleable: false },
+    props: { level: 6 },
     icon: RiH6,
   },
   {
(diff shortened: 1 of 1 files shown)


hiro-nikaitou:
Addressed CodeRabbit feedback in the updated commit (2a091b7):

The fix now takes a cleaner approach instead of removing `isToggleable: false` from items:

1. **Kept `isToggleable: false`** in all non-toggleable heading items — this preserves correct `propsMatch` selection state so toggleable and non-toggleable headings are properly distinguished.

2. **Fixed the schema filter** — `filteredItems` now only passes props that actually exist in the block's `propSchema` to `editorHasBlockWithType`. Props not in the schema (like `isToggleable` when `allowToggleHeadings: false`) are filtered out before the schema check.

This fix is semantically equivalent to the original for end users, but doesn't break selection highlighting.

Issue #2889, opened 2026-07-09
mobile: deletion of bullet points with backspace not working


judithroth:
### What’s broken?

When you work with lists on desktop, you can use the backspace key to delete the whole entry, including the bullet point. On mobile (Firefox on Android) it does not work:

https://github.com/user-attachments/assets/7bec79ee-9a8e-4bc3-8b30-b31359c1845c

### What did you expect to happen?

It should work the same as on a desktop and remove the bullet point

### Steps to reproduce

1. Create a new document or use the BlockNote demo page
2. Create a list with some bullet points
3. Try to use backspace to delete the bullet

### Environment

Firefox on Android

### Contribution

- [ ] I'd be interested in contributing a fix for this issue

### Sponsor

- [x] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

Gxrvish:
Hello @judithroth 

I tried reproducing the bug on same android and Firefox, But couldn't reproduce it.

https://photos.app.goo.gl/AnsaRQFns9CcksUC7

EDIT:
Can you tell me version of your firefox and is it a Google keyboard installed?

YousefED:
I also can't seem to reproduce this unfortunately. Could you share some more browser / keyboard details?

judithroth:
> I also can't seem to reproduce this unfortunately. Could you share some more browser / keyboard details?

Hi @YousefED 
thanks for looking into this!

Here are the device details:

Browser: Firefox
152.0.5 (Build #2016171143), d6222c84b8b1ad3cc36c5749afd40e05ebfd20d6
GV: 152.0.5-20260706120035
AS: 152.0.1
OS: Android 15

OS:/e/OS (based on Android)
Android 15
4.0-a15-20260610633983-official-FP5

Keyboard:
I see the same behavior with FUTO (version 0.1.29.1-playstore) and the pre-installed Android keyboard.

Gxrvish:
Thanks for the info @judithroth , I am now able to reproduce the bug using FUTO keyboard.
[...]

PR #2892, opened 2026-07-11
fix: patch prosemirror-view to widen Android backspace workaround to Gecko

YousefED:
Some Android keyboards (AOSP LatinIME, FUTO) don't dispatch Backspace key events — presses only surface as `beforeinput` (`deleteContentBackward`), so keymap-based backspace handling never runs. prosemirror-view has a workaround for backspace presses that have no DOM effect, but it's gated to Chrome Android, leaving Firefox broken (#2889).

This patches prosemirror-view (via pnpm `patchedDependencies`) to widen that gate to `(chrome || gecko) && android`, for testing in a preview environment. If confirmed, the fix should be upstreamed to prosemirror-view and this patch dropped.

Alternative synchronous variant (cancels the `beforeinput` event and routes it through the keymap directly) is available at 3e0df7a22.


YousefED:
being fixed upstream https://code.haverbeke.berlin/prosemirror/prosemirror/issues/1578

PR #2896, opened 2026-07-16
fix: Plain content issues

matthewlipski:
# Summary

This PR fixes some leftover issues from #2868:

1. Inline content had its type updated to support `content: "plain"`, but the type was not propagated fully and the implementation was missing.
2. Typing for `ReactBlockSpec.ts` was not updated to support plain text content.

## Rationale

The aforementioned issues are blocking for #2857.

## Changes

Fixed the 2 issues listed above.

## Impact

N/A

## Testing

TODO

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A


packages/core/src/api/nodeConversions/blockToNode.ts
```diff
@@ -64,7 +64,11 @@ function styledTextToNodes<T extends StyleSchema>(
       ? [...schema.nodes[blockType].allowedMarks(marks)]
       : marks;
 
-  const parseHardBreaks = !blockType || !schema.nodes[blockType].spec.code;
+  // Plain content nodes hold raw text — including newlines —
+  // rather than inline content, so they can't contain `hardBreak` nodes. Keep
+  // newlines as text characters for them instead of splitting into hard breaks.
+  const parseHardBreaks =
+    !blockType || !isPlainContentNodeType(schema, schema.nodes[blockType]);
```
nperez0111, line 71: should still be for code too no?

nperez0111 (approved):

packages/core/src/api/nodeConversions/blockToNode.ts
matthewlipski: I think the only case it matters is if someone creates a block with `code: true` and `content: "styled"`, which is unlikely to ever happen since code should be `content: "plain"`. But if someone does do that, we can just parse line breaks normally. We were just checking for code before since `content: "plain"` didn't exist.

PR #2911, opened 2026-07-22
fix: `getCellSelection` throwing error in positions (BLO-1193)

### LINKED ISSUE #2748: RangeError: Position -1 out of range when clicking left margin with table
### What’s broken?

When having a static formatting toolbar, and a table as the first block of the editor, clicking in the left margin of it will crash the editor (the whole client to feel more precise) - You have to click the empty space, avoid clicking the plus sign or the drag icon.

An easier way to reproduce it is having a table as the first block, then create a second block beneath it and nest it. Then you click the margin on the nested block (there's plenty of space to click now) the crash occurs again.

I tested a possible fix in `packages/core/src/extensions/TableHandles/TableHandles.ts` in `getCellSelection`:

```ts
<...>
return editor.transact((tr) => {
        <...>
        } else {
          const fromCellPos =
            selection.$from.pos - selection.$from.parentOffset - 1;
          const toCellPos =
            selection.$to.pos - selection.$to.parentOffset - 1;

          // Opt-out when the selection is not pointing into cells
          if (fromCellPos <= 1 || toCellPos <= 1) {
            return undefined;
          }

          $fromCell = tr.doc.resolve(fromCellPos);
          $toCell = tr.doc.resolve(toCellPos);
        }
<...>
```

And that worked, but I am not sure if that's the right approach, seems buggy to select a magic number like 1 to be the threshold of `fromCellPos` and `toCellPos`.

I'd expect it not to crash if I clicked anywhere, really.

### Steps to reproduce

1. Create a new document with static toolbar (there's an example in the playground with it);
2. With an empty document, create a table as the first block in the editor;
3. Create another block and then nest it;
4. When hovering the nested block you'll see the plus sign and the drag button a bit to the right, click on the empty space to its left;
5. See the whole thing crashes, with error `RangeError: Position -1 out of range`;

### BlockNote version

I could reproduce with v0.50.0, but I know it also happened with v0.48.1

### Environment

Chrome v140.0.7339.80 - Ubuntu 22.04

### Additional context

Following videos done with example static-formatting-toolbar from playground

I'll link a video of the problem:

The easiest way to reproduce:

https://github.com/user-attachments/assets/a50e1c75-cd0d-4fae-a81b-b298103e94eb

A bit harderbut still able:

https://github.com/user-attachments/assets/10b588f5-bb2b-4b9f-b2c1-e449d58a4d7e

matthewlipski:
# Summary

`TableHandlesExtension.getCellSelection` throws an error when the selection is within a table block but not actually a cell selection. For example, the selection can be a gap cursor at the start of the table.

In the static formatting toolbar example, the text alignment buttons call `getCellSelection` on selection change. In this example, if a table is at the start of the document, you can move the selection to a gap cursor above the table using the up arrow key or clicking in certain places to the side of the table. The text alignment buttons call `getCellSelection` which errors.

Closes #2748 

## Rationale

This is an uncommon error, but one that should be fixed either way.

## Changes

- Added additional checks before & after resolving cell positions to ensure they actually point to cells.

## Impact

N/A

## Testing

Added unit tests.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

It seems strange that a button in the formatting toolbar should even need the table handles extension, so we may want to fix this also.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table selection handling for cursor positions outside table cells, including gap cursors near tables and nested blocks.
  * Prevented invalid cell selections when the cursor is not inside a table cell.

* **Tests**
  * Added coverage for valid table-cell selections and non-table cursor positions.

### DIFF
--- a/packages/core/src/blocks/Table/TableExtension.test.ts
+++ b/packages/core/src/blocks/Table/TableExtension.test.ts
@@ -1,3 +1,4 @@
+import { GapCursor } from "@tiptap/pm/gapcursor";
 import { TextSelection } from "prosemirror-state";
 import { CellSelection } from "prosemirror-tables";
 import {
@@ -10,6 +11,7 @@ import {
 } from "vite-plus/test";
 
 import { BlockNoteEditor } from "../../editor/BlockNoteEditor.js";
+import { TableHandlesExtension } from "../../extensions/TableHandles/TableHandles.js";
 import type { PartialBlock } from "../defaultBlocks.js";
 
 /**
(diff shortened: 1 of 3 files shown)


nperez0111 (approved):

GuiLeme:
Hello, everyone! Thanks for the fix. Do you have an estimated release date for the version that includes this fix?

PR #2912, opened 2026-07-22
fix: ignore useless block/inline content mutations (BLO-1224)

### LINKED ISSUE #2818: Dark reader extension memory leak
### What’s broken?

Lot of users are using dark reader extensions, we got some complains about performance issues when they were using it.

After investigation, I noticed that 2 blocks give issues with a dark reader extension:
- having a code block (with languages setup) with a dark reader extension, make the tab to be totally unusable. https://www.blocknotejs.org/examples/theming/code-block
- the toggle blocks seems to not be supported neither https://www.blocknotejs.org/examples/custom-schema/toggleable-blocks

To works with a dark reader extension.

### Steps to reproduce

On Chrome, I used Dark reader, https://chromewebstore.google.com/detail/dark-reader/eimadpbcbfnmbkopoojfekhnkhdbieeh?hl=en-US&utm_source=ext_sidebar

### BlockNote version

v0.51.3

### Environment

Chrome

### Additional context

_No response_

### Contribution

### Sponsor

matthewlipski:
# Summary

This PR makes blocks & inline content ignore all mutations which don't modify the block/inline content's content. By default, ProseMirror handles all DOM mutations to a node. This is to allow changes in the view to propagate to the state. For BlockNote, this isn't really needed - you're expected to update the state (e.g. by updating a block's props) and let those changes trigger a re-render to propagate them to the view.

This has a number of benefits. For one, interactive elements that are view-only and aren't tied to e.g. a block's props won't cause re-renders. Also, browser extensions may modify inline styles on elements in the editor. Because these are DOM mutations, ProseMirror picks them up and triggers a re-render, and ends up causing an infinite re-render loop.

Closes #2818 

## Rationale

We expect most people to create custom blocks using React, so these changes don't really apply there. So this PR is mostly to prevent issues with browser extensions like DarkReader.

## Changes

- Added `ignoreNonContentMutations`.

## Impact

N/A

## Testing

Added unit tests.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary editor re-renders caused by irrelevant DOM changes in block and inline content views.
  * Improved compatibility with browser extensions and other tools that rewrite DOM attributes or structure.
  * Preserved updates to actual content and selection changes while ignoring unrelated mutations.

* **Tests**
  * Added coverage for content, selection, attribute, and non-content DOM mutations.

### DIFF
--- a/packages/core/src/schema/blocks/createSpec.ts
+++ b/packages/core/src/schema/blocks/createSpec.ts
@@ -7,6 +7,7 @@ import {
 } from "@tiptap/pm/model";
 import { NodeView } from "@tiptap/pm/view";
 import { mergeParagraphs } from "../../blocks/defaultBlockHelpers.js";
+import { ignoreNonContentMutations } from "../nodeViewMutations.js";
 import {
   Extension,
   ExtensionFactoryInstance,
@@ -270,6 +271,11 @@ export function addNodeAndExtensionsToSpec<
           if (blockImplementation.meta?.selectable === false) {
             applyNonSelectableBlockFix(typedNodeView, this.editor);
           }
+
+          // Ignores DOM mutations that don't affect the block's content, so
+          // that browser extensions which rewrite the DOM (e.g. Dark Reader)
+          // can't trigger an infinite re-render loop that freezes the tab.
+          ignoreNonContentMutations(typedNodeView);
 
           // See explanation for why `update` is not implemented for NodeViews
           // https://github.com/TypeCellOS/BlockNote/pull/1904#discussion_r2313461464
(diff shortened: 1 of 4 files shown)


nperez0111 (commented):
This is probably fine, but I'd want to see how it handles something like a data-open attribute on the block element. Like is the behavior different before & after this change? Does this change make it more reasonable?

matthewlipski:
> This is probably fine, but I'd want to see how it handles something like a data-open attribute on the block element. Like is the behavior different before & after this change? Does this change make it more reasonable?

It only affects mutations for block content elements, so elements matching `.bn-block[data-open]` will work the same as before. I.e. updating the attribute directly will cause a re-render to ensure the view is in sync with the underlying state.

For attributes on block content elements (`.bn-block-content`), updating a HTML attribute won't trigger a re-render. I guess the one potential issue is that if a `data-open` attribute is rendered from an `open` prop, updating the attribute will cause it to be out of sync with the underlying prop. Though in all of the docs and examples we're pretty explicit that you should update a block's props if you want it to render something else, so this shouldn't be something people run into.

PR #2918, opened 2026-07-25
fix(core): prefer longest suggestion menu trigger over shadowing shor…

Gxrvish:
# Summary

  Fix multi-character suggestion menu triggers (e.g. `img:`) that never opened when a shorter registered trigger (e.g. the default emoji `:`) shadowed them. Fixes https://github.com/TypeCellOS/BlockNote/issues/2905.

  ## Rationale

  `SuggestionMenuController` with `triggerCharacter="img:"` never fired. The default emoji picker registers `:`, and `handleTextInput` matched the first trigger in insertion order — so the single-char `:` always won and consumed the input before `img:` was ever checked. `img#` (no competing trigger) and `:` alone both worked, which masked the bug.

  ## Changes

  - Sort candidate triggers by length descending in `handleTextInput` so the most specific (longest) trigger matches first.
  - Fix an off-by-one in the multi-character prefix lookup: `from - triggerChar.length` grabbed one extra char, so a trigger preceded by other text on the same line (e.g. `hello img:`) never matched. Now `from - (triggerChar.length - 1)`.

  ## Impact

  - Scoped to `@blocknote/core` `SuggestionMenu.handleTextInput`.
  - Multi-char triggers that collide on their final char with a shorter trigger now resolve to the longer one.
  - Single-char triggers and programmatic menu opening are unaffected (single-char path unchanged; sort is stable for equal lengths).

  ## Testing

  - Added 2 unit tests in `SuggestionMenu.test.ts`:
    - `img:` preferred over a `:` trigger registered first (shadowing case).
    - `img:` matches mid-line (`hello img` + `:`), covering the off-by-one fix.
  - Verified both new tests fail on unfixed source and pass with the fix.
  - `pnpm lint` (full, type-aware) clean.
  - Pre-existing unrelated failures on this machine: `chatHandlers.test.ts` (needs Node 22 iterator helpers; running Node 20) and `htmlBlocks.test.ts` (needs LLM provider keys). Neither touches this change.

  ## Screenshots/Video

  N/A — non-visual logic fix.

  ## Checklist

  - [x] Code follows the project's coding standards.
  - [x] Unit tests covering the new feature have been added.
  - [x] All existing tests pass. 
  - [ ] The documentation has been updated to reflect the new feature 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved suggestion menu matching so longer, multi-character triggers take precedence over shorter triggers.
  * Multi-character triggers now work correctly when typed after existing text within a paragraph.
[...]

PR #2930, opened 2026-07-29
fix: suggestion menu behaviour (BLO-1283, BLO-955)

### LINKED ISSUE #2090: Suggestion menu not opening when text is selected
I have a suggestion menu on '@' trigger its not working if I select a text and then enter the '@' character, is there any workaround or functions to get this use case working?
Also i can see this behaviour in block note official examples 

Steps to reproduce:
Type and select some characters
Then enter '@'

Expectation:
Suggestion menu should be opened

Actual:
No suggestion menu opened

### LINKED ISSUE #2905: `:` is not supported on multi-character trigger for suggestion menus
### What’s broken?

When defining a `SuggestionMenuController` with a multi-character trigger, colons prevent the menu from ever being triggered.

The menu should open fine with a colon.

### Steps to reproduce

```tsx
<SuggestionMenuController
      triggerCharacter={"img:"}
      getItems={...}
      suggestionMenuComponent={...}
    />
```

Note that both `img#`  and `:` alone work too. 

### BlockNote version

v0.51.4

### Environment

Firefox 152.0.4, Fedora 44

### Additional context

_No response_

### Contribution

### Sponsor

matthewlipski:
# Summary

Thanks to @Gxrvish for much of the code in this PR, originally from #2918!

This PR fixes 2 issues r.e. suggestion menus:

1. Suggestion menus with multi-character triggers don't work properly. There is no prioritization of longer triggers, so given e.g. triggers `:` and `img:`, when `img:` is typed by the user, the `:` trigger suggestion menu may prioritized with no way to control this behaviour. Also, for longer triggers, checking if they are entered while the text cursor is at the start of the document will throw errors.
2. Inputting a suggestion menu trigger while a selection is active clears the selection but doesn't open the suggestion menu.

Issue 1 is fixed using basically the same code as proposed in #2918. Issue 2 is fixed by removing a check which ensures the selection is empty before opening a suggestion menu. Given that this was an intentional check we were making, this is more a change in behaviour rather than a bug fix.

Both issues are included in this PR as they touch basically the exact same snippet of code.

Closes #2905 
Closes #2090 

## Rationale

While a suggestion menu's `triggerCharacter` suggests that a single character should be used, there is not much added complexity to make longer strings work.

As for the suggestion menu opening while a selection is active, this seems to be more in-line with the behaviour of other products.

## Changes

- Removed check for empty selection when determining if a suggestion menu should be opened.
- Made suggestion menu triggers get ordered by length when determining if a suggestion menu should be opened.
- Added handling for edge case when a suggestion menu trigger is longer than the text before the cursor, so that we don't attempt to fetch text from an invalid position.
- Made a small update to docs to note that `triggerCharacter` also accepts longer strings.

## Impact

N/A

## Testing

Added unit tests.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

- **New Features**
  - Suggestion menus now support multi-character triggers such as `img:` and `ref:`.
  - Longer matching triggers take precedence over shorter overlapping triggers.
  - Shorter triggers remain available when a longer trigger does not match the surrounding text.

- **Documentation**
  - Updated guidance to explain multi-character suggestion-menu triggers and matching behavior.

### DIFF
--- a/packages/core/src/extensions/SuggestionMenu/SuggestionMenu.ts
+++ b/packages/core/src/extensions/SuggestionMenu/SuggestionMenu.ts
@@ -338,44 +338,41 @@ export const SuggestionMenu = createExtension(({ editor }) => {
         },
 
         props: {
-          handleTextInput(view, from, to, text) {
-            // only on insert
-            if (from === to) {
-              const doc = view.state.doc;
-              // Sort triggers by length (longest first) so that a more specific
-              // multi-character trigger (e.g. "img:") is preferred over a shorter
-              // one (e.g. ":") that would otherwise shadow it by matching first.
-              const triggers = [...suggestionMenus.entries()].sort(
-                ([a], [b]) => b.length - a.length,
-              );
-              for (const [trigger, menuOptions] of triggers) {
-                const input =
-                  doc.textBetween(
-                    // If the cursor is at the very start of the document, and we are checking if
-                    // an n-character trigger has been entered (e.g. "img:"), we need to get the n
-                    // characters before the text cursor to compare do so. However, the text cursor
-                    // may be near the start of the document and not have n characters before it.
-                    Math.max(0, from - trigger.length - text.length),
-                    from,
-                  ) + text;
-                if (trigger === input) {
-                  // Check the per-suggestion-menu filter before activating.
-                  if (
-                    menuOptions.shouldOpen &&
-                    !menuOptions.shouldOpen(view.state.tr)
-                  ) {
-                    continue;
-                  }
-                  view.dispatch(view.state.tr.insertText(text));
-                  view.dispatch(
-                    view.state.tr
-                      .setMeta(suggestionMenuPluginKey, {
-                        triggerCharacter: input,
-                      })
-                      .scrollIntoView(),
-                  );
-                  return true;
+          handleTextInput(view, from, _to, text) {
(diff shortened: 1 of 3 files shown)


Gxrvish:
@matthewlipski can you atleast give credits to original author?
Thanks!🙏

packages/core/src/extensions/SuggestionMenu/SuggestionMenu.ts
```diff
@@ -332,34 +338,41 @@ export const SuggestionMenu = createExtension(({ editor }) => {
         },
 
         props: {
-          handleTextInput(view, from, to, text) {
-            // only on insert
-            if (from === to) {
```
nperez0111, line 337: I think we still want this if statement no? We don't want this to trigger on a backspacing

nperez0111 (changes requested):

packages/core/src/extensions/SuggestionMenu/SuggestionMenu.ts
matthewlipski: `handleTextInput` only fires when there is actually text being inserted, so it doesn't fire on backspace/delete/enter/etc.

nperez0111 (approved):

PR #2931, opened 2026-07-29
fix: HTML export/parse round trip ignoring empty blocks (BLO-873)

### LINKED ISSUE #986: Empty paragraphs are being removed
**Describe the bug**
Empty paragraphs are being removed, Which is sometimes crucial for the formatting of the text.

**To Reproduce**

Load the following content

```
<p class="bn-inline-content">asdfasdf</p><p class="bn-inline-content">asd</p><p class="bn-inline-content">f</p><p class="bn-inline-content">asdf</p><p class="bn-inline-content"></p><p class="bn-inline-content"></p><p class="bn-inline-content"></p><p class="bn-inline-content"></p><p class="bn-inline-content"></p><p class="bn-inline-content"></p><p class="bn-inline-content"></p><p class="bn-inline-content"></p>
```

Will result in not preserving the empty lines

Is there a hidden option or something?
Thanks!

matthewlipski:
# Summary

This PR fixes empty paragraphs getting dropped when running to round trip of `blocksToHTMLLossy` -> `tryParseHTMLToBlocks`. The fix is to add a Unicode object replacement character to empty blocks' inline content elements when exporting to lossy HTML. This causes the blocks to be parsed instead of ignored without introducing any other changes in behaviour. Markdown export/parse works the same, and the character don't affect rendering the lossy HTML. For good measure though, the character is also stripped when parsing HTML to ensure full round-trip consistency.

Closes #986 

## Rationale

Dropping empty blocks is a loss of information from the original document that I believe is never desirable for consumers. For this reason, Ive also opted to not make this behaviour configurable.

## Changes

See above.

## Impact

N/A

## Testing

Added unit test case and updated existing HTML export snapshots.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved empty paragraphs and other empty inline-content blocks during HTML export and re-import.
  * Prevented blank blocks between content from being dropped in round trips.
  * Prevented internal placeholder characters from appearing in Markdown or parsed content.

* **Tests**
  * Added coverage for HTML containing an empty paragraph between two populated paragraphs.


nperez0111 (approved):

nperez0111:
We'll need to get this passing again before merging

PR #2938, opened 2026-08-03
fix: stop stale node view positions crashing the editor

### LINKED ISSUE #2937: RangeError: Position N out of range — getBlockFromPos resolves an unvalidated getPos() during nodeView mount
### What's broken?

`getBlockFromPos` calls `doc.resolve(pos)` with no bounds check against `doc.content.size`, so when a node view is constructed with a `getPos()` that points past the end of the current document, ProseMirror throws a raw `RangeError: Position N out of range` out of the render path.

We hit this in production (`RangeError: Position 2604 out of range`). It reached our route-level React error boundary, so the user lost their editing session to a full-page error screen.

The relevant code, from the 0.45.0 build we run:

```js
function getBlockFromPos(getPos, editor, view, type) {
  const pos = getPos();
  if (pos === undefined) throw new Error("Cannot find node position");
  const id = view.state.doc.resolve(pos).node().attrs.id;   // <-- unvalidated
  if (!id) throw new Error("Block doesn't have id");
  ...
```

**This is still unguarded on the current release.** `doc.resolve(pos)` in 0.52.1 has no bounds check and no try/catch:
https://cdn.jsdelivr.net/npm/@blocknote/core@0.52.1/src/schema/blocks/internal.ts

### Where it comes from

The stack shows the throw happening while ProseMirror is reconciling the doc, inside a React node view being mounted synchronously:

```
prosemirror ResolvedPos.resolve            <- RangeError: Position 2604 out of range
  <- ResolvedPos.resolveCached
  <- getBlockFromPos
  <- (React block-spec render)
  <- ReactRenderer ctor -> flushSync(render)
  <- ReactNodeViewRenderer.mount
  <- NodeViewDesc.create
  <- ViewTreeUpdater.addNode
  <- iterDeco
  <- NodeViewDesc.updateChildren   (x several, nested)
```

So a node view is created during `updateChildren`, its React render runs synchronously via `flushSync`, and it reads a `getPos()` that still reflects the pre-change document. If the document shrank in that transaction, the position is now out of range and the `resolve` throws from inside render.

A stale position during mount should degrade rather than throw from the render path — e.g. bounds-check `pos` against `doc.content.size` and return `undefined`/`null` so the node view can skip the frame and re-render on the next transaction, or throw a typed BlockNote error that consumers can catch meaningfully.

### Steps to reproduce

**I don't have a minimal reproduction** — this is a low-frequency production crash (a handful of occurrences), and I couldn't reduce it to a reliable local repro. What I can offer instead is the exact code path above and the observation that the guard is absent in current code, which is reproducible by inspection.

Filing anyway because the two closest existing reports were both closed for lack of a repro, and the underlying missing bounds check is still there:

- #2621 — "RangeError: Position X out of range when deleting after multi-column block" — closed via #2629 with the note *"I couldn't replicate this bug exactly, but did find a related bug that might be causing this."*
- #2682 — "Block doesn't have id during Editor.mount" — the very next line of the same function, same mount/`updateChildren` path; closed as not reproducible.

Both suggest the family is real and hard to pin down. A defensive bounds check would turn all of them from a crash into a recoverable no-op regardless of which trigger produced the bad position.

### BlockNote version

`@blocknote/core`, `@blocknote/react`, `@blocknote/mantine`, `@blocknote/xl-multi-column` 0.45.0 (the missing guard verified present in 0.52.1 as linked above).

### Environment

Chrome 150 / macOS, production build. React 19.2.3, @tiptap/core 3.17.0. No Yjs / collaboration. Custom blocks via `createReactBlockSpec`, with multiple editor instances on a page.

nperez0111:
# Summary

Upgrades every `@tiptap/*` dependency from `^3.13.0` (resolving to 3.22.4) to `^3.29.2`, and stops a node view's stale `getPos()` from throwing out of the render path and tearing down the consumer's React tree.

Fixes #2937

## Rationale

`getPos()` is derived from ProseMirror's view-desc tree, but `EditorView.updateStateInner` assigns the new state *before* it reconciles that tree — so anything rendering part-way through reconciliation reads a position that no longer lines up with `view.state.doc`, and resolving it threw `RangeError: Position N out of range` (or `Node should be a bnBlock, but is instead: doc` when the position was in range but pointed at the wrong node). Upstream considers this TipTap's problem ([ProseMirror/prosemirror#1532](https://github.com/ProseMirror/prosemirror/issues/1532)) and TipTap considers `flushSync` unavoidable, so we recover on our side instead; TipTap's own partial fix ([ueberdosis/tiptap#8106](https://github.com/ueberdosis/tiptap/pull/8106), in 3.28.0) only catches the `TypeError` thrown inside `posBeforeChild` and doesn't cover either shape above.

## Changes

- Bumps all `@tiptap/*` ranges to `^3.29.2`, including the `pnpm-workspace.yaml` overrides that actually pin `@tiptap/core` and `@tiptap/pm`, plus our direct `prosemirror-*` dependencies to match what `@tiptap/pm@3.29.2` declares (without which `prosemirror-model` and `prosemirror-view` each resolved to two copies and broke type-checking).
- Adds core `getBlockFromNodeView`, used by `addNodeView`, which falls back to building the block from the node alone — correct `type`/`props`/`content`, with a freshly generated `id` that belongs to no block, since this is only reached once a re-entrant dispatch has superseded the document.
- Adds react `useNodeViewBlock`, used by `createReactBlockSpec`, which falls back to the last block it resolved, seeded from the one core resolved at construction.

## Impact

No API change: `block` stays a plain `Block` for consumers of `createReactBlockSpec`, never `undefined`. Both fallbacks are silent and transient — ProseMirror rebuilds the node view against the current document immediately after, so a stale frame is invisible where a throw is not. Neither fallback is reached during normal editing (0 hits across the 828-test e2e suite), so there is no cost on the happy path.

## Testing

- `tests/src/unit/react/staleNodeViewPos.test.tsx` reproduces the crash end-to-end by dispatching re-entrantly from a node view's mount effect. Verified as a real regression test: bypassing the fallback makes both recovery cases fail with `Node should be a bnBlock, but is instead: doc`, while the normal-case test still passes.
- `internal.test.ts` and `useNodeViewBlock.test.tsx` cover each branch, including that the standalone block's synthetic id never collides with a real one.
- Full suite green on 3.29.2: lint, build, 882 unit tests, and all six E2E shards in CI. (Four E2E failures I saw locally turned out to be environment-specific and do not reproduce in CI.)

## Checklist

## Additional Notes

The reproduction only fires because it sets `isEditorContentInitialized = true` by hand. BlockNote never sets it, having replaced TipTap's `PureEditorContent` with its own mounting, so on `@tiptap/react` >= 3.22 it takes the deferred-microtask path and avoids the mount-time `flushSync` by accident rather than by design — worth revisiting alongside `EditorContent.tsx`, which still notifies portal subscribers synchronously rather than adopting upstream's 3.28.0 batching.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved React node view stability when document changes cause temporarily stale or invalid positions.
  * Preserved the last valid block during position resolution failures, helping prevent rendering errors and keeping the editor usable.

* **New Features**
  * Added a public hook for resolving the current block associated with a React node view.

* **Chores**
  * Updated Tiptap and ProseMirror packages to newer versions.

* **Tests**
  * Added coverage for stale positions, fallback recovery, and block resolution behavior.


YousefED:
@nperez0111 nice find. Is this workaround (costing quite a bit of code and extra edge cases to consider), better than returning nothing? Wouldn't it automatically recover on a next cycle / is the result of the "fault calls" even used?

Didn't dive deep into this, just checking in from a maintenance perspective

PR #2939, opened 2026-08-03
feat: Mobile formatting toolbar (BLO-1292)

### LINKED ISSUE #280: Undesired behavior when marking text on mobile devices
**Describe the bug**
Usually when marking text on a Mac, the "header" appears, which also happens on Chrome on Android. However, then also the usual Chrome/Android line comes (for copying etc.), which collides with it

**To Reproduce**
Use a mobile device, preferably Android, but perhaps also iOS, and mark some text.

**Misc**
- Browser: Chrome

Let me know if you need some further information. 🙂

### LINKED ISSUE #938: Format toolbar goes offscreen on mobile

### LINKED ISSUE #2122: iOS Safari: On text selection two overlapping menus

matthewlipski:
# Summary

This PR completes the experimental mobile formatting toolbar and adds it to the default UI. The toolbar opens when the virtual keyboard opens and sits above it.

Closes #280 
Closes TypeCellOS/BlockNote#938 
Closes TypeCellOS/BlockNote#2122

The browser APIs for dealing with the virtual keyboard and positioning elements around it are super half baked, so there's quite a lot of complexity in how this works. This complexity is almost entirely related to the [visual viewport](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport).

### Viewport Background Context

The browser has two main viewports - the [layout and visual viewport](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/CSSOM_view/Viewport_concepts#layout_and_visual_viewports).

The layout viewport is what we generally refer to when talking about the browser viewport. Put simply, it's the box in which the browser renders a web page and determines the layout of that page.

The visual viewport is a box that lies within the layout viewport. In most cases (especially on desktop), it has the exact same dimensions as the layout viewport and is not really something you need to worry about. In this state, scrolling will move both the layout and visual viewports simultaneously around the page. However, using the pinch-to-zoom gesture on a trackpad or touchscreen shrinks the visual viewport. Now, scrolling will instead move the visual viewport within the layout viewport, and only once it reaches an edge will it move the layout viewport too.

The OS-level virtual keyboard shrinks the visual viewport the same way as pinch-to-zoom does. This means you can't just set `position: fixed` on the toolbar. Doing so will anchor it to the layout viewport, which the visual viewport can still move around in while the keyboard remains open. Therefore, the toolbar will not stay anchored to the virtual keyboard.

### Default "Scrolling Document" Mode

The default mobile formatting toolbar implementation works in any app, but works quite differently between iOS and Android.

**iOS**

Unfortunately, while `position: fixed` allows us to easily anchor an element to the layout viewport, no there is equivalent to doing the same for the visual viewport. Therefore, we are forced to update the position dynamically as the visual viewport scrolls and resizes. While this does work, it's quite laggy and jittery as each event has to issue a repaint on the next tick. Using a CSS transition makes it slightly less choippy, and this is the best we can do as a drop-in solution.

**Android**

Under `<head>` is a `<meta name="viewport">` element. It contains attributes which specify some viewport behaviour. One of those attributes is [`interactive-widget`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/viewport#interactive-widget), which is specifically for determining how the virtual keyboard should interact with the browser viewports. As mentioned above, by default it resizes the visual viewport, but not the layout viewport, which is equivalent to `interactive-widget: resizes-visual`. Changing it to `interactive-widget: resizes-content` makes it resize both viewports, allowing us to set `position: fixed` on the toolbar to anchor it to the virtual keyboard. However, this changes nothing for pinch-to-zoom, in which case we fallback to the same implementation as iOS.

Overall though, this is still a better experience than on iOS as it removes the lag and jitter on scroll, at least while fully zoomed out. The only reason we don't use it on iOS is because it doesn't support the `interactive-widget` attribute.

### Opt-in "Scroll Container" Mode

In order to remove the scrolling lag and jitter, the formatting toolbar must have the same position at all times. The only way to do this is by having the `<html>` and `<body>` elements be non-scrollable, then use a descendant element which contains all scrollable content on the page, and fills the full width/height of the page. Then, by listening to scroll and resize events on the visual viewport, we can dynamically resize this element to match its size, basically causing the layout viewport to resize in sync with the visual viewport. Therefore, as long as we render the formatting toolbar outside this scroll container, we can use `position: fixed` as both viewports are synced.

The net result is that scrolling is always smooth, though resizing the visual viewport is still choppy. The bigger caveat is that `<html>` and `<body>` elements must be non-scrollable, which is not a common pattern and may require fairly significant changes to an app's DOM structure. Hence why you have to opt-in, which is done by simply adding the `bn-scroll-container` class name to the element that should hold the page's scrollable content.

## Rationale

The desktop formatting toolbar has a pretty annoying issue on mobile where the OS will display its own floating menu on top of it for things like cut, copy, and paste. A formatting toolbar that sits on top of the virtual keyboard is also a more common mobile UX pattern that's better for one-handed reachability.

## Changes

* Added `interactive-widget: resizes-content` to playground & each example.
* Added `portalRoot` prop to `Menu` and `ToolbarSelect` components.
* Fixed `getActiveStyles` not returning marks which are active but not in the document.
* Added `isTouchDevice` helper.
* Made `AddCommentButton`/`AddTiptapCommentButton`/`CreateLinkButton` only show when the selection isn't collapsed.
* Made formatting toolbar dropdowns render in a portal on mobile to avoid getting clipped by scrollable container.
* Added `MobileFormattingToolbarController`.
* Renamed existing `FormattingToolbarController` to `DesktopFormattingToolbarController`.
* Added new `FormattingToolbarController` which automatically switches between desktop and mobile variants.
* Added `useVirtualKeyboard` helper hook.
* Added `hide` middleware by default to `GenericPopover`.
* Added `UIModeContext` for components to be able to render different things based on device type (mobile/desktop).

## Impact

N/A

## Testing

We could add e2e tests for this, but getting them to work would require stubbing most of the things we'd actually want to test since there is no virtual keyboard in the test environment.

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **New Features**
  * Added a mobile formatting toolbar that appears when the on-screen keyboard is open.
  * Improved toolbar positioning during scrolling, keyboard use, and pinch-zooming.
  * Toolbar menus and popovers now remain visible instead of being clipped by horizontal scrolling.
  * Added a complete mobile toolbar example and usage documentation.
* **Bug Fixes**
  * Improved touch-device focus behavior and menu placement across supported UI themes.
  * Preserved editor focus when closing formatting popovers.
* **Refactor**
  * Replaced the experimental mobile toolbar implementation with the standard mobile experience.

## Summary by CodeRabbit

* **New Features**
  * Added a mobile formatting toolbar that appears when the on-screen keyboard is open.
  * Improved toolbar positioning, scrolling, and dropdown visibility on mobile devices.
  * Added support for customizing mobile toolbar rendering and portal placement.
  * Added a complete mobile formatting toolbar example.
* **Bug Fixes**
  * Improved touch interactions, focus handling, and popover placement across supported UI themes.
  * Prevented toolbar buttons and menus from being clipped or incorrectly focused on mobile browsers.
* **Documentation**
  * Documented mobile toolbar behavior, scrolling limitations, and configuration options.

## Changes since the first review round

- **Viewport-meta check moved and made development-only.** The warning about a missing `interactive-widget=resizes-content` moved from `useVirtualKeyboard` into `BlockNoteViewEditor` (`viewportMeta.ts`), so it reaches developers on desktop and with a custom formatting toolbar alike, and it returns early under `process.env.NODE_ENV === "production"`; the library build keeps the expression for the consumer's bundler. The react package declares the `process` global in `vite-env.d.ts` the same way core does.
- **Example in the playground, via CSS.** The mobile toolbar example gets `playground: true`. "Scrolling document" mode sets one class on `body`, and the example's own `style.css` neutralises any page-level scroll container in that mode and any nested one in the other, both `position: static`; no React reaches into the playground's root. Library CSS untouched.
- `isTouchDevice()` is `navigator.maxTouchPoints > 0 && matchMedia("(pointer: coarse)")`, i.e. "primary input is touch", cached per session; it also selects the mobile toolbar in `FormattingToolbarController`.
- Dev origins in `docs/next.config.ts` restricted to the private `172.16/12` range (review finding).

## Testing

The android e2e instance (added in #3029) runs `tests/src/end-to-end/mobile/` stub-free with an Android UA and touch emulation, the keyboard emulated as a visual-viewport resize; `mobileToolbar.test.tsx` and `skinFocus.test.tsx` in the layers above pin the toolbar's behaviour per skin. Verified on a Fairphone 5 (Chrome) and the iOS simulator during the release review.

## Summary by CodeRabbit

* **New Features**
  * Added a mobile formatting toolbar for editing with the on-screen keyboard.
  * Added stable toolbar positioning during scrolling, zooming, and keyboard interactions.
  * Added mobile portal support for menus and popovers to prevent clipping and preserve editor focus.
  * Added a mobile formatting toolbar example with scroll-mode switching.

* **Bug Fixes**
  * Improved touch-device controls and editor focus after popovers close.
  * Formatting and comment controls now hide when no text is selected.

* **Documentation**
  * Added mobile compatibility and viewport configuration guidance.

* **Chores**
  * Replaced the former experimental mobile toolbar example.


nperez0111:
Is there a way to implement this such that it requires no changes to an existing application (i.e. not require a specific parent div)?

 - If there is a way to do it, what are the pros/cons of it?
 - If there is not a way to do it, can we somehow constrain this to only happen when the toolbar needs to be shown (so it doesn't impact the whole app the entire time, just during the formatting toolbar opening)? Like could we only lock the body when we need to?

nperez0111:
Just for UX reasons, I think we should add padding on the top & bottom to these buttons, since I don't feel confident hitting them properly on mobile. Maybe a bit more on the X & Y too, but the top & bottom will already make a large difference:

image

packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  if (!isVirtualKeyboardOpen(viewport)) {
+    return null;
+  }
+
+  const Component = props.formattingToolbar || FormattingToolbar;
+
+  return (
+    <MobileFormattingToolbarPortalContext.Provider value={toolbarElement}>
+      <div className="bn-mobile-formatting-toolbar" ref={setToolbarElement}>
```
nperez0111, line 50: I wonder if we could just pass a ref around rather than the element itself. It feels like this is just asking for extra re-renders

nperez0111: Or skip this and just use a unique id and querySelector for it when you need it

packages/react/src/components/FormattingToolbar/useVisualViewportRect.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+
+function readVisualViewport(): VisualViewportRect {
+  const vp = window.visualViewport;
+  return {
+    top: vp?.offsetTop ?? 0,
+    left: vp?.offsetLeft ?? 0,
+    width: vp?.width ?? window.innerWidth,
+    height: vp?.height ?? window.innerHeight,
+    scale: vp?.scale ?? 1,
```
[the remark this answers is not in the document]
nperez0111: ^ Yea this is relevant, will need some guards for these when SSR'ing

packages/react/src/components/FormattingToolbar/useVisualViewportRect.ts
```diff
@@ -0,0 +1,88 @@
...
+ * relies on for positioning and keyboard detection.
+ */
+export function useVisualViewportRect(): VisualViewportRect {
+  const [rect, setRect] = useState(readVisualViewport);
+
+  useEffect(() => {
+    const html = document.documentElement;
+
+    const vp = window.visualViewport;
+    const update = () => {
```
nperez0111: Don't these fire a lot and cause a bunch of re-renders? Do we need the rect in React state? If we use the CSS version, this is just overhead right? Like it should be not listened to in that case?

packages/react/src/components/FormattingToolbar/useVisualViewportRect.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Does not lock document scroll. For the smoother "non-scrolling document"
+ * behavior, the host app opts in with CSS (see
+ * {@link MobileFormattingToolbarController}). This is what that controller
+ * relies on for positioning and keyboard detection.
+ */
+export function useVisualViewportRect(): VisualViewportRect {
+  const [rect, setRect] = useState(readVisualViewport);
+
+  useEffect(() => {
```
nperez0111: Should be a layout effect, is to do with reading elements

packages/react/src/components/FormattingToolbar/useVisualViewportRect.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Whether the on-screen keyboard is open, from a visual-viewport snapshot. We
+ * compare `height * scale` — the zoom-invariant layout-equivalent height, so
+ * pinch-zoom (which also shrinks `height`) doesn't count — against the tallest
+ * value seen, treating a drop of more than 150px as open: comfortably above
+ * URL-bar show/hide (~60-100px) and below any real keyboard (~250px+).
+ */
+export function isVirtualKeyboardOpen(viewport: VisualViewportRect): boolean {
+  const layoutHeight = viewport.height * viewport.scale;
+  maxLayoutViewportHeight = Math.max(maxLayoutViewportHeight, layoutHeight);
+  return maxLayoutViewportHeight - layoutHeight > 150;
```
nperez0111: This works on ios, does it work on android?

packages/react/src/hooks/useIsMobile.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const useIsMobile = () => {
+  const [isMobile, setIsMobile] = useState(false);
+
+  useEffect(() => {
+    setIsMobile(isTouchDevice());
+  }, []);
+
+  return isMobile;
+};
```
nperez0111: This should not be a hook, it is just reading `isTouchDevice` it should not have to trigger re-renders or anything like that

packages/react/src/components/FormattingToolbar/DefaultButtons/ColorStyleButton.tsx
```diff
@@ -136,7 +138,7 @@ export const ColorStyleButton = () => {
   }
 
   return (
-    <Components.Generic.Menu.Root>
+    <Components.Generic.Menu.Root portalRoot={portalRoot}>
```
nperez0111: I see a lot of passing around of this portalRoot, but I don't understand why it is being passed around?

Shouldn't it be based on editor.portalElement? Why does it need to be prop-drilled through everywhere instead of just using the context?

nperez0111 (changes requested):

matthewlipski:
R.e. adding padding to buttons to make them easier to tap, imo we should just scale up the whole toolbar with a transform as the text and icons are kinda small too on mobile.

packages/react/src/components/FormattingToolbar/DefaultButtons/ColorStyleButton.tsx
matthewlipski: By default, the Mantine UI creates popovers/dropdowns as siblings of the target element. It's a problem on mobile as the formatting toolbar is scrollable, so the popovers get clipped (desktop doesn't show this behaviour). That's why we need a portal here.

The reason that it portals to the formatting toolbar container element and not `editor.portalElement`, is because  the `Components.Generic.~` components are used across the editor, not just the formatting toolbar. With the `portalRoot` prop, we're only affecting the formatting toolbar elements and also only portalling them up 1 level in the DOM tree, so it's not a breaking change. Putting all popovers/dropdowns in an `editor.portalElement` portal would affect them across the editor, breaking a bunch of CSS and e2e test selectors. This is not an unsolvable problem by any means, but does require changing quite a lot of loc, which I didn't want to commit to doing at the time.

With that context, wdyt?

packages/react/src/hooks/useIsMobile.ts
matthewlipski: Wouldn't it be an issue for SSR? Then it would always be false on the server render and would not update on the client render. I could be wrong though

nperez0111: This is for the formatting toolbar right? So nothing would actually render in SSR right?

packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx
matthewlipski: I think this is ok tbh, the `toolbarElement` state only updates when the toolbar is opened or closed, so when opening/closing the virtual keyboard. Which is necessary as any popovers/menus have to re-render so they can be portalled to the `toolbarElement`.

We could make `toolbarElement` a ref, but then we'd still need a state for each dropdown to ensure a re-render to apply the portal.

examples/03-ui-components/14-mobile-formatting-toolbar/src/DummyUI.tsx
```diff
@@ -0,0 +1,78 @@
...
+        aria-expanded={open}
+        onClick={() => setOpen((v) => !v)}
+      >
+        <span />
+        <span />
+        <span />
+      </button>
+      {open && (
+        <nav className="hamburger-menu">
+          <a href="#">Home</a>
```
YousefED: Is the hamburger menu required for testing / demonstrational purposes?

if so, let's change to "some", "dummy", "buttons" or similar to clarify these menus are fake and people don't think they're broken

examples/03-ui-components/14-mobile-formatting-toolbar/src/style.css
```diff
@@ -0,0 +1,133 @@
...
+  font: 600 18px/1.2 sans-serif;
+}
+
+.prose p {
+  margin: 0 0 8px;
+  font: 14px/1.6 sans-serif;
+  color: #333;
+}
+
+/* A top-level wrapper div is the scroll container (the document itself doesn't
```
YousefED: move up, because this is the actual fix

examples/03-ui-components/14-mobile-formatting-toolbar/src/style.css
```diff
@@ -0,0 +1,133 @@
...
+   scroll — `html`/`body` are locked with `overflow: hidden` above), pinned to
+   the visual viewport rectangle via the `--bn-vv-*` variables the mobile
+   toolbar controller publishes, so it sits directly above the keyboard on iOS —
+   where the layout viewport doesn't resize and can be left with a nonzero
+   `offsetTop`. */
+.scroll-host {
```
YousefED: We could consider shipping this class with BlockNote, so implementors only need to add a single class to their existing DOM? wdyt?

packages/ariakit/src/toolbar/ToolbarSelect.tsx
```diff
@@ -20,14 +20,22 @@ export const ToolbarSelect = forwardRef<
...
   const setValue = (value: string) => {
     items.find((item) => item.text === value)!.onClick?.();
   };
 
   return (
-    <AriakitSelectProvider value={selectedItem.text} setValue={setValue}>
+    <AriakitSelectProvider
+      value={selectedItem.text}
+      setValue={setValue}
+      placement={"bottom"}
```
YousefED: does this placement negatively effect non-mobile setups?

packages/mantine/src/menu/Menu.tsx
```diff
@@ -36,11 +44,20 @@ export const Menu = (props: ComponentProps["Generic"]["Menu"]["Root"]) => {
...
       position={position}
-      returnFocus={false}
+      // Only when portalling into the mobile toolbar: don't move focus into the
+      // dropdown on open, since on mobile that blurs the editor's
+      // contentEditable and dismisses the on-screen keyboard.
+      // `withInitialFocusPlaceholder={false}` drops the focusable placeholder
+      // Mantine otherwise autofocuses. On desktop we leave Mantine's defaults
+      // (focus trapping/return) intact for accessibility.
+      trapFocus={portalRoot ? false : undefined}
+      returnFocus={portalRoot ? false : undefined}
```
YousefED: `returnFocus` used to be `false` when not portalling, now it will be `undefined`. intended?

packages/mantine/src/toolbar/ToolbarButton.tsx
```diff
@@ -57,11 +57,17 @@ export const ToolbarButton = forwardRef<HTMLButtonElement, ToolbarButtonProps>(
...
+          if (isTouchDevice()) {
+            event.preventDefault();
+            return;
+          }
+
+          // Needed as Safari doesn't focus button elements on mouse down
+          // unlike other browsers.
           if (isSafari()) {
-            (e.currentTarget as HTMLButtonElement).focus();
+            (event.currentTarget as HTMLButtonElement).focus();
```
YousefED, line 75: tested whether this path still works well on safari on desktop?

packages/mantine/src/toolbar/ToolbarSelect.tsx
```diff
@@ -26,18 +30,38 @@ export const ToolbarSelect = forwardRef<
 
   return (
     <MantineMenu
-      withinPortal={false}
+      withinPortal={!!portalRoot}
+      portalProps={portalRoot ? { target: portalRoot } : undefined}
+      position={"bottom-start"}
```
YousefED: this changes desktop behavior. desired?

packages/mantine/src/toolbar/ToolbarSelect.tsx
```diff
@@ -26,18 +30,38 @@ export const ToolbarSelect = forwardRef<
...
       transitionProps={{
         exitDuration: 0,
       }}
       disabled={isDisabled}
-      middlewares={{ flip: true, shift: true, inline: false, size: true }}
+      // Don't move focus into the dropdown on open: on mobile that blurs the
+      // editor's contentEditable and dismisses the on-screen keyboard.
+      // `withInitialFocusPlaceholder={false}` drops the focusable placeholder
+      // Mantine otherwise autofocuses.
+      trapFocus={false}
```
YousefED: tested accessibility on desktop?

packages/react/src/editor/PortalContext.ts
```diff
@@ -0,0 +1,8 @@
+import { createContext } from "react";
+
+/**
+ * Context to override the portal root of dropdowns/popovers in `ComponentsContext`. If not present
+ * or left undefined, the default render location of the dropdown/popover will be used, which is
+ * library-specific.
+ */
+export const PortalContext = createContext<HTMLElement | null>(null);
```
YousefED: Is this separate PortalContext needed? Maybe it's clearer to revert this to a Prop, if possible at all?

I think it now causes confusion as there's both a `PortalContext` and a `blockNoteContext.editor?.portalElement`. Naming doesn't make clear what the difference is and when which one should be used. So if we want to stick to context, we should at least improve naming

packages/core/src/util/browser.ts
```diff
@@ -28,3 +28,10 @@ export function mergeCSSClasses(...classes: (string | false | undefined)[]) {
 
 export const isSafari = () =>
   /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+
+export const isTouchDevice = () =>
```
YousefED: very micro, but maybe we should globally cache the result of this (it's unlikely to change). I'm not sure what the cost of `matchMedia` is

packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!keyboardOpen) {
+    return null;
+  }
+
+  const Component = props.formattingToolbar || FormattingToolbar;
+
+  return (
+    <PortalContext.Provider value={toolbarElement}>
+      <div className="bn-mobile-formatting-toolbar" ref={setToolbarElement}>
```
YousefED: Do we need this element, or can we just have these sub components also Portal to the existing editor `portalElement` we already use?

packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx
```diff
@@ -0,0 +1,55 @@
+import { FC, useState } from "react";
+
+import { PortalContext } from "../../editor/PortalContext.js";
+import { FormattingToolbarProps } from "./FormattingToolbarProps.js";
+import { FormattingToolbar } from "./FormattingToolbar.js";
+import { useVirtualKeyboard } from "./useVirtualKeyboard.js";
+
+/**
```
YousefED, line 22: I might have missed this, but;

- what if the device keyboard is showing because a different BlockNote editor is selected?
- what if the device keyboard is showing because an <input> is opened.

Can you add these scenarios to the example?

YousefED (commented):
- On my Android, there's a gap between the toolbar and the device keyboard
- ~What should happen when no text is selected and the Link or Comment button is clicked? Or should they be disabled?~
- what can we add in testing (e2e or unit tests)
- ~The way it's currently set up, the examples that customize the FormattingToolbar will opt-out of the mobile formatting toolbar. So I suppose either the default FormattingToolbarController need to be a proxy to a mobile and desktop variant, or all examples need to be updated~
- ~did you also test on other touch devices like a tablet? Did you test both landscape and portrait mode of mobile?~


fyi, Claude also had this feedback (didn't fully review this):

Findings
1. ~Toolbar shows whenever the viewport shrinks, regardless of editor focus/selection (High) — [MobileFormattingToolbarController.tsx:72-82](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx#L72-L82) gates only on keyboardOpen. The experimental version also required useExtensionState(FormattingToolbarExtension). Reproduced: with activeElement === BODY and an empty editor, shrinking the viewport 300px shows a full 12-button toolbar (screenshot [mobile-kb-unfocused.png](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/.claude/skills/playwright-cli/screenshots/mobile-kb-unfocused.png)). Concretely: typing in any other input on the page (search box, the comments composer — which disables its own formatting toolbar, so the main editor's appears) brings up the main editor's toolbar. Corollary: N editors on a page → N stacked toolbars; reproduced on collaboration/suggestion-gallery (3 editors → 3 .bn-mobile-formatting-toolbar nodes at the same fixed position). Suggest gating on keyboardOpen && editor focused (or the FormattingToolbar extension state / editor.isFocused()).~

2. ~Desktop regression in Mantine ToolbarSelect (Medium) — [ToolbarSelect.tsx:30-45](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/packages/mantine/src/toolbar/ToolbarSelect.tsx#L30-L45) sets trapFocus/returnFocus/withInitialFocusPlaceholder=false and position="bottom-start" unconditionally, unlike Menu.tsx which scopes them to portalRoot. Mantine 9.1.1 Menu defaults are trapFocus: true, withInitialFocusPlaceholder: true, and MenuTarget has no arrow-key handling — keyboard nav only works once focus is inside the dropdown. So on desktop, click-to-open the block type select no longer allows Arrow-key navigation. Also the dropdown alignment changes from centered to start-aligned on desktop. Both should be portalRoot ? false : undefined like Menu.tsx.~

3. ~Float-equality baseline reset breaks on pinch-zoom (Medium-low) — [useVirtualKeyboard.ts:30](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/packages/react/src/components/FormattingToolbar/useVirtualKeyboard.ts#L30) layoutWidth !== baselineLayoutWidth compares vv.width * vv.scale; I checked 7% of arbitrary scales give (390/s)*s !== 390. A pinch gesture with the keyboard open will almost surely hit one → max resets to the keyboard-open height → toolbar disappears until the keyboard is closed and reopened. Use a tolerance (e.g. Math.abs(...) > 1).~

4. ~Touch devices with no virtual keyboard lose the formatting toolbar entirely (Medium, design question) — [BlockNoteDefaultUI.tsx:123-130](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/packages/react/src/editor/BlockNoteDefaultUI.tsx#L123-L130) replaces the desktop controller on pointer: coarse. iPad/Android tablets with a Bluetooth keyboard (no trackpad → still coarse), iOS floating/undocked keyboard, or interactive-widget=overlays-content never shrink the visual viewport → no toolbar ever. Worth deciding if that's acceptable, or falling back to the floating toolbar when keyboard is closed.~

5. Leftovers / small things

[Menu.tsx:207](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/packages/mantine/src/menu/Menu.tsx#L207) flip = !position?.startsWith("top") + comment about "the mobile toolbar's color menu" — no call site passes a top* position anymore (ColorStyleButton passes none); stale, and it silently changes behavior for custom UIs that pass position="top".
[ToolbarSelect.tsx:389](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/packages/mantine/src/toolbar/ToolbarSelect.tsx#L389) truncated comment "Prevents focus shift on mo".
Mantine select now bottom-start, Ariakit select now bottom — inconsistent; and FileReplaceButton losing position="bottom" is a no-op (Mantine/Ariakit default to bottom) — fine, just noting.
[examples.gen.tsx](vscode-webview://13ebsmp5jilfnhd6v7jgv0rauv9pad5s7krn75bkvdgul96btso4/playground/src/examples.gen.tsx) readme string is stale vs. the new README (needs vp run gen); example style.css comment references DemoChrome but the file is DummyUI.tsx.
ExperimentalMobileFormattingToolbarController removed from the public export — intended, but a release note is warranted.
onPointerDown uses isTouchDevice() rather than event.pointerType === "touch"; the latter is more precise and avoids the global.

YousefED:
On desktop, with the example I can create this bug (select text, then scroll down):

image

Is this a regression, or pre-existing?

packages/react/src/editor/PortalContext.ts
matthewlipski: Yea I talked with @nperez0111 about it and he preferred to omit the prop. Will fix the naming

packages/react/src/components/FormattingToolbar/useVirtualKeyboard.ts
```diff
@@ -1,5 +1,51 @@
...
 
+// Styles for the opt-in "non-scrolling document" setup. A host app that adds the
+// `bn-scroll-host` class to its scroll container gets a fixed-position element
+// pinned to the visual viewport (via the `--bn-vv-*` variables published below),
+// with the document itself locked so the browser can't scroll `<html>`/`<body>`
+// — which is what makes the mobile toolbar jitter. Scoping the lock with `:has`
+// means it only kicks in when a scroll host is actually present, so importing
+// this hook never changes scrolling for apps that don't opt in.
+const SCROLL_HOST_STYLE_ID = "bn-scroll-host-styles";
+const SCROLL_HOST_STYLES = `
```
YousefED: @matthewlipski why not just add this to the stylesheet?

nperez0111:
Using a fixed positioning strategy in floating ui likely will resolve this like it did here: https://github.com/TypeCellOS/BlockNote/pull/3002

docs/content/docs/react/components/formatting-toolbar.mdx
```diff
@@ -38,3 +38,60 @@ The first element in the default Formatting Toolbar is the Block Type Select, an
...
+
+### Scrolling document
+
+This is what you get without any changes to your app. The toolbar follows the visible area above the keyboard as the page scrolls. Mobile browsers only report visual viewport changes after the fact, so the toolbar can lag or jitter slightly while the page is scrolling. If that matters for your app, switch to a scroll container.
+
+### Scroll container
+
+In this layout, `<html>` and `<body>` are locked and all page content lives inside a single scroll container that BlockNote keeps aligned with the visual viewport. Since the document never scrolls, the toolbar can stay at a truly fixed position and the lag/jitter disappears. Since the document no longer scrolls, this comes with some potential trade-offs. Browser gestures that rely on document scrolling, like pull-to-refresh, may stop working and browser UI elements like the address bar, which normally hides and reappears as you scroll, may stay fixed. Note that these trade-offs are browser-dependent - some will have neither, while others will have both.
```
YousefED: duplicate `Since the document no longer scrolls,`

docs/content/docs/react/components/formatting-toolbar.mdx
```diff
@@ -38,3 +38,60 @@ The first element in the default Formatting Toolbar is the Block Type Select, an
...
+<div className="bn-scroll-host">{/* nav, editor, page content... */}</div>
+```
+
+<Callout type="warning">
+  Your app should only ever have a single `bn-scroll-host` element. It's pinned
+  to the visual viewport with `position: fixed`, so multiple hosts would overlap
+  each other. Wrap all your scrollable page content in one host.
+</Callout>
+
+That's all the setup needed. BlockNote injects the following styles for you:
```
YousefED: `injects` and `publishes on `<html>`:` is out-of-date, as it should be in the stylesheet and is not runtime injected anymore

docs/content/docs/react/components/formatting-toolbar.mdx
```diff
@@ -38,3 +38,60 @@ The first element in the default Formatting Toolbar is the Block Type Select, an
 <Example name="ui-components/formatting-toolbar-block-type-items" />
 
 Here, we use the `FormattingToolbar` component but keep the default buttons (we don't pass any children). Instead, we pass our customized Block Type Select items using the `blockTypeSelectItems` prop.
+
+## Mobile Formatting Toolbar
+
+On touch devices, BlockNote's default UI replaces the floating Formatting Toolbar with a mobile Formatting Toolbar that sits just above the on-screen keyboard. It shows the same items as the regular Formatting Toolbar and is enabled by default - there's nothing to set up. Open any of the examples above on a phone to see it.
+
+The mobile Formatting Toolbar works with two page layouts. Which one you get is decided purely by your app's CSS:
```
YousefED: `Which one you get is decided purely by your app's CSS:` obsolete and incorrect

examples/03-ui-components/14-mobile-formatting-toolbar/src/DummyUI.tsx
YousefED: the text of the buttons hasn't been updated. It's nice that the code is now clearer, but a user just playing with the example could still think "why doesn't this button do anything?

examples/03-ui-components/14-mobile-formatting-toolbar/src/App.tsx
```diff
@@ -0,0 +1,63 @@
...
+  // the class - a real app would apply it unconditionally, the switch is only
+  // here so you can compare both.
+  const [scrollMode, setScrollMode] = useState<
+    "scrolling-document" | "scroll-container"
+  >("scroll-container");
+
+  return (
+    <div
+      className={
+        scrollMode === "scroll-container" ? "bn-scroll-host" : undefined
```
YousefED: let's unify scroll-host and scroll-container between code and documentation? Can we think of a single term?

packages/mantine/src/menu/Menu.tsx
```diff
@@ -36,11 +40,14 @@ export const Menu = (props: ComponentProps["Generic"]["Menu"]["Root"]) => {
...
-      middlewares={{ flip: true, shift: true, inline: false, size: true }}
+      withinPortal={!!portalRoot}
+      portalProps={portalRoot ? { target: portalRoot } : undefined}
+      // Do not move focus to dropdown when portaled (mobile), as it blurs the
+      // editor's contentEditable and dismisses the on-screen keyboard.
+      trapFocus={portalRoot ? false : undefined}
+      middlewares={{ flip, shift: true, inline: false, size: true }}
       onChange={onOpenChange}
       position={position}
-      returnFocus={false}
```
YousefED, line 43: this was deleted, intentional?

packages/react/src/components/FormattingToolbar/DefaultButtons/ColorStyleButton.tsx
```diff
@@ -136,7 +138,9 @@ export const ColorStyleButton = () => {
   }
 
   return (
-    <Components.Generic.Menu.Root>
+    <Components.Generic.Menu.Root
+      portalRoot={uiMode === "mobile" ? editor.portalElement : undefined}
```
YousefED: let's document why we only do this on mobile

packages/react/src/components/FormattingToolbar/DefaultButtons/CreateLinkButton.tsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useState } from "react";
```
YousefED, line 1: I'm unable to create a link on mobile. is this the known issue?

packages/react/src/components/FormattingToolbar/DefaultButtons/FileCaptionButton.tsx
```diff
@@ -88,6 +90,7 @@ export const FileCaptionButton = () => {
     <Components.Generic.Popover.Root
       open={popoverOpen}
       onOpenChange={setPopoverOpen}
+      portalRoot={uiMode === "mobile" ? editor.portalElement : undefined}
```
YousefED: document why we only do this on mobile. Also, have you tested this button / dialog?

packages/react/src/components/FormattingToolbar/DefaultButtons/FileRenameButton.tsx
```diff
@@ -88,6 +90,7 @@ export const FileRenameButton = () => {
     <Components.Generic.Popover.Root
       open={popoverOpen}
       onOpenChange={setPopoverOpen}
+      portalRoot={uiMode === "mobile" ? editor.portalElement : undefined}
```
YousefED: (same, let's document this.. skipping that comment from now on)

packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx
```diff
@@ -0,0 +1,79 @@
...
+ */
+export const MobileFormattingToolbarController = (props: {
+  formattingToolbar?: FC<FormattingToolbarProps>;
+}) => {
+  const editor = useBlockNoteEditor();
+  const keyboardOpen = useVirtualKeyboard();
+
+  // Whether this editor holds focus, kept in sync via its `focus`/`blur`
+  // events so the toolbar shows/hides as focus enters or leaves the editor.
+  const [focused, setFocused] = useState(() => editor.isFocused());
```
YousefED, line 64: don't we have sth like `useEditorState` for this?

packages/react/src/components/Popovers/GenericPopover.tsx
```diff
@@ -134,16 +135,19 @@ export const GenericPopover = (
...
-    ...restFloatingOptions,
-  });
+  const { refs, floatingStyles, context, middlewareData } =
+    useFloating<HTMLDivElement>({
+      whileElementsMounted: mergeWhileElementsMounted(
+        autoUpdate,
+        props.useFloatingOptions?.whileElementsMounted,
+      ),
+      middleware: [...(middleware ?? []), hide()],
+      ...restFloatingOptions,
```
YousefED, line 149: what's the use of `hide` again (which use-case from user perspective?)

YousefED:
Main issue atm is that positioning still doesn't work well:
- on Chrome (Android) with address bar at bottom
- on Brave (Android), always
- When switching "modes" in the example (might be example specific, lets fix the basics first)

Discussing this with Matthew via DM

matthewlipski:
> Using a fixed positioning strategy in floating ui likely will resolve this like it did here: #3002

This almost works, it looked very promising when I first tried it but unfortunately the sizing gets messed up when scrolling. The issue is also related to the visual viewport scrolling within the layout viewport.

https://github.com/user-attachments/assets/35fee16d-d743-4b72-b32c-63f53db2fea5

packages/react/src/components/FormattingToolbar/DefaultButtons/CreateLinkButton.tsx
matthewlipski: See #2982

packages/react/src/components/Popovers/GenericPopover.tsx
matthewlipski: Makes the popover not visible when the reference element is also not visible. Allows for popovers to overflow the editor if more space is available, but stay hidden if the reference gets clipped by overflow. Basically a nicer UX alternative to keeping the popover within the same clipping container as the reference.

docs/content/docs/react/components/formatting-toolbar.mdx
matthewlipski: Fixed - removed the duplicated clause.

docs/content/docs/react/components/formatting-toolbar.mdx
matthewlipski: Reworded to "Which one you get depends on your app's layout:".

docs/content/docs/react/components/formatting-toolbar.mdx
matthewlipski: Updated - the CSS rules (the scroll lock + `.bn-scroll-host` positioning) now ship in BlockNote's stylesheet, so I reworded "injects" to reflect that. The `--bn-vv-*` variables are still set on `<html>` at runtime by JS (they're dynamic and track the visual viewport, so they can't live in a static stylesheet) - I've clarified that distinction in the text.

examples/03-ui-components/14-mobile-formatting-toolbar/src/DummyUI.tsx
matthewlipski: Renamed the hamburger links to "Dummy link 1-4" so they clearly read as placeholders.

examples/03-ui-components/14-mobile-formatting-toolbar/src/App.tsx
matthewlipski: Unified everything on **scroll host** (matching the public `bn-scroll-host` class): the `scrollMode` state value is now `"scroll-host"`, the toggle label is "Pinned scroll host", and the docs/README use "scroll host" throughout instead of "scroll container".

packages/mantine/src/menu/Menu.tsx
matthewlipski: Not intentional - it was accidentally dropped during the refactor and has since been restored in 22c7645, so `returnFocus={false}` is still there on the current branch.

packages/react/src/components/FormattingToolbar/DefaultButtons/ColorStyleButton.tsx
matthewlipski: Documented - added a comment explaining that the mobile toolbar scrolls horizontally (clipping the inline dropdown), so on mobile we portal it to `editor.portalElement` to escape the clip, and that a set `portalRoot` also prevents focus moving into the dropdown (which would dismiss the keyboard). Desktop has no such clipping so keeps the default inline rendering.

packages/react/src/components/FormattingToolbar/DefaultButtons/FileRenameButton.tsx
matthewlipski: Documented (same comment added here).

packages/react/src/components/FormattingToolbar/DefaultButtons/FileCaptionButton.tsx
matthewlipski: Documented. On testing: the file caption/rename dialogs use the exact same portal path as the color menu, so the clipping/keyboard behaviour is shared - but I haven't given these two specific dialogs a dedicated on-device pass yet. I'll verify the popover isn't clipped and the keyboard stays up before this merges; if you already hit a concrete issue with them on mobile, let me know what you saw.

packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx
matthewlipski: `useEditorState` only subscribes to the editor's transaction/selection/change/mount events - not its `focus`/`blur` events - so it can't observe focus changes (focusing doesn't dispatch a transaction). On top of that, the blur path here needs bespoke logic: a deferred `requestAnimationFrame` re-check of `document.activeElement` against `isWithinEditor`, so the toolbar stays up while a toolbar popover's input pulls focus into the portal. That can't be expressed as a pure selector, so `useEditorState` isn't a clean fit and the direct `on("focus"/"blur")` listeners are simpler here.

examples/03-ui-components/14-mobile-formatting-toolbar/src/App.tsx
matthewlipski: Correction: settled on **scroll container** as the single term (not "scroll host"). The public class is now `bn-scroll-container` too (renamed from `bn-scroll-host`), so the class name, the `scrollMode` state value, the docs, and the README all use "scroll container" consistently.

packages/react/src/components/FormattingToolbar/DefaultButtons/CreateLinkButton.tsx
YousefED: This is the bug the stack on top of this PR fixes — #3030: with a lone text field outside a `<form>`, Android's IME assigns Enter the "Next" action (advance focus, no key event at all), so the popover never hears anything. `Form.Root` becoming a real form is what makes the IME offer a submitting action; verified on a device. Covered by `end-to-end/form/` and `end-to-end/mobile/linkSubmit.test.tsx`.

packages/react/src/components/FormattingToolbar/MobileFormattingToolbarController.tsx
YousefED: Resolved by #3028 in the stack: the controller's hand-rolled focus/blur wiring (including its private `editor._tiptapEditor` reach) is replaced by a new `useEditorFocus({ includeEditorUI: true })` hook — the state counterpart to `useEditorFocusChange`, same split as `useEditorState`/`useEditorChange`.

examples/03-ui-components/14-mobile-formatting-toolbar/.bnexample.json
```diff
@@ -1,5 +1,5 @@
 {
-  "playground": true,
+  "playground": false,
```
YousefED: imo it would still be nice to access this through the playground. Can't we have it adjust the scroll-container option of the playground? (it's just a classname right?)

packages/react/src/components/FormattingToolbar/useVirtualKeyboard.ts
```diff
@@ -67,6 +100,8 @@ export function useVirtualKeyboard(): boolean {
   const [open, setOpen] = useState(isVirtualKeyboardOpen);
 
   useLayoutEffect(() => {
+    warnIfViewportMetaMisconfigured();
```
YousefED: are we logging this only on mobile? not sure that's useful, because not many people see the console on mobile

packages/ariakit/src/toolbar/ToolbarSelect.tsx
```diff
@@ -40,6 +40,7 @@ export const ToolbarSelect = forwardRef<
         className={mergeCSSClasses("bn-ak-popover", className || "")}
         ref={ref}
         gutter={4}
+        portalElement={portalRoot ?? undefined}
```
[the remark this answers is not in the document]
YousefED: Handled in #3030 on this stack: the select popover gets `autoFocusOnShow={!preventFocusOnOpen}` (the mobile toolbar sets the prop), and `tests/src/end-to-end/mobile/skinFocus.test.tsx` pins per skin that opening the block type select keeps focus in the editor.

examples/03-ui-components/14-mobile-formatting-toolbar/src/style.css
```diff
@@ -0,0 +1,155 @@
+html,
+body {
+  margin: 0;
+}
```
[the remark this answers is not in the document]
YousefED: Skipped on purpose. In the scroll-container mode the container is position: fixed and fills the viewport, so the document itself does not overflow and a root overflow lock has nothing to lock; overscroll-behavior: contain already stops the chaining. It is also the one lock iOS Safari does not honour reliably (an overflow: hidden root still rubber-bands), which is why the container is fixed in the first place. The example's own style.css only has to neutralise the playground's page-level scroll container per mode.

YousefED (approved):

nperez0111 (approved):

PR #2975, opened 2026-08-16
fix(core): bind Mod-a to select all the document

Asodariyasujal:
BlockNote had no `Mod-a` binding, so select-all was left to the browser's native `contenteditable` handling and ProseMirror had to rebuild a document selection from the DOM selection it produced.

That fails when a block puts non-editable content first, which check list items do: the checkbox div sits ahead of the `<p>` holding the block's content. So in a document starting with a check list item, ProseMirror could not map the DOM selection to a valid position and dropped it, leaving the caret in place - Backspace then only edited that one block instead of clearing the document.

Now `Mod-a` sets an `AllSelection` itself, which selects every block type reliably and deletes down to a single empty paragraph.

Also stops `getNearestBlockPos` warning for the positions at the very start and end of the doc, which is where an `AllSelection` ends.

## Summary by CodeRabbit

* **New Features**
  * Added a keyboard shortcut for selecting all content in the editor using Ctrl/Cmd+A.
  * Select-all works across multi-block documents, including paragraphs, checklists, and math blocks.

* **Bug Fixes**
  * Improved handling of selections at document boundaries, ensuring the first or last block is selected correctly.

* **Tests**
  * Expanded coverage for keyboard shortcuts, full-document selection, and clearing selected content.

### DIFF
--- a/packages/core/src/api/getBlockInfoFromPos.ts
+++ b/packages/core/src/api/getBlockInfoFromPos.ts
@@ -126,6 +126,31 @@ export function getNearestBlockPos(doc: Node, pos: number) {
     node = $pos.node(depth);
   }
 
+  // The doc's boundary positions (0 and `doc.content.size`) sit around the
+  // `blockGroup` holding the top-level blocks, so they're outside every block.
+  // Unlike the positions handled below, they're expected rather than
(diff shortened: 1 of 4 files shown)


Asodariyasujal:
@coderabbitai review

Issue #2595, opened 2026-03-27
Typing/echo lag with many blocks (~50k chars total); plain long paragraphs are fine


Lusice520:
### What’s broken?

When a document contains many blocks, typing starts to feel delayed: the text appears in the editor a bit after the IME/input has already produced it.

This is not just an IME-specific issue on one input method. I can reproduce the same visible echo delay with different input methods, and the same content also lags in the official BlockNote demo.

What seems important is the number of blocks, not only the total character count:

- A long document with relatively few large paragraph blocks is much smoother.
- A document with many `heading` blocks starts lagging around ~50k characters total.
- A document with many `bulletListItem` / `numberedListItem` blocks also starts lagging around ~50k characters total.

So the current suspicion is that performance degrades with block count / block-level rendering cost, rather than with plain text length alone.

Also relevant:
- The same content reproduced in plain Tiptap does **not** show the same lag.
- The lag is about typing echo / visible editor update delay, not initial load time or scrolling.

### What did you expect to happen?

Typing should remain responsive even when the document has many blocks.
The visible editor echo should keep up with input instead of lagging behind.

### Steps to reproduce

 Open BlockNote with the latest version.
2. Create a document that has many blocks instead of a few long paragraphs.
3. For example, create:
   - many H1 blocks totaling around ~50k characters, or
   - many list item blocks totaling around ~50k characters.
4. Start typing continuously.
5. Notice that the visible editor echo/caret update falls behind input.

Control case:
1. Create another document with similar total text length, but keep it as a small number of long paragraph blocks.
2. Typing is noticeably smoother there.

Additional comparison:
- The same content also reproduces in the official BlockNote demo.
- The same content in Tiptap is much smoother / does not show the same lag.

### BlockNote version

0.47.3

### Environment

- React 18 - Reproduced on the latest BlockNote as of March 27, 2026

### Additional context

This does not currently look like:
- app-specific autosave logic
- a floating toolbar integration
- one specific IME implementation

It looks more like a BlockNote core/editor performance issue related to many blocks.

### Contribution

- [ ] I'd be interested in contributing a fix for this issue

### Sponsor

- [ ] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

PR #2981, opened 2026-08-18
perf(core): resolve block changes from changed range only

nperez0111:
# Summary

Resolve block changes from only the range a transaction touched, instead of snapshotting the entire document on every transaction.

## Rationale

`getBlocksChangedByTransaction` (behind `editor.getChanges()`) walked the whole document and ran a `nodeToBlock()` conversion for every block, twice, per transaction. Because apps typically read `getChanges()` inside `onChange` on every keystroke, this made typing lag in large documents (O(document-size) per keypress). The existing `#2595` perf test never subscribed to `onChange`, so this path was uncovered.

## Changes

- `getBlocksChangedByTransaction` now diffs only the blocks overlapping the changed range (mapped into both the old and new document) rather than snapshotting every block.
- Added a shared `getChangedRange()` helper: a drop-in for ProseMirror's `changedRange()` that also covers attribute-only steps (`AttrStep`) and mark steps, at the same O(steps) cost.
- `PreviousBlockType` now uses `getChangedRange()`, fixing a latent bug where attribute-only changes (e.g. a heading's `level`, a numbered list `index`) were silently missed by its ranged diff — breaking those CSS transitions.

## Impact

Behavior of `getChanges()` is unchanged (all 25 snapshot tests pass as-is); only the cost model changes, from O(document-size) to O(changed-range) per transaction. `PreviousBlockType` now correctly animates attribute-only block changes it previously dropped.

## Testing

`vp run test` (full `@blocknote/core` suite, 748 passing) plus lint/type-check pass. Added a `getChanges()`-in-`onChange` performance regression test (previously-uncovered path) and a `PreviousBlockType` test for the `AttrStep` (heading-level) case that fails without the fix.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

Relates to #2595 — the earlier plugin optimizations closed that issue, but the `getChanges()`/`onChange` path (which apps hit on every keystroke) still snapshotted the whole document. This PR completes that fix and adds the regression test the original lacked.

The `PreviousBlockType` change is a correctness fix (not just perf) surfaced while investigating the same optimization; it is isolated to that file and its test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of document changes, including heading-level and other block attribute updates.
  - More accurately limits change tracking to affected document ranges.
  - Unchanged transactions are handled without unnecessary document processing.

- **Performance**
  - Improved efficiency when detecting changes in large documents, helping maintain responsive editing as document size grows.

- **Tests**
  - Added coverage for attribute-only updates and change-detection performance across small and large documents.

### DIFF
--- a/packages/core/src/api/getBlocksChangedByTransaction.test.ts
+++ b/packages/core/src/api/getBlocksChangedByTransaction.test.ts
@@ -2,7 +2,10 @@ import { describe, expect, it, beforeEach } from "vite-plus/test";
 
 import { setupTestEnv } from "./blockManipulation/setupTestEnv.js";
 import { getBlocksChangedByTransaction } from "./getBlocksChangedByTransaction.js";
+import { getBlockInfo } from "./getBlockInfoFromPos.js";
+import { getNodeById } from "./nodeUtil.js";
 import { BlockNoteEditor } from "../editor/BlockNoteEditor.js";
+import { PartialBlock } from "../blocks/defaultBlocks.js";
 
 const getEditor = setupTestEnv();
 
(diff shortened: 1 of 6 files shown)


YousefED:
sweet!

PR #2983, opened 2026-08-18
fix: update selected inline content styling

matthewlipski:
# Summary

This PR removes the border from selected inline content and changes the background to a more visible color. Selected block styling remains the same.

## Rationale

The current selected inline content styling looks off.

## Changes

- Added CSS rule.

## Impact

N/A

## Testing

N/A

## Screenshots/Video

Before:

image

After:

image

## Checklist

## Summary by CodeRabbit

- **Style**
  - Improved visual highlighting for selected inline elements with a darker translucent background.
  - Removed the inset border from inline selections while preserving existing styling for other selected elements.

### DIFF
--- a/packages/core/src/editor/Block.css
+++ b/packages/core/src/editor/Block.css
@@ -45,6 +45,11 @@ BASIC STYLES
   pointer-events: none;
 }
 
+.bn-inline-content .ProseMirror-selectednode::after {
+  background-color: rgba(100, 160, 255, 0.3);
+  box-shadow: none;
+}
+
 .bn-block-content::before {
   content: "";
   margin-right: 0;


YousefED:
Nice. Tested on both dark and light mode? seems like those should be different colors, right?

nperez0111:
I wonder if it would've been better as an outline rather than a border?

matthewlipski:
> Nice. Tested on both dark and light mode? seems like those should be different colors, right?

Color is the same for both because it's a translucent overlay, so it looks different on light and dark mode because the light/dark background is still visible behind it.

> I wonder if it would've been better as an outline rather than a border?

Tried that but wasn't a fan. For selected inline content to look good with a border you'd need to add padding, else the border can overlap content inside. The added padding then looks weird when the inline content isn't selected, as there is additional empty space before and after. I briefly considered adding the padding only when selected but having text around the inline content get repositioned because of it felt janky.

PR #2985, opened 2026-08-19
docs: name the mobile toolbar layouts + add layout toggle to the example

YousefED:
Follow-up to #2939, targeting its branch. Docs/example-only — no behavior changes in `@blocknote/react` (two JSDoc comments reworded).

## Why

The opt-in "smooth scrolling" setup was described three different ways (the docs called it a "workaround", the code comments "non-scrolling document", the example `.scroll-host`), which made the docs section hard to follow. This PR gives the two page layouts names and restructures the section around them:

- **Scrolling document** (default) – the page scrolls, BlockNote repositions the toolbar as you scroll (slight lag/jitter).
- **Pinned scroll container** (opt-in) – `html`/`body` locked, a container pinned to the visual viewport scrolls instead, toolbar is truly fixed.

"Pinned scroll container" names the one thing the host app has to add; "locked document" would name the precondition instead, and "jank-free mode" the outcome.

### Docs – `formatting-toolbar.mdx`
Simple-first rewrite of **Mobile Formatting Toolbar**: intro (on by default on touch devices, nothing to set up) → two-bullet summary naming both layouts → `### Scrolling document` → `### Pinned scroll container` with the same two CSS rules as before, the `--bn-vv-*` vars (now also listing `--bn-vv-scale`) and the actionable trade-off (pull-to-refresh etc.). Headings give stable anchors (`#scrolling-document`, `#pinned-scroll-container`).

### The embedded example broke the docs page → `docs: false`
The docs render examples **inline** (`dynamic()` import, no iframe), so this example's page-level CSS leaked into the docs page: `html, body { overflow: hidden }`, the full-viewport `position: fixed` `.scroll-host`, `.bn-container { height: 300px }` (hits the other two examples on the page) and `.prose h2/p` (collides with fumadocs' typography class). On the #2939 Vercel preview the lorem ipsum and demo editor are fixed over the prose and the page can't scroll.

Since a pinned-scroll-container layout inherently owns the whole page, it can't be embedded. Following the `editor-scrollable` / `comments.mdx` precedent, the example is now `docs: false` and the docs link to `https://playground.blocknotejs.org/ui-components/mobile-formatting-toolbar?hideMenu=true`. If you'd rather keep an embed, it needs an iframe-based one.

### Example – layout toggle
A "Pinned scroll container" switch in the demo nav bar toggles the layout so both can be compared on a device. It's pure CSS: `App.tsx` toggles a `pinned-scroll-container` class on `<html>` (on `<html>` rather than `<body>` so the lock rule can scope both `html` and `body` without `:has()`), and `style.css` scopes the two rules under it, labelled "step 1" / "step 2". Defaults to on; comments say a real app applies the rules unconditionally. Verified in iPhone emulation: class/overflow/`position` flip correctly and the toolbar still lands above the keyboard with the host resized to the visual viewport.

Known/inherent: with the playground menu visible, the full-viewport fixed `.scroll-host` sits under the playground sidebar (hence `?hideMenu=true` in the docs link).

### Playground – `.mantine-AppShell-root { width: 100vw }` → `100%`
Separable. `100vw` includes the vertical scrollbar, so any playground example taller than the viewport shows a horizontal scrollbar with classic/always-visible scrollbars (this example exposes it when the toggle is off; pinned mode only masks it because the document can't scroll). Reproduced with real scrollbars (`clientWidth` 1265 vs `scrollWidth` 1280) and verified fixed.

### Housekeeping
- README, JSDoc in `MobileFormattingToolbarController` / `useVirtualKeyboard`, and example comments aligned with the naming (also fixes the stale `DemoChrome` → `DummyUI.tsx` reference).
- `examples.gen.tsx` regenerated via `vp run gen` — it was already stale vs. the example README on the branch.

## Verification
- `node docs/validate-links.mjs` → 0 errors; docs page renders (200) with the new anchors, no overlay.
- `pnpm run lint` clean.
- Playwright emulation (iPhone 13, Chromium) for the toggle and toolbar placement; desktop with real scrollbars for the playground fix.

### DIFF
--- a/docs/content/docs/react/components/formatting-toolbar.mdx
+++ b/docs/content/docs/react/components/formatting-toolbar.mdx
@@ -41,15 +41,22 @@ Here, we use the `FormattingToolbar` component but keep the default buttons (we 
 
 ## Mobile Formatting Toolbar
 
-On mobile, BlockNote's default UI automatically shows a dedicated formatting toolbar pinned just above the on-screen keyboard - no setup needed. It renders the same items as the regular Formatting Toolbar, but stays anchored to the keyboard so it's always reachable while editing on a touch device. Try it in any of the previous examples to see it in action!
+On touch devices, BlockNote's default UI replaces the floating Formatting Toolbar with a mobile Formatting Toolbar that sits just above the on-screen keyboard. It shows the same items as the regular Formatting Toolbar and is enabled by default - there's nothing to set up. Open any of the examples above on a phone to see it.
 
-Due to browser limitations, scrolling the page can cause the mobile Formatting Toolbar to appear laggy or jittery. BlockNote offers a workaround for these limitations, which you can see below.
+The mobile Formatting Toolbar works with two page layouts. Which one you get is decided purely by your app's CSS:
 
-<Example name="ui-components/mobile-formatting-toolbar" />
+- **Scrolling document** (the default): the page scrolls as usual and BlockNote repositions the toolbar as you scroll.
+- **Pinned scroll container**: the document itself doesn't scroll; a container pinned to the visual viewport scrolls instead, and the toolbar never has to move.
 
-Here, the lag/jitter is eliminated, at the cost of `<body>` and its ancestors no longer being scrollable. Instead, all scrollable page content must be in a scrollable container that's a descendant of `<body>`.
+### Scrolling document
 
-To set this up, first lock scrolling on the document itself. This prevents the browser from scrolling `<html>`/`<body>`, which is what causes the toolbar to jitter:
+This is what you get without any changes to your app. The toolbar follows the visible area above the keyboard as the page scrolls. Mobile browsers only report visual viewport changes after the fact, so the toolbar can lag or jitter slightly while the page is scrolling. If that matters for your app, switch to a pinned scroll container.
+
+### Pinned scroll container
+
+In this layout, `<html>` and `<body>` are locked and all page content lives inside a single scroll container that BlockNote keeps aligned with the visual viewport. Since the document never scrolls, the toolbar can stay at a truly fixed position and the lag/jitter disappears. The trade-off is that browser gestures which rely on document scrolling, like pull-to-refresh, no longer work.
+
+Setting it up takes two CSS rules. First, lock scrolling on the document:
 
 ```css
 html,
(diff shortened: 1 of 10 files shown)


docs/content/docs/react/components/formatting-toolbar.mdx
```diff
@@ -41,15 +41,22 @@ Here, we use the `FormattingToolbar` component but keep the default buttons (we
...
 
-Here, the lag/jitter is eliminated, at the cost of `<body>` and its ancestors no longer being scrollable. Instead, all scrollable page content must be in a scrollable container that's a descendant of `<body>`.
+### Scrolling document
 
-To set this up, first lock scrolling on the document itself. This prevents the browser from scrolling `<html>`/`<body>`, which is what causes the toolbar to jitter:
+This is what you get without any changes to your app. The toolbar follows the visible area above the keyboard as the page scrolls. Mobile browsers only report visual viewport changes after the fact, so the toolbar can lag or jitter slightly while the page is scrolling. If that matters for your app, switch to a pinned scroll container.
+
+### Pinned scroll container
+
+In this layout, `<html>` and `<body>` are locked and all page content lives inside a single scroll container that BlockNote keeps aligned with the visual viewport. Since the document never scrolls, the toolbar can stay at a truly fixed position and the lag/jitter disappears. The trade-off is that browser gestures which rely on document scrolling, like pull-to-refresh, no longer work.
```
matthewlipski: 2 small things:

- The other main trade-off is that the browser UI may no longer show/hide on scroll
- Different browsers may have both of these issues, only 1 of them, or neither. So I would change the framing of the last sentence and say these trade-offs may occur depending on the browser used.

docs/content/docs/react/components/formatting-toolbar.mdx
```diff
@@ -74,4 +81,6 @@ Then, make your scroll container (`.scroll-host` in the demo) the element that a
 }
 ```
 
-BlockNote keeps the `--bn-vv-*` variables up to date as the keyboard opens/closes and the user zooms or pans, so both the toolbar and your scroll container stay aligned with the visual viewport without any JavaScript on your end.
+BlockNote keeps `--bn-vv-top`, `--bn-vv-left`, `--bn-vv-width`, and `--bn-vv-height` (plus `--bn-vv-scale`, the pinch-zoom factor) up to date as the keyboard opens and closes and as the user pans or zooms, so the scroll container always lines up with the visible area above the keyboard without any JavaScript on your end.
```
matthewlipski: Perhaps worth saying that these variables track the [visual viewport](https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport)

matthewlipski (commented):
Good changes, left 2 comments

docs/content/docs/react/components/formatting-toolbar.mdx
YousefED: sounds good

docs/content/docs/react/components/formatting-toolbar.mdx
YousefED: Would be a good addition, if you can find a good way to phrase this. "Browser UI" is very generic / not clear to me as a reader what this is

matthewlipski (approved):

PR #2986, opened 2026-08-20
fix(core): resolve the table drag position against the current document

### LINKED ISSUE #2921: Table drag: view.tablePos is never remapped through tr.mapping, throws on a concurrent edit mid-drag
## Summary
While building a table-drag visual-feedback example (#2920), a reviewer asked about behavior when two users edit a table at the same time. Investigating turned up a crash bug in `TableHandlesExtension` itself, unrelated to that PR's own code.

## Root cause
In `packages/core/src/extensions/TableHandles/TableHandles.ts`, `TableHandlesView.tablePos` is set once per `mousemove`:

```ts
this.tablePos = pmNodeInfo.posBeforeNode + 1;
```

It's then used directly, without ever being remapped through `tr.mapping`, in several places - notably the drop-cursor `decorations()`:

```ts
const tableResolvedPos = state.doc.resolve(view.tablePos + 1);
```

`mousemove` doesn't fire on the dragged-over element while a native HTML5 drag is in progress, so `tablePos` (and the `state.block` content snapshot used later by `dropHandler`) can't refresh during a drag. If any transaction changes the document elsewhere while a drag is active - a concurrent local edit, or another collaborator's change over Yjs - `tablePos` goes stale.

## Reproduction
1. Start dragging a table row (mousedown on the row handle, then move over another row without releasing).
2. While the drag is still in progress, dispatch any transaction that changes the document before the table (e.g. `editor._tiptapEditor.view.dispatch(editor._tiptapEditor.view.state.tr.insertText("X", 1, 1))` to insert into a heading above the table).
3. The next `dragover` recomputes the drop-cursor decoration from the now-stale `tablePos`, which resolves into the wrong node and throws:

```
RangeError: Index 1 out of range for <"...heading text...">
  at posAtIndex (prosemirror-model)
  at TableHandles.ts:695 (decorations)
```

This throws out of `viewDecorations` → `updateStateInner` → `dispatchTransaction`, i.e. it breaks the transaction dispatch for the *other* user's edit, not just the drag.

## Suggested fix
Same pattern already used correctly elsewhere for stored positions across transactions: remap `tablePos` (and any other stored position/snapshot used mid-drag) through `tr.mapping` in an `appendTransaction`/plugin `apply`, rather than only refreshing it on `mousemove`. Happy to open a PR for this if useful, but wanted to report it as its own issue since it's unrelated to the table-drag-visualization example in #2920 where it was found.

## Environment
Found against `main` while working from a local build of `packages/core`/`packages/react`/`packages/mantine`.

---

## Update — after #2920 (`225c1ccf4`)

Recording where this stands, since #2920 changes the symptom but not the cause.

**The crash path is now guarded.** #2920 moves the drag decoration build into `getTableDragDecorations` and calls it inside a single `try`/`catch` in the plugin's `decorations` prop, covering every `resolve`/`posAtIndex` in both the highlight and drop-cursor branches. It also verifies the resolved node is actually a table before indexing into it. With that in place the reproduction above no longer throws — the decorations are skipped for that state, so the other user's transaction dispatches normally instead of `dispatchTransaction` blowing up.

**What this issue still covers:** `view.tablePos` is captured on `mousemove` and never refreshed against the current document. That is the whole of it.

**Correction.** An earlier version of this section also claimed `view.state.block` goes stale and that `dropHandler` therefore works from a stale snapshot, so a concurrent edit could land a wrong drop. That was wrong, and I withdraw it. `TableHandlesView.update()` already re-resolves the block on every view update:

```ts
this.state.block = this.editor.getBlock(this.state.block.id)!;
```

`tablePos` is the only piece that isn't refreshed, which is what the title of this issue says. Apologies for the noise.

So the impact splits cleanly by version. Before #2920: the `RangeError` above, which escapes `dispatchTransaction`. After #2920: the guard catches it, so the drop cursor silently stops rendering for the rest of the drag while the drop itself still lands correctly. A visual bug rather than a correctness one — but still worth fixing, since the guard is a backstop, not a fix.

Note the stack trace above cites `TableHandles.ts:695`; that line number predates #2920 and has since moved.

Per @nperez0111's comment below, this will come as its own PR.

mustafa-yilmaz:
# Summary

Fixes #2921. `TableHandlesView.tablePos` was captured on `mousemove` and then used as-is. `mousemove` doesn't fire while a native HTML5 drag is in progress, so any transaction that changed the document mid-drag left it stale, and the next `decorations()` call resolved it into whatever node now occupied that position:

```
RangeError: Index 1 out of range for <paragraph("…")>
```

That escapes `viewDecorations` → `updateStateInner` → `dispatchTransaction`, so it breaks the dispatch of the *other* edit, not just the drag.

## Rationale

The trigger is any document change during a drag — a concurrent local edit, or another collaborator's change arriving over Yjs. Since the failure surfaces in the transaction dispatch of the edit that arrived, a remote user's typing can break the editor for whoever happens to be dragging a table row at the time.

## Changes

- Adds `TableHandlesView.getTablePos(doc)`, which resolves the position against the document it's about to be used with: it keeps the cached `tablePos` when that still points at the table, and re-resolves it from the table's block ID when it doesn't.
- Routes all three consumers through it: the drop-cursor `decorations()`, the `tablePos` put on the drag-start transaction meta, and `setCellSelection()`.

Re-resolving by ID rather than remapping through `tr.mapping` was a deliberate choice, and differs from the suggestion in the issue. A mapping follows the position when surrounding content shifts, but not when the table node is replaced outright — which is what Yjs does when, for example, a cell colour is changed from the side menu. The block ID survives both. The cached position is validated first, so the common path is a single `resolve` rather than a document walk.

Worth noting for review: `view.state.block` is *not* affected. `TableHandlesView.update()` already re-resolves it from its ID on every view update, so `tablePos` was the only piece left unrefreshed. An earlier comment of mine on the issue claimed otherwise; it's corrected there.

## Impact

Behaviour is unchanged when nothing edits the document mid-drag, which is the overwhelmingly common case — the cached position validates and is returned as-is.

No public API change. `getTablePos` is added to `TableHandlesView`, and `tablePos` remains a public property.

This overlaps with #2920, which wraps the drag decoration build in a `try`/`catch` so a stale position skips the decorations rather than throwing. That is a backstop for the same failure; this PR removes the cause, so the backstop stops being load-bearing. The two are independent and can land in either order — this branch is written against `main` without #2920 — but they touch the same file and the second to merge will need a small conflict resolution.

## Testing

Adds an end-to-end regression test in `tests/src/end-to-end/tables/tables.test.tsx`. It drives a real drag via Playwright's mouse, dispatches an insert into the paragraph above the table while the drag is in progress, and then asserts:

- the drop cursor still renders afterwards, so the position was re-resolved rather than merely swallowed by a guard, and
- the concurrent edit itself landed, i.e. its transaction dispatched normally.

The test reproduces the `RangeError` above without the fix and fails on chromium and webkit; it passes with it. Drag tests skip on firefox, matching the existing convention in that file.

Also run: the full browser e2e suite (the failures present are pre-existing on this base and unrelated — keyboard handlers, AI selection, comments, y-prosemirror — all confirmed failing on a clean tree too), and `@blocknote/core`'s unit tests, 695 passing.

## Checklist

- [ ] The documentation has been updated to reflect the new feature — n/a, internal bug fix with no documented surface.

## Additional Notes

Branched off `v0.52.1` rather than current `main`, so it is behind by a number of commits. Happy to rebase if that's preferred before review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table row dragging when document edits occur during an active drag.
  * Preserved drop-cursor positioning and existing edits during concurrent changes.
  * Improved table selection and drag behavior when table positions become outdated.

* **Tests**
  * Added coverage for dragging table rows while editing content above the table.

### DIFF
--- a/packages/core/src/extensions/TableHandles/TableHandles.ts
+++ b/packages/core/src/extensions/TableHandles/TableHandles.ts
@@ -1,3 +1,4 @@
+import type { Node } from "prosemirror-model";
 import { EditorState, Plugin, PluginKey, PluginView } from "prosemirror-state";
 import {
   CellSelection,
@@ -526,6 +527,44 @@ export class TableHandlesView implements PluginView {
 
     return true;
   };
+  /**
+   * The position just before the table node, resolved against `doc`.
+   *
+   * `tablePos` is only refreshed on `mousemove`, which doesn't fire while a
+   * native drag is in progress, so any transaction that changes the document
+   * mid-drag leaves it stale - resolving it then lands in the wrong node, or
+   * past the end of the document, and throws (#2921).
+   *
+   * Rather than mapping the stored position through every transaction, it's
+   * checked against the document it's about to be used with and re-resolved
+   * from the table's block ID when it no longer points at that table. That
+   * also covers the table node being replaced outright rather than moved,
+   * which a mapping wouldn't follow - it happens when collaborating.
+   */
+  getTablePos(doc: Node): number | undefined {
+    if (this.tableId === undefined) {
+      return undefined;
+    }
+
+    // `tablePos` sits just inside the block container, so the node it resolves
+    // into is the container, which carries the ID to check against. A false
+    // negative here only costs the lookup below, so the cheap check is enough.
+    if (this.tablePos !== undefined && this.tablePos <= doc.content.size) {
+      try {
+        if (doc.resolve(this.tablePos).parent.attrs.id === this.tableId) {
+          return this.tablePos;
+        }
+      } catch {
+        // Out of range for this document - re-resolve below.
+      }
+    }
+
+    const posInfo = getNodeById(this.tableId, doc);
+    this.tablePos = posInfo && posInfo.posBeforeNode + 1;
+
+    return this.tablePos;
+  }
+
   // Updates drag handles when the table is modified or removed.
   update() {
     if (!this.state || !this.state.show) {
@@ -645,9 +684,16 @@ export const TableHandlesExtension = createExtension(({ editor }) => {
             if (
               view === undefined ||
               view.state === undefined ||
-              view.state.draggingState === undefined ||
-              view.tablePos === undefined
+              view.state.draggingState === undefined
             ) {
+              return;
+            }
+
+            // Resolved against the state being rendered, rather than read from
+            // the last `mousemove`, so a document change mid-drag can't leave
+            // the decorations pointing at a position that no longer exists.
+            const tablePos = view.getTablePos(state.doc);
+            if (tablePos === undefined) {
               return;
             }
 
@@ -681,7 +727,7 @@ export const TableHandlesExtension = createExtension(({ editor }) => {
             }
 
             // Gets the table to show the drop cursor in.
-            const tableResolvedPos = state.doc.resolve(view.tablePos + 1);
+            const tableResolvedPos = state.doc.resolve(tablePos + 1);
 
             if (view.state.draggingState.draggedCellOrientation === "row") {
               const cellsInRow = getCellsAtRowHandle(
@@ -816,7 +862,7 @@ export const TableHandlesExtension = createExtension(({ editor }) => {
             view!.state!.draggingState!.draggedCellOrientation,
           originalIndex: view!.state!.colIndex,
           newIndex: view!.state!.colIndex,
-          tablePos: view!.tablePos,
+          tablePos: view!.getTablePos(tr.doc),
         }),
       );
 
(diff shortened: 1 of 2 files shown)


mustafa-yilmaz:
Closing this — it's redundant. #2921 was already fixed on `main` by #2972 (`b9cd9c4f`), and I missed that before opening this.

`b9cd9c4f` added `TableHandlesView.getTablePos(doc)`, which looks the table up by ID on each use, and `decorations()` already calls it. That is the same fix as this PR, arrived at independently — same method name, same signature, same reasoning, including the point that ProseMirror computes decorations for a transaction before it calls the plugin view's `update`.

My mistake was branching from `v0.52.1` rather than current `main`, and then not re-checking that the bug still reproduced there before writing the fix.

I did verify rather than just assume. The end-to-end regression test on this branch reproduces the original `RangeError` — it drives a real drag, dispatches an insert into the paragraph above the table mid-drag, and asserts the drop cursor still renders afterwards. Against `b9cd9c4f` with no other changes, it passes on chromium and webkit.

Sorry for the noise, @nperez0111. Closing #2921 as fixed too.

PR #2987, opened 2026-08-20
docs: migrate demo to yhub collaboration on stable yjs 13

nperez0111:
# Summary

Migrates the public `/demo` page's real-time collaboration from `y-partykit` to `yhub.teleportal.tools`, using a plain `y-websocket` `WebsocketProvider` on stable `yjs` v13 (not the experimental `@y/y` v14 stack used by the existing YHub versioning examples).

## Rationale

Requested migration to yhub as the demo's collaboration backend, while staying on the stable `yjs` v13 API surface. While testing, found that the demo's `collaboration` option was passed directly into `useCreateBlockNote()` instead of being wrapped in `withCollaboration()`, which is required since the yjs/blocknote-core decoupling refactor. As a result, real-time sync was silently broken on the live demo (ProseMirror updated locally, but nothing ever reached the underlying `Y.Doc`, so collaboration between clients never occurred) — this is the root cause behind BLO-1317. Fixed the wiring alongside the provider swap.

## Changes

- Replaced `YPartyKitProvider` with `y-websocket`'s `WebsocketProvider`, connecting to `wss://yhub.teleportal.tools/api/ws/v1`.
- Wrapped the editor's options in `withCollaboration()` (from `@blocknote/core/yjs`) so the Yjs sync/cursor/undo extensions actually attach.
- Removed the now-unused `y-partykit` dependency from `docs/package.json`.

## Impact

Fixes real-time collaboration on the public demo page, which was silently non-functional. No API changes to BlockNote packages themselves.

## Testing

Manually verified in two browser tabs against the live `yhub.teleportal.tools` server: late-joining clients receive existing content, and concurrent edits propagate live between tabs. Ran `pnpm run lint` clean.

## Checklist

## Additional Notes

Fixes BLO-1317: https://linear.app/blocknote/issue/BLO-1317/demo-on-website-is-broken

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collaboration reliability in the demo editor.
  * Ensured collaboration connections are properly cleaned up when leaving the editor.
  * Improved handling of room identifiers containing special characters.

### DIFF
--- a/docs/app/demo/_components/DemoEditor.tsx
+++ b/docs/app/demo/_components/DemoEditor.tsx
@@ -10,7 +10,7 @@ import {
   CommentsExtension,
   DefaultThreadStoreAuth,
 } from "@blocknote/core/comments";
-import { YjsThreadStore } from "@blocknote/core/yjs";
+import { YjsThreadStore, withCollaboration } from "@blocknote/core/yjs";
 import { filterSuggestionItems } from "@blocknote/core/extensions";
 import "@blocknote/core/fonts/inter.css";
 import * as locales from "@blocknote/core/locales";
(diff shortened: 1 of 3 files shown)


YousefED:
@nperez0111 should we throw an error (or even better, fix the typings), so that we would have caught the issue at compile time? I assume other consumers will also run into this

PR #2989, opened 2026-08-20
Fix versioning diff re-render over suggestion docs; surface recovered sync errors

YousefED:
## The bug

Editing "Version 2" in the suggestion gallery's versioning mode silently stopped updating the Diff pane whenever the rendered diff contained a **moved block** — broken since the versioning mode shipped, and invisible because the failure was swallowed (see "Observability" below).

Root cause: `removeAndInsertBlocks` re-resolved target ids via `getNodeId(node, tr.doc)` **while deleting**. A suggested-deletion copy's id is positional (`middle-1` = "the deletion-marked node with 1 same-id node before it"), so deleting the live block shifted the copy's index mid-walk and ids captured from `editor.document` no longer matched (`Blocks with the following IDs could not be found: middle-1`). This fired on every preview re-entry (`clearDocumentForConfigure` → `removeBlocks(editor.document)`) — no user editing required.

## Fixes

- **`removeAndInsertBlocks`** resolves ids against the doc the ids came from (the pre-removal snapshot). Unit test + e2e coverage: the versioning e2e now **re-enters the preview after a follow-up edit for every scenario** — the exact operation that broke (red before the fix on both move scenarios, all 3 browsers).
- **Change tracking** (`getBlocksChangedByTransaction`) excludes suggested-deletion subtrees from its snapshots: they duplicate a live block's id and are positional render artifacts, so diffing them across before/after docs emitted phantom delete+insert pairs for untouched copies.
- **PDF exporter React keys**: styled-text runs were keyed by their text and block fragments by `block.id` — same-text/empty runs and empty ids aliased siblings to one key. Positional keys are correct here (single-pass transform). Found by the new console guard.

## Known limitation, documented instead of fixed

The editor's `blockCache` (keyed by node object) can serve stale/aliased ids for suggested-deletion blocks — only reachable while *editing* suggestion-mode docs, which is experimental and whose fake-id scheme is slated for rework. A NOTE at the cache and two `it.fails` tests pin the contract that rework must satisfy.

## Observability (why this stayed invisible for two months)

The gallery re-rendered the diff synchronously inside the Yjs `update` observer — i.e. inside the *typing* editor's transaction commit — so the throw unwound into that editor's y-prosemirror last-resort catch and became a `console.warn`, misattributed and invisible to tests. Three layers address the class:

- **`@y/prosemirror` 2.0.0-7** ships an `onInternalError` hook (upstreamed via yjs/y-prosemirror#273); the pnpm patch shrinks to: threading the option through `syncPlugin`, the pre-existing type/export fixes, and the ported drop-invalid-nodes hunks (yjs/y-prosemirror#258) that 2.0.0-7 doesn't include.
- **`onYSyncInternalError`** (new, `@blocknote/core/y`): module-scoped observer registry fed by that hook, so harnesses can watch every editor.
- **e2e console guard** (`vitestSetup.browser.ts`): any test now fails on `console.error`, on known swallowed-error warn patterns, or on an observed internal error (with the original stack). Validated by forcing `applyDelta` to throw in a mounted editor.
- The gallery (and the comments thread store, which had the same shape: whole-doc mark updates inside `observeDeep`) now defer their heavy work out of the observer chain via coalesced microtasks — failures surface as real uncaught errors and CRDT forwarding is never interrupted.

## Validation

- Core unit suite fully green (748 passed; 2 `it.fails` documenting the blockCache limitation).
- Full y-prosemirror e2e folder: 128 passed, identical to pre-change baseline; versioning suite green on chromium/webkit/firefox.
- Failure-mode check: with the fix temporarily reverted, the gallery now surfaces the original error loudly (uncaught, correct stack) instead of a silent stale pane.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved collaboration versioning with more reliable block identification and accurate change tracking.
  - Fixed large-diff scenarios that could crash during block insertion or deletion.
  - Prevented rendering artifacts from appearing as duplicate changes.
  - Improved PDF exports by preventing duplicate element-key issues.

- **New Features**
  - Added a way to observe recovered synchronization errors.
  - Deferred and coalesced collaboration updates for smoother rendering and safer cleanup.

- **Tests**
  - Expanded coverage for version previews, follow-up edits, block changes, and synchronization errors.

### DIFF
--- a/examples/07-collaboration/14-suggestion-gallery/src/App.tsx
+++ b/examples/07-collaboration/14-suggestion-gallery/src/App.tsx
@@ -466,11 +466,35 @@ function VersionMerge({
         setup.attrs,
       );
     renderDiff();
-    setup.afterDoc.on("update", renderDiff);
+    // Defer the re-render out of the Yjs observer chain. `afterDoc`'s update
+    // events fire synchronously inside the *typing* editor's Y transaction
+    // commit, so rendering the Diff directly in the handler runs enterPreview
+    // within that editor's sync machinery — a failure there gets caught by
+    // its y-sync last-resort catch and reduced to a console warning (which
+    // once hid a stale-diff bug for two months), and it interrupts the
+    // remaining observers mid-forward. A microtask lets the CRDT forwarding
+    // complete untouched, coalesces update bursts into one render, and makes
+    // a render failure surface as a real uncaught error.
+    let renderQueued = false;
+    let disposed = false;
+    const scheduleRenderDiff = () => {
+      if (renderQueued) {
+        return;
+      }
+      renderQueued = true;
+      queueMicrotask(() => {
+        renderQueued = false;
+        if (!disposed) {
+          renderDiff();
+        }
+      });
+    };
+    setup.afterDoc.on("update", scheduleRenderDiff);
 
     return () => {
+      disposed = true;
       offs.forEach((off) => off());
-      setup.afterDoc.off("update", renderDiff);
+      setup.afterDoc.off("update", scheduleRenderDiff);
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
--- a/packages/core/src/api/nodeConversions/nodeToBlock.ts
+++ b/packages/core/src/api/nodeConversions/nodeToBlock.ts
@@ -22,6 +22,7 @@ import {
   getBlockInfoWithManualOffset,
   getNodeId,
 } from "../getBlockInfoFromPos.js";
+
 import {
   getBlockCache,
   getBlockSchema,
@@ -405,6 +406,12 @@ export function nodeToBlock<
     throw Error("Node should be a bnBlock, but is instead: " + node.type.name);
   }
 
+  // NOTE: the cache (keyed by node object) is not safe for suggested-deletion
+  // blocks: their ids are positional (see `getNodeId`), so an unchanged node
+  // object can change id when the doc around it changes — the cache would then
+  // serve a stale (or aliased) id. Accepted for now: suggestion rendering is
+  // experimental and its fake-id scheme is slated for rework. See the
+  // `it.fails` tests in nodeToBlock.test.ts.
   const cachedBlock = blockCache?.get(node);
 
   if (cachedBlock) {
--- a/packages/core/src/y/comments/YjsThreadStoreBase.ts
+++ b/packages/core/src/y/comments/YjsThreadStoreBase.ts
@@ -37,13 +37,35 @@ export abstract class YjsThreadStoreBase extends ThreadStore {
   }
 
   public subscribe(cb: (threads: Map<string, ThreadData>) => void) {
+    // Deferred out of the Yjs observer chain: `observeDeep` fires inside the
+    // transaction that changed the threads (a local comment edit, or a remote
+    // update mid-apply on the provider's chain). Subscribers do real work —
+    // the comments extension walks the whole doc and dispatches mark updates —
+    // and running that synchronously inside the commit means a subscriber
+    // failure unwinds into the sync machinery and gets misattributed there
+    // (see the suggestion gallery's deferred `renderDiff` for the same
+    // pattern). The microtask also coalesces observer bursts into a single
+    // callback and moves the `getThreads()` materialization out of the
+    // committing transaction.
+    let queued = false;
+    let unsubscribed = false;
     const observer = () => {
-      cb(this.getThreads());
+      if (queued) {
+        return;
+      }
+      queued = true;
+      queueMicrotask(() => {
+        queued = false;
+        if (!unsubscribed) {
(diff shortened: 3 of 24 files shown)


packages/core/src/y/comments/YjsThreadStoreBase.ts
```diff
@@ -37,13 +37,35 @@ export abstract class YjsThreadStoreBase extends ThreadStore {
...
+    // Deferred out of the Yjs observer chain: `observeDeep` fires inside the
+    // transaction that changed the threads (a local comment edit, or a remote
+    // update mid-apply on the provider's chain). Subscribers do real work —
+    // the comments extension walks the whole doc and dispatches mark updates —
+    // and running that synchronously inside the commit means a subscriber
+    // failure unwinds into the sync machinery and gets misattributed there
+    // (see the suggestion gallery's deferred `renderDiff` for the same
+    // pattern). The microtask also coalesces observer bursts into a single
+    // callback and moves the `getThreads()` materialization out of the
+    // committing transaction.
```
nperez0111, line 49: was this observed or speculative?

examples/07-collaboration/14-suggestion-gallery/src/App.tsx
```diff
@@ -466,11 +466,35 @@ function VersionMerge({
...
+      }
+      renderQueued = true;
+      queueMicrotask(() => {
+        renderQueued = false;
+        if (!disposed) {
+          renderDiff();
+        }
+      });
+    };
+    setup.afterDoc.on("update", scheduleRenderDiff);
```
nperez0111, line 492: If we really do need this (which I do question design-wise if we should), then we should at least extract this to a separate utility that is a thunk, taking in the callback to execute, and returns a function which will defer the execution of that function until the next microtask.

YousefED: I do think it's an issue that the error of 1 editor breaks the other (because the error is triggered in the listener).

I'd say we either need:
- a `try / catch` around the handler, and log + rethrow errors there manually (and / or call `reportError`?)
-  the current solution (a microtask to decouple them)
- have this handled at y-prosemirror level

Without any of these, we don't notice the error, but just get broken behavior (a stale diff editor that's not updated anymore). 

fyi, The way to reproduce this issue is shown in the video at `Move paragraph up` in [this](https://docs.numerique.gouv.fr/docs/12715032-d314-4d0a-9002-a92a343b712e/) doc

Preferred solution?

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -405,6 +406,12 @@ export function nodeToBlock<
     throw Error("Node should be a bnBlock, but is instead: " + node.type.name);
   }
 
+  // NOTE: the cache (keyed by node object) is not safe for suggested-deletion
+  // blocks: their ids are positional (see `getNodeId`), so an unchanged node
+  // object can change id when the doc around it changes — the cache would then
+  // serve a stale (or aliased) id. Accepted for now: suggestion rendering is
+  // experimental and its fake-id scheme is slated for rework. See the
+  // `it.fails` tests in nodeToBlock.test.ts.
```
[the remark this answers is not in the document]
YousefED: we explicitly don't yet want to fix this, because the code path shouldn't be used for static diffs when implementing the next phase (live track changes / suggestions), we should revisit this

packages/core/src/y/comments/YjsThreadStoreBase.ts
YousefED: It was observed in the gallery (see the other comment thread).

Question is whether we want to guard this handler as well (and if so, how).

Errors in `this.getThreads()` would currently not surface, (or just as a console.warn)

examples/07-collaboration/14-suggestion-gallery/src/App.tsx
nperez0111: I see no other way here than for y-prosemirror to change. If you catch errors, you should inspect the error you got and handle them if they are the one you expected. Catching all errors and swallowing them is bad practice

tests/vitestSetup.browser.ts
```diff
@@ -32,3 +33,105 @@ beforeEach(() => {
...
+//     our pnpm patch threads it through syncPlugin). The observer receives
+//     the original Error with its stack; y-prosemirror's own warning still
+//     fires too, so a swallowed sync error shows up as both entries.
+// Uncaught exceptions (window "error" events) are already failed by vitest
+// itself; this guard only covers failures something caught and logged.
+//
+// Allowlist deliberate console output per-pattern below, with a comment
+// saying which test produces it and why it is expected.
+
+const CONSOLE_ERROR_ALLOWLIST: RegExp[] = [
```
nperez0111, line 60: I think this whole setup is overly heavy-handed.

PR #2990, opened 2026-08-20
Table-safe diff rendering + id-aware block pairing

YousefED:
Stacked on #2989.

## Fix 1: tables in diff mode

- **`fixTables` is now blocked while the editor is read-only** (a `filterTransaction` on `fixTablesKey` meta in `TableExtension`, mirroring AIExtension's ai-writing gate). A rendered diff shows inconsistent tables *on purpose* (deleted/inserted row and column copies); letting prosemirror-tables "repair" them:
  - rewrote the displayed diff,
  - **wrote the normalization back into the suggestion Y-doc** through the read-only merged viewer's binding (the phantom empty cells removed from the `tables.concurrent` snapshots were exactly these artifacts), and
  - could crash the merge outright — the gallery's "Delete row vs add column" scenario (`lib0 'Unexpected case'`) **no longer crashes**, verified in-browser; its `knownCrash` flag is removed.
- The table-dimension pairing rule in `blockMatchNodes` (replace-when-both-dims-changed) is removed along with its helper; with the gate in place it guarded nothing, and removing it produced zero behavior changes in the suite.

## Fix 2: pair blocks by id in `blockMatchNodes`

Containers with **different block ids never pair in place** (enforced only when both sides carry ids, so id-less external content still pairs by shape; delta attrs are op-wrapped — the values are compared). Pairing different blocks diffed one into the other, which rendered e.g. a full-document replacement as edits *inside* the first deleted block. Now the replacement renders as its own inserted block below the deleted document — resolving the documented issue on the "Delete a whole document" scenario (note removed).

The `DiffVersioningExtension` test fixture gives its before/after docs a stable id: they model one block at two points in time, the only shape real versioning inputs have (snapshots preserve ids).

## Gallery notes

The six concurrent table scenarios now share one low-severity note: concurrent structural table edits aren't conflict-free — a pre-existing collaboration limitation (not diff-specific); a dedicated table CRDT in Yjs + BlockNote is the follow-up. The resolved "Merge cells" phantom-column and "Delete a whole document" placement notes are removed.

## Validation

- y-prosemirror e2e, all three browsers: 387 passed / 0 failed; snapshot churn reviewed hunk-by-hunk (all removed content = fixTables write-back artifacts, plus the intended remove-all shape change).
- Core y + api unit suites: 472 passed.
- In-browser: delete-whole-document diff placement, and the formerly-crashing "Delete row vs add column" suggestion merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved read-only table behavior by preventing unintended automatic table corrections.
  - Improved collaborative editing comparisons so blocks with different identities are handled correctly.
  - Improved version comparisons to recognize edits to the same block more accurately.
- **Documentation**
  - Added consistent guidance for known low-severity limitations when merging concurrent table changes.
  - Clarified feedback for scenarios involving simultaneous table edits, row additions, and column changes.

### DIFF
--- a/examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
+++ b/examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
@@ -90,6 +90,20 @@ export const IMG_SRC_BASE =
 export const IMG_SRC_NEW =
   "data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='100' height='100'><rect width='100' height='100' fill='%234ecdc4'/></svg>";
 
+// Shared note for the concurrent table scenarios: structural table merges are
+// a known pre-existing limitation of the collaboration layer, independent of
+// the suggestion/diff rendering.
+const CONCURRENT_TABLE_FEEDBACK: Feedback[] = [
+  {
+    severity: "low",
+    note:
+      "Concurrent structural edits to the same table don't merge " +
+      "conflict-free — simultaneous row/column changes can combine into an " +
+      "inconsistent table. A follow-up is a dedicated table CRDT in Yjs + " +
+      "BlockNote. (Pre-existing collaboration issue, not specific to diffs.)",
+  },
+];
+
 // Shared 2×2 table baseline used by most of the table scenarios.
 const TABLE_2X2 = {
   id: "table",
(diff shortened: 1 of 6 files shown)


nperez0111 (approved):
This looks great

Issue #2853, opened 2026-06-15
Improve table and videos accessibility in BlockNote


YousefED:
### Discussed in https://github.com/TypeCellOS/BlockNote/discussions/2766

<div type='discussions-op-text'>

<sup>Originally posted by **Ovgodd** May 21, 2026</sup>
# allow authors to define an accessible table title

## Feature Request

This issue was identified during the Docs accessibility audit on a page containing table-related content.

Audited page:
https://docs.numerique.gouv.fr/docs/0d998bef-1b69-47ce-bd18-83d4d849c734/

BlockNote tables do not currently provide a way for authors to define an accessible table title.

A table title helps users understand the purpose of the table before navigating its content, especially screen reader users.

## What did you expect to happen?

Authors should be able to define an accessible table title in the table editing UI.

This title should be exposed in the generated markup, preferably using `<caption>`.

Example:

```html
<table>
  <caption>Project members</caption>
  <thead>
    <tr>
      <th>Name</th>
      <th>Role</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Alice</td>
      <td>Designer</td>
    </tr>
  </tbody>
</table>
```

Alternative implementations could use aria-label or aria-labelledby if <caption> is not suitable.

Without an accessible table title, users may need to navigate the table content before understanding what the table is about.

A dedicated table title improves comprehension for screen reader users and helps authors create more accessible structured content.

## BlockNote version

0.51.4</div>

PR #2992, opened 2026-08-20
feat: Typst-powered accessible PDF export (PDF/UA-1)

### LINKED ISSUE #1979: Tables columns size are not respected in export
**Describe the bug**
Tables columns size are not respected in export
Ex : https://docs.numerique.gouv.fr/docs/b222be3e-654e-48f5-90d2-4fb0585a0236/
See result
[questions-frequentes-offreurs_140gojx.pdf](https://github.com/user-attachments/files/22120951/questions-frequentes-offreurs_140gojx.pdf)

**To Reproduce**
Create a doc with a table with 3 columns
Adjust width of columns 2 small, 1 large
See that widths are not respected

**Misc**

- Node version:
- Package manager:
- Browser:

### LINKED ISSUE #1980: PDF export: large tables content overflow
**Describe the bug**
When exporting a large table to PDF, the table does not split across multiple pages. Instead, the text from oversized cells overflows into adjacent cells, making the content unreadable.

**To Reproduce**
1. Create a table large enough that it cannot fit on a single PDF page.
2. Export the document to PDF.
3. Open the exported file and observe that the content overlaps and is not displayed properly.
Reproduced [here](https://blocknote-git-feature-file-exporters-typecell.vercel.app/interoperability/converting-blocks-to-pdf)
Image

**Expected behavior**
Large tables should automatically split across multiple PDF pages, with the content preserved in a readable format.

**Misc**
- Node version: v20.19.0
- Package manager: npm 10.8.2
- Browser: Version 138.0.7204.93

### LINKED ISSUE #2806: Export PDFs as tagged PDFs
### What’s broken?

After a Docs accessibility audit, we identified an issue with PDF exports.

The PDF exported from Docs does not contain structure tags.

Bookmarks, the document title and the language declaration are present and correct, but the PDF is not tagged.

Without a tag structure, assistive technologies cannot properly understand the document structure. A screen reader may render the content as an unstructured text flow, or may not be able to read it correctly at all.

This is blocking for blind users who need to read exported PDF documents.

A possible solution would be to use or configure a PDF export engine that supports tagged PDF output.

It may be interesting to look at this resource:

```text
https://typst.app/docs/reference/pdf/#pdf-ua
```

However, the final solution should ensure that the exported PDF is usable with assistive technologies and not only visually correct.

The exported PDF should be a tagged PDF.

The tag tree should reflect the document structure, including at least:

```text
<Document>
<H1>…<H6>
<P>
<L> / <LI>
<Link>
<Figure> with Alt attribute
```

The exported structure should preserve the semantics already present in the document: headings, paragraphs, lists, links, images with alternative text, etc.

- the PDF contains a valid tag tree;
- headings are exposed as headings;
- paragraphs are exposed as paragraphs;
- lists are exposed as lists;
- links are exposed as links;
- figures are exposed with their alternative text when needed.

### Steps to reproduce

1. Export a document from Docs to PDF format.
2. Open the PDF in Adobe Acrobat.
3. Check the Tags panel, or verify the PDF with PAC or a similar accessibility tool.
4. Observe that the document has no tag structure.

### BlockNote version

0.51.1

### LINKED ISSUE #2866: PDF export : image ratio not kept
### What’s broken?

Hi, 

original issue [here](https://github.com/suitenumerique/docs/issues/2436).

It seems that xl-pdf-exporter is trying to fit an image on the end of the page by modifying its aspect ratio instead of just inserting a page break, making the image on pdf export weird sometimes. 

Doc example (https://www.blocknotejs.org/demo#jbm5s)

Image

PDF export : 

Image

I'll try to see if I can solve it but my skills in PDF and JS are rather limited. 

Image should be pushed to the next page instead of tinkering its aspect ratio

### Steps to reproduce

1. Create a new document
2. Make it so the image is at the end of a page
3. See its aspect ratio changing

### BlockNote version

_No response_

### Environment

_No response_

### Additional context

_No response_

### Contribution

### Sponsor

YousefED:
> [!NOTE]
> Layer 1 of **stack #3023** (reviewable and mergeable per layer): this PR builds the Typst pipeline on typst.ts; #3020 replaces typst.ts with our own compiler and reworks the export API (`toPDF`), #3021 fixes editor-rendering parity, #3022 drops the react-pdf math/diagram mappings. Statements below describe this layer; where a follow-up landed later in the stack, it is marked.

## What

Replaces the react-pdf based PDF export with a [Typst](https://typst.app)-powered pipeline that produces **accessible, tagged PDF/UA-1** documents, entirely client-side (Typst compiled to wasm). Closes the gap described in #2806: react-pdf emits no logical structure (`/StructTreeRoot`), so exported PDFs were invisible to screen readers; Typst 0.15 emits a full tag tree (`Document › Hn › P › L›LI › Table›TH/TD › Figure+Alt › Link › Code`) and the output validates with veraPDF (`--flavour ua1`, 0 failed checks).

## Architecture

- **`@blocknote/xl-typst-exporter`** (new): the pure Typst source exporter — `TypstExporter`, `typstDefaultSchemaMappings`, custom-mapping helpers. No compile/PDF dependencies; usable standalone (e.g. compile server-side with the Typst CLI or node compiler).
- **`@blocknote/xl-pdf-exporter`** (root, breaking): the new Typst-based `PDFExporter` (`toBlob`/`toBytes`, plus inherited `toTypst`), composed from `compileTypstToTaggedPdf` (wasm) + `declarePdfUA` (pure-JS PDF/UA declaration via @cantoo/pdf-lib, default-on with an opt-out). Re-exports the Typst layer, so one package covers PDF consumers — and **one custom-block Typst mapping serves both `.typ` and PDF export**.
- **`@blocknote/xl-pdf-exporter/react-pdf`** (deprecated): the previous react-pdf exporter, unchanged API, exposed only via this subpath (react/react-pdf are optional peers now, so its dependency tree stays out of Typst-only installs). To be removed after a deprecation window.
- Math blocks export as native Typst equations (real text, via tex2typst), diagrams as vector SVG with labels as text — both with alt text. CJK via font fallback lists; emoji via a dedicated emoji-font channel.
- **Zero-config output matches the editor**: the package embeds a default font set (Inter, Geist Mono, New Computer Modern Math, Noto Color Emoji) as lazily-loaded chunks — `fonts` and `emojiFont` each default to it independently, and **no fonts come from a CDN** (Typst's stock fonts are opt-in fallbacks via `preloadDefaultFonts: true`; the compiler wasm is the only CDN default, bundleable via `getModule`). The math fonts are load-bearing: the wasm embeds no fonts, so fontless math fails to compile.
- Also included: ODT tables gain colspan/rowspan support (span attributes + ODF's required `covered-table-cell` placeholders); the docs site's `/demo` PDF action runs on the new exporter; the pdf-ua example needs no font wiring anymore (12MB of vendored fonts deleted); all packages' build tasks declare `types/**` outputs (stale-declaration cache replays).

## Breaking changes

- The `@blocknote/xl-pdf-exporter` root now exports the Typst-based `PDFExporter`; existing consumers keep the old behavior by changing one import to `@blocknote/xl-pdf-exporter/react-pdf`. Custom blocks need Typst mappings for the new exporter (math-block/diagram-block ship them from their `typst-exporter` subpaths).
- Breakage is loud (type-incompatible mappings / missing exports), never silent.

## Testing & CI

- Unit: `.typ` golden snapshots; PDF/UA declaration unit tests; full-pipeline conformance test gated on **veraPDF** (digest-pinned container in CI via `tests/scripts/install-pdf-tooling.sh`).
- Visual: per-page PNG snapshots rasterized through a **digest-pinned poppler container** (`--platform linux/amd64`) so baselines are byte-stable across environments. Both PDF gates (veraPDF conformance + visual snapshots) pass on CI.
- Browser (3 engines): wasm compile path, concurrency serialization, math/diagram image e2e; fully offline (fonts + wasm bundled, CDN font preloading genuinely disabled when opted out).
- Docs: new [Typst export] page, rewritten [PDF export] page, updated math/diagram/supported-formats pages; live PDF/UA example with editor-matching bundled fonts.

## Fixes

- Closes #2806 — the whole point: exports are tagged and declare PDF/UA-1 (veraPDF-validated).
- Closes #2866 — Typst sizes images by width only (aspect ratio always preserved), and figures are unbreakable, so an image that doesn't fit moves to the next page whole instead of being squeezed.
- Closes #1980 — tables split across pages natively (verified: a 100-row table compiles to multiple pages, cell content wraps).
- Closes #1979 — `columnWidths` map to fixed-width column tracks (verified: `[300, 100, 100]` → `columns: (225.0pt, 75.0pt, 75.0pt)`).

(The last three describe the react-pdf exporter's behavior, which remains unchanged behind the deprecated `/react-pdf` subpath — they're fixed by the new default exporter.)

## Follow-ups

- `alt` field on media blocks (#2853); heading-level normalization for UA's first-heading-must-be-H1 rule.
- ~~Native `--pdf-standard ua-1` once the *web* compiler exposes it — removes `declarePdfUA` and @cantoo/pdf-lib.~~ **Done in #3020** (own compiler; the compile itself validates ua-1). #3020 also renames `toBlob`/`toBytes` to a single `toPDF`.
- Remove the `/react-pdf` subpath at the end of the deprecation window — its never-shipped math/diagram mappings are already dropped in #3022.

## Summary by CodeRabbit

* **New Features**
  * Added Typst-based document export, including accessible tagged PDF/UA-1 output.
  * Added support for exporting math, diagrams, images, tables, merged cells, styling, lists, and multi-column layouts.
  * Added standalone Typst export and a PDF/UA demonstration.
  * Added offline compiler support and bundled fonts to reduce CDN reliance.

* **Bug Fixes**
  * Improved diagram fonts, image handling, line breaks, table merging, list numbering, and file-fetch error reporting.

* **Documentation**
  * Added Typst and PDF/UA export guidance.
  * Marked the existing React-PDF export path as deprecated.


docs/content/docs/features/export/pdf.mdx
```diff
@@ -16,98 +20,239 @@ It's possible to export BlockNote documents to PDF, completely client-side.
   [Business subscription](/pricing).
 </Callout>
 
-First, install the `@blocknote/xl-pdf-exporter` and `@react-pdf/renderer` packages:
+First, install the `@blocknote/xl-pdf-exporter` package:
 
 ```bash
-npm install @blocknote/xl-pdf-exporter @react-pdf/renderer
+npm install @blocknote/xl-pdf-exporter
 ```
```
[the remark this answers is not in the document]
YousefED: Deliberate: the docs site consistently uses `npm install` in end-user installation instructions (the docx/odt/email exporters and every getting-started page all do). The vp/pnpm rule governs commands run inside this repo, not what library consumers are told - changing only these two pages would break docs-wide consistency.

examples/05-interoperability/11-converting-blocks-to-pdf-ua/index.html
```diff
@@ -0,0 +1,14 @@
+<html lang="en">
```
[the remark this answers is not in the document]
YousefED: Pre-existing template convention, out of scope for this PR: all 102 examples' `index.html` files come from the shared template (`packages/dev-scripts/examples/template-react/index.html.template.tsx`) without a doctype. A local edit would be overwritten on regeneration, and fixing the template would touch every example in the repo.

examples/05-interoperability/11-converting-blocks-to-pdf-ua/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@blocknote/example-interoperability-converting-blocks-to-pdf-ua",
+  "description": "AUTO-GENERATED FILE, DO NOT EDIT DIRECTLY",
+  "type": "module",
+  "private": true,
+  "version": "0.12.4",
+  "scripts": {
+    "start": "vite",
+    "dev": "vite",
+    "build:prod": "tsc && vite build",
```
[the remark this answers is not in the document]
YousefED: Rejected: `tsc && vite build` is the standard scaffold in all 102 generated examples and targets end users who download the example as a standalone Vite project - the contributor-facing `vp run lint` guideline doesn't apply there. Pre-existing template convention, out of scope for this PR.

examples/05-interoperability/11-converting-blocks-to-pdf-ua/vite.config.ts
```diff
@@ -0,0 +1,35 @@
...
+            // Comment out the lines below to load a built version of blocknote
+            // or, keep as is to load live from sources with live reload working
+            "@blocknote/core": path.resolve(
+              __dirname,
+              "../../packages/core/src/",
+            ),
+            "@blocknote/react": path.resolve(
+              __dirname,
+              "../../packages/react/src/",
+            ),
```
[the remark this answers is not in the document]
YousefED: Known pre-existing issue, out of scope: the template has carried `// TODO: the ../../ paths are broken` since before this PR's merge-base, and all 102 examples share the same paths. Fixing it would regenerate 102 files and change dev-alias behavior repo-wide - deliberately deferred.

packages/dev-scripts/examples/template-react/testDocumentBlocks.ts.template.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const template = () => {
+  const sharedFile = path.resolve(
+    path.dirname(fileURLToPath(import.meta.url)),
+    "../../../../shared/testDocumentBlocks.ts",
+  );
+  const source = fs.readFileSync(sharedFile, "utf-8");
+  return `// AUTO-GENERATED FILE, DO NOT EDIT DIRECTLY.
+// Generated from shared/testDocumentBlocks.ts — run \`npm run gen\` to update.
+${source}`;
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9: the template header now says `pnpm run gen`, and the five generated `testDocumentBlocks.ts` fixtures were regenerated to match.

packages/diagram-block/src/block/helpers/render/DiagramBlockPreviewWithPopup.tsx
```diff
@@ -62,15 +73,18 @@ export const useMermaidSVG = (source: string) => {
     return () => {
       stale = true;
     };
-  }, [source]);
+  }, [source, fontFamilyElement]);
```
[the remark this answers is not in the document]
YousefED: Deliberate tradeoff, documented in the code ("Read the font at render time, not at mount..."): the computed font is re-read on every effect run, so a theme change without a source change causes only transient cosmetic staleness that self-corrects on the next edit. A full fix needs computed-style observation (MutationObserver / theme-signal plumbing), which is disproportionate here.

packages/math-block/src/pdf-exporter/index.tsx
```diff
@@ -171,4 +171,10 @@ export function createInlineMathMapping(options?: {
  * options - see {@link createInlineMathMapping}. Browser-only; when
  * exporting elsewhere, use the factory to pass a `rasterize` function.
  */
+/**
+ * @deprecated Mappings for the react-pdf based exporter
+ * (`@blocknote/xl-pdf-exporter/react-pdf`), which is deprecated - use the
+ * Typst mappings from `@blocknote/math-block/typst-exporter` with the
+ * Typst-based `PDFExporter` instead.
+ */
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9: `mathBlockMapping` now carries the same `@deprecated` JSDoc as `inlineMathMapping` and diagram-block's `diagramBlockMapping`, pointing to the Typst mapping from `@blocknote/math-block/typst-exporter`.

packages/math-block/src/typst-exporter/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  const result = latexToTypstMath(source);
+  if (result.error !== undefined) {
+    return `#align(center)[#${errorText(exporter, source)}]`;
+  }
+
+  // A block equation is a display equation, centered by default (matching the
+  // editor and the other exporters).
+  return `#math.equation(block: true, alt: ${strLit(source)}, $ ${
+    result.math
+  } $)`;
```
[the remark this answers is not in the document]
YousefED: Deliberate, documented in the JSDoc: the equation carries the LaTeX source as its alt text - a recognized textual representation - because PDF/UA requires a non-empty one. Generating localized natural-language descriptions would require a speech-rule-engine dependency plus i18n, disproportionate for this exporter.

packages/xl-pdf-exporter/package.json
```diff
@@ -41,24 +39,26 @@
...
     }
   },
   "scripts": {
     "dev": "vp dev",
     "lint": "vp lint src",
     "test": "vp test --run",
     "test-watch": "vp test watch",
-    "clean": "rimraf dist && rimraf types",
-    "email": "email dev"
+    "clean": "rimraf dist && rimraf types"
```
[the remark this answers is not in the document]
YousefED: Rejected: `"lint": "vp lint src"` is the standard per-package script in 17 other packages - it is what the root `vp run lint` fans out to. The guideline applies to commands contributors run, not to package script internals.

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
```diff
@@ -0,0 +1,188 @@
...
+    const newFonts = fontList(options).filter(
+      (f) => !configured!.fonts.has(fontFingerprint(f)),
+    );
+    // getModule is compared by its *resolved* source (the URL/module it
+    // returns), not by function identity - callers typically pass a fresh
+    // closure per call around the same URL. Omitting it after a first call
+    // that had one means "reuse what's loaded" and is fine.
+    const moduleChanged =
+      options.getModule !== undefined &&
+      options.getModule() !== configured.moduleSource;
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9: `getModule` results are now canonicalized before comparison (`URL` -> `href`; strings already compare by value), so the documented `() => new URL(...)` pattern no longer rejects the second compile. The browser suite's shared `getModule` now returns a fresh `URL` per call, making every sequential compile in it a regression test.

packages/xl-pdf-exporter/src/pdfua/pdfua.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  });
+
+  it("matches the per-page visual snapshot", async () => {
+    requireTool(
+      "docker",
+      ["info"],
+      "Start Docker - the visual snapshot rasterizes through a digest-pinned poppler container so the PNG baselines are byte-stable across environments.",
+    );
+    const dir = mkdtempSync(join(tmpdir(), "bn-visual-"));
+    const pages = rasterizePdf(tagged, dir);
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9: the visual snapshot now rasterizes `Buffer.from(ua)` - the final `declarePdfUA` output. The post-process is metadata-only so the PNG baselines are unchanged, but a post-process regression that altered page content would now fail the snapshot.

packages/xl-pdf-exporter/src/pdfua/postProcess.ts
```diff
@@ -0,0 +1,155 @@
...
+ * long-term replacement is the compiler's own `--pdf-standard ua-1` flag
+ * once the wasm binding exposes it (see the TODO in index.ts).
+ */
+function withPdfUaIdentifier(xmp: string): string {
+  // Already identified - matched as the actual element/attribute form, not
+  // as a bare substring (document text merely *mentioning* "pdfuaid:part"
+  // must not skip the declaration).
+  if (/<pdfuaid:part[\s>]|pdfuaid:part\s*=/.test(xmp)) {
+    return xmp;
+  }
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9: an existing identifier is accepted only when it declares part 1 (element or attribute form); any other existing `pdfuaid:part` now throws - per the module's contract (never return a wrong declaration as success), and rather than regex-rewriting foreign values. Regression tests added for part-2 and empty elements.

packages/xl-typst-exporter/src/defaultSchema/blocks.ts
```diff
@@ -0,0 +1,246 @@
...
+ */
+async function imageFigure(
+  props: MediaProps,
+  exporter: TypstExporter<any, any, any>,
+): Promise<string> {
+  const { url, caption } = props;
+  if (!url) {
+    return imagePlaceholderFigure(props);
+  }
+  const alt = caption || props.name || url;
```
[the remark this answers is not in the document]
YousefED: Deliberate: the fallback chain uses the most informative text available (caption > name > url) - a URL distinguishes images and usually ends in a descriptive filename, unlike a fixed "Image" label, and PDF/UA only requires non-empty alt. `imagePlaceholderFigure` uses "Image" only because that path has no URL at all.

packages/xl-typst-exporter/src/defaultSchema/blocks.ts
```diff
@@ -0,0 +1,246 @@
...
+    const lines = [
+      ...(headerCells.length
+        ? [`  table.header(${headerCells.join(", ")}),`]
+        : []),
+      ...rows
+        .slice(headerRows)
+        .map(
+          (cells) => `  ${cells.map((c) => renderCell(c, false)).join(", ")},`,
+        ),
+    ];
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9: body cells in header columns now get the header treatment (bold, colspan-aware), matching the editor and the docx/react-pdf/email exporters. Typst's `table.header` only models header rows, so there is no per-cell TH tag to add (noted in a code comment). Regression test added.

packages/xl-typst-exporter/src/util.ts
```diff
@@ -0,0 +1,84 @@
...
+  kind: "text" | "background",
+): string | undefined {
+  if (!name || name === "default") {
+    return undefined;
+  }
+  const colors = exporter.options.colors as Record<
+    string,
+    { text: string; background: string } | undefined
+  >;
+  return colors[name]?.[kind];
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9, at the boundary instead of optional chaining (which would silently drop all colors): the `TypstExporter` constructor now defaults per key (`options?.colors ?? COLORS_DEFAULT`, etc.), so an explicitly-undefined option keeps the default and the color lookups stay non-optional. Test added.

playground/src/examples.gen.tsx
```diff
@@ -1137,13 +1138,13 @@ export const examples = {
...
         },
-        title: "Exporting documents to PDF",
+        title: "Exporting documents to PDF (react-pdf, deprecated)",
         group: {
           pathFromRoot: "examples/05-interoperability",
           slug: "interoperability",
         },
         readme:
-          'This example exports the current document (all blocks) as an PDF file and downloads it to your computer.\n\n**Try it out:** Edit the document and click "Download .pdf" at the top to download the PDF file.',
+          '> **Deprecated:** this example uses the react-pdf based exporter\n> (`@blocknote/xl-pdf-exporter/react-pdf`), which is deprecated and will be\n> removed after a few releases. Use the Typst-based `PDFExporter` instead -\n> see the "Exporting documents to tagged PDF (PDF/UA)" example - which\n> produces accessible, tagged PDF/UA-1 output.\n\nThis example exports the current document (all blocks) as an PDF file and downloads it to your computer.\n\n**Try it out:** Edit the document and click "Download .pdf" at the top to download the PDF file.',
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9 - in the source `examples/05-interoperability/05-converting-blocks-to-pdf/README.md` and the regenerated `examples.gen.tsx`.

shared/util/binaryFileSnapshotUtil.ts
```diff
@@ -63,6 +71,15 @@ export async function toMatchBinaryFileSnapshot(
...
   if (option === "none" || (option === "new" && fileBuffer !== undefined)) {
+    if (options.writeActualOnMismatch) {
+      // `name.ext` -> `name.actual.ext`; an extension-less path gets a plain
+      // `.actual` suffix (the no-match case must never fall through to the
+      // baseline path itself).
+      const actualPath = /\.[^./]+$/.test(filepath)
+        ? filepath.replace(/(\.[^./]+)$/, ".actual$1")
+        : `${filepath}.actual`;
+      fs.writeFileSync(actualPath, buffer);
+    }
```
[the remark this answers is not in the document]
YousefED: Fixed in d23319fc9: the `.actual` diagnostic write now creates `path.dirname(actualPath)` first, so a missing baseline directory can no longer mask the snapshot mismatch error with ENOENT.

.github/workflows/update-e2e-screenshots.yml
```diff
@@ -0,0 +1,62 @@
...
+          node-version-file: ".node-version"
+          cache: true
+
+      - name: Install dependencies
+        run: vp install
+
+      # `-u` must come after the filter, and each browser runs separately so a
+      # failure in one doesn't stop the others' baselines from updating.
+      - name: Update screenshots (chromium)
+        run: HOME=/root vp test -c vite.config.browser.ts --run --browser chromium ${{ inputs.filter }} -u
```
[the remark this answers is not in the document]
YousefED: Fixed in 21db06234: the `inputs.filter` value is now passed via `env` (`$FILTER`) instead of template expansion into the `run:` script, so the input can no longer inject into the generated shell code.

.github/workflows/update-e2e-screenshots.yml
```diff
@@ -0,0 +1,62 @@
...
+        working-directory: tests
+        continue-on-error: true
+      - name: Update screenshots (firefox)
+        run: HOME=/root vp test -c vite.config.browser.ts --run --browser firefox ${{ inputs.filter }} -u
```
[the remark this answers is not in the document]
YousefED: Fixed in 21db06234: the `inputs.filter` value is now passed via `env` (`$FILTER`) instead of template expansion into the `run:` script, so the input can no longer inject into the generated shell code.

.github/workflows/update-e2e-screenshots.yml
```diff
@@ -0,0 +1,62 @@
...
+        working-directory: tests
+        continue-on-error: true
+      - name: Update screenshots (webkit)
+        run: HOME=/root vp test -c vite.config.browser.ts --run --browser webkit ${{ inputs.filter }} -u
```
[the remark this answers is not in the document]
YousefED: Fixed in 21db06234: the `inputs.filter` value is now passed via `env` (`$FILTER`) instead of template expansion into the `run:` script, so the input can no longer inject into the generated shell code.

YousefED:
@nperez0111 would welcome a first look at this (I'll do another pass + review the docs), but feedback welcome at this point

.github/workflows/update-e2e-screenshots.yml
```diff
@@ -0,0 +1,68 @@
+# Regenerates the Linux screenshot baselines for the browser (e2e) suite in
+# the same Playwright container the e2e jobs use, and uploads them as an
+# artifact to download and commit. Linux baselines can otherwise only be
+# produced locally through docker; this is the push-button alternative.
+name: Update e2e screenshots
```
nperez0111: I don't get why we'd need this, you can re-gen with docker, why is that different?

docs/content/docs/features/blocks/math.mdx
```diff
@@ -121,34 +121,29 @@ import {
...
-```
+The LaTeX is converted to Typst's math notation. Most KaTeX commands are
+covered; a valid formula using an uncovered command fails the export loudly
+(rather than silently exporting something else), while invalid LaTeX renders
+the error placeholder as in other formats.
 
-Math blocks require the `@react-pdf/math` package (a peer dependency of the PDF mapping).
+The deprecated [react-pdf exporter](/docs/features/export/pdf#deprecated-the-react-pdf-exporter)'s
+mappings remain available from `@blocknote/math-block/pdf-exporter` during its
+deprecation window.
```
nperez0111, line 146: I don't see reason in preserving this for math which is new

docs/content/docs/features/export/pdf.mdx
```diff
@@ -16,98 +20,252 @@ It's possible to export BlockNote documents to PDF, completely client-side.
...
+keeps the default emoji font (and vice versa), and an explicit empty array
+disables one entirely:
+
+```typescript
+const blob = await exporter.toBlob(editor.document, {
+  // Font bytes (Uint8Array) to load into the compiler.
+  fonts: [myBodyFont, myMonoFont],
+  // An emoji-capable font. Browsers give the compiler no access to OS
+  // fonts, so without one emoji render as missing glyphs (and fail PDF/UA).
+  emojiFont: myEmojiFont,
```
nperez0111, line 186: May be worth looking into the [OffscreenCanvas API](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) I think you can put an emoji there and screenshot it to bytes using system fonts

docs/AGENTS.md
nperez0111, line 1: not useful

docs/CLAUDE.md
nperez0111, line 1: not useful

packages/dev-scripts/vite.config.ts
```diff
@@ -18,6 +18,9 @@ export default defineConfig({
           { pattern: "examples/*/*/README.md", base: "workspace" },
           { pattern: "examples/*/*/src/**", base: "workspace" },
           { pattern: "packages/dev-scripts/examples/**", base: "workspace" },
+          // The testDocumentBlocks template copies this file verbatim into
+          // the examples - edits to it must re-run generation.
+          { pattern: "shared/testDocumentBlocks.ts", base: "workspace" },
```
nperez0111, line 23: I'd prefer that the gen script owns this rather than having this be the source of truth, it is awkward like this

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
```diff
@@ -0,0 +1,200 @@
...
+   *
+   * If omitted, Typst.ts loads its matching compiler wasm from its CDN — handy
+   * for demos, but bundle/self-host it for production.
+   */
+  getModule?: InitOptions["getModule"];
+  /**
+   * Extra fonts (as bytes) to load into the compiler — typically the body fonts
+   * the exporter references (e.g. Inter, Geist Mono).
+   */
+  fonts?: Uint8Array[];
```
nperez0111, line 19: how do you know which fonts? It's just an array

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * The emoji source: an emoji-capable font (or fonts) as bytes. Unlike
+   * react-pdf's image-based `emojiSource`, Typst renders emoji from a font, so
+   * this is how you supply one. Required for *fully* PDF/UA-1 conformant output
```
nperez0111, line 23: ah that's why you can't use OffscreenCanvas then

nperez0111: We can't fallback to something like system-ui or something which should have emoji?

packages/xl-pdf-exporter/src/pdfua/defaultFonts.ts
```diff
@@ -0,0 +1,67 @@
+import { loadFileBuffer } from "@shared/util/fileUtil.js";
+
+/**
+ * The family name of the bundled emoji font; {@link PDFExporter} declares it
+ * as the default `emojiFontFamily` so multi-codepoint emoji shape correctly
+ * with zero configuration.
+ */
+export const DEFAULT_EMOJI_FONT_FAMILY = "Noto Color Emoji";
```
nperez0111, line 8: default per os? Since I think this is android right?

packages/xl-pdf-exporter/src/index.ts
```diff
@@ -1 +1,179 @@
...
+    });
+    // TODO: collapse this into one step once the web compiler supports native
+    // PDF/UA-1 export. typst's node compiler already accepts
+    // `pdfStandard: "ua-1"`, which emits a veraPDF-conformant PDF/UA-1
+    // directly (verified: 0 failed checks) and validates accessibility at
+    // compile time (errors on missing alt text) — making `declarePdfUA` (and
+    // the `@cantoo/pdf-lib` dependency) unnecessary. `@myriaddreamin/
+    // typst.ts`'s browser `pdf()` does not yet expose a PDF standard option,
+    // so the post-process is still required for the wasm path.
+    return withDeclaration ? declarePdfUA(taggedPdf) : taggedPdf;
```
nperez0111, line 163: maybe we should upstream?

tests/scripts/install-pdf-tooling.sh
nperez0111, line 1: I don't love this. I'm all for tests and all. But this is a bit too far, having a script in the repo that validates this once in a while is fine but I don't think this should be in CI. It adds weight and only increases time for something unlikely to change or break

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
nperez0111, line 1: It's pretty hard to follow how the compiler wasm is loaded, but I wonder on the different ways of packaging this

docs/AGENTS.md
YousefED: Removed in 92fe8a476 (and gitignored — it's a Next.js postinstall artifact, so it regenerates locally but can't ride into commits again).

docs/CLAUDE.md
YousefED: Removed in 92fe8a476, same as AGENTS.md — gitignored alongside it.

.github/workflows/update-e2e-screenshots.yml
YousefED: Agreed — removed in 92fe8a476. It existed for CI-side regeneration while local docker was unavailable; with the docker-only e2e policy, `vp run e2e:updateSnaps` covers it. (build.yml's include-hidden-files failure-artifact fix is separate and stays.)

docs/content/docs/features/export/pdf.mdx
YousefED: We did sth like this earlier for the react pdf exporter if I'm not mistaken.

Downside; rasterized emojis are weaker for Accessibility and copy-paste. So I'd opt to keep as-is

packages/dev-scripts/vite.config.ts
YousefED: Agree, but I'm not sure how we can do this? The vite.config here must know what dependencies are a cache input?

tests/scripts/install-pdf-tooling.sh
YousefED: Agree it's overkill to run on every build, but on the other hand I would like it tested somehow. Happy to change this, but not sure about the right solution;
- only on main / on release?
- only on the periodic job (iirc we have a once in 24 hours job)
- ... other suggestions?

packages/xl-pdf-exporter/src/pdfua/defaultFonts.ts
YousefED: We need to embed the font files themselves, because wasm doesn't have access to system fonts. The apple font is not freely available afaik

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
YousefED: Wasm can't access system fonts, so we always need font source files I think

packages/xl-pdf-exporter/src/index.ts
YousefED: opened a PR upstream, but also opened #3020 to remove the dependency

tests/scripts/install-pdf-tooling.sh
YousefED: Removed in #3020:
- (for veraPDF): most of the conformance is tested by typst anyway
- for pdf2ppm; we already had tests that use pdf-in-browser-rendering (pdf.js) combined with browser screenshot tests, so use that instead

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
YousefED: should be simpler in #3020

docs/content/docs/features/blocks/math.mdx
YousefED: #3022

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
YousefED: The compiler reads each font file's internal name table, so the files carry their own identity - the `fontFamily` options just have to match what the files declare, and a mismatch surfaces as an `unknown font family` entry in the result's `compileWarnings`. The bundled defaults' name-to-file pairing is pinned by the "keeps the bundled default fonts in sync" test, and `TypstCompiler.fontFamilies()` exposes what actually loaded. The docs' Fonts section on the PDF page now walks through this.

packages/xl-pdf-exporter/src/pdfua/compileBrowser.ts
```diff
@@ -0,0 +1,200 @@
...
+      (f) => !configured!.fonts.has(fontFingerprint(f)),
+    );
+    // getModule is compared by its *resolved* source (the URL/module it
+    // returns, canonicalized via `moduleSourceKey`), not by function
+    // identity - callers typically pass a fresh closure per call around the
+    // same URL. Omitting it after a first call that had one means "reuse
+    // what's loaded" and is fine.
+    const moduleChanged =
+      options.getModule !== undefined &&
+      moduleSourceKey(options.getModule()) !== configured.moduleSource;
```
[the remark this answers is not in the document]
YousefED: Superseded in #3020: `compileBrowser.ts` (and `getModule`) are removed - the own compiler takes the wasm as bytes or a URL directly, so the promise-identity comparison no longer exists.

PR #2996, opened 2026-08-21
docs: file upload example (BLO-1311)

matthewlipski:
# Summary

This PR fixes the file upload example and related file upload code as tmpfiles changed their API and it no longer works for our use. To prevent issues like that happening in the future, we now just save the file as a base64 URL.

Closes @2967

## Rationale

The example is broken.

## Changes

- Updated docs demo.
- Updated file upload example.
- Updated Uppy file panel example.
- Updated and renamed `uploadFilesToTmpFilesDotOrg_DEV_ONLY` to `uploadFile_DEV_ONLY`.

## Impact

N/A

## Testing

N/A - TODO?

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **New Features**
  * File uploads in development examples now use local base64 data URLs, enabling image previews without an external temporary file service.
  * Upload guidance now explains how to replace the local approach with a production backend.

* **Bug Fixes**
  * Improved handling of file-reading failures during uploads.

* **Documentation**
  * Updated examples and file-upload documentation to reflect local data URL encoding.


docs/app/demo/_components/utils.ts
```diff
@@ -1,3 +1,5 @@
+import { uploadFile_DEV_ONLY } from "@blocknote/core";
```
nperez0111: I wouldn't really want this exported, I get that it is a shared util, but not one that needs to be actually packaged. Just copy-paste in the places that need it

packages/core/src/blocks/index.ts
```diff
@@ -20,7 +20,7 @@ export * from "./Code/helpers/parse/parsePreCode.js";
 export * from "./Code/helpers/render/createCodeBlock.js";
 export * from "./Code/helpers/toExternalHTML/createPreCode.js";
 export * from "./ToggleWrapper/createToggleWrapper.js";
-export * from "./File/helpers/uploadToTmpFilesDotOrg_DEV_ONLY.js";
+export * from "./File/helpers/uploadFile_DEV_ONLY.js";
```
nperez0111: oh we had this exported before? That is strange to me

nperez0111 (commented):
Can we change it from being exported? Otherwise fine to move off the actual upload path

packages/core/src/blocks/index.ts
matthewlipski: Should I still remove the export then? Or keep as it was?

nperez0111: I'd remove it

PR #2997, opened 2026-08-21
feat(core): container block API for nested blocks

nperez0111:
Part 1 of 3 of the container blocks stack (1: core API ← you are here, 2: multi-column migration, 3: docs & examples).

Replaces #2697, split into reviewable stacked PRs.

## What this adds

A first-class API for **container blocks**: custom blocks that hold other blocks as children, declared with a new `children` config on `BlockConfig`:

```ts
const Callout = createBlockSpec(
  {
    type: "callout",
    content: "inline", // optional: containers can also have their own content
    children: {
      allow: "any",          // or a list of block types / containers
      min: 1,                // structural minimum, maintained by repair
      default: [{ type: "paragraph" }],
      whenEmptied: "unwrap", // or "refill"
      boundary: "open",      // or "isolated" / "sealed"
    },
  },
  { render: ... },
);
```

- **Two container shapes.** A *pure* container holds children directly in its PM node. A *content-bearing* container (e.g. a toggle: own inline title + child blocks) compiles to two generated PM nodes (`<type>__content`, `<type>__children`) behind one block type.
- **Validation & schema invariants** (`validateChildren.ts`, `assertSchemaInvariants.ts`): child configs are checked at schema build time, with reachability checks for `placement: "containerOnly"` blocks.
- **Repair** (`fixContainer.ts`): removals that empty a container below `min` either unwrap it or refill it from `default`, applied by the block manipulation API and the keyboard handlers.
- **Generic keyboard behavior** (`KeyboardShortcutsExtension.ts`): the previous hardcoded columnList Backspace/Delete/Enter handlers are generalized to any container, driven by schema navigation (`containerNav.ts`) and the `boundary` config (`sealed` containers never leak or swallow content implicitly).
- **Serialization & parsing**: internal/external HTML round-trips for both container shapes, `data-children-of` markers so non-content UI text in a render never parses back as document content, `parse`/`parseContent`/`runsBefore` support for containers.
- **Block manipulation API**: `insertBlocks` placements (`"start"`/`"end"`), `updateBlock` conversions into/out of containers, container-aware `moveBlocks`/`nestBlock`/`mergeBlocks`.
- **UI**: side-menu handling for containers (`sideMenuContainerGeometry.ts`, `containerUI.ts`), React node-view support (`ReactBlockSpec`, `useNodeViewBlock`), `BlockPopover` fixes.
- New `@blocknote/core/internal` entry point for the container machinery that integrations (e.g. `xl-multi-column`) need but that isn't public API.

## Legacy multi-column compatibility

`@blocknote/xl-multi-column` is untouched here; its hand-written `column`/`columnList` PM nodes keep working through a handful of small shims, each marked with a `// Legacy` comment:

- `fixColumnList.ts` kept and re-exported from the root
- `fixContainer` falls back to `fixColumnList` for config-less column nodes
- `blockToNode` keeps the plain-`create` path (invalid column structures still throw on insert)
- the internal HTML serializer keeps the old `bnBlock` path
- `UniqueID` still assigns ids to `columnList`/`column`
- `Exporter.isContainerBlock` and `containerUI` still recognize the legacy types
- `fragmentToBlocks` keeps the old single-column flattening rule

The next PR in the stack migrates multi-column onto the container API and deletes every one of these shims.

## Testing

- Full node test suite: 15/15 packages green (incl. `xl-multi-column`'s existing tests, unchanged, against the new core).
- Full Docker browser suite (chromium/firefox/webkit, unit + e2e incl. the multi-column drag/drop e2e tests): green.
- New test coverage: `children.test.ts`, `containers.test.ts` / `containers.browser.test.ts`, `contentContainers.*`, `containerParse.browser.test.ts`, `insertPlacement.test.ts`, `sideMenuContainerGeometry.browser.test.ts`, `ReactBlockSpec.container.browser.test.tsx`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable container blocks with child rules, placement constraints, boundaries, defaults, and child limits.
  * Blocks can be inserted at the start or end of containers, as well as before or after existing blocks.
  * Improved container-aware editing, dragging, selection, side-menu positioning, and keyboard interactions.
  * Added container rendering, HTML round-tripping, export support, and React content references.
  * Added an internal package entry point for container integrations.

* **Bug Fixes**
  * Improved container repair, merging, conversion, and empty-container handling.
  * Prevented invalid schema configurations, unsafe collaboration updates, and invalid container splits.

### DIFF
--- a/packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
@@ -1,4 +1,4 @@
-import { Fragment, Slice } from "prosemirror-model";
+import { Fragment, Node, NodeType, Slice } from "prosemirror-model";
 import type { Transaction } from "prosemirror-state";
 import { ReplaceStep } from "prosemirror-transform";
 import { Block, PartialBlock } from "../../../../blocks/defaultBlocks.js";
@@ -8,10 +8,94 @@ import {
   InlineContentSchema,
   StyleSchema,
 } from "../../../../schema/index.js";
+import { isContainerBlockNode } from "../../../../schema/blocks/children.js";
 import { blockToNode } from "../../../nodeConversions/blockToNode.js";
 import { nodeToBlock } from "../../../nodeConversions/nodeToBlock.js";
 import { getNodeById } from "../../../nodeUtil.js";
 import { getPmSchema } from "../../../pmUtil.js";
+import {
+  descendToFirstInsertionPos,
+  descendToLastInsertionPos,
+} from "../../containers/containerNav.js";
+
+/**
+ * Where blocks go relative to a reference block. `"before"`/`"after"` make them
+ * siblings of it; `"start"`/`"end"` nest them inside it, as its first or last
+ * children.
+ *
+ * The nested placements cover containers that have no children to point at:
+ * a `min: 0` container that is currently empty has no child block to insert
+ * before or after.
+ */
+export type BlockPlacement = "before" | "after" | "start" | "end";
+
+/**
+ * Resolves a `placement` against a reference block into the document position
+ * a node of `nodeType` should be inserted at, or `null` when the reference
+ * block cannot take it there.
+ *
+ * Shared by `insertBlocks` and the move commands, so "does this block fit
+ * here?" is answered in one place. The answer comes from the schema's content
+ * matches rather than from a hand-written rule, so a container's `children`
+ * config decides it.
+ *
+ * `wrapIn` is set when the position only becomes valid once the nodes are
+ * wrapped: a regular block with no children yet has no `blockGroup` for them
+ * to go in, so one is created around them.
+ */
+export function getInsertionPos(
+  doc: Node,
+  reference: { node: Node; posBeforeNode: number },
+  placement: BlockPlacement,
+  nodeType: NodeType,
+): { pos: number; wrapIn?: NodeType } | null {
+  const { node, posBeforeNode } = reference;
+
+  if (placement === "before" || placement === "after") {
+    const pos =
+      placement === "before" ? posBeforeNode : posBeforeNode + node.nodeSize;
+    const $pos = doc.resolve(pos);
+
+    return $pos.parent.contentMatchAt($pos.index()).matchType(nodeType)
+      ? { pos }
+      : null;
+  }
+
+  // A container holds its children itself, or, when it has content of its
+  // own, in its generated `__children` node, which the descent helpers step
+  // into. The helpers ignore sealed boundaries by default, which is correct
+  // here: an explicit `insertBlocks` placement is an intentional crossing.
+  if (isContainerBlockNode(node)) {
+    const pos =
+      placement === "start"
+        ? descendToFirstInsertionPos(node, posBeforeNode, nodeType)
+        : descendToLastInsertionPos(node, posBeforeNode, nodeType);
+
+    return pos === null ? null : { pos };
+  }
+
+  // A regular block keeps its children in a `blockGroup` that only exists once
+  // it has some.
+  const blockGroupType = nodeType.schema.nodes["blockGroup"];
+  if (node.type.name !== "blockContainer" || !blockGroupType) {
+    return null;
+  }
+
+  const blockGroupPos = posBeforeNode + 1 + node.firstChild!.nodeSize;
+
+  if (node.childCount < 2) {
+    return blockGroupType.contentMatch.matchType(nodeType)
+      ? { pos: blockGroupPos, wrapIn: blockGroupType }
+      : null;
+  }
+
+  const pos =
+    placement === "start"
+      ? descendToFirstInsertionPos(node.lastChild!, blockGroupPos, nodeType)
+      : descendToLastInsertionPos(node.lastChild!, blockGroupPos, nodeType);
+
+  return pos === null ? null : { pos };
+}
 
 export function insertBlocks<
   BSchema extends BlockSchema,
@@ -21,7 +105,7 @@ export function insertBlocks<
   tr: Transaction,
   blocksToInsert: PartialBlock<BSchema, I, S>[],
   referenceBlock: BlockIdentifier,
-  placement: "before" | "after" = "before",
+  placement: BlockPlacement = "before",
 ): Block<BSchema, I, S>[] {
   const id =
     typeof referenceBlock === "string" ? referenceBlock : referenceBlock.id;
@@ -37,14 +121,30 @@ export function insertBlocks<
     throw new Error(`Block with ID ${id} not found`);
   }
 
-  let pos = posInfo.posBeforeNode;
-  if (placement === "after") {
-    pos += posInfo.node.nodeSize;
+  if (nodesToInsert.length === 0) {
+    return [];
   }
 
-  tr.step(
-    new ReplaceStep(pos, pos, new Slice(Fragment.from(nodesToInsert), 0, 0)),
+  const target = getInsertionPos(
+    tr.doc,
+    posInfo,
+    placement,
+    nodesToInsert[0].type,
   );
+  if (!target) {
+    throw new Error(
+      `Cannot insert a block of type "${blocksToInsert[0].type ?? "paragraph"}" ` +
+        (placement === "before" || placement === "after"
+          ? `${placement} block with ID ${id}: its parent does not accept it.`
+          : `at the ${placement} of block with ID ${id}: the block does not accept it as a child.`),
+    );
+  }
+
+  const fragment = target.wrapIn
+    ? Fragment.from(target.wrapIn.create(null, nodesToInsert))
+    : Fragment.from(nodesToInsert);
+
+  tr.step(new ReplaceStep(target.pos, target.pos, new Slice(fragment, 0, 0)));
 
   // Now that the `PartialBlock`s have been converted to nodes, we can
   // re-convert them into full `Block`s.
--- a/packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
@@ -51,6 +51,11 @@ export function getInsertionPos(
 ): { pos: number; wrapIn?: NodeType } | null {
   const { node, posBeforeNode } = reference;
 
+  const descend = (holder: Node, pos: number) =>
+    placement === "start"
+      ? descendToFirstInsertionPos(holder, pos, nodeType)
+      : descendToLastInsertionPos(holder, pos, nodeType);
+
   if (placement === "before" || placement === "after") {
     const pos =
       placement === "before" ? posBeforeNode : posBeforeNode + node.nodeSize;
@@ -66,10 +71,7 @@ export function getInsertionPos(
   // into. The helpers ignore sealed boundaries by default, which is correct
   // here: an explicit `insertBlocks` placement is an intentional crossing.
   if (isContainerBlockNode(node)) {
-    const pos =
-      placement === "start"
-        ? descendToFirstInsertionPos(node, posBeforeNode, nodeType)
-        : descendToLastInsertionPos(node, posBeforeNode, nodeType);
+    const pos = descend(node, posBeforeNode);
 
     return pos === null ? null : { pos };
   }
@@ -89,10 +91,7 @@ export function getInsertionPos(
       : null;
   }
 
-  const pos =
-    placement === "start"
-      ? descendToFirstInsertionPos(node.lastChild!, blockGroupPos, nodeType)
-      : descendToLastInsertionPos(node.lastChild!, blockGroupPos, nodeType);
+  const pos = descend(node.lastChild!, blockGroupPos);
 
   return pos === null ? null : { pos };
 }
--- a/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
@@ -14,7 +14,8 @@ import {
   getNodeId,
 } from "../../../getBlockInfoFromPos.js";
 import { getNodeById } from "../../../nodeUtil.js";
-import { insertBlocks } from "../insertBlocks/insertBlocks.js";
+import { flattenNonInsertableBlocks } from "../../containers/fixContainer.js";
+import { getInsertionPos, insertBlocks } from "../insertBlocks/insertBlocks.js";
 import { removeAndInsertBlocks } from "../replaceBlocks/replaceBlocks.js";
 
 type BlockSelectionData = (
@@ -131,16 +132,6 @@ function updateBlockSelectionFromData(
   tr.setSelection(selection);
 }
 
-// Replaces top-level `column` blocks with their children, as a `column` is not
-// a valid block outside a `columnList`. Other blocks are returned as-is.
-function flattenColumns(
-  blocks: Block<any, any, any>[],
-): Block<any, any, any>[] {
-  return blocks.flatMap((block) =>
-    block.type === "column" ? block.children : [block],
-  );
-}
-
 /**
  * Removes the given blocks from the editor, then inserts them before/after a
  * reference block.
@@ -169,10 +160,12 @@ export function moveBlocks(
     // </column>
     // When the non-empty block is moved up, the column is seen as empty and
     // collapsed in the removal step, so the following insertion fails.
-    removeAndInsertBlocks(tr, blocks, [], { fixColumns: false });
+    removeAndInsertBlocks(tr, blocks, [], { fixContainers: false });
     insertBlocks<any, any, any>(
       tr,
-      flattenColumns(blocks),
+      // Blocks that can't stand on their own outside their container (e.g. a
+      // `column` outside its `columnList`) are replaced by their children.
+      flattenNonInsertableBlocks(blocks, editor.pmSchema),
       referenceBlock,
       placement,
     );
@@ -207,12 +200,33 @@ export function moveSelectedBlocksAndSelection(
   });
 }
 
-// Checks if a block is in a valid place after being moved. This check is
-// primitive at the moment and only returns false if the block's parent is a
-// `columnList` block. This is because regular blocks cannot be direct children
-// of `columnList` blocks.
-function checkPlacementIsValid(parentBlock?: Block<any, any, any>): boolean {
-  return !parentBlock || parentBlock.type !== "columnList";
+// Checks if a regular block would be in a valid place after being moved
+// before/after `referenceBlock`. A regular block nests under any non-container
+// block (it goes into that block's `blockGroup`), but a container block (e.g. a
+// `columnList`) only accepts what its content expression allows.
+//
+// Deferred to `getInsertionPos` so that "can a block go here?" has exactly
+// one answer, shared with `insertBlocks`, and comes from the schema rather
+// than from a rule restated here.
+function checkPlacementIsValid(
+  editor: BlockNoteEditor<any, any, any>,
+  referenceBlock: Block<any, any, any>,
+  placement: "before" | "after",
+): boolean {
+  return editor.transact((tr) => {
+    const posInfo = getNodeById(referenceBlock.id, tr.doc);
+    if (!posInfo) {
+      return false;
+    }
+    return (
+      getInsertionPos(
+        tr.doc,
+        posInfo,
+        placement,
+        editor.pmSchema.nodes["blockContainer"],
+      ) !== null
+    );
+  });
 }
 
 // Gets the placement for moving a block up. This has 3 cases:
@@ -253,8 +267,8 @@ function getMoveUpPlacement(
     return undefined;
   }
 
-  const referenceBlockParent = editor.getParentBlock(referenceBlock);
-  if (!checkPlacementIsValid(referenceBlockParent)) {
+  if (!checkPlacementIsValid(editor, referenceBlock, placement)) {
+    const referenceBlockParent = editor.getParentBlock(referenceBlock);
     return getMoveUpPlacement(
       editor,
       placement === "after"
@@ -305,8 +319,8 @@ function getMoveDownPlacement(
     return undefined;
   }
 
-  const referenceBlockParent = editor.getParentBlock(referenceBlock);
-  if (!checkPlacementIsValid(referenceBlockParent)) {
+  if (!checkPlacementIsValid(editor, referenceBlock, placement)) {
+    const referenceBlockParent = editor.getParentBlock(referenceBlock);
     return getMoveDownPlacement(
       editor,
       placement === "before"
--- a/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
+++ b/packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
@@ -1,3 +1,4 @@
+import { NodeType } from "prosemirror-model";
 import {
   NodeSelection,
   Selection,
@@ -212,21 +213,31 @@ function checkPlacementIsValid(
   editor: BlockNoteEditor<any, any, any>,
   referenceBlock: Block<any, any, any>,
   placement: "before" | "after",
+  nodeType: NodeType,
 ): boolean {
   return editor.transact((tr) => {
     const posInfo = getNodeById(referenceBlock.id, tr.doc);
     if (!posInfo) {
       return false;
     }
-    return (
-      getInsertionPos(
-        tr.doc,
-        posInfo,
-        placement,
-        editor.pmSchema.nodes["blockContainer"],
-      ) !== null
-    );
+    return getInsertionPos(tr.doc, posInfo, placement, nodeType) !== null;
   });
+}
+
+// The PM node type `insertBlocks` validates a destination against: the first
+// flattened block's own node type when it's a container (e.g. `callout`),
+// otherwise the generic `blockContainer` wrapper. Mirrors what `moveBlocks`
+// inserts (`flattenNonInsertableBlocks` + `insertBlocks`), so the placement
+// pre-check agrees with the insertion instead of always assuming a regular
+// block.
+function movedNodeType(
+  editor: BlockNoteEditor<any, any, any>,
+  block: Block<any, any, any>,
+): NodeType {
+  const blockContainer = editor.pmSchema.nodes["blockContainer"];
+  const first = flattenNonInsertableBlocks([block], editor.pmSchema)[0];
+  const nodeType = first?.type ? editor.pmSchema.nodes[first.type] : undefined;
+  return nodeType?.isInGroup("bnBlock") ? nodeType : blockContainer;
 }
 
 // Gets the placement for moving a block up. This has 3 cases:
@@ -241,6 +252,7 @@ function checkPlacementIsValid(
 // the block is already at the top of the document.
 function getMoveUpPlacement(
   editor: BlockNoteEditor<any, any, any>,
+  nodeType: NodeType,
   prevBlock?: Block<any, any, any>,
   parentBlock?: Block<any, any, any>,
 ):
@@ -267,10 +279,11 @@ function getMoveUpPlacement(
     return undefined;
   }
 
-  if (!checkPlacementIsValid(editor, referenceBlock, placement)) {
+  if (!checkPlacementIsValid(editor, referenceBlock, placement, nodeType)) {
     const referenceBlockParent = editor.getParentBlock(referenceBlock);
     return getMoveUpPlacement(
       editor,
+      nodeType,
       placement === "after"
         ? referenceBlock
         : editor.getPrevBlock(referenceBlock),
@@ -293,6 +306,7 @@ function getMoveUpPlacement(
 // the block is already at the bottom of the document.
 function getMoveDownPlacement(
   editor: BlockNoteEditor<any, any, any>,
+  nodeType: NodeType,
   nextBlock?: Block<any, any, any>,
   parentBlock?: Block<any, any, any>,
 ):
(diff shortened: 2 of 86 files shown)


packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
```diff
@@ -8,10 +8,93 @@ import {
...
+ * Shared by `insertBlocks` and the move commands, so "does this block fit
+ * here?" is answered in one place. The answer comes from the schema's content
+ * matches rather than from a hand-written rule, so a container's `children`
+ * config decides it.
+ *
+ * `wrapIn` is set when the position only becomes valid once the nodes are
+ * wrapped: a regular block with no children yet has no `blockGroup` for them
+ * to go in, so one is created around them.
+ */
+export function getInsertionPos(
```
YousefED, line 46: (nice to have)

this function is exported in @blocknote/core now, which afaik doesn't need to be. Maybe a good opportunity to introduce the "index"-file-per-folder pattern of exporting?

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
```diff
@@ -8,10 +8,93 @@ import {
...
+    const pos = descend(node, posBeforeNode);
+
+    return pos === null ? null : { pos };
+  }
+
+  // A regular block keeps its children in a `blockGroup` that only exists once
+  // it has some.
+  const blockGroupType = nodeType.schema.nodes["blockGroup"];
+  if (node.type.name !== "blockContainer" || !blockGroupType) {
+    return null;
```
YousefED, line 83: should this be an error?

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
```diff
@@ -8,10 +8,93 @@ import {
...
+  }
+
+  const blockGroupPos = posBeforeNode + 1 + node.firstChild!.nodeSize;
```
YousefED, line 86: for safety and readability, can't we use `getBlockInfo` for this? Then we don't need to do the manual position calculations, etc

packages/core/src/api/blockManipulation/containers/containerNav.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (opts?.respectSealed && isSealed(container)) {
+    return null;
+  }
+  const endPos = containerBeforePos + 1 + container.content.size;
+  if (container.contentMatchAt(container.childCount).matchType(nodeType)) {
+    return endPos;
+  }
+  const lastChild = container.lastChild;
+  if (lastChild && isContainerNode(lastChild.type)) {
+    return descendToLastInsertionPos(
```
YousefED, line 34: Is it desirable to descend like this? Or would it be prefered to throw an error? 

The way I understand it, this would allow inserting a regular block into a ColumnList block (by inserting it into the last column). Isn't it cleaner + clearer / less code to just throw an error in that case?

packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts
```diff
@@ -8,10 +8,93 @@ import {
...
+/**
+ * Where blocks go relative to a reference block. `"before"`/`"after"` make them
+ * siblings of it; `"start"`/`"end"` nest them inside it, as its first or last
+ * children.
+ *
+ * The nested placements cover containers that have no children to point at:
+ * a `min: 0` container that is currently empty has no child block to insert
+ * before or after.
+ */
+export type BlockPlacement = "before" | "after" | "start" | "end";
```
YousefED, line 30: would "as-first-child" or "as-last-child" be clearer than "start" / "end"?

packages/core/src/api/blockManipulation/containers/fixContainer.ts
```diff
@@ -0,0 +1,347 @@
...
+    .forEach(({ id }) => {
+      const target = getNodeById(id, tr.doc);
+      if (!target) {
+        return;
+      }
+      fixContainer(tr, target.posBeforeNode);
+    });
+}
+
+export function flattenNonInsertableBlocks<
```
YousefED, line 324: what does this function do? Not super clear to me from the name

packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
```diff
@@ -207,12 +201,43 @@ export function moveSelectedBlocksAndSelection(
...
+  });
+}
+
+// The PM node type `insertBlocks` validates a destination against: the first
+// flattened block's own node type when it's a container (e.g. `callout`),
+// otherwise the generic `blockContainer` wrapper. Mirrors what `moveBlocks`
+// inserts (`flattenNonInsertableBlocks` + `insertBlocks`), so the placement
+// pre-check agrees with the insertion instead of always assuming a regular
+// block.
+function movedNodeType(
```
YousefED, line 233: I don't really understand what this function does (the comment above also doesn't very clearly describe it)

packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts
```diff
@@ -369,20 +399,28 @@ export function moveBlocksDown(
 ) {
   editor.transact(() => {
     let sourceBlock: Block<any, any, any> | undefined;
+    // The block whose position anchors the move (the last of a selection when
+    // moving down) vs. the first block that gets inserted, which is what the
+    // placement check must validate against.
```
YousefED, line 404: maybe I'm misunderstanding this, but shouldn't the placement validate against all blocks that are being moved?

packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts
```diff
@@ -84,10 +88,10 @@ export function removeAndInsertBlocks<
 
     const $pos = tr.doc.resolve(pos - removedSize);
 
-    if ($pos.node().type.name === "column") {
-      columnListPositions.add($pos.before(-1));
-    } else if ($pos.node().type.name === "columnList") {
-      columnListPositions.add($pos.before());
+    for (const container of getAncestorContainers($pos.doc, $pos.pos)) {
```
YousefED, line 91: maybe nice to make sure this behavior (and the change) is covered in tests?

packages/core/src/api/blockManipulation/commands/splitBlock/splitBlock.ts
```diff
@@ -36,7 +37,15 @@ export const splitBlockTr = (
 
   const info = getBlockInfo(nearestBlockContainerPos);
 
-  if (!info.isBlockContainer) {
+  if (!info.isWrappedBlock) {
```
YousefED, line 40: add to unit test?

packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.test.ts
```diff
@@ -181,7 +181,7 @@ describe("Test updateBlock", () => {
       getNodeById("heading-with-everything", getEditor().prosemirrorState.doc)!,
     );
 
-    if (!info.isBlockContainer) {
+    if (!info.isWrappedBlock) {
```
YousefED, line 184: generic comment about this testfile:

I think updateBlock has a lot of cases for what happens when we change from 1 type of block to another, that are also covered by tests here (e.g.: image to paragraph, table to paragraph, etc)

Do we need additional unit tests for container related functionality?

packages/core/src/api/blockManipulation/containers/containerNav.ts
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Seal handling for the navigation helpers below. By default the helpers
+ * ignore seals. The block manipulation API crosses them freely, since an
+ * explicit placement is an intentional crossing. Gesture code (keyboard
+ * merges and moves) opts in with `respectSealed`, so content never
+ * implicitly crosses a sealed boundary.
+ */
+type SealOpts = { respectSealed?: boolean };
+
+export function descendToLastInsertionPos(
```
YousefED, line 19: can we have jsdoc for these helpers?

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -6,8 +6,11 @@ import {
   BlockImplementation,
   BlockSchema,
   InlineContentSchema,
+  isContainerType,
   StyleSchema,
 } from "../../../../schema/index.js";
+import { fillContainerAttributes } from "../../../../schema/blocks/containerAttributes.js";
```
YousefED, line 12: as far as I can see there are no additions to the internal / external html / markdown test cases

packages/core/src/api/getBlockInfoFromPos.ts
```diff
@@ -20,13 +26,17 @@ export type BlockInfo = {
...
+      // its own.
 
       /**
-       * The Prosemirror node that holds block.children. For non-blockContainer, this node will be the same as bnBlock.
+       * The Prosemirror node that holds block.children. For a container block,
+       * this node is the same as bnBlock.
        */
       childContainer: SingleBlockInfo;
-      isBlockContainer: false;
+      blockContent?: undefined;
```
YousefED, line 38: not sure, should this be `never` or `undefined`?

packages/core/src/api/pmUtil.ts
```diff
@@ -67,7 +68,16 @@ export function isPlainContentNodeType(
   schema: Schema,
   nodeType: NodeType,
 ): boolean {
-  if (getBlockSchema(schema)[nodeType.name]?.content === "plain") {
+  const blockSchema = getBlockSchema(schema);
+  // A content-bearing container's content lives in a generated node, so it
+  // isn't a key in the block schema. Resolve it back to the block it belongs
+  // to.
+  const blockType =
```
YousefED, line 75: is this needed? the container will never have plain content, right?

packages/core/src/yjs/extensions/FixUpSchema.ts
```diff
@@ -25,7 +25,15 @@ export const FixUpSchemaExtension = createExtension(({ editor }) => {
 
       // create a copy that we can mutate (otherwise, assigning attrs is not safe and corrupts the pm state)
       const jsonNode = JSON.parse(JSON.stringify(ret.toJSON()));
-      jsonNode.content[0].content[0].attrs.id = "initialBlockId";
+      // The first fill of the doc's blockGroup is guaranteed to be a
```
YousefED, line 28: is this needed? how could a custom schema omit an `id`?

packages/core/src/schema/blocks/types.ts
```diff
@@ -80,6 +86,98 @@ export interface BlockConfigMeta<
...
+ * slot. Ordered multi-slot bodies (a `sequence` of slots) can be added later
+ * as a sibling form.
+ */
+export type ChildrenConfig = {
+  /** What may appear as a child. See {@link ChildrenAllow}. */
+  allow: ChildrenAllow;
+  /** @default 1 */
+  min?: number;
+  /** @default unbounded */
+  max?: number;
```
YousefED, line 123: as discussed, `max` might not be necessary (until we see a very useful use-case, maybe omit to reduce maintenance burden)

packages/core/src/api/nodeConversions/contentContainers.test.ts
```diff
@@ -0,0 +1,307 @@
+// @vitest-environment node
+import type { Node, Schema } from "@tiptap/pm/model";
```
YousefED, line 2: I'm not sure about the naming / location of this file. e.g.: there's no `contentContainers.ts` source file

If it's a unit test, it should be in a corresponding blockToNode / nodeToBlock test. If it's more of an integration test, it should be `/tests` directory?

(maybe we also need a clearer policy around this / update the .skill file)

YousefED (commented):
Did a first pass, still need to process some of it (it's a biggie!).

Overall looks pretty neat. My main concern is that it changes / adds quite a bunch of code to core areas of the library, and wondering whether we can simplify things.

Two options I see;

a) Remove the possibility for a rich content slot. This is a big part of what makes this PR significantly more complex than multi-columns
b) See if we can consolidate the default blocks / blockcontainers into the new setup. (also see comment by Claude below)

My other main point of feedback is that I think we should closely review testing coverage. E.g.:
- containers should probably be covered in `tests/src/unit/` subtests
- some of the `blockManipulation` code has changed significantly, but the corresponding tests have not been updated to tests the new functionality

YousefED (commented):
additional claude feedback:
---

# Review: PR #2997 — container block API (`container-blocks/core` → `main`)

**Verdict:** The architecture is solid and the legacy column port is careful — renames are complete repo-wide, the `__content`/`__children` facade is tight and well-tested, exporters and keyboard handlers were genuinely generalized, and legacy shims are correctly marked. But the review found **21 confirmed correctness bugs**, concentrated in two systematic gaps: **the sealed-boundary contract is enforced per-call-site and four gesture paths forgot it**, and **content-bearing containers are second-class wherever code uses `isContainerNode` (pure-only) instead of `isContainerBlockNode`**. Every finding below survived an adversarial verification pass; most were reproduced with actual editor tests. (12 finder agents → 72 candidates → deduped and verified; 3 candidates were refuted and dropped.)

## High severity

**1. The sealed boundary leaks through four gesture paths** — all reproduced end-to-end through the real keymap:

- [mergeBlocks.ts:290](packages/core/src/api/blockManipulation/commands/mergeBlocks/mergeBlocks.ts#L290): `mergeBlocksCommand` calls `getBottomNestedBlockInfo` without `{ stopAtSealed: true }`, so Backspace at the start of a block merges its text into a block *inside* a sealed container nested under the previous sibling — violating that helper's own doc comment.
- [moveBlocks.ts:212](packages/core/src/api/blockManipulation/commands/moveBlocks/moveBlocks.ts#L212) (`checkPlacementIsValid`, plus `getMoveUpPlacement`/`getMoveDownPlacement` at 253/307): Shift-Mod-ArrowUp/Down move blocks **into** a sealed container, and a sealed container's last child **out** — no `isSealed` consultation anywhere in the path.
- [KeyboardShortcutsExtension.ts:589](packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts#L589): Delete at the end of a block whose first nested child is a *content-bearing* sealed container dissolves it (title merged, children lifted) — the branch checks `isWrappedBlock` but never `isSealed`, and runs before `selectSealedSiblingCommand`.
- [KeyboardShortcutsExtension.ts:842](packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts#L842): the Delete parent-climb branch deletes a sealed content-bearing container wholesale — the climb guards `isSealed` on ancestors climbed out of, but never on the found next block itself.

In all four, *pure* sealed containers are only incidentally protected by `!isWrappedBlock`. Root cause (flagged independently by the design audit): seal-respecting is opt-in per call site with the unsafe behavior as default ([containerNav.ts:12-17](packages/core/src/api/blockManipulation/containers/containerNav.ts#L12-L17)) — consider inverting the default so navigation helpers respect seals unless an API entry point explicitly opts out.

**2. `allow: "containers"` schemas pass validation, then overflow the stack at editor creation** — [validateChildren.ts:379](packages/core/src/schema/blocks/validateChildren.ts#L379): `validateNoCycles` skips wildcard `containers === true` configs, but `allowTerm` compiles them to the self-including `anyContainer` group, so `assertContainersAreFillable`'s `fillBefore` recurses forever. Verified empirically twice (independent repro scripts): mutual wildcards always crash, and even the supported `{tabs: containers, card: any}` crashes depending on **registration order**. This is exactly the failure the validator's header promises to catch statically.

**3. `doc.check()` on `initialContent` breaks the PR's own compatibility promise** — [BlockNoteEditor.ts:568](packages/core/src/editor/BlockNoteEditor.ts#L568): legacy persisted documents (canonical case: a single-column `columnList`, which old column bugs produced and `fixColumnList` repaired lazily) now throw in the constructor before any repair path can run. The PR body's compat section says invalid legacy structures throw "on insert" and otherwise keep working — this check silently moves that throw to initial load, bricking previously-loadable documents. Either run the repair pass pre-check or scope the check to freshly-authored content.

**4. Enter in a container title silently destroys text** — [KeyboardShortcutsExtension.ts:1121](packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts#L1121): the handler hardcodes the new first child as `blockContainer>paragraph`. For a container whose `children.allow` excludes regular blocks (a combination `validateChildren` permits), reproduction showed **no error**: the fitter drops the unfittable child and the `tr.delete` has already removed the title tail — the keystroke deletes "lo" from "Hello" with nothing inserted anywhere. Derive the child from the children node's `contentMatch`/`children.default` instead.

**5. Converting to/from a container resets the block's props** — [updateBlock.ts:183](packages/core/src/api/blockManipulation/commands/updateBlock/updateBlock.ts#L183): the full-replace arm spreads only `{content, children, ...block}`, so `updateBlock("p-0", {type: "toggle"})` on a red-background paragraph yields default props and a regenerated id (reproduced; the heading control keeps both). The id half is codified in [contentContainers.test.ts:186](packages/core/src/api/blockManipulation/containers/contentContainers.test.ts#L186); the props drop is codified nowhere. The arm predates the PR, but the PR routes the mainstream "turn into toggle/callout" flow through it — the blast radius grew from column edge cases to a headline feature.

**6. Insertion validation misses ProseMirror's tail re-match** — [insertBlocks.ts:64](packages/core/src/api/blockManipulation/commands/insertBlocks/insertBlocks.ts#L64) and [containerNav.ts:61](packages/core/src/api/blockManipulation/containers/containerNav.ts#L61): `contentMatchAt(i).matchType(type)` alone (no `matchFragment` + `validEnd`, i.e. not `canReplaceWith`). Reproduced: `insertBlocks(..., "before"|"start")` into a full `max: 1` container throws raw `Invalid content` instead of the friendly error, and `moveBlocksDown` next to one crashes mid-command (transaction aborts, so no corruption). The end-side checks are fine; `insertPlacement.test.ts` only covers `"end"`, which is why this was untested.

**7. Content-bearing containers are skipped by the `isContainerNode` family** — the recurring near-miss between `isContainerNode` (pure only) and `isContainerBlockNode` (both shapes):

- [containerNav.ts:33](packages/core/src/api/blockManipulation/containers/containerNav.ts#L33)/[65](packages/core/src/api/blockManipulation/containers/containerNav.ts#L65): insertion descent refuses to enter a content-bearing child container — `insertBlocks(..., "end")` throws "does not accept it as a child" where the pure-container analog descends (codified in [insertPlacement.test.ts:131](packages/core/src/api/blockManipulation/commands/insertBlocks/insertPlacement.test.ts#L131)).
- [containerNav.ts:89-97](packages/core/src/api/blockManipulation/containers/containerNav.ts#L89-L97): `getFirstLeafBlock` treats one as a leaf, so Delete pulls the whole subtree out where a nested pure container yields its deepest leaf (behavioral asymmetry, not data loss).
- [BlockPopover.tsx:36](packages/react/src/components/Popovers/BlockPopover.tsx#L36): the zero-rect anchoring fix guards on `isContainerNode`, so content-bearing React containers still anchor to the `display: contents` contentDOM → popover/drag-handle at (0,0) — the exact bug the branch fixes for pure containers. Note `isContainerBlockNode` takes a `Node` and currently only exports via `@blocknote/core/internal`.
- [fixContainer.ts:24-38](packages/core/src/api/blockManipulation/containers/fixContainer.ts#L24-L38): `isEmptyContainerChild` never counts an empty content-bearing container as empty (conservative direction, but it makes `whenEmptied` unable to fire for containers of content-bearing children — and enables the next finding).

**8. `unwrapContainer` corrupts content-bearing survivors** — [fixContainer.ts:215](packages/core/src/api/blockManipulation/containers/fixContainer.ts#L215): stripping "one node level" from a `containerOnly` survivor exposes its raw `__content`/`__children` pair, invalid in `blockGroup`. Reproduced: single survivor → raw `Invalid content for node blockGroup` aborting the whole `removeBlocks`; the multi-survivor branch instead silently drops the title via the fitter.

**9. A selection starting mid-title loses the title text with no marker** — [nodeToBlock.ts:598](packages/core/src/api/nodeConversions/nodeToBlock.ts#L598): `prosemirrorSliceToSlicedBlocks`'s open-boundary recursion assumes the cut runs through `__children`; when it runs through `__content`, the partial title vanishes and `blockCutAtStart` is never set (reproduced; regular blocks keep partial content + marker). The end-open splice behavior is test-codified as intended — this start-open variant is the uncodified gap. Affects `getSelectionCutBlocks` consumers (xl-ai selection context).

**10. `children.default` with explicit ids stamps duplicate ids** — [blockToNode.ts:419](packages/core/src/api/nodeConversions/blockToNode.ts#L419): seeding/refill converts defaults via `blockToNode`, which honors `id`; reproduced two containers sharing a child id (UniqueID's dedupe only acts within one transaction's changed ranges). Either strip ids from defaults or reject them in `validateDefault`.

## Medium severity

- **Backspace-into-container skips source repair** — [KeyboardShortcutsExtension.ts:281-292](packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts#L281-L292): raw `tr.delete`/`tr.insert` without `fixContainersById`, so the source container gets PM schema-padding instead of its configured refill/unwrap (reproduced; observable when `min ≥ 2`). The parallel branches use `moveBlockOutAndPlaceCaret` for exactly this.
- **docx exporter breaks on array-returning container mappings** — [docxExporter.ts:143](packages/xl-docx-exporter/src/docx/docxExporter.ts#L143): `ret.push(self as Table)` skips the `Array.isArray` spread applied to every other block; the mapping type explicitly permits `Paragraph[]`, and the branch now runs for every custom container.
- **Side-menu anchor over-matches** — [SideMenu.ts:305](packages/core/src/extensions/SideMenu/SideMenu.ts#L305): `querySelector('[data-node-type="blockOuter"]')` is depth-first over all descendants, so a container whose first child is a container anchors to a deeply-nested block. Scope to direct children (`:scope > …` or the PR's own `getDirectChildBlocks`).
- **Tab/Shift+Tab across columns became a no-op** — [nestBlock.ts:22](packages/core/src/api/blockManipulation/commands/nestBlock/nestBlock.ts#L22): the widened `childContainer` predicate resolves the block range *at* the columnList, so sink/lift preconditions never hold where `main` nested/lifted the whole columnList. Reachability caveat: default-toolbar setups swallow Tab for multi-block selections; bites `tabBehavior: "prefer-indent"`, toolbar-less, and programmatic callers.
- **Pre-existing (not a PR regression), surfaced while verifying:** [replaceBlocks.ts:81](packages/core/src/api/blockManipulation/commands/replaceBlocks/replaceBlocks.ts#L81): when `blocksToRemove` aren't in document order, the stale `tr.insert(pos, …)` plus adjusted delete **silently destroys the inserted block** (reproduced: `replaceBlocks(['p-2','p-0'], [NEW])` → doc ends up `[p-1]`). Same code exists on `main`. Also pre-existing: multi-block pastes into container titles mangle the container (independent of this PR's retype change, which verification cleared).

## Minor

- [validateChildren.ts:130](packages/core/src/schema/blocks/validateChildren.ts#L130): `whenEmptied` and `placement` string values are never validated (unlike `allow`/`boundary`), so `"unwarp"` silently means `"refill"` (reproduced).

## Quality (fact-checked, all 18 held)

*Duplication / single-owner gaps:* the `data-children-of`/`data-content-type`/`bn-inline-content` literals are written independently in 3 files across 2 packages with no shared constant; the "container type incl. legacy columns" predicate is spelled 3× ([containerUI.ts:46](packages/core/src/api/blockManipulation/containers/containerUI.ts#L46), [Exporter.ts:91](packages/core/src/exporter/Exporter.ts#L91), [extensions.ts:66](packages/core/src/editor/managers/ExtensionManager/extensions.ts#L66)); the orphaned-content-becomes-paragraph policy 3× (each hidden behind a cast); `ReactBlockSpec`'s container static-render block is pasted twice (already diverging on `isFileBlock`); `unwrapContainer`/`refillContainer` share a duplicated below-min preamble; `getFirstLeafBlock`/`descendToFirstInsertionPos` duplicate the descent walk with asymmetric seal opts; the Backspace seal-probe runs the full descent twice to recover one bit ([KeyboardShortcutsExtension.ts:245-263](packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts#L245-L263) — a discriminated return type would fix it); `getChildrenConfig` vs `isContainerType` gives one predicate three spellings.

*Efficiency:* `getContainerUIInfo` rebuilds sets+selector from all blockSpecs **per mousemove** (schema is immutable — memoize per editor); `ContainerNodeView` does an O(doc) `editor.getBlock(id)` on every render when the cached `nodeToBlock(props.node, …)` fallback would be O(1); `removeEmptyChildren` is O(n²) (iterate `container.forEach` + delete back-to-front); `fixContainersById` does one full-doc scan per ancestor; the undepped `useLayoutEffect` re-stamps all attributes every render (memoize last-applied tuple).

*Conventions (CLAUDE.md):* `mergeIntoContainerContent` is a new exported `const` arrow with an `any`-typed `dispatch`; `blockSchema[block.type as any]` in [serializeBlocksExternalHTML.ts:287](packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts#L287) hides that `PartialBlock.type` can be `undefined` (would throw in `isContainerType`); gratuitous `as any`/`Record<string, any>` casts in [extensions.ts:78](packages/core/src/editor/managers/ExtensionManager/extensions.ts#L78) and [Exporter.ts:94](packages/core/src/exporter/Exporter.ts#L94) where typed access compiles.

*Leftovers:* the `coords.left + 50` "bit hacky" probe survives even though this PR's own `rectIndexAtCursor` can resolve the child from measured rects; `isContainerNode` reaches `index.ts` through a `fixContainer.ts` re-export hop that already forced an import-cycle workaround comment.

## Verified clean (for the record)

`isBlockContainer`→`isWrappedBlock` rename complete repo-wide; xl-multi-column imports and legacy shims all route correctly; `@blocknote/core/internal` wiring complete; no stale `fixColumns` callers; test changes are mechanical/additive with no weakened assertions; `mergeIntoContainerContent` position math correct.

---

**Suggested priority:** the sealed-boundary family and `updateBlock` props reset are contract violations in the new API's headline features; the stack overflow and `doc.check()` compat break hurt schema authors and existing users on upgrade; the `isContainerNode` family is one systematic fix (use `isContainerBlockNode` at the four sites).

---

# Architecture: should regular blocks migrate to the container model?

Most of the 21 confirmed bugs aren't isolated mistakes; they're the tax of **three block shapes coexisting** in one document model:

1. **Regular blocks**: shared `blockContainer` node → `blockContent` child + optional generic `blockGroup`
2. **Pure containers**: one per-type node holding children directly
3. **Content-bearing containers**: per-type outer node → generated `__content` + `__children`

Nearly every finding family maps to a seam between these: the `isContainerNode` vs `isContainerBlockNode` vs `isWrappedBlock` near-misses (BlockPopover, containerNav descent, `isEmptyContainerChild`), the two-arm duplication in `updateBlock`/`blockToNode`, the serializer re-building region DOM by hand, and `unwrapContainer` guessing how many node levels to strip per shape. So the instinct is right: fewer shapes would have prevented most of this class.

**What full migration would mean.** "Regular block = content-bearing container" is coherent: `paragraph` becomes `{content: 'inline', children: {allow: 'any'}}`, `blockContainer`/`blockGroup` disappear, and every block is `outer > __content? + __children?`. It would even be a feature win — children rules (`allow`/`min`/`max`/`whenEmptied`) would become expressible per block type for *all* blocks, not just containers. But it's a **persisted-format break**, and that's the real cost, not code churn:

- **Yjs/collab docs** share the XmlFragment structure across clients on different versions — you can't atomically migrate a live collab document, so you'd need a format version + migration story.
- HTML serialization and clipboard formats change with it (the `bn-block-outer` structure vs the `data-content-type`/`data-children-of` scheme).
- The shared `blockContainer` wrapper is what makes `paragraph → heading` a cheap, **id- and props-preserving** `setNodeMarkup`. The review confirmed the container conversion path (full-replace) currently loses both — under naive unification *every* type change takes that path, so identity preservation in the replace path must be solved first anyway.

**The pragmatic path — unify the code, not (yet) the format.** The key observation is that shape 1 is already expressible as a degenerate case of shape 3: `blockContainer` *is* an outer node, its `blockContent` child *is* the `__content` slot (just type-polymorphic instead of generated per type), and `blockGroup` *is* the `__children` holder. The PR half-built this facade (`isWrappedBlock`, `getContainerChildrenHolder`, the naming helpers in `children.ts`) but the consumers still branch per shape instead of normalizing. Concretely:

1. **One regions accessor** — something like `getBlockRegions(node) → {outer, content?, childrenHolder?}` that all three shapes resolve into — consumed by `getBlockInfoFromPos`, containerNav, `updateBlock`, `blockToNode`, and both serializers. That collapses the duplicated arms and makes the predicate near-misses structurally impossible (there's no "which shape am I" question left to get wrong).
2. **Generate `blockContainer`/`blockGroup` from the container machinery** in `children.ts`/`createSpec.ts` so parse rules, DOM markers, and priorities have one owner — several reuse findings (marker literals in three files, the container-predicate triplication) fall out of that for free.
3. **Migrate the semantically-container defaults onto the API first**: multi-column (already planned — the legacy shims mark the sites), then toggle/quote/callout-style blocks. That deletes bespoke keyboard handlers and exercises the container API against real defaults without touching paragraph/heading documents.
4. **Reserve node-level unification for an explicit format-versioned change** later — after step 1, it becomes a mostly mechanical migration rather than an architectural one, and step 3 will have shown what the container model still can't express.

Bottom line: don't merge regular blocks into container nodes in this PR or the next — the collab-format compatibility and identity-preservation costs are real. But treat the accessor-level unification (step 1) as near-term work, because the review shows the branch-per-shape code is where the bugs actually breed, and that fix needs no format change at all.

nperez0111:

PR #2998, opened 2026-08-21
feat(xl-multi-column): migrate columns onto the container block API

nperez0111:
Part 2 of 3 of the container blocks stack (1: core API, 2: multi-column migration ← you are here, 3: docs & examples). Stacked on #2997.

## What this does

Migrates `@blocknote/xl-multi-column` from hand-written ProseMirror nodes onto the container block API from the previous PR, and deletes the legacy compatibility shims that PR carried for it.

- `column` / `columnList` are now regular `createBlockSpec` container blocks (`pm-nodes/Column.ts` and `pm-nodes/ColumnList.ts` deleted):
  - `columnList`: `children: { allow: ["column"], min: 2, whenEmptied: "unwrap" }`
  - `column`: `placement: "containerOnly"`, so it can only ever live inside a `columnList`
  - both `meta.draggable: false`, matching the previous side-menu behavior
- Column keyboard behavior, repair (collapsing emptied columns, unwrapping single-column lists), drag & drop, and HTML round-trips now come entirely from the generic container machinery; the bespoke column code paths in core are gone.
- Core cleanup: `fixColumnList.ts` deleted along with every `// Legacy` shim from the previous PR (`blockToNode`, internal HTML serializer, `UniqueID` types, `Exporter.isContainerBlock`, `containerUI`, `fragmentToBlocks`, `fixContainer`).
- Column resizing moves fully into `ColumnResizeExtension` (widths are no longer a schema prop concern of core).

## Behavior notes

- Conversion snapshots change shape slightly (attribute order, `data-children-of` markers on the children hosts, `data-width` now handled by the resize extension).
- `insertBlocks` with a partial `columnList` (missing columns/children) now auto-fills from the container config instead of throwing, matching every other container block.

## Testing

- `fixColumnLists.test.ts` → `fixContainer.test.ts` (same scenarios against the generic repair), new `enter.test.ts` and `columnResize.test.ts`.
- Multi-column e2e suite (`tests/src/end-to-end/multicolumn`) extended for the migrated behavior; full suite green.

## Summary by CodeRabbit

* **New Features**
  * Improved multi-column editing, including more reliable resizing, dragging, dropping, and block placement.
  * Columns now support flexible widths and improved container behavior.
  * Added support for unwrapping column content when columns are removed or emptied.

* **Bug Fixes**
  * Fixed resize state after columns or documents are replaced.
  * Improved drag-handle positioning for blocks inside columns.
  * Improved Enter-key behavior within multi-column layouts.


nperez0111:

PR #2999, opened 2026-08-21
docs: container block docs and examples

nperez0111:
Part 3 of 3 of the container blocks stack (1: core API, 2: multi-column migration, 3: docs & examples ← you are here). Stacked on #2998.

## What this adds

- **Docs**: a new [Container Blocks](docs/content/docs/features/custom-schemas/container-blocks.mdx) feature page covering the `children` config, both container shapes, boundaries, repair, parsing, and exporter mappings; small pointer updates to the Custom Blocks and Manipulating Blocks pages.
- **Example: Container Block** (`09-container-block`): a Notion-style callout that holds child blocks, plus the "string prop slot" pattern for its title.
- **Example: Table Built From Container Blocks** (`12-container-table`): BlockNote's table rebuilt as four container blocks (`table`/`tableRow`/`tableCell`/`tableHeader`) with `boundary: "sealed"` cells that can hold arbitrary blocks - no `prosemirror-tables` involved.
- Regenerated `playground/src/examples.gen.tsx` and the lockfile entries for the two new example packages.

### DIFF
--- a/docs/content/docs/features/custom-schemas/container-blocks.mdx
+++ b/docs/content/docs/features/custom-schemas/container-blocks.mdx
@@ -1,0 +1,257 @@
+---
+title: Container Blocks
+description: Learn how to create custom blocks that hold other blocks as their body
+---
+
+# Container Blocks
+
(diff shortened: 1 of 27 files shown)


nperez0111:

PR #3009, opened 2026-08-25
refactor: remove content-container support

nperez0111:
## What this PR is

A **decision artifact** for the container blocks API, stacked on `container-blocks/examples-docs`:

- **Merge it** → content containers (`content: "inline" | "plain"` combined with `children`) are **cut from v1**. Containers are always `content: "none"`; an editable title/caption is a string prop rendered as an input (the pattern the callout example and docs already use).
- **Close it** → content-container support **stays** as currently implemented on the base branch.

The diff is the exact price of keeping the feature: everything below exists *only* to support combining a block's own content with children.

## What's removed (34 files, +149 / −2,531)

**Schema compilation** — a content container compiled to three ProseMirror nodes: the block node plus generated `<type>__content` (group `containerContent`) and `<type>__children` nodes. Deleted: `buildContentContainerNode`, the generated node names and their name-collision validation, the `containerContent` group, and the `extraNodes` plumbing through the spec type and extension manager.

**Predicates** — `isContentContainerNode` and the `isContainerBlockNode` wrapper are gone; every `isContainerNode(x) || isContentContainerNode(x)` disjunction (containerNav, fixContainer, insertBlocks, conversions) collapses to the pure-container predicate. `getContainerChildrenHolder` and `getContentContainerNodeTypes` (which existed to locate the generated nodes) are gone with it.

**Behavior** — the keyboard branches for Enter-splits-the-content-head-into-a-first-child and Backspace-merges-the-first-child-back-into-it (`mergeIntoContainerContent`), the `fixContainer` guard that waited for the content head to empty, the splitBlock/mergeBlocks/updateBlock content-container arms, and the content-container paths in `blockToNode` / `nodeToBlock` / `fragmentToBlocks`.

**DOM contract** — the `[data-content-type]` / `[data-children-of]` sibling-region rendering in the internal HTML serializer and the React node view. `data-children-of` itself stays: pure containers still use it to scope their round-trip parse rule.

**Validation** — `children` + any `content` other than `"none"` is now a schema-creation error with a message pointing at the string-prop pattern, keeping the door open to re-add content containers later without an API change.

**Tests & docs** — the contentContainers unit/browser/conversion suites and fixture, the content-container cases in the parse/React/insert-placement tests (a new test asserts the inline+children error), and the "Containers with their own content" / two-regions docs sections. The string-prop editable-title pattern is now the documented pattern.

## Untouched

Pure containers (`content: "none"` + `children`) work exactly as before: callout, the container-table example, and xl-multi-column (columnList/column) need no changes. React `contentRef` for containers keeps working.

## Verification

- `vp run lint` / `vp run format`: clean
- core unit tests: 838 passed
- xl-multi-column: 86 passed
- react: 4 passed
- all package browser suites (chromium/firefox/webkit): 186 passed
- `node docs/validate-links.mjs`: 0 errors


nperez0111:

PR #3007, opened 2026-08-25
refactor(core): simplify BlockInfo with derived positions and consolidated producers

nperez0111:
Part of the container-blocks stack: sits between #2997 (core container API) and #2998 (multi-column migration).

## What changed

**Field renames** — `BlockInfo` drops the internal jargon:

| Before | After |
|---|---|
| `bnBlock` | `block` |
| `blockContent` | `content` |
| `childContainer` | `children` |
| `isWrappedBlock` | `hasContent` |

(ProseMirror group name strings like `"bnBlock"` are unchanged — only the TS API is renamed.)

**Precomputed derived fields** — callers kept re-deriving the same positions with hand arithmetic (`blockContent.beforePos + 1` at 25+ sites, `afterPos - 1` at ~12, empty-inline-block checks at ~6). `BlockInfo` now carries them directly:

- `contentStart` / `contentEnd` — inside edges of the content node
- `children.childrenStart` / `children.childrenEnd` — inside edges of the child container
- `contentKind: "inline" | "none" | "table" | "other"` — replaces raw `spec.content` string checks
- `isContentEmpty`

These stay a discriminated union on `hasContent`, so the existing narrowing guards keep working.

**Producer consolidation** — 6 overlapping producers → 4, named by what you have; the module doc comment carries the decision table:

| You have | Call |
|---|---|
| a PM node + its before-pos | `getBlockInfoFromNode(node, beforePos)` |
| a position just before a block | `getBlockInfoAt(doc, posBeforeBlock)` |
| an arbitrary position | `getBlockInfoNearPos(source, pos)` |
| a selection | `getBlockInfoFromSelection(source)` |

`getBlockInfo` and `getBlockInfoFromResolvedPos` are deleted; `getBlockInfoWithManualOffset`, `getBlockInfoAtNearest`, and `getBottomNestedBlockInfo` are renamed (`getBlockInfoFromNode`, `getBlockInfoNearPos`, `getLastDescendantBlockInfo`).

**Navigation helpers moved** — `getParentBlockInfo` / `getPrevBlockInfo` / `getNextBlockInfo` / `getLastDescendantBlockInfo` move from `mergeBlocks.ts` to `getBlockInfoFromPos.ts` (they're generic navigation, not merge logic) and become public.

## Testing

[...]

PR #3010, opened 2026-08-25
refactor(core): simplify the BlockInfo API and make it the single vocabulary for block/children plumbing

nperez0111:
Top of the container-blocks stack, on top of #3014. Subsumes #3007 (now closed) — the BlockInfo simplification and the BlockInfo consolidation are one API-simplification change, so they ship as a single layer.

**Rebased onto the rebased base** (#3014 is now linear on current `main`, which landed the Typst PDF export stack). No conflicts: the commits carry identical changes, with context shifted only where `main` touched `docxExporter.ts`. Two commits are gone from this PR:

- `ci: retrigger checks` — an empty commit; the force-push retriggers CI on its own.
- `refactor(blocks): extract isDocumentFragment for SSR safety` — **moved down into #3014**, as an earlier revision of this description already said it should be. It fixes `containerRootDOM`, which the base PR introduces, and the base's `Build` job was red without it.

This PR is all about simplifying the block-plumbing APIs: fewer names, fewer producers, no hand arithmetic, one vocabulary. It is a pure refactor apart from the two behavior changes called out below.

## One vocabulary: `BlockInfo`

"Where do this block's children live?" was answered in ~8 different vocabularies (`BlockInfo.children`, write-target helpers, `fixContainer`'s private repair targets, raw `beforePos + 1` walks, structural slice-walk branches, …). This PR consolidates them down to two: `getBlockRegions` resolves block shape (container vs `blockContainer`) in one place, and `BlockInfo` is the position-annotated view everything downstream reads.

- **One shape resolver**: `getBlockRegions(node) → { outer, content?, childrenHolder? }`, consumed by `getBlockInfoFromNode`. No other code asks "which shape am I".
- **Deleted synonym vocabularies**: `childrenHolder.ts`, `ChildrenWriteTarget`, `fixContainer`'s private `{ blockPos, childrenStart }` repair targets, and the seal-variant descend trio (now one `descendToInsertionPos(info, nodeType, edge, opts)`).
- **Deleted helpers that were bare property reads**: `getChildrenConfig` (= `config.children`), `isContainerType` (= `config.children !== undefined`), `isPlaceableAnywhere` (= `config.placement !== "containerOnly"`), `isInsertableChild`, `flattenNonInsertableBlocks`, `seedRefillChildren`, `seedDefaultChildren`, `createContainerChildrenNode`, `createExplicitChildrenNode`; inlined `deleteBlockCollapsingSingletonGroup` and the table-caret `±4` arithmetic.

## Simpler `BlockInfo` shape

**Field renames** — drops the internal jargon (ProseMirror group name strings like `"bnBlock"` are unchanged; only the TS API is renamed):

| Before | After |
|---|---|
| `bnBlock` | `block` |
| `blockContent` | `content` |
| `childContainer` | `children` |
| `isWrappedBlock` | `hasContent` |

**Precomputed derived fields** — callers kept re-deriving the same positions with hand arithmetic (`blockContent.beforePos + 1` at 25+ sites, `afterPos - 1` at ~12, empty-inline-block checks at ~6). `BlockInfo` now carries them directly:

- `contentStart` / `contentEnd` — inside edges of the content node
- `children.childrenStart` / `children.childrenEnd` — inside edges of the child container
- `contentKind: "inline" | "plain" | "none" | "table" | "other"` — replaces raw `spec.content` string checks
- `isContentEmpty`

These stay a discriminated union on `hasContent`, so the existing narrowing guards keep working.

## Fewer producers

6 overlapping producers → 4, named by what you have; the module doc comment carries the decision table:

| You have | Call |
|---|---|
| a PM node + its before-pos | `getBlockInfoFromNode(node, beforePos)` |
| a position just before a block | `getBlockInfoAt(doc, posBeforeBlock)` |
| an arbitrary position | `getBlockInfoNearPos(source, pos)` |
| a selection | `getBlockInfoFromSelection(source)` |

`getBlockInfo` and `getBlockInfoFromResolvedPos` are deleted; `getBlockInfoWithManualOffset`, `getBlockInfoAtNearest`, and `getBottomNestedBlockInfo` are renamed (`getBlockInfoFromNode`, `getBlockInfoNearPos`, `getLastDescendantBlockInfo`).

## Navigation helpers moved & fixed

`getParentBlockInfo` / `getPrevBlockInfo` / `getNextBlockInfo` / `getLastDescendantBlockInfo` move from `mergeBlocks.ts` to `getBlockInfoFromPos.ts` (they're generic navigation, not merge logic) and become public.

**Behavior fix**: `getParentBlockInfo` now has block-model semantics — a block inside a column parents to the column, not the columnList. Fixes the Delete-at-end climb running its seal check on the wrong node for container children. Regression test added.

**Named positions over raw arithmetic**: `childrenStart/End`, `contentStart/End` replace the surviving `±1`/`nodeSize` sites; the Backspace-into-table caret now anchors on the actual previous block's content region, fixing a latent off-by-2 for nested tables.

## Callers get shorter

The point of the derived fields is that call sites stop doing structural work:

- `setSelection` drops its manual `TableMap.get` / `positionAt` table-edge arithmetic and its `content === "none"` checks for `blockEdgePos(info, "start" | "end")` (−40 lines).
- `setTextCursorPosition`'s 55-line content-type switch collapses into a single `blockEdgeSelection(...)` call (−45 lines).
- `canNestBlock` / `canUnnestBlock` now dry-run the real command (`editor.canExec(...)`) instead of re-deriving its preconditions by hand — the hand-written copies had already drifted from the command.
- `moveBlocks` gets a local recursive `dissolveContainerOnlyBlocks`, replacing the exported `flattenNonInsertableBlocks`.

## Schema validation reduced

`validateChildren.ts` goes from 392 → 115 lines and `assertSchemaInvariants.ts` is deleted outright. Both were re-checking things ProseMirror's own schema/content-expression machinery already enforces at node-creation time (`allow`-permits-nothing, unknown types in `allow`, `default` count vs min/max, `default` not permitted, containerOnly reachability, container-runs-before ordering, fillability).

What is kept is what ProseMirror can *not* tell you, or only tells you far from the cause: `max < min`, a regular block listed in `allow`, and cycles in the children graph.

## API rename

`insertBlocks` placements `"start"` / `"end"` → `"first-child"` / `"last-child"` (docs, tests, jsdoc updated). The cursor-placement `"start"` / `"end"` vocabulary of `setTextCursorPosition` is unrelated and unchanged.

## Testing

- Core unit suite: 852 passed | 9 skipped
- xl-multi-column: 87 passed | 5 skipped; react: 4 passed
- Exporters: typst 17, pdf 14, email 22, docx 10, odt 3; math-block 35, diagram-block 29
- `@blocknote/server-util`: 5 passed — the two SSR serialization failures noted in an earlier revision are fixed by the `isDocumentFragment` commit, now in the base PR
- e2e exporter suites in the repo's Linux docker image, chromium/firefox/webkit: 7 passed | 2 skipped
- Type-aware lint across the monorepo: clean

## Note on scope

An earlier revision of this branch also carried a rewrite of `KeyboardShortcutsExtension.ts`, a `blockToNode.ts` → `contentToNodes.ts` file split, a `ReactBlockSpec` node-view merge, and a SideMenu geometry cleanup. Those are unrelated to the block/children plumbing and have been dropped from this PR; only the mechanical adaptation of those files to the new `BlockInfo` API remains.


nperez0111:
Folded into #3014, which now carries the whole stack as one reviewable diff — GitHub marked this merged automatically once its commits became reachable from the base branch. `compare` reports the two branches `identical`, so nothing here is unreviewed; it all sits in #3014.

Reason for combining: reviewing the two separately cost ~2,900 lines of write-then-rewrite churn — this PR deleted 297 of the 392 lines `validateChildren.ts` was introduced with, and all 98 of `assertSchemaInvariants.ts`. As one PR the diff shows only the net result. The commits are preserved in #3014 if the intermediate reasoning is wanted.

PR #3014, opened 2026-08-26
feat(core): container blocks — core API, multi-column migration, docs & examples

nperez0111:
Recreates the container blocks stack, which was merged to `main` prematurely and has since been rolled back (`main` is back at 57190dab). GitHub does not allow reopening merged PRs, so this single PR replaces the four that made up the stack:

| original | title | review thread |
|---|---|---|
| #2997 | feat(core): container block API for nested blocks | still readable there |
| #2998 | feat(xl-multi-column): migrate columns onto the container block API | still readable there |
| #2999 | docs: container block docs and examples | still readable there |
| #3009 | refactor: remove content-container support | still readable there |

The commits are the same in content as what was on `main` (new SHAs — the original merge was a rebase-merge), apart from the rebase adaptations described below. #3010 is stacked on top of this branch.

**Rebased onto current `main`** (`f0d80aea0`), which landed the Typst-powered PDF export stack (#2992, #3020, #3021, #3022). The branch is linear on top of `main`, no merge commit. That rebase brought one conflict and one new exporter to support — see part 5.

**#3010 is folded in here** (parts 6 and 7). Reviewed separately, the two PRs cost ~2,900 lines of write-then-rewrite churn between them — #3010 deleted 297 of the 392 lines `validateChildren.ts` was introduced with, and all 98 of `assertSchemaInvariants.ts`. As one PR the diff shows only the net result; the commits are preserved if the intermediate reasoning is wanted.

---

## 1. Core: container block API (was #2997)

A first-class API for **container blocks**: custom blocks that hold other blocks as children, declared with a new `children` config on `BlockConfig`:

```ts
const Callout = createBlockSpec(
  {
    type: "callout",
    children: {
      allow: "any",          // or a list of block types / containers
      min: 1,                // structural minimum, maintained by repair
      default: [{ type: "paragraph" }],
      whenEmptied: "unwrap", // or "refill"
      boundary: "open",      // or "sealed"
    },
  },
  { render: ... },
);
```

- **Validation & schema invariants** (`validateChildren.ts`, `assertSchemaInvariants.ts`): child configs are checked at schema build time, with reachability checks for `placement: "containerOnly"` blocks.
- **Repair** (`fixContainer.ts`): removals that empty a container below `min` either unwrap it or refill it from `default`, applied by the block manipulation API and the keyboard handlers.
- **Generic keyboard behavior** (`KeyboardShortcutsExtension.ts`): the previously hardcoded columnList Backspace/Delete/Enter handlers are generalized to any container, driven by schema navigation (`containerNav.ts`) and the `boundary` config (`sealed` containers never leak or swallow content implicitly).
- **Serialization & parsing**: internal/external HTML round-trips, `data-children-of` markers so non-content UI text in a render never parses back as document content, `parse`/`parseContent`/`runsBefore` support for containers.
- **Block manipulation API**: `insertBlocks` placements (`"start"`/`"end"`), `updateBlock` conversions into/out of containers, container-aware `moveBlocks`/`nestBlock`/`mergeBlocks`.
- **UI**: side-menu handling for containers (`sideMenuContainerGeometry.ts`, `containerUI.ts`), React node-view support (`ReactBlockSpec`, `useNodeViewBlock`), `BlockPopover` fixes.
- New `@blocknote/core/internal` entry point for the container machinery that integrations need but that isn't public API.

## 2. Multi-column migration (was #2998)

Migrates `@blocknote/xl-multi-column` from hand-written ProseMirror nodes onto the container block API, and deletes the legacy compatibility shims the first part carried for it.

- `column` / `columnList` are now regular `createBlockSpec` container blocks (`pm-nodes/Column.ts` and `pm-nodes/ColumnList.ts` deleted):
  - `columnList`: `children: { allow: ["column"], min: 2, whenEmptied: "unwrap" }`
  - `column`: `placement: "containerOnly"`, so it can only ever live inside a `columnList`
  - both `meta.draggable: false`, matching the previous side-menu behavior
- Column keyboard behavior, repair, drag & drop, and HTML round-trips now come entirely from the generic container machinery; the bespoke column code paths in core are gone.
- Core cleanup: `fixColumnList.ts` deleted along with every `// Legacy` shim (`blockToNode`, internal HTML serializer, `UniqueID` types, `Exporter.isContainerBlock`, `containerUI`, `fragmentToBlocks`, `fixContainer`).
- Column resizing moves fully into `ColumnResizeExtension` (widths are no longer a schema prop concern of core).
- `insertBlocks` with a partial `columnList` now auto-fills from the container config instead of throwing, matching every other container block.

## 3. Docs & examples (was #2999)

- A new [Container Blocks](docs/content/docs/features/custom-schemas/container-blocks.mdx) feature page covering the `children` config, boundaries, repair, parsing, and exporter mappings; pointer updates to the Custom Blocks and Manipulating Blocks pages.
- **Example: Container Block** (`09-container-block`): a Notion-style callout that holds child blocks, plus the "string prop slot" pattern for its title.
- **Example: Table Built From Container Blocks** (`12-container-table`): BlockNote's table rebuilt as four container blocks (`table`/`tableRow`/`tableCell`/`tableHeader`) with `boundary: "sealed"` cells that can hold arbitrary blocks — no `prosemirror-tables` involved.

## 4. Content containers cut from v1 (was #3009)

#3009 was a decision artifact, and merging it took the decision: **content containers are cut from v1**. Containers are always `content: "none"`; an editable title/caption is a string prop rendered as an input.

That removal (34 files, +149 / −2,531) is included here:

- **Schema compilation**: `buildContentContainerNode`, the generated `<type>__content` / `<type>__children` nodes, the `containerContent` group, and the `extraNodes` plumbing are gone.
- **Predicates**: `isContentContainerNode` / `isContainerBlockNode` removed; every `isContainerNode(x) || isContentContainerNode(x)` disjunction collapses to the pure-container predicate.
- **Behavior**: the Enter-splits-content-head and Backspace-merges-first-child branches (`mergeIntoContainerContent`), plus the content-container arms of splitBlock/mergeBlocks/updateBlock and `blockToNode` / `nodeToBlock` / `fragmentToBlocks`.
- **DOM contract**: the `[data-content-type]` / `[data-children-of]` sibling-region rendering. `data-children-of` itself stays — pure containers still use it to scope their round-trip parse rule.
- **Validation**: containers are declared `content: "none"`; combining `children` with any other `content` is a schema-creation error. (Part 8 relaxed this to "ignored" and review reverted that: the relaxed state throws at first render, far from the config, while the public `Block` type still promises the content.) The door stays open to re-add content containers later without an API change.

Pure containers (`content: "none"` + `children`) are unaffected: callout, the container-table example, and xl-multi-column need no changes. React `contentRef` for containers keeps working.

## 5. Container support in the Typst & PDF exporters

`main` replaced the react-pdf PDF export with a Typst-backed one (`@blocknote/xl-typst-exporter` + `@blocknote/xl-pdf-exporter`, where `PDFExporter extends TypstExporter`). It hardcoded `columnList` layout in `transformBlocks` and left `column` / `columnList` as never-invoked stubs, so every other container fell into the generic path: its mapping was called with empty children, and the children were then appended *after* it, indented — the same bug the DOCX, ODT, react-pdf and email exporters were fixed for in part 1.

- `TypstExporter.transformBlocks` now recognizes containers via `isContainerBlock` and passes their rendered children to the mapping, which owns their placement. Because `PDFExporter` extends it and `xl-pdf-exporter` re-exports the Typst mappings, one custom-block Typst mapping serves both `.typ` and PDF export.
- `columnList` / `column` become real, overridable mappings. Typst takes a grid's track sizes on the grid rather than the cell, so a `column` returns the Typst *value* its parent needs — `(width: 1.4, body: [...])` — and `columnList` assembles the grid from those pairs in Typst code, keeping the fractional tracks proportional. Same technique the DOCX mapping already uses (a width-carrying table cell). The cell bodies are byte-identical to before; only the assembly around them changed.
- The email exporter's columns were the one merge conflict: `main` added hardcoded `columnList`/`column` flattening plus no-op mappings, this branch handles containers generically. Resolved in favor of the generic path, with the email `column`/`columnList` mappings now placing their children — `main`'s email snapshots pass unmodified, so the generic path produces byte-identical output to the hardcoded flattening.
- Docs: the container page's interop section now names the Typst and PDF exporters, and `export/typst.mdx` gains a "Container blocks" section (`export/pdf.mdx` already links there).

### Exporter schemas must contain the document's block types

Container detection is schema-driven (`Exporter.isContainerBlock` reads the block's spec), so an exporter constructed with a schema that lacks a container type in the document falls back to the generic path: the mapping is called with no children and the children are appended after it. Two test setups were doing exactly that with the shared test document, which contains columns — the email exporter's unit tests and `tests/.../exporterTestUtil.tsx`'s `schema()`. Both now include `ColumnBlock` / `ColumnListBlock`, matching every other exporter test in the repo.

This is worth knowing about as an API sharp edge: the failure is silent (correct-looking export, misplaced children), and in the Typst exporter's case the column mapping's non-content output leaks into the document as visible text. Making it a loud error is a follow-up worth considering.

### Also in this PR

`refactor(blocks): extract isDocumentFragment for SSR safety` moves down from #3010, where its description already said the fix belonged here: `containerRootDOM`'s `instanceof DocumentFragment` is introduced by this PR's container work and is not cross-realm safe under SSR, which was failing this branch's `@blocknote/server-util` tests (and the `Build` job) before the fix.

## 6. BlockInfo consolidation (was #3010)

One vocabulary for block/children plumbing: `getBlockRegions` resolves block shape in one place and `BlockInfo` is the position-annotated view everything downstream reads, replacing ~8 overlapping vocabularies. Field renames (`bnBlock` → `block`, `blockContent` → `content`, `childContainer` → `children`, `isWrappedBlock` → `hasContent`), precomputed derived positions (`contentStart`/`contentEnd`, `children.childrenStart`/`childrenEnd`, `contentKind`, `isContentEmpty`) replacing hand arithmetic at 40+ sites, and 6 producers down to 4. `getParentBlockInfo` gains block-model semantics (a block inside a column parents to the column). `validateChildren.ts` drops to 115 lines and `assertSchemaInvariants.ts` is deleted, both re-checking what ProseMirror's own content expressions already enforce. `insertBlocks` placements `"start"`/`"end"` → `"first-child"`/`"last-child"`.

Net effect: source deletions exceed additions across every subsystem it touches — the container API is the same size as before this layer, with fewer names.

## 7. Follow-up fixes

Container behaviour and API sharp edges found while reviewing the combined diff:

- **`Exporter.isContainerBlock` answered "not a container" for any type missing from the exporter's schema.** Silently wrong for a container: children get appended after it rather than placed by its mapping, and a mapping whose output is only meaningful to its parent (the Typst `column`'s `(width, body)` pair) lands in the document as visible text. It broke two test setups, each surfacing only as a confusing snapshot diff. It now takes the block: a *childless* block of an unknown type is unambiguous (nothing to place), which keeps working the pattern where a block package ships only a mapping that reads the block's JSON — math and diagram both rely on it. A block *with* children throws rather than guessing.
- **The last type-name special case in a generic path is gone**: the ODT exporter branched on `columnList`/`column` to reset nesting depth, calling them "legacy" though they are ordinary container blocks now. That reset belongs to every container — nesting depth renders there as literal `<text:tab>` indentation, which a mapping placing its own children governs itself.
- **`insertBlocks` validated only the first node's type**, so a batch whose first block fit and whose later blocks didn't got through and failed later inside `tr.step` with a ProseMirror-level message. The whole fragment is now checked where the nodes are known.
- **`updateBlock` carried every child into the new node**, so converting a block into a container that holds only containers built a node that failed `check()` and aborted the update. It now keeps the subset the new type can hold.
- **External HTML now marks the children region** (`data-children-of`) as the internal serializer does, so a container's parse rule can scope itself and a render's non-content UI text doesn't reparse as document content.
- **`BlockPopover`** searched for any `[data-node-type]` descendant, anchoring to a nested child container; the search is scoped to the block's own type. **`SideMenu`** no longer assumes a container has element children.
- `applyContainerAttributes(containerRootDOM(...), ...)` appeared at four call sites in `createSpec.ts`, two byte-identical — folded into one `markContainerRoot`.
- The side menu's pure rect arithmetic moved out of the browser test suite into a node test; the walks that measure live layout stay. (`containerParse.browser.test.ts` deliberately stays a browser test — it exercises HTML parsing, which validating against jsdom's parser would defeat.)

## 8. Rebased onto current `main`, and the fixes that came with it

Rebased onto `main` at `63c2389e3` (was `f0d80aea0`), linear, no merge commit.

One conflict: `main` landed #3039, which is the same `CustomBlockNoteSchema.extend()` fix the first commit below carried. Resolved in favour of `main`'s — that commit's `schema.ts` hunk is now empty, and the branch keeps only its `validateChildrenConfigs` call in that file.

New commits since the last review pass:

- **`refactor(core): read block facts from the spec that declares them`** — a pass walking the stack from the schema up to the side menu, lifting each "why" to the layer that owns it. `contentKind` is derived from the block spec's `content` instead of by string-matching the generated content expression, dropping both the `"other"` catch-all and the position layer's knowledge of `tableRow+`. `isContainerOnly` replaces two different ad-hoc answers to "can this stand outside its container?" in `moveBlocks` and `fragmentToBlocks`. `blockToNode`/`fragmentToBlocks` read the children config off the node spec, as `fixContainer` already did, so neither needs an editor to convert. A hand-written node (`createBlockSpecFromTiptapNode`) whose content expression contradicts its config is rejected once, when the schema is built.
- **`refactor(schema): consolidate container attribute application`** — `applyContainerAttributes` and `fillContainerAttributes` were the same function twice, differing only in whether an existing attribute wins; they become one with a `mode` (`"overwrite"` for a live node view, which must clear the attribute of a prop back at its default; `"fill"` for HTML serialization, which renders a fresh element and lets the render's own attribute stand). `getParseRules`' two-nested-ternaries-deep `getContent` becomes `containerChildrenParser` / `blockContentParser`. The `resolveChildren` memo cache and `isBlockGroupInsertable` are gone.
- **`fix(core): enable nesting containers with restricted child types`** — Tab and Shift-Tab across two columns did nothing. `sinkItem`/`liftItem` resolve their range at the *deepest* ancestor holding child blocks, and a `columnList` qualifies — but it holds only `column`s, so the range landed at the columns themselves, where `sinkItem`'s `nodeBefore` is a `column` and `liftItem`'s parent is a `blockGroup`; both bailed. The range predicate now also asks whether the candidate can hold the node being moved, so the walk continues out to the `blockGroup` the list sits in and the list indents as a unit. Two more container fixes rode along: external HTML strips `data-id` from a container's root (a container's `toExternalHTML` writes its ID the way its node view does, but external HTML carries no IDs — a regular block's is filtered out — so pasting back was duplicating IDs rather than minting fresh ones), and `children` combined with a non-`"none"` `content` is ignored rather than rejected (see the amendment below).
- **`refactor(core): read the previous block's content kind from BlockInfo`** — the Backspace handler's caret placement was the last site string-matching a generated content expression (`"tableRow+"`, `""`). It reads `BlockInfo.contentKind`.
- **`fix(react): re-stamp container attributes on every commit`** — the layout effect stamping BlockNote's attributes onto a React block's author-owned root element was keyed on `[block, selected]`, so a root the author's component swapped on a re-render of its own went unstamped. Stamping is cheap and idempotent, so it runs on every commit rather than trying to predict when the root changed.

### `children.max` is cut

`children` no longer takes a `max`. It was the one child-count control nothing in the repo used — no `columnList`, `callout`, or container-table cell declared one — and the only two things exercising it were its own tests. Its ceiling had to be re-enforced by hand in every path that adds a child (`insertBlocks`, `updateBlock`'s conversion filter), each an opportunity to get it wrong, and each one that was.

It is removed everywhere rather than deprecated: the config field, `ResolvedChildren.max`, the `{min,max}` quantifier arm of the compiled content expression, and the `max < min` validation. `min` is unchanged, and a container is still unbounded above — which is what every container in the repo already was. Should a real ceiling be wanted later, it can come back as an additive change.

### Amendment to part 4

Part 4 says `children` combined with any `content` other than `"none"` is a schema-creation error. Part 8 relaxed that to "ignored", on the grounds that the error only ever fired on a config whose `content` had no effect — but that doesn't hold: `nodeToBlock` throws at runtime (`Block "X" declares inline content but holds children`) the moment such a block renders, far from the config, and the public `Block` type still promises content it never has. Following review, the schema-creation error is back, centralized in `validateChildrenConfigs` so both `createBlockSpec` and `createBlockSpecFromTiptapNode` are covered. `checkNodeMatchesConfig`'s old container arm stays removed — it checked that a hand-written container node joined the container groups, which nothing in the repo builds — and its remaining arm, plus the node-name check now consolidated into it, still guards hand-written nodes.

## Verification

Re-run in full after the rebase onto `63c2389e3` and the `children.max` removal.

- `vp run lint` (type-aware): clean
- `vp run test`, all 17 tasks: **2,266 passed | 286 skipped**, zero failures. core 873 | 9 skipped; the shared `tests` package 937 | 2 skipped; xl-multi-column 94 | 5 skipped; react 4; server-util 5; exporters typst 17, pdf 14, email 22, docx 10, odt 3; math-block 35, diagram-block 29
- the full browser + e2e suite in the repo's Linux docker image (`tests/docker-run.sh`), chromium/firefox/webkit: **1,016 passed | 52 skipped** across 174 files, including the `typstPdfImages` per-page visual snapshot, which matches `main`'s committed chromium-linux images untouched
- `keyboardhandlers > Check Delete before shallower block`, previously noted as flaky under 3-browser parallel load, passed in all three browsers on this run
- new Typst container tests: the column grid keeps proportional tracks and compiles against the real Typst compiler; a custom container's children render *inside* its output; an unmapped container throws
- The new `columnList` markup was verified to be render-identical, not just plausible: `main`'s `#grid(...)` form and the new `#{ let cols = …; grid(…) }` form compile to byte-identical PDFs
- The `children.max` removal was folded into the commits that introduced it rather than appended, so no commit in this PR adds `max`. To make that safe the removal was built and tested at the tip first and its tree hash recorded; the rebuilt branch's final tree is byte-identical to the tested one.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom container blocks with nested children, configurable boundaries, defaults, validation, and placement rules.
  * Added interactive Callout and container-based table examples.
  * Added container-aware HTML, Markdown, DOCX, ODT, email, PDF, and Typst export support.
  * Added first-child and last-child insertion options.
* **Bug Fixes**
  * Improved editing, selection, movement, merging, splitting, dragging, and side-menu behavior in nested and sealed containers.
* **Documentation**
  * Expanded container-block and exporter guides.
* **Tests**
  * Added coverage for container behavior, parsing, exports, keyboard interactions, and multi-column editing.


docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+  {
+    // Child blocks mount into the element you attach `contentRef` to.
+    render: (props) => <div className="callout" ref={props.contentRef} />,
+  },
+);
+```
+
+`children: { allow: "any" }` accepts any block, requires at least one, and never throws. When a container is created without children, BlockNote fills it with whatever its schema requires.
+
+A container block always declares `content: "none"`: its body is its children. Any other `content` is ignored — the node is built from `children` alone. For an editable title or caption, use a string prop rendered as an `<input>`, as the demo below does.
```
YousefED, line 34: do we throw when content is passed (either to the block instance or a different value than `none` is passed to the schema?)

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+      "id": "para-1",
+      "type": "paragraph",
+      "content": [{ "type": "text", "text": "Hello", "styles": {} }],
+      "children": []
+    }
+  ]
+}
+```
+
+### Where children render
```
YousefED, line 54: I'm not sure about whether all of this section is relevant for consumers. Feels like we're mixing API explanation and internals a bit

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+<Callout type="warn">
+  _`whenEmptied` never destroys typed text: only empty children are dropped._
+</Callout>
+
+## Defaults and refilling
+
+`default` is an insertion template: a container inserted without an explicit `children` array is created with those blocks. Omit it and BlockNote fills the container with empty blocks its schema accepts.
+
+The same template drives `whenEmptied: "refill"`. When a refill container's non-empty children drop below `min`, say `k` remain, BlockNote appends `default[k]` through `default[min - 1]` at the end, falling back to empty blocks where `default` is absent or has no entry for a position. A checklist with `min: 2` and a two-entry `default` that loses its second item gets `default[1]` back, not a bare paragraph.
```
YousefED, line 108: this feels like an overly complex explanation (omit or simplify?)

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+## Boundaries
+
+`boundary` declares whether editing gestures cross a container's edge. On an open edge they move blocks across it: Backspace at the start of the first child moves that child out, and Enter on an empty last child escapes below the container. A sealed edge blocks all of that, so the container behaves as a single unit.
```
YousefED, line 112: should we be able to escape with shift+enter just like in code blocks?

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+| Value | Crosses the edge | Use for |
+| --- | --- | --- |
+| `"open"` (default) | The caret and editing gestures. | Containers that are part of the surrounding flow of text, like a callout or the columns of a `columnList`. |
+| `"sealed"` | Nothing implicitly. The caret won't wander in, and from outside the container selects and deletes as one unit. | Compartments that should stay put, like a table cell. |
```
YousefED, line 117: do we have other examples than a table cell? Conflating container blocks with tables is confusing as long as tables are not built on top of container blocks

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+`allow` takes one of four forms:
+
+```typescript
+allow: "any" | "blocks" | "containers" | string[]
+```
+
+- `"any"`: any regular block, plus any container placeable anywhere.
+- `"blocks"`: regular blocks only, no containers.
+- `"containers"`: any anywhere-placeable container, no regular blocks.
```
YousefED, line 139: when would "containers" be useful"?

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+- `string[]`: only the named container block types.
+
+The wildcard forms (`"any"`, `"containers"`) exclude `placement: "containerOnly"` types: a `column` never shows up inside your callout just because the callout accepts "any" block. A containerOnly type appears only where a parent names it in an array.
```
YousefED, line 142: references to `column` and `columnList` throughout this doc are a bit confusing. They assume the reader understands the difference between both and how the multi-column package works

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+The array form is exact because each container block type is distinct in the schema, while every regular block (paragraph, heading, code block) shares one underlying type. So "only headings" is not something the schema can enforce yet. Naming a regular block type in the array is a startup error; per-type filtering of regular blocks is not yet supported, and the array is where it will land later with no API change.
```
YousefED, line 144: this seems like implementation explanation / excuse, I don't think it adds anything as long as the API explanation  already makes clear that you can only name specific container block types

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+// Siblings of the reference block:
+editor.insertBlocks([{ type: "paragraph" }], calloutId, "before");
+editor.insertBlocks([{ type: "paragraph" }], calloutId, "after");
+
+// Nested inside it, as its first or last child:
+editor.insertBlocks([{ type: "paragraph" }], calloutId, "first-child");
+editor.insertBlocks([{ type: "paragraph" }], calloutId, "last-child");
+```
+
+The nested placements are what addresses a container with no children to point at. A `min: 0` container that is currently empty has no child block to insert before or after. Whether a block fits is answered by the schema, so it's your `children` config that decides.
```
YousefED, line 177: this paragraph is confusing: 

> The nested placements are what addresses a container with no children to point at.

then why would you need the option of both first / last child?

> so it's your `children` config that decides.

not sure what this sentence means

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+## Validation
```
YousefED, line 179: is this relevant to consumers, or an implementation detail and consumers should just see a descriptive error message as soon as they configure something wrong?

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+- a `placement: "containerOnly"` block that no container's `allow` array names, or `placement: "containerOnly"` on a regular block;
+- container cycles: a container that (transitively) requires a child that requires it back could never be created. An `allow` that permits regular blocks breaks the cycle, since they're always satisfiable. `allow: "containers"` with `min: 1` is the same problem, since the container counts as a container itself.
+
+Documents are checked too. `initialContent` that doesn't fit the schema throws when the editor is created, rather than loading in a broken state. This matters when you change a `children` config on a schema whose documents are already saved somewhere: a stored document that no longer fits, say a `columnList` left with a single column, now fails at load. Migrate those documents before shipping the change.
+
+## Parsing HTML into a container
+
+Containers parse like any other custom block. The default rule matches `[data-node-type="<type>"]` so BlockNote's own HTML round-trips, and `implementation.parse` recognizes foreign HTML. Both work exactly as described for [custom blocks](/docs/features/custom-schemas/custom-blocks).
```
YousefED, line 193: if both work as described in custom-blocks, do we need a separate section for this?

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+<Callout type="warn">
+  _`allow` does not filter what a user pastes. Content your container rejects is
+  placed after the container rather than dropped. `allow` constrains the
+  document model, not the parser._
+</Callout>
+
+## Interop behavior
+
+Containers serialize to a `<div data-node-type="...">` with their children nested inside, and round-trip losslessly. For lossy targets you place the children yourself: return a `childrenDOM` from `toExternalHTML` (this is how toggles export as `<details>`), and give container blocks an explicit mapping in the DOCX, ODT, email, Typst, and PDF exporters, which throw on a missing one. That mapping receives the container's rendered children as its last argument and decides where they go — the exporters do not append them after the container's own output. Markdown flattens containers, exporting their children in order.
```
YousefED, line 205: `data-node-type` is an implementation detail that I don't think we should surface to the user (we don't do it anywhere else, so I don't see why we should do it here)

YousefED (commented):
first feedback, docs only (TLDR; maybe we can shorten it and strip some implementation details?)

packages/core/src/pm-nodes/README.md
```diff
@@ -99,6 +99,10 @@ We use Prosemirror "groups" to help organize this schema. Here is a list of the
 
 _Note that the last two groups, `bnBlock` and `childContainer`, are not used anywhere in the schema. They are however helpful while programming. For example, we can check whether a node is a `bnBlock`, and then we know it corresponds to a BlockNote Block. Or, we can check whether a node is a `childContainer`, and then we know it's a container of a BlockNote Block's `children`. See `getBlockInfoFromPos` for an example of how this is used._
 
+## Relation to container blocks
```
YousefED, line 102: not sure this is useful as is. Better to revise the entire document to also explain container nodes?

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * How few children the container may hold. Compiled into its content
+   * expression, so ProseMirror refills it on the way down; `whenEmptied`
+   * decides what happens when it can't be met.
+   */
+  min: number;
+  /** What to seed a fresh or refilled container with, if not a paragraph. */
+  default: readonly ChildDefault[] | undefined;
+  /**
+   * What to do with a container that lost its last children: put `default`
```
YousefED, line 27: isn't this the other way around? i.e.: column uses unwrap, and callout refill?

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+   * back (a `column`, which has to keep existing), or replace the container
+   * with them (a `callout` the user emptied, which should get out of the way).
+   */
+  whenEmptied: "refill" | "unwrap";
+  /**
+   * Whether editing gestures may cross the container's edge. A `column` is
+   * `"open"`: backspace at its start merges into the block above. A table
+   * cell is `"sealed"`: nothing implicitly moves in or out of it.
+   */
+  boundary: "open" | "sealed";
```
YousefED, line 37: note to self; I want to review how well this generalizes. which operations does "sealed" block and which not? Should this always be a single "takes-it-all" configuration?

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+};
+
+export const CHILD_CONTAINER_GROUP = "childContainer";
```
YousefED, line 40: we still have references in the codebase to the string "childContainer", replace with const

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+
+export const BLOCK_GROUP_CHILD_GROUP = "blockGroupChild";
+
+// Joined by every container block placeable anywhere (`placement` other than
+// `"containerOnly"`). It's what the `allow` container wildcards (`"any"`,
+// `"containers"`) compile to: a containerOnly type only ever lives where a
+// container names it explicitly, so it stays out of the group.
+export const ANY_CONTAINER_GROUP = "anyContainer";
```
YousefED, line 48: what's the use of this from a dev perspective? can we come up with scenarios where this is relevant? or premature?

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+
+// Whether `type` is a node that holds child blocks directly: a container
+// block's own node. A container is a child-holding node that is itself a
+// block; `blockGroup` also holds children but is not a block (it's regular
+// blocks' nesting machinery), so the `bnBlock` check excludes it.
+export function isContainerNode(type: NodeType): boolean {
```
YousefED, line 54: maybe a more descriptive name would be `isContainerBlockNode`?

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+ * defined only in terms of the container that holds it (a `column`), so it can
+ * never stand where a regular block goes.
+ *
+ * The schema encodes this by keeping such types out of
+ * `BLOCK_GROUP_CHILD_GROUP`, which is how ProseMirror enforces it while
+ * matching content. This answers the same question from the declaration
+ * itself, for code reasoning about the block rather than about what PM will
+ * match.
+ */
+export function isContainerOnly(type: NodeType): boolean {
```
YousefED, line 137: (see above; is this useful?)

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+  const steps = Math.round((priority - DEFAULT_SPEC_PRIORITY) / 10);
+
+  return Math.min(
+    CONTAINER_PRIORITY_BAND.max,
+    Math.max(CONTAINER_PRIORITY_BAND.min, CONTAINER_NODE_PRIORITY + steps),
+  );
+}
+
+export function resolveChildren(children: ChildrenConfig): ResolvedChildren {
+  return {
```
YousefED, line 167: throw an error if min === 0 and `whenEmptied` is specified? Or is `min: 0` not allowed? (in that case, throw an error when min === 0?)

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (terms.length === 0) {
+      throw new Error(
+        "Container `allow` permits nothing. A container must accept at least one block or container type; drop `children` entirely for a block that holds none.",
+      );
+    }
+
+    allowed = terms.length === 1 ? terms[0] : `(${terms.join(" | ")})`;
+  }
+
+  return allowed + (min === 0 ? "*" : min === 1 ? "+" : `{${min},}`);
```
YousefED, line 252: tbd, do we want to support min 0 (see above)

packages/core/src/api/exporters/html/util/serializeBlocksInternalHTML.ts
```diff
@@ -170,9 +175,27 @@ function serializeBlock<
     ret.contentDOM.appendChild(ic);
   }
 
-  const pmType = editor.pmSchema.nodes[block.type as any];
+  if (isContainer) {
```
YousefED, line 178: Not following this entirely;

- line 176+ in the old file don't have an equivalent anymore for regular blocks as far as I see? (it's now blocked by `isContainer === false`?
- why is `applyContainerAttributes` necessary? We didn't have it for columns for example. When I remove that call, all unit tests still pass

packages/core/src/api/exporters/html/util/serializeBlocksInternalHTML.ts
```diff
@@ -159,6 +161,9 @@ function serializeBlock<
     editor as any,
   );
 
+  const blockConfig = editor.schema.blockSchema[block.type as any];
+  const isContainer = blockConfig.children !== undefined;
```
YousefED, line 165: we have a helper for this now, right?

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -256,26 +263,57 @@ function serializeBlock<
...
   } else {
+    // Asked of the block config rather than of its ProseMirror node. See the
+    // same check in `serializeBlocksInternalHTML`.
+    if (editor.schema.blockSchema[block.type as any].children !== undefined) {
+      // Container blocks own their outer DOM. Make sure the attributes
+      // needed to parse the HTML back (the type marker and non-default
+      // props, in the same `data-*` convention `propsToAttributes` reads)
+      // are present even when the block's render didn't add them.
+      // Author-set attributes win.
+      applyContainerAttributes(
```
YousefED, line 292: when I remove this line, unit tests still pass, let's validate why / if it's necessary (also see internalhtml)

packages/core/src/schema/blocks/containerAttributes.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // Emit the reserved markers after the prop loop so a prop whose name
+  // kebab-cases to `data-node-type` or `data-id` can't overwrite them.
+  attributes["data-node-type"] = blockType;
+  if (id) {
+    attributes["data-id"] = id;
+  }
+
+  if (mode === "fill") {
```
YousefED, line 48: Curious if this is needed, see comments in HTML serializers

packages/core/src/schema/blocks/containerParse.browser.test.ts
```diff
@@ -0,0 +1,357 @@
+import { Fragment } from "prosemirror-model";
```
YousefED, line 1: this is not colocated to a source file like our regular unit tests rule. If I'm not mistaken, it's tests behavior that's currently in the /tests/unit directory. Can we fold it into the existing test setup?

packages/core/src/schema/blocks/types.ts
```diff
@@ -106,8 +209,44 @@ export interface BlockConfig<
...
+   * - `"containerOnly"`: only inside a container that names this type in its
+   *   `children.allow` array (e.g. a `column` inside a `columnList`).
+   *
+   * Only meaningful for container blocks; regular blocks are always placeable
+   * anywhere.
+   */
+  placement?: "anywhere" | "containerOnly";
+}
+
+declare module "prosemirror-model" {
```
YousefED, line 240: Do we need this? Type augmentations like this are a bit hacky imo. Can't we read the details from the BlockNote schema?

packages/core/src/schema/blocks/internal.ts
```diff
@@ -244,6 +283,10 @@ export function createBlockSpecFromTiptapNode<
       type: config.type as T["type"],
       content: config.content,
       propSchema,
+      ...(config.children !== undefined ? { children: config.children } : {}),
```
YousefED, line 286: as a safety guard, maybe throw an error here if `config.children` is defined, but `content !== "none"`?

packages/core/src/schema/blocks/types.ts
```diff
@@ -590,19 +733,31 @@ export type BlockImplementation<
...
     contentDOM?: HTMLElement;
+    /**
+     * The block author's own root element, when it isn't `dom` itself. React
+     * renders a node view through wrapper elements of its own, so the element
+     * ProseMirror is handed is not the one the author wrote. This points at
+     * the author's element, which container attributes (`data-node-type`,
+     * `data-id`, prop `data-*`) are applied to.
+     * @default dom
+     */
+    rootDOM?: HTMLElement | null;
```
YousefED, line 744: TBD: Still investigating, but at first sight I find it difficult to completely understand what this is needed for

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -81,223 +261,374 @@ export function getParseRules<
...
+    selectable: blockImplementation.meta?.selectable ?? true,
+    // Deliberately not `isolating`, not even for a sealed container. PM only
+    // honours that flag while no selection spans the edge, and nothing stops
+    // one being made: given a spanning slice, `Fitter` refuses to open into
+    // the container and wraps the content in a spurious `blockGroup` instead,
+    // so a copy-paste across the edge corrupts the document. Seals bind
+    // editing gestures, and those are enforced by BlockNote's own `isSealed`
+    // guards (see `containerNav.ts` and `KeyboardShortcutsExtension.ts`),
+    // which need no help from the schema.
+    defining: true,
```
YousefED, line 301: double check, we never need `isolating` on container nodes?

Let's throw errors if any unsupported `meta` fields are set?

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -81,223 +261,374 @@ export function getParseRules<
...
+          blockImplementation as unknown as BlockImplementation<
+            TName,
+            TProps,
+            "none"
+          >,
+          priority,
+        )
+      : buildRegularNode(blockConfig, blockImplementation, priority);
+
+  if (builtNode.name !== blockConfig.type) {
```
YousefED, line 614: could be moved to `checkNodeMatchesConfig`

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -81,223 +261,374 @@ export function getParseRules<
...
+ * Checks a hand-written node's content expression against the `content` its
+ * spec declares — the one `getBlockInfoFromPos` reports as the block's
+ * `contentKind`, without looking at the node. A generated node's expression
+ * comes from that same config, so this only bites on a hand-written one
+ * (`createBlockSpecFromTiptapNode`).
+ */
+function checkNodeMatchesConfig(node: Node, blockConfig: BlockConfig) {
+  // A container's node holds blocks rather than content, and its expression
+  // comes from `children` instead.
+  if (blockConfig.children !== undefined) {
```
YousefED, line 564: check that `content` === `none`?

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -81,223 +261,374 @@ export function getParseRules<
...
+}
+
+export function containerRootDOM(output: {
+  dom: HTMLElement | DocumentFragment;
+  rootDOM?: HTMLElement | null;
+}): HTMLElement | null {
+  if (output.rootDOM !== undefined) {
+    return output.rootDOM;
+  }
+  if (isDocumentFragment(output.dom)) {
```
YousefED, line 339: does it make sense to have a container with a `DocumentFragment` as it's `dom`? Because as far as I understand, containers don't have a generated-wrapper, which would mean there's no wrapper element of they'd specifify a DocumentFragment only?

Is this covered in tests / scenarios where this is actually used?

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -81,223 +261,374 @@ export function getParseRules<
...
+  const update = typedNodeView.update?.bind(typedNodeView);
+  if (update) {
+    typedNodeView.update = (node, decorations, innerDecorations) => {
+      if (node.type.name !== blockConfig.type) {
+        return false;
+      }
+      if (update(node, decorations, innerDecorations) === false) {
+        return false;
+      }
+      applyContainerAttributes(
```
YousefED, line 411: The whole attribute handling (also non-container nodes) across react / core can be streamlined a lot I think, and maybe container attributes could be part of that.

But maybe better to keep as a follow up

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -45,22 +57,190 @@ export function applyNonSelectableBlockFix(nodeView: NodeView, editor: Editor) {
...
+
+  return (node, schema) => {
+    const parsed =
+      parseContent({ el: node as HTMLElement, schema }) ??
+      DOMParser.fromSchema(schema).parse(node as HTMLElement, {
+        topNode: schema.nodes["blockGroup"].create(),
+        preserveWhitespace: true,
+      }).content;
+
+    // A container holds blocks, so any inline run `parseContent` returned is
```
YousefED, line 112: when I just do `return parsed;` here everything still passes. So either the stuff below is not necessary, or we need to cover it in tests

YousefED (commented):
This time reviewed `/core/src/schema/*`

docs/content/docs/features/custom-schemas/container-blocks.mdx
YousefED: Zoomed out; I think it might be useful to have examples of a `sealed` container outside of the Table example (I don't think we're not really migrating tables to this?)

If we can't come up with good examples, we should not expose the API imo.

examples/06-custom-schema/12-container-table/src/Table.tsx
```diff
@@ -0,0 +1,346 @@
...
+// no explicit children seeds it from `children.default`: a header row plus
+// two body rows, three columns wide.
+export const createTable = createReactBlockSpec(
+  {
+    type: "table",
+    propSchema: {},
+    content: "none",
+    children: {
+      allow: ["tableRow"],
+      default: [
```
YousefED, line 239: Do we know use-cases where setting `default` is actually useful (instead of relying on the underlying pm schema default)?

In this example, I'd argue that this logic (a 3x3 table) should be part of the place where you insert a table (e.g.: slash menu)

(Trying to minimize the API surface to prevent us from having to support "nice to have, but not sure anyone really needs it" features in the long term - until proven otherwise)

packages/xl-multi-column/src/blocks/Columns/index.ts
```diff
@@ -1,28 +1,82 @@
...
-    node: Column,
-    type: "column",
+    type: "column" as const,
+    propSchema: {
+      width: {
+        default: COLUMN_WIDTH_DEFAULT,
+      },
+    },
     content: "none",
+    children: { allow: "any" },
```
YousefED, line 17: I think this should be allow: "blocks"

The previous implementation had `"blockContainer+"`, which means you intentionally couldn't have a `ColumnList` inside `Column`.

This would mean however you can't have a Callout inside a Column now either, which might be a bit of an issue in the API design (low-prio, I'd argue)

packages/core/src/api/blockManipulation/containers/fixContainer.ts
```diff
@@ -0,0 +1,287 @@
...
+  containerPos: number,
+  type: NodeType,
+  config: ResolvedChildren,
+) {
+  // Emptied children are dropped unconditionally, even when the container
+  // sits at or above `min` afterwards: for an unwrap container (a
+  // columnList), a child the user emptied is done for — an emptied third
+  // column disappears rather than lingering. This deliberately differs from
+  // `refillContainer`, which leaves empty children alone at or above `min`.
+  removeEmptyChildren(tr, containerPos);
```
YousefED, line 133: (co-op with Claude):

I'm confused that an *ancestor's* `whenEmptied: "unwrap"` decides what happens to nested blocks. A column that holds only an empty paragraph gets deleted here because its *parent* (`columnList`) is `unwrap`; the column's own policy (`refill`, the default) never gets a say. So `whenEmptied` really means two things: what a container does to its own emptied children, and what happens to the container itself below `min`. Only the second is documented, and I'd expect a block's fate to follow its own config, not its parent's. Is this intended, and if so shouldn't it be a separate, named option?

---

Side note, related cases that break (both also break on `main`'s `fixColumnList`, so inherited, but this generalizes them to every `unwrap` container):

- `replaceBlocks(["c2"], [<column with one empty paragraph>])` in a 3-column list: the replacement column is deleted by the same call that inserted it.
- `removeBlocks(["b"])` in column 3 while column 2 holds only an empty paragraph: column 2 is swept away too, the list is left with one column and unwraps entirely.

examples/06-custom-schema/09-container-block/src/Callout.tsx
```diff
@@ -0,0 +1,104 @@
...
+      title: {
+        default: "",
+      },
+    },
+    content: "none",
+    // `children: { allow: "any" }` is the entire container declaration: any
+    // block is allowed, at least one is required, and BlockNote fills the
+    // callout with an empty paragraph when it's created. `min` / `default` /
+    // `whenEmptied` / `boundary` tune this.
+    children: { allow: "any" },
```
YousefED, line 39: UX-wise, I think this should have `whenEmptied: "unwrap"` (to match Notion for example).

Which brings me to the follow-up question: is `whenEmptied: "refill"` (the default) actually useful? Can we come up with examples where this is desired?

examples/06-custom-schema/09-container-block/src/Callout.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+// The Callout block. Declared with `content: "none"` plus the `children`
+// config: the block hosts arbitrary child blocks in its body, exposed at
+// runtime as `block.children`.
+//
+// The callout's title shows a related pattern: content that shouldn't be
+// part of the rich-text document (no formatting, comments, or multiplayer
+// cursors needed) can live in a plain string prop, edited through a regular
+// <input> rendered inside the block.
+export const createCallout = createReactBlockSpec(
```
YousefED, line 22: We now have Alert (and ReactAlert) and Callout. Can we consolidate this? (e.g.: name this Alert Container and keep rest of code similar to Alert? maybe extend Alert with a title as well)

packages/core/src/schema/blocks/types.ts
YousefED: If I'm not mistaken, one of the main reasons for this is to prevent an extra level of dom wrappers around React node views, specifically?

Although I don't like the nesting layers for regular React Custom Blocks, I maybe rather have React Container blocks "suffer" from the same issue vs. adding an extra `RootDOM` option that's only relevant for container blocks (and thus adds another code path, so to say

Then again, I might be overlooking some things here

PR #3016, opened 2026-08-27
feat: add a dedicated alt text field to the image block (#2822)

### LINKED ISSUE #2822: Missing alternative text field for images
### What’s broken?

When inserting an image in the Docs editor, no field allows providing alternative text. Only a caption is offered, which does not substitute for alternative text.

A dedicated alternative text field, distinct from the caption, must be available in the image editing panel. Its value must be reflected in the alt attribute of the generated  element.

### Steps to reproduce

Open a Docs document
Insert an image via / or the toolbar
Open the image editing panel
Notice that no alternative text field is available, only a caption field
Possible solution

Add an "Alternative text" field in the image editing panel, distinct from the caption field.
Additional context / Screenshots

The caption is visible content associated with the image; it does not replace alternative text, which is transmitted to assistive technologies and must describe the content or function of the image.

WCAG Criteria

[Criterion 1.1: Does each informative image have alternative text?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#1.1)

Impact

A blind user navigating with a screen reader inserts an image and has no way to provide alternative text. The image will be rendered without accessible description for users relying on assistive technologies.

### BlockNote version

0.51.4

mmaxence:
Closes #2822.

Follows the acceptance criteria from the issue: alt text is stored on the image block as an optional prop, there is a toolbar button to add or edit it, and it controls the `alt` attribute both in the editor and on HTML export.

Until now the image block reused the file name as the `alt` attribute, so there was no way to set a real accessibility label. This adds a dedicated `alt` prop that mirrors how `caption` is already plumbed through the block.

What changed:
- New optional `alt` prop on the image block (defaults to `""`).
- In-editor render and external HTML export now use `alt || name || ""`, keeping the file-name fallback so existing content is unaffected.
- `parseImageElement` reads the `` `alt` attribute into the prop, so alt text survives paste and round-trips.
- New "Edit alt text" formatting-toolbar button next to the caption button, with the dictionary key added to every locale.
- The button uses a small custom "ALT" icon (inline SVG paths, `currentColor`), since there is no fitting alt-text glyph in the icon set.

The button only shows for blocks that expose an `alt` prop, which in the default schema is images.

Verification: full-monorepo typecheck passes, and the `core`, `react`, and `server-util` test suites pass with the image-block snapshots regenerated.

## Summary by CodeRabbit

* **New Features**
  * Added support for editing image and file alternative text directly from the formatting toolbar.
  * Image alternative text is preserved when importing, rendering, and exporting content.
  * Added localized alternative-text controls across supported languages.

* **Accessibility**
  * Images now use explicitly provided alternative text, with sensible fallback behavior when none is supplied.
  * Improved support for creating descriptive text for images and files.


mmaxence:
Pushed 4cbb03c to fix the HTML round-trip issue: `parseImageElement` was reading both `name` and `alt` from the single `` attribute, while export writes `alt || name` into it, so setting alt text and round-tripping replaced the file name with the alt text. The parser now reads the `alt` attribute into `alt` only, and `name` is preserved through the block serialization. I updated the one affected markdown round-trip snapshot, which previously encoded the bug (`name` had become the alt text).

On the alt-only-block point: the button is intentionally shown only for blocks exposing both `url` and `alt`, which in the default schema is images. Happy to broaden that if you'd prefer it to cover alt-only blocks too.

PR #3028, opened 2026-08-31
feat(core,react): public focus API with editor-UI-aware tracking

YousefED:
First layer of a 4-PR stack (focus API → test infra → link popover → Android Enter). Together, the stack supersedes #3025.

## What

`editor.isFocused()` and `onFocusChange` only saw the content area, so focus moving into the editor's own UI — a toolbar popover's input — read as a blur. Fine for a desktop toolbar that unmounts anyway; the mobile toolbar has to stay up while the user types a URL into the popover it opened.

- `includeEditorUI` option on `isFocused()` / `onFocusChange()`: treats everything portalled into `editor.portalElement` (and siblings of the content area) as part of the editor, and defers the blur decision until focus has settled — at `focusout` time `document.activeElement` reads as `<body>`, so the destination isn't knowable yet.
- `useEditorFocus` (state, via `useSyncExternalStore`) and `useEditorFocusChange` (side effect) — the same split as `useEditorState` vs `useEditorChange`.
- `MobileFormattingToolbarController` drops its 30-line private reach into `editor._tiptapEditor` for one hook call.

## Behaviour notes for review

- The new hooks hold their callback in a ref; `useEditorChange` / `useEditorSelectionChange` are converted to the same latest-ref pattern for consistency. They no longer resubscribe when the callback identity changes — the latest callback is simply invoked. Typed consumers can't observe a difference; `useEditorSelectionChange` keeps forwarding the (undocumented) editor argument so untyped callers don't break.
- The focus unsubscribe is reference-counted and now idempotent — a double unsubscribe used to permanently kill tracking for all later subscribers (proven red-first in the regression test).

## Tests

`EventManager.browser.test.ts` (12 tests × 3 engines) pins the DOM contract this rests on — the documented focus event order, `<body>` during focusout — plus dedupe, multi-editor independence, and unsubscribe semantics. Sabotage-checked: breaking the tracker's dedupe fails 2 tests on all 3 engines. `useEditorFocus.browser.test.tsx` (colocated with the hooks) covers them (15 tests).

## Summary by CodeRabbit

- **New Features**
  - Added focus tracking for editor content and registered portalled UI.
  - Added the `useEditorFocus` React hook and public focus configuration options.
  - Added portal registration APIs and improved body-level mobile formatting toolbars.

- **Bug Fixes**
  - React editor hooks now use the latest callbacks without unnecessary resubscriptions.
  - Focus transitions between editor content and portalled UI are tracked more reliably.

- **API Changes**
  - Replaced direct portal-element access with explicit registration and unregistration.
  - Removed the mobile toolbar portal context and `useEditorFocusChange` exports.


tests/src/end-to-end/focus/useEditorFocus.test.tsx
```diff
@@ -0,0 +1,196 @@
+import {
```
YousefED: this is more a unit test (tests a specific function) then an e2e test. see if we can change this to either a pure unit tests, or if it needs the browser, a browser unit test (next-to-source-located .browser.test.tsx)

YousefED: Moved to `packages/react/src/hooks/useEditorFocus.browser.test.tsx` in 38328449a (git tracks it as a rename). It does need a real browser — settled-focus semantics and `document.activeElement` transitions aren't reproducible in jsdom — so it's a colocated browser unit test now, running in the same suite as `EventManager.browser.test.ts`. Porting required dropping the mantine dependency (react can't depend on a skin): it renders `BlockNoteViewRaw` with plain react-dom instead. Still 15 tests across the three engines.

packages/core/src/editor/BlockNoteEditor.ts
```diff
@@ -805,11 +815,17 @@ export class BlockNoteEditor<
...
       return false;
     }
-    return this.prosemirrorView?.hasFocus() || false;
+    const contentFocused = this.prosemirrorView?.hasFocus() || false;
+    if (!options?.includeEditorUI) {
+      return contentFocused;
+    }
+    const active =
+      typeof document !== "undefined" ? document.activeElement : null;
+    return contentFocused || (!!active && this.isWithinEditor(active));
```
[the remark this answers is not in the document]
YousefED: Fixed in ecb085e2a: when the mount element's parent is `<body>`/`<html>`, the boundary falls back to the content element itself, so a bare-body mount no longer classifies the whole page as editor UI. `BlockNoteView` setups keep the parent boundary (needed so UI rendered as siblings of the content counts). Covered by a new case in `EventManager.browser.test.ts` asserting an outside input isn't within a body-mounted editor and that `includeEditorUI` focus blurs correctly there.

packages/react/src/hooks/useEditorFocus.ts
```diff
@@ -0,0 +1,76 @@
...
+ * `includeEditorUI: true` to instead get "is the user still interacting with
+ * this editor", which counts toolbars, menus and popovers as focused and only
+ * changes once focus movement has settled.
+ *
+ * @param options - See `editor.onFocusChange`.
+ * @param editor - The BlockNote editor instance. If omitted, uses the editor
+ * from the nearest `BlockNoteContext`.
+ */
+export function useEditorFocus(
+  options?: { includeEditorUI?: boolean },
```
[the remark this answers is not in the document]
YousefED: Done in 38328449a — though via the named type rather than `Parameters<...>`: `EditorFocusOptions` already existed in core (both `isFocused` and `onFocusChange` use it), it just wasn't exported publicly. It is now, and the hook imports it.

packages/react/src/hooks/useEditorFocusChange.ts
```diff
@@ -0,0 +1,50 @@
...
+) {
+  const editorContext = useBlockNoteContext();
+  const resolvedEditor = editor ?? editorContext?.editor;
+
+  // Latest-ref pattern: the subscription lives as long as the editor does,
+  // while the callback stays current without retriggering the effect.
+  const callbackRef = useRef(callback);
+  useEffect(() => {
+    callbackRef.current = callback;
+  });
```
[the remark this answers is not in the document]
YousefED: Fixed in fc3585ed4: the callback refs in all three latest-ref hooks now update via an isomorphic layout effect (extracted to a shared util from the inline copy `useEditorState` already had), so a layout effect emitting an editor event right after commit can no longer reach the previous render's callback.

packages/core/src/editor/BlockNoteEditor.ts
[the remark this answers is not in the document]
YousefED: Update: reverted after discussion — we optimize for React, where `BlockNoteView` always provides a wrapper element, so the bare-`<body>` mount only arises in vanilla-JS setups. The behaviour stays documented in the `isWithinEditor` comment; we'll fix it when it becomes an actual issue.

packages/react/src/hooks/useEditorFocus.ts
```diff
@@ -0,0 +1,76 @@
...
+ * `includeEditorUI: true` to instead get "is the user still interacting with
+ * this editor", which counts toolbars, menus and popovers as focused and only
+ * changes once focus movement has settled.
+ *
+ * @param options - See `editor.onFocusChange`.
+ * @param editor - The BlockNote editor instance. If omitted, uses the editor
+ * from the nearest `BlockNoteContext`.
+ */
+export function useEditorFocus(
+  options?: EditorFocusOptions,
```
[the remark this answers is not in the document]
YousefED: Fixed in c4ebbfeab: the cache is now keyed by both inputs (editor + `includeEditorUI`), so an input change re-reads live during that same render — the same read the first render already did — instead of rendering one frame computed for the old inputs. Proven red-first with the transition you suggested: flipping the option while focus sits in the editor's UI rendered a stale `false` frame on all three engines with the old cache; the test records every rendered value so the frame can't hide behind the later correction.

packages/core/src/editor/managers/EventManager.ts
```diff
@@ -51,15 +83,94 @@ export class EventManager<
...
+  private computeUIFocused(): boolean {
+    const active =
+      typeof document !== "undefined" ? document.activeElement : null;
+    return (
+      this.editor.isFocused() ||
+      (!!active && this.editor.isWithinEditor(active))
+    );
+  }
+
+  private settleUIFocus(event: FocusEvent) {
```
nperez0111: All of this could be done more simply, a set of registered callbacks. computeUIFocused is duplicated with editor.isFocused. The timeouts are a smell. we can listen for mount & unmount to attach and remove this handler. We don't need to detach if the subscribers go to 0, it costs almost nothing if it no-ops.

packages/react/src/hooks/useEditorChange.ts
```diff
@@ -20,13 +22,25 @@ export function useEditorChange(
...
 
+  // Latest-ref pattern: the subscription lives as long as the editor does,
+  // while the callback stays current without resubscribing on re-renders.
+  const callbackRef = useRef(callback);
+  // Layout-effect timing, not passive: a layout effect elsewhere can
+  // trigger an editor event right after commit, and the subscription must
+  // not invoke the previous render's callback then.
+  useIsomorphicLayoutEffect(() => {
+    callbackRef.current = callback;
+  });
```
nperez0111, line 33: better pattern, I like it

packages/react/src/hooks/useEditorFocus.ts
```diff
@@ -0,0 +1,104 @@
...
+        },
+        { includeEditorUI },
+      );
+    },
+    [resolvedEditor, includeEditorUI],
+  );
+
+  const getSnapshot = useCallback(() => focused.current!.value, []);
+
+  return useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
```
nperez0111, line 99: Can we not use useEditorState for this?

packages/react/src/hooks/useEditorFocusChange.ts
```diff
@@ -0,0 +1,54 @@
...
+ * popover's input). Pass `includeEditorUI: true` to instead observe "is the
+ * user still interacting with this editor" — floating UI counts as focused,
+ * and the callback fires only on settled changes of that combined state.
+ *
+ * @param callback - Function called with the editor and `{ focused, event }`.
+ * @param editor - The BlockNote editor instance. If omitted, uses the editor
+ * from the nearest `BlockNoteContext`.
+ * @param options - See `editor.onFocusChange`.
+ */
+export function useEditorFocusChange(
```
nperez0111: Do we need this one if we have useEditorFocus

YousefED: it's not used atm, it's there for consistency with other hooks.

I'm neutral on keeping / removing

packages/react/src/hooks/useEditorFocus.ts
YousefED: focus info is not in the editor state, we could maybe extend the hook with support for it, but I think it will get quite messy

packages/core/src/editor/managers/EventManager.ts
YousefED: - unfortunately, we need the timeout because focus doesn't instantly go from one element to another (but first it blurs, body becomes active, then the new element)
- agree that the subscriber count tracking might be an over-optimization

packages/core/src/editor/managers/EventManager.ts
```diff
@@ -51,15 +83,94 @@ export class EventManager<
...
+    if (typeof document === "undefined") {
+      return;
+    }
+    this.uiFocused = this.computeUIFocused();
+    // On focusin the new element already holds focus, so the state can be
+    // read immediately.
+    const onFocusIn = (event: FocusEvent) => this.settleUIFocus(event);
+
+    // On focusout it can't: `document.activeElement` is still the outgoing
+    // element (and passes through `<body>` mid-handoff), and some UI
```
matthewlipski, line 144: Weird behaviour, does this not cause any visual weirdness on mobile with the virtual keyboard? A focus change to `body` would dismiss it I would think.

packages/core/src/editor/managers/EventManager.ts
```diff
@@ -51,15 +83,94 @@ export class EventManager<
...
+
+  private uiFocusSubscriberCount = 0;
+
+  private uiFocusSettleHandle: ReturnType<typeof setTimeout> | undefined;
+
+  private detachUIFocusTracker: (() => void) | undefined;
+
+  private computeUIFocused(): boolean {
+    const active =
+      typeof document !== "undefined" ? document.activeElement : null;
```
matthewlipski: Why do we need to check if `document` is defined? SSR?

packages/core/src/editor/managers/EventManager.ts
matthewlipski: Huh, nvm this is indeed just how focus works.

packages/core/src/editor/managers/EventManager.ts
YousefED: Landed most of this. computeUIFocused was indeed a verbatim duplicate of isFocused({ includeEditorUI }) — gone, the manager calls it now. Reference counting is gone too, with a middle ground on attach: the tracker attaches on the first subscriber ever and detaches only on editor destroy — that deletes the count, the ==1/==0 transitions and the whole double-unsubscribe guard (unsubscribe is a bare off(), naturally idempotent), while a page that never subscribes still installs no document listeners (multi-editor pages would otherwise run one no-op handler per editor on every focus event page-wide). On the timeout: it's load-bearing rather than a smell — activeElement passes through <body> mid-handoff and the ariakit/shadcn link popovers restore focus asynchronously; a microtask is verifiably too early (those popover tests go red) and rAF doesn't run in background tabs. The comment above the handler carries that evidence.

packages/react/src/hooks/useEditorFocusChange.ts
YousefED: Removed — it had no consumers, and the split reads better as one rule: focus-as-state is useEditorFocus; a side effect on focus changes subscribes directly with editor.onFocusChange. Its browser test went with it.

packages/react/src/hooks/useEditorFocus.ts
YousefED: Adding the concrete why-not: useEditorState's store is fed by on: "all" | "mount" | "selection" | "change" — transaction/lifecycle events. Focus is a DOM-side signal with settling semantics that never produces a transaction, so folding it in means a second event source in a transaction-snapshot store, re-rendering every on:"all" consumer on focus flips, and still needing the settling machinery underneath. useEditorFocus is the same pattern (useSyncExternalStore + input-keyed snapshot cache) applied to the focus store — a sibling of useEditorState rather than something to nest inside it.

packages/core/src/editor/managers/EventManager.ts
```diff
@@ -51,15 +83,94 @@ export class EventManager<
...
+  }
+
+  private attachUIFocusTracker() {
+    if (typeof document === "undefined") {
+      return;
+    }
+    this.uiFocused = this.computeUIFocused();
+    // On focusin the new element already holds focus, so the state can be
+    // read immediately.
+    const onFocusIn = (event: FocusEvent) => this.settleUIFocus(event);
```
matthewlipski: It's not entirely clear to me why we need to listen for both `focusIn` and `focusOut` events. Isn't one of each fired per focus change?

packages/core/src/editor/managers/EventManager.ts
matthewlipski: Yeah agree on the subscriber count, I think it's trading off readability for pretty negligible performance gains.

packages/react/src/hooks/useEditorFocus.ts
nperez0111: Not true, I tried it:

```js
  const isFocused = useEditorState({
    editor,
    selector: ({ editor }) => editor.isFocused(),
  });
```
Does correctly report the focus state of the editor, because tiptap already has a prosemirror plugin that adds a transaction for each blur & focus event. A bunch of things in Tiptap already rely on this: https://github.com/ueberdosis/tiptap/blob/92c6d734dc895a20dd34350bd14feba3862bf31e/packages/core/src/extensions/focusEvents.ts#L18-L39

So, this is no more complex than what I wrote above. Selector results are memoized, so this would not re-render the component unless the value flipped, which is just about optimal.

packages/core/src/editor/managers/EventManager.ts
nperez0111: Yea, SSR, this is fine I think

packages/react/src/hooks/useEditorFocus.ts
YousefED: Ah, cool! Wasn't aware of that. If we were to follow that pattern, I suppose we'd also need to fire a transaction for Chrome (`includeEditorUI`) focus changes, right? First thoughts;

- Pro: it would provide a consistent API (`useEditorState` for any getters that BlockNoteEditor exposes)
- Con: we'd "pollute / misuse" (strongly worded..) the Prosemirror transaction system for things that are not directly editor (contenteditable surface) related.

What's your recommended approach here?

nperez0111: We don't need to add document-level events to editor transactions. We already have the event handler. I've just gone ahead and added my commit of how this would work. Much simpler

nperez0111:
This may have messed up the stack, but I added a few commits with the simplifications I was looking for @YousefED

packages/react/src/hooks/useEditorState.ts
```diff
@@ -130,6 +130,10 @@ class EditorStateManager<
         this.subscribers.forEach((callback) => callback());
       };
 
+      if (on === "focus") {
+        return this.editor.onFocusChange(fn, { includeEditorUI: true });
```
YousefED: I think this would now not work with "all". besides that it makes sense

nperez0111: I didn't bother to map "all" and I think it is enough of an edge case to not care. "all" would capture focus already, it just does not include the editor UI

nperez0111:
~~focus -> editor-ui-focus?~~ that feels overly pedantic

Instead, maybe, "all" should also just wire up editor.onFocusChange(fn, {includeEditorUI: true}) so that it will always trigger re-renders. Right now the onFocusChange only reports state transitions so it wouldn't be too bad.

nperez0111:
I'm unsure about #2 since that callback already does the timeout stuff which should normalize it?

nperez0111 (approved):
This is good now. I see why it needed the state to not flip now, so the custom hook is justified for this

nperez0111 (approved):

PR #3029, opened 2026-08-31
test: mobile e2e instance and the test-layer ladder

YousefED:
Second layer of the stack on #3028. Puts the test harness in place so the two fix layers above arrive with their tests rather than their scaffolding.

## What

- **Android-emulated browser instance** for `end-to-end/mobile/`: real touch emulation and a mobile UA, so `isTouchDevice()` is genuinely true and prosemirror-view takes its Android code paths — no platform stubs. Per-instance `contextOptions` are silently ignored by the runner, so emulation goes through the provider, and `ensureTouchEmulation` asserts it actually took effect. This layer registers the instance with only `mobile/**`; the behavioural suites join in the Android-Enter layer, whose fix is what makes them pass there.
- **`imeComposition` browser command**: drives Chromium's real IME pipeline over CDP (`Input.imeSetComposition`). Synthetic `CompositionEvent`s are untrusted and never touch the DOM, so they can't reproduce what a mobile keyboard does.
- **The test-layer ladder** documented in the testing skill: six layers ordered by cost, placement decided red-first at the highest rung where the test can fail, no duplicate coverage below. It also carries the short manual mobile release checklist.

## What deliberately isn't here

Earlier revisions of this layer carried a real-device suite — first on BrowserStack, then a local emulator/simulator suite (Playwright `_android` + adb for Android, Appium/XCUITest for iOS) with its own CI. It was removed after a fix-by-fix audit showed **every fix in this stack has red-first proof at the emulated layer**, while the device CI added real maintainer load (the iOS job turned out to have never executed a test — a silent probe-skip made it green — and Appium startup on hosted runners was still unstable when we pulled it). What only real devices can observe — the IME's own action-key choice, real-keyboard viewport resize, iOS Safari zoom behavior — is a two-minute manual release checklist (testing skill), and the finished suite is parked on `mobile/emulator-layer` (#3034) for revival if a bug class ever demands it.

## Summary by CodeRabbit

- **Tests**
  - Added validation for Android touch emulation.
  - Added Chromium IME composition testing, including text commits, selection, and replacement scenarios.
  - Mobile end-to-end tests now run in a dedicated Android-like browser environment, while desktop browser runs exclude mobile-only tests.

- **Documentation**
  - Added guidance for choosing appropriate test layers and prioritizing regression coverage.
  - Added a manual checklist for validating device-specific mobile behavior.


.github/workflows/device-tests.yml
```diff
@@ -0,0 +1,56 @@
+name: Device tests
```
YousefED: let's not run this as cron, but on any PR as part of the normal CI

tests/device/lib/editorPage.ts
```diff
@@ -0,0 +1,149 @@
...
+  await session.waitFor("selection + link button", applyAndCheck, 25_000);
+}
+
+/**
+ * Opens the create-link popover from the mobile toolbar and waits for its URL
+ * input to hold focus. A mis-aimed tap (iOS chrome-offset guessing) can hit
+ * the keyboard's accessory bar and collapse the whole editing state, so each
+ * attempt rebuilds editing + selection from scratch before tapping.
+ */
+export async function openLinkPopover(session: DeviceSession): Promise<void> {
```
YousefED: should be specific to the link-related tests I think, or in a helper file next to those tests

tests/device/lib/gestures.ts
```diff
@@ -0,0 +1,191 @@
...
+  text: string,
+): Promise<void> {
+  if (session.platform === "android") {
+    await session.elementValue(editorCss, text);
+  } else {
+    await session.typeKeys(text);
+  }
+}
+
+export async function typeAndSubmit(
```
YousefED: why do we use dispatchevent here instead of clicking the enter button on keyboard?

tests/device/lib/tunnel.ts
```diff
@@ -0,0 +1,141 @@
...
+  );
+  if (!res) {
+    throw new Error(
+      `No app server at ${targetOrigin()}. Start the playground (\`pnpm run dev\`) ` +
+        `or point DEVICE_TEST_TARGET at a running server.`,
+    );
+  }
+}
+
+function startProxy(): http.Server {
```
YousefED: can we simplify / clean this? (even if that'd mean using hono for example)

tests/device/lib/tunnel.ts
```diff
@@ -0,0 +1,141 @@
...
+    } catch (error) {
+      res.writeHead(502);
+      res.end(String(error));
+    }
+  });
+  server.listen(PROXY_PORT, "127.0.0.1");
+  return server;
+}
+
+async function ensureLocalBinary(): Promise<void> {
```
YousefED: why do we have this? Does this run in CI?

- If so, is that the recommended browserstack set up?
- if it doesn't run in CI, we should separate it to a file that makes that clear

tests/device/lib/webdriver.ts
```diff
@@ -0,0 +1,210 @@
+/**
+ * Dependency-free WebDriver REST client for BrowserStack real-device sessions.
+ *
+ * Deliberately not WebdriverIO/Appium-client based: the handful of endpoints
+ * we need (session, execute, element, actions, screenshot) are stable W3C
```
YousefED: is this needed? would an extra dependency reduce a lot of code?

tests/src/utils/ensureTouchEmulation.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Restores touch *detection* for the android instance if a previously run
```
YousefED: can't we configure this per pw "device / target"? this sees ugly

tests/vite.config.browser.ts
```diff
@@ -151,12 +153,53 @@ export default defineConfig(
...
+              // suggestion menus) are added alongside the fix that makes them
+              // pass under this emulation — running them here first would
+              // just be reporting a known editor bug as a test failure.
+              //
+              // Keep iframe-screenshotting suites (the exporters'
+              // `screenshotFull` previews) out permanently: Playwright's
+              // element-screenshot path for iframe elements drops the
+              // context's touch emulation for later files (see
+              // utils/ensureTouchEmulation.ts).
+              include: ["./src/end-to-end/mobile/**/*.test.tsx"],
```
YousefED, line 207: if this only includes `mobile/...` tests, why did we make changes to copypaste and keyboardhandlers?

.github/workflows/device-tests.yml
YousefED: Done in 7b40f4ef8: runs on PRs and pushes to main (workflow_dispatch kept for filtered manual runs), cron dropped. Fork PRs have no secrets, so the suite self-skips and the job is a green no-op there. Since device minutes are metered, a superseding push cancels the in-flight PR run.

tests/device/lib/editorPage.ts
YousefED: Moved in 45b00f681 — `tests/device/linkPopover.ts`, next to the tests, in the layer that adds them (#3030). `selectFirstWord` went along too: its verify script waits on the link button, so it was link-coupled despite the generic name. Shared lib keeps only editor-generic helpers.

tests/device/lib/gestures.ts
YousefED: Because no automation channel can press the on-screen keyboard's Enter — W3C actions are clamped to the viewport, this driver has no UiAutomator gestures, and `mobile: shell` is blocked (see the README). But the question caught something real: the dispatched keydown only worked while the popovers had key handlers. After #3030 moves submission to the form's `submit` event, an untrusted synthetic keydown does nothing — the helper was silently broken. It now calls `form.requestSubmit()` (45b00f681), the browser's own submission path, which exercises the popover's real `onSubmit` wiring. The IME's own action-key choice remains a manual release check.

tests/device/lib/tunnel.ts
YousefED: Simplified past hono — deleted (25114c3ac). The proxy existed only to satisfy Vite's `allowedHosts` check, and the playground config already whitelists a hostname for the docker e2e setup, so `bs-local.com` joins that list and devices browse the dev server directly through the tunnel. ~60 lines gone, and the CodeQL stack-trace finding with them.

tests/device/lib/tunnel.ts
YousefED: Yes, it runs in CI — same self-managed daemon as locally, which is a supported BrowserStack setup (their GitHub Action wraps the same BrowserStackLocal binary). I kept self-managed for parity: a CI tunnel failure reproduces identically on a laptop. The file header now says this explicitly (25114c3ac). Happy to switch CI to the official action if you prefer their lifecycle handling.

tests/device/lib/webdriver.ts
YousefED: Investigated: the client is 210 lines covering 8 stable W3C routes (session, execute, element value, actions, orientation, screenshot) plus the artifacts/annotation helpers, which webdriverio doesn't replace. webdriverio would swap ~120 of those lines for a multi-package dependency tree (its client, protocol, and utils packages plus transitive deps). The header documents the trade-off; my read is the hand-rolled client is the smaller maintenance surface, but it's a judgment call — say the word and I'll switch.

tests/src/utils/ensureTouchEmulation.ts
YousefED: It is configured per instance — the playwright provider's contextOptions on the android instance set the real emulation. This util existed for the one way it gets lost mid-session: Playwright's iframe-element screenshot path permanently drops the context's touch emulation. Agreed the self-healing stubs were ugly (and worse, dishonest: detection passed while actual touch input behaved like a desktop). It now asserts and throws naming the cause (0b60a12ac); the android instance's include list is what actually prevents the situation.

tests/vite.config.browser.ts
YousefED: Stacking artifact — you're right. The skips only matter once the android instance runs those suites, which happens in #3031 alongside the Enter fix that makes them passable there. Moved: this layer no longer touches copypaste/keyboardhandlers (85b8d132c); #3031 carries them with the include widening.

.env.sample
```diff
@@ -1,2 +1,13 @@
 export NX_SELF_HOSTED_REMOTE_CACHE_SERVER=https://cache.nickthesick.com
-export NX_SELF_HOSTED_REMOTE_CACHE_ACCESS_TOKEN=g8@ucL8em4*Z9TKXDY9OEX@!upf^Nz9
\ No newline at end of file
+export NX_SELF_HOSTED_REMOTE_CACHE_ACCESS_TOKEN=g8@ucL8em4*Z9TKXDY9OEX@!upf^Nz9
```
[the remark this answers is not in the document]
YousefED: Blanked in 78d8f9563. Note the value has been public in git history since the nx 21 upgrade (#1868, July 2025), so if it's a live credential, rotating it is the actual fix — blanking the sample only stops new copies. @YousefED flagging for the rotation decision.

tests/device/lib/gestures.ts
```diff
@@ -0,0 +1,191 @@
...
+): Promise<void> {
+  if (session.platform === "android") {
+    // A WebDriver Enter key event converges on the same production code path
+    // as the soft keyboard's Enter here: prosemirror-view ignores Enter
+    // keydowns on Android Chrome entirely, so handling proceeds through the
+    // `beforeinput` (insertParagraph) the browser emits — the exact path the
+    // IME takes and where #3001-class bugs live. (BrowserStack blocks the
+    // higher-fidelity options: `mobile: shell` needs an insecure-feature
+    // opt-in and `clickGesture` isn't allowlisted.)
+    await session.typeKeys("\uE007");
```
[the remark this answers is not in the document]
YousefED: Documented in the function body: on Android Chrome, prosemirror-view ignores Enter keydowns entirely, so a W3C Enter converges on the same beforeinput (insertParagraph) handling path the IME takes — and the higher-fidelity channels are blocked (mobile: shell needs an insecure opt-in, clickGesture isn't allowlisted). The 229-without-keypress sequence itself is covered by a synthetic-InputEvent test in the emulated android instance (androidEnter.test.tsx), proven red with the interception removed; the IME's own behaviour stays on the manual release checklist.

tests/device/lib/tunnel.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (existsSync(BINARY)) {
+    return;
+  }
+  const platform = process.platform === "darwin" ? "darwin-x64" : "linux-x64";
```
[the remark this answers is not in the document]
YousefED: Fixed in 25114c3ac: explicit platform/arch map — darwin (x64 under Rosetta 2, noted), linux-x64, linux-arm64 — and an unsupported host now fails before downloading, with guidance.

.github/workflows/device-tests.yml
```diff
@@ -0,0 +1,56 @@
...
+          node-version: 22
+          cache: "pnpm"
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Start playground dev server
+        run: |
+          pnpm run dev &
+          for i in $(seq 1 120); do
```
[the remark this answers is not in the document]
YousefED: Fixed in 7b40f4ef8 (`for _ in`).

tests/device/lib/webdriver.ts
YousefED: Follow-up: swapped in e827fdf93 — the standalone `webdriver` package (WebdriverIO's W3C protocol client, not the framework). It owns the session envelope, request handling and protocol quirks; DeviceSession keeps only the domain layer, same public API, so gestures and tests didn't move. Its edgedriver/geckodriver postinstall downloads are denied via allowBuilds since we only drive the remote hub. Runtime-unverifiable without credentials — covered by the recertification run.

tests/device/lib/tunnel.ts
YousefED: Follow-up: the surprise was justified. BrowserStack's documented Node.js integration is their `browserstack-local` package, which downloads and manages the right daemon itself — adopted in 2daf821b0; the custom binary fetch, platform/arch map and daemon spawning are gone. (Their *Playwright* SDK was also considered: it reaches real Android but iOS only as Playwright-WebKit on macOS, not real iPhones — and real iOS Safari is this suite's reason to exist, which keeps us on the Appium/WebDriver product.)

tests/device/lib/gestures.ts
YousefED: Follow-up: it now does hit the Enter key (d10475e47) — a W3C protocol Enter, which is *trusted* input, so the browser runs its default action and the whole real path is exercised: key press → implicit form submission → the popover's submit handling. requestSubmit() was skipping the key half. The unreachable thing remains the on-screen keyboard's own action button, which stays a manual release check.

tests/device/lib/webdriver.ts
YousefED: Final form, after checking where the docs actually stand (452be0cb1): the client is now selenium-webdriver — the one BrowserStack's Automate Node.js docs and samples use, with auth in bstack:options as they show (devices.ts already had that shape). The bare `webdriver` protocol package from my previous reply worked but appears in no BrowserStack doc, so it's gone. On browserstack-node-sdk specifically: it layers on selenium-webdriver but wraps supported runners (Jest/Mocha) and manages the tunnel + platform matrix from a yml — adopting it would mean a second test runner in a vitest-standardized repo, for capabilities we already have via the official browserstack-local binding and devices.ts. Left as a deliberate non-adoption, noted in the commit message; worth revisiting if Test Observability becomes interesting.

YousefED:
The previous push's `ios-simulator` jobs on the upper layers ran 0 tests and then hung to the 45-minute job timeout. Diagnosis from the logs: on a cold macos-15 runner, the first `xcrun simctl` invocation can exceed the availability probe's 10s timeout, so the iOS target silently dropped out (0 tests — a false green), and the timed-out probe's orphaned child kept its stdio pipes open, which wedged the process at exit. Locally `xcrun` is always warm, which is why the suite passed everywhere else.

Fixed in this push, three ways:

- the probe now spawns with `stdio: "ignore"` (nothing to wedge on) and a generous timeout;
- the workflow warms the simulator toolchain in its own step before the suite, so the cold-start cost is paid visibly;
- on CI, a `DEVICE_FILTER`-selected target whose probe still fails now **fails the job loudly** instead of skipping to a 0-test green — per the fail-loudly convention.

.github/workflows/emulator-tests.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Run device suite on the emulator
+        uses: reactivecircus/android-emulator-runner@a421e43855164a8197daf9d8d40fe71c6996bb0d # v2.38.0
+        with:
+          api-level: 35
+          arch: x86_64
+          target: google_apis
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -no-audio -no-boot-anim
+          disable-animations: true
+          script: DEVICE_FILTER=local-android vp run test:device
```
nperez0111: Wait so this runs on an "actual" android & ios. This is asking for reproducibility issues in the future between CI and local dev. I'm not sure that I like this direction.

YousefED:
Update on the two comments above: the device layer discussed there has been removed from this stack entirely (fix-by-fix audit in the PR body — every fix is red-first provable on the emulated instance, and the device CI's cost/benefit didn't hold up: the iOS job had silently never executed a test, and Appium startup on hosted runners was still unstable). The finished suite — including the probe and Appium fixes from those comments — is parked on mobile/emulator-layer (#3034).

.claude/skills/testing-skill/SKILL.md
```diff
@@ -7,6 +7,33 @@ description: Instructions for writing, running, and updating unit/end-to-end tes
 
 In most cases, once a feature, bug fix, or other modification has been written, it will need to have tests added, or existing tests updated.
 
+## The test-layer ladder
+
+Six layers, ordered by cost (speed, determinism, machinery). Prefer the highest workable rung:
+
+1. **Unit tests** — node, colocated in `packages/*` (`vp run test`)
+2. **Cross-package integration tests** — node, `tests/src/unit`: full documents through real pipelines (parsing, format conversion, export)
```
YousefED: I think 3 should be listed above 2. Yes, you mention it's a different axis, but I still think co-located should have priority

.env.sample
```diff
@@ -1,2 +1,2 @@
 export NX_SELF_HOSTED_REMOTE_CACHE_SERVER=https://cache.nickthesick.com
-export NX_SELF_HOSTED_REMOTE_CACHE_ACCESS_TOKEN=g8@ucL8em4*Z9TKXDY9OEX@!upf^Nz9
\ No newline at end of file
+export NX_SELF_HOSTED_REMOTE_CACHE_ACCESS_TOKEN=
```
YousefED: remove this change

tests/src/utils/ensureTouchEmulation.ts
```diff
@@ -0,0 +1,28 @@
...
+ * previews) rewrites the device-metrics override and permanently drops the
+ * context's touch emulation — `navigator.maxTouchPoints` becomes 0 for every
+ * later test file. The android instance therefore keeps such suites out of
+ * its include; touch-dependent tests call this in `beforeEach` so that if the
+ * include ever regresses, the run fails naming the cause instead of silently
+ * testing a desktop context that merely claims to be mobile.
+ */
+export function ensureTouchEmulation() {
+  if (
+    navigator.maxTouchPoints === 0 ||
```
YousefED, line 17: where is the code that sets pointer coarse and maxtouchpoints?

.github/workflows/emulator-tests.yml
YousefED: Agreed — this concern is why the layer is gone: the device suite and its CI were removed from the stack entirely after a fix-by-fix audit showed every fix has red-first proof on the emulated instance (details in the PR body). The finished suite is parked on mobile/emulator-layer (#3034) should a bug class ever demand real-OS input; until then mobile CI is deterministic CDP emulation only.

.claude/skills/testing-skill/SKILL.md
YousefED: Reordered: colocated browser unit tests are now rung 2, cross-package integration rung 3, with the corollary reworded to say the ordering is by colocation, not machinery cost.

.env.sample
YousefED: Dropped — the .env.sample change is out of the PR; the file is untouched relative to the base again.

.claude/skills/testing-skill/SKILL.md
```diff
@@ -7,6 +7,33 @@ description: Instructions for writing, running, and updating unit/end-to-end tes
...
+3. **Browser unit tests** — colocated `*.browser.test.{ts,tsx}`: one unit that genuinely needs real DOM/rendering
+4. **Browser e2e** — `tests/src/end-to-end`: UI flows in real browsers (Docker)
+5. **Mobile-emulated e2e** — `end-to-end/mobile/**` and the suites on the `android` instance: only for behavior that differs under mobile conditions (touch, viewport, UA, emulated IME)
+6. **Device suite** — _parked_: a real emulator/simulator suite (one session interface, Android via Playwright `_android` + adb, iOS via Appium/XCUITest) lives on the `mobile/emulator-layer` branch (PR #3034). It left the active stack because every fix it guarded is red-first provable on the emulated instance; revive it only for a bug class that emulation demonstrably cannot observe (the IME's own action-key choice, real-keyboard viewport resize, real iOS Safari focus/zoom) — until then those are the manual checklist below.
```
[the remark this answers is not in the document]
YousefED: Added to the ladder: rung 5 now carries the restriction (no iframe-element screenshots / screenshotFull in android-instance suites, with the pointer to ensureTouchEmulation.ts).

tests/src/utils/ensureTouchEmulation.ts
YousefED: Nothing in our code sets them directly — they come from Playwright's browser-context creation: the android instance passes provider-level contextOptions with hasTouch: true and isMobile: true (tests/vite.config.browser.ts lines 183-191), and Playwright translates hasTouch into CDP touch emulation (Emulation.setTouchEmulationEnabled), which is what makes navigator.maxTouchPoints > 0 and (pointer: coarse) match. The subtlety that motivated this helper: those options only take effect at the provider level — the runner silently ignores per-instance contextOptions — and Playwright's iframe-element screenshot path can permanently drop the emulation afterwards, so this asserts it rather than trusting the config.

tests/device/.cache/BrowserStackLocal
nperez0111, line 1: rm binaries

nperez0111 (approved):

tests/device/.cache/BrowserStackLocal
YousefED: Purged from history, not just tipped off — the daemon binary (34.6 MB) and its zip (9.8 MB) had been swept into a later commit by a broad git add and outlived the whole BrowserStack removal. Amended out of the introducing commit; nothing regenerates .cache/ since the browserstack-local integration is gone entirely.

nperez0111 (approved):

matthewlipski (approved):

Issue #714, opened 2024-04-25
[Collaboration] Make yjs provider composable


bojieyang:
**Is your feature request related to a problem? Please describe.**

Currently the collaboration structure only support one yjs provider. But the yjs providers are composable. 

For example, I may want to compose y-indexeddb and y-partykit/provider to support offline editing and automatic sync to the server when it is online. yjs support this feature in [it's document](https://docs.yjs.dev/getting-started/allowing-offline-editing), but it can't used in blocknote.
 

**Describe the solution you'd like**

Allow set multi providers in the collaboration structure or other way to make the provider composable .

YousefED:
I'm pretty sure this is supported; just add multiple providers to the document

PR #3030, opened 2026-08-31
fix(ui): commit popover forms through submit, not a key handler

YousefED:
Third layer of the stack, on #3029.

## The bug

Creating a link on Android didn't work: the URL never became a link and focus jumped to the next editor. A mobile IME picks what its Enter key *does* — with a lone text field outside a form it picks "Next": advance focus, **no key event at all**. A popover listening for Enter has nothing to hear. Being inside a real `<form>` is what makes the IME offer a submitting action instead (verified on a physical device, with `enterkeyhint` ruled out as the cause) — and `Form.Root` was a `<div>`, so its `onSubmit` could never fire.

## The fix

`Form.Root` renders a real `<form>`; submission runs off its `submit` event. Three consequences, each pinned by tests:

- **HTML only submits implicitly with a submit button or exactly one field** — the two-field link *edit* form would still reach nothing. So every form declares its submit control explicitly: `Form.Root` requires `submitButton: ReactElement | "none"`, with `ScreenReaderOnlySubmit` in `@blocknote/react` as the standard control (visually hidden — clipped, not `display:none` — so assistive technology keeps a labelled control; `tabIndex={-1}` so sighted keyboard users never land on an invisible tab stop; CSS and dictionary in one copy instead of per skin). The embed tab passes its real button as the `submitButton` — inside the form, one commit path for click/Enter/IME-action, one labelled action — which also forced `FilePanel.Button` to take an explicit `type` (the skins disagreed on the default, which is what had exiled the button from the form in the first place).
- **No composition guard is needed on this path** — the IME consumes the confirming Enter (it reaches the page as keyCode 229, which triggers no default action), so implicit submission cannot fire mid-composition; the `isComposing` checks the old *keydown* handlers carried don't transfer to the submit path. The tests pin the native contract against the real popover: accepting a candidate does not submit, Enter afterwards does.
- **One submission path** makes the five Enter key handlers redundant; removed. `EmbedTab` had no form at all and gains one; the AI prompt menu's handler and `onSubmit` disagreed on whether Enter picks the highlighted suggestion or submits the raw text, and now share one decision.

Also: `TextInput` loses its `onSubmit` prop — every skin forwarded it onto the `<input>`, where `submit` never fires, so it was dead since #652 (and plausibly why the gap went unnoticed).

## Notes for review

- **Breaking, for release notes**: `Form.Root` requires `submitButton` (replaces the earlier optional opt-out), `FilePanel.Button` requires `type`, and the new dictionary key `generic.form_submit` (the submit control's accessible name) makes hand-rolled dictionaries a compile error until added. The 23 non-English translations are machine-generated and unreviewed.
- **Coverage limit, stated where it lives**: emulation proves submission works with *no key event at all* (`mobile/linkSubmit.test.tsx`) — exactly what the IME's submitting action does — but which action the IME itself offers is an OS decision no emulation can observe. That half is on the manual release checklist (testing skill); `linkSubmit.test.tsx`'s header documents the boundary. A device suite automating it exists parked on `mobile/emulator-layer` (#3034).
- Platform facts asserted rather than assumed, now in their own `end-to-end/platform/` directory (raw `createElement` fixtures, nothing BlockNote — the README states the contract): implicit-submission rules and composition behaviour, per engine, on the desktop engines and the emulated-android instance alike. All fixes proven red-first (details in commit messages).

## What fixes the keyboard collapse on focus

Asked in review: the description covers form submission, but the popover no longer collapses the virtual keyboard when focus moves into its input. Which change does that?

The one-line `hideDetached` gate in the Mantine `Popover` adapter (`packages/mantine/src/popover/Popover.tsx`). Mantine's `Popover` defaults `hideDetached` to `true`: when the target is considered detached, it sets `display: none` on the still-mounted dropdown. The on-screen keyboard opening resizes the viewport, Mantine treats that as detached, the dropdown goes `display: none`, a hidden ancestor cannot hold focus, so the URL input blurs, the toolbar unmounts with it, and the keyboard closes. That is the whole cascade. `hideDetached={preventFocusOnOpen ? false : undefined}` turns it off for exactly the popovers that live in the mobile toolbar (`preventFocusOnOpen` is set for those and nothing else); desktop keeps Mantine's default. Diagnosed in #3025, carried here with the popover work. Ariakit and shadcn never had the behaviour, which is why the skinParity suite only pins the Mantine flow through `mobileToolbar.test.tsx`.

## Focus on the mobile toolbar, all three skins

This layer also makes `preventFocusOnOpen` hold on the ariakit and shadcn skins (only Mantine honoured it), and settles what takes focus where. The model is written up in `packages/react/src/editor/focus-management.md`; the short version:

| Surface | Desktop | Mobile toolbar |
|---|---|---|
| Popover (link form, file panels, emoji) | only an `autoFocus` input | same |
| Menu (colors, drag handle, table) | the menu, on open (library) | nothing |
| Select (block type) | the list, on open (library) | nothing |
| Toolbar button, select trigger | the button, on click; most hand it back | nothing, on a tap |

- **Popovers never take focus from the UI library, on any device.** Mantine already had `trapFocus` off, Ariakit `autoFocusOnShow` off, shadcn now passes `initialFocus={false}`. BlockNote owns focus in its popovers through `useAutoFocus` on the input that wants it. That leaves nothing for `preventFocusOnOpen` to do on `Popover.Root`, so the prop is removed there and from the buttons that passed it (breaking for custom skins that read it: it was unreleased).
- **Ariakit menus and select** (`preventFocusOnOpen` set by the mobile toolbar): `autoFocusOnShow={() => false}` on the menu (a plain `false` defers to a store flag Ariakit's MenuButton sets on click), `autoFocusOnShow={!preventFocusOnOpen}` on the select popover, items and options do not focus on hover; the popover gets `unmountOnHide`, since Ariakit keeps closed popover content mounted and `useAutoFocus` fired once on the hidden URL input and never again, which is what broke desktop ariakit link creation too.
- **shadcn menus and select**: Base UI has no `initialFocus` on Menu or Select, by design upstream (mui/base-ui#2143, #714), so the popup hands focus straight back to the editor inside the same focus dispatch (`onFocusCapture`, `packages/shadcn/src/lib/preventFocusOnOpen.ts`). Traced on a real Android device: without it the popup focuses, the keyboard closes, the toolbar and popup are gone about 150 ms after opening.
- **One tap guard**, `preventFocusOnTap` exported from `@blocknote/react`, replaces the per-skin copies on toolbar buttons, dropdown triggers, menu items and select options: on touch it cancels the focus move of a tap while the click still fires, on Safari it focuses the button like other browsers do. The ariakit and shadcn toolbar buttons spread the library's props first and forward its injected `onMouseDown`, so the guard is never replaced and Base UI's mousedown-opened menus still open.
- **Mantine `hideDetached` is off unconditionally** (replaces the gate described in the section below): BlockNote's `GenericPopover` already hides floating UI on a hidden reference, and on the phone the keyboard's viewport resize made Mantine hide the dropdown and blur its input.
- **`PortalElementOverride` attaches, themes and registers its root in an insertion effect.** A parent's layout effect runs after its children's, and Ariakit picks the mount point of its eagerly rendered popovers in a layout effect, re-parenting a still-detached root to `document.body`, outside the editor's registered UI; on the mobile toolbar that took the link popover with it and the toolbar unmounted on the URL input's focus.
- Every prop that a mutation sweep proved load-bearing carries a one-line "How-to-test" comment: the manual repro and the `skinFocus` case that covers it.

Tests: `tests/src/end-to-end/mobile/skinFocus.test.tsx` on the android instance, per skin: tapping a button never moves focus onto it, the link button hands focus to the URL input, opening the block type select or the colors menu keeps focus in the editor, picking from each leaves it there. Each case was red on the skin it pins before the change. The portal pin test in `portalElements.test.tsx` fails with a layout effect.

## Summary by CodeRabbit

* **New Features**
  * Added native form submission support across editor popovers, including Enter-key and mobile IME submission.
  * Added accessible, visually hidden submit controls for multi-field forms.
  * Added localized submit labels across supported languages.

* **Bug Fixes**
  * Improved mobile focus handling to keep the on-screen keyboard open and prevent unwanted scrolling.
  * Improved autofocus behavior when opening popovers.
  * Improved toolbar and menu interactions across supported UI skins.

* **Style**
  * Increased touch-device input sizing for improved usability.


packages/core/src/editor/managers/StyleManager.ts
```diff
@@ -183,7 +183,26 @@ export class StyleManager<
    */
   public getSelectedLinkUrl() {
     return this.editor.transact((tr) => {
-      return this.getLinkMarkAtPos(tr.selection.from)?.href;
+      const { from, to, empty } = tr.selection;
+      if (empty) {
+        return this.getLinkMarkAtPos(from)?.href;
```
YousefED: maybe keep this as is, I don't think it's critical for the current bugfix this PR is made for

packages/mantine/src/form/TextInput.tsx
```diff
@@ -29,26 +28,37 @@ export const TextInput = forwardRef<
 
   assertEmpty(rest);
 
+  // Focus with `preventScroll`, rather than the native `autofocus`: these
```
YousefED: doesn't mantine's autofocus (with data-autofocus) already do preventscroll by default?

packages/mantine/src/popover/Popover.tsx
```diff
@@ -13,6 +13,12 @@ export const Popover = (
 ) => {
   const { open, onOpenChange, position, portalRoot, children, ...rest } = props;
 
+  // A `portalRoot` is only passed by the mobile toolbar, which renders its
+  // popovers into its own container — so it doubles as "this popover belongs
+  // to the mobile toolbar", which is what the two behaviours below actually
+  // depend on. Named here so the reason isn't hidden behind an unrelated prop.
+  const isMobileToolbarPopover = !!portalRoot;
```
YousefED: add a TODO here indicating this should be cleaned up once we've settled on a good portalling solution (that's still pending slack discussions)

packages/mantine/src/blocknoteStyles.css
```diff
@@ -257,6 +257,19 @@ on touch devices (e.g. the mobile formatting toolbar). */
   font-size: 12px;
 }
 
+/* On touch devices, enlarge the form-popover inputs (e.g. the link popover's
+   URL field). The 16px font-size is load-bearing: iOS Safari auto-zooms the
+   page when focusing an input with a smaller computed font-size, and that zoom
+   perturbs the visual viewport the mobile toolbar positions itself from. The
+   taller min-height also gives a comfortable tap target. */
+@media (pointer: coarse) {
```
YousefED, line 266: Note; this comes from @matthewlipski 's earlier PR

packages/react/src/components/FilePanel/DefaultTabs/EmbedTab.tsx
```diff
@@ -73,17 +55,26 @@ export const EmbedTab = <
...
+          placeholder={dict.file_panel.embed.url_placeholder}
+          value={currentURL}
+          onChange={handleURLChange}
+          data-test={"embed-input"}
+        />
+      </Components.Generic.Form.Root>
       <Components.FilePanel.Button
         className={"bn-button"}
-        onClick={handleURLClick}
+        onClick={embedURL}
```
YousefED: I think this button should be placed inside the form and it should be a submit button

packages/shadcn/src/form/Form.tsx
```diff
@@ -1,10 +1,32 @@
...
-  const { children, ...rest } = props;
+  const { children, onSubmit, omitSubmitButton, ...rest } = props;
+  const dict = useDictionary();
 
   assertEmpty(rest);
 
-  return <>{children}</>;
+  return (
+    <form
+      onSubmit={(event) => {
```
YousefED, line 12: should be usecallback?

packages/shadcn/src/form/Form.tsx
```diff
@@ -1,10 +1,32 @@
...
+      }}
+    >
+      {children}
+      {/*
+        Gives the form a submit button, which is what makes Enter submit it at
+        all once a caller renders more than one field (see the `onSubmit`
+        contract in `ComponentsContext`). Visually hidden rather than absent,
+        so assistive technology still has a labelled control to activate.
+      */}
+      {!omitSubmitButton && (
```
YousefED: I think we can architect this better; what about this;

- a form has a required prop `submitButton` which can be an `HTMLxxxElement | "none"` (none=explicit opt-out
- in blocknote/react, we create an `ScreenReaderOnlySubmit` button. This way we can move the CSS and dictionary usage all to blocknote/react

good practice?

tests/src/end-to-end/platform/compositionSubmit.test.tsx
```diff
@@ -0,0 +1,106 @@
+import { afterEach, describe, expect, test } from "vite-plus/test";
```
YousefED, line 1: is this test file needed / recommended to keep?

tests/src/end-to-end/platform/implicitSubmit.test.tsx
```diff
@@ -0,0 +1,135 @@
+import { afterEach, describe, expect, test } from "vite-plus/test";
```
YousefED, line 1: same. this file required / recommended to keep or not?

tests/src/end-to-end/mobile/mobileToolbar.test.tsx
```diff
@@ -0,0 +1,200 @@
+import App from "@examples/01-basic/testing/src/App";
```
YousefED, line 1: this also has link specific things. Should this be tested here or in linksubmit file? What's the clean split / organization? Also, are all tests needed?

packages/core/src/editor/managers/StyleManager.ts
YousefED: Agreed, and verified: with this hunk reverted (worktree, uncommitted), all 18 link/popover/submit tests pass on chromium, firefox AND webkit in Docker — the boundary-disagreement case the fix guards against is not observable by any current test, so it fails the red-first bar for this PR. Reverted here; it can come back in its own PR with a test that actually goes red on the affected engine.

packages/mantine/src/form/TextInput.tsx
YousefED: It would — but only inside a focus trap. Mantine's preventScroll focusing lives in useFocusTrap (@mantine/hooks: focusElement.focus({ preventScroll: true }) on the [data-autofocus] element), and that only runs when trapFocus is active. Popover's trapFocus defaults to false, and we explicitly keep it off for mobile-toolbar popovers (trapping would blur the contentEditable and dismiss the keyboard) — so data-autofocus is never read in these popovers, and the native autofocus attribute path scrolls without preventScroll. The manual focus({ preventScroll: true }) is load-bearing.

packages/mantine/src/popover/Popover.tsx
YousefED: TODO added (worktree).

packages/mantine/src/blocknoteStyles.css
YousefED: Noted — provenance comment added: the rule now carries '(From #2982.)'. Related: with this stack landing, #2982 itself looks fully superseded (its keyboard-closing fix is covered by the hideDetached/trapFocus handling here) — worth closing it with a pointer when this merges.

packages/react/src/components/FilePanel/DefaultTabs/EmbedTab.tsx
YousefED: Done, via the submitButton architecture from your other comment: the visible embed button is now the form's submitButton — inside the <form>, type='submit', no onClick — so clicking, Enter, and the IME action key all commit through the single submit event, and assistive technology sees exactly one labelled action. The 'skins disagree on the default type' problem is fixed at the root: FilePanel.Button now takes an explicit required type ('button' | 'submit'; shadcn hardcoded submit while Mantine defaults to button, which is what made the old outside-the-form hack necessary). The structural test now asserts the inverse: exactly one submit control in the form, and it is the visible labelled button.

packages/shadcn/src/form/Form.tsx
YousefED: Not needed here: the handler goes to a native <form> element, so its identity per render has no observable effect — nothing memoized receives it, and DOM props are just reassigned. useCallback would add indirection without a consumer that benefits. (All three skins share this inline pattern.)

packages/shadcn/src/form/Form.tsx
YousefED: Concur — implemented (worktree, uncommitted) with two refinements. (1) Type: submitButton: ReactElement | 'none' — ReactNode absorbs string literals so the 'none' sentinel wouldn't type-check as a distinct case (the linter flagged it); ReactElement also matches your HTMLxxxElement instinct. (2) Kept your requiredness: every Form.Root caller now explicitly decides its submit affordance — the compiler forces the decision instead of a boolean opt-out defaulting silently. blocknote/react gains ScreenReaderOnlySubmit (owns the dictionary lookup) and the clip CSS moved to react's styles.css — single copy; the three skin Forms collapse to rendering {submitButton === 'none' ? null : submitButton} with no dictionary import. Callsites updated (link edit, file caption/rename, AI prompt menu get ScreenReaderOnlySubmit; embed tab passes its real button). Two breaking changes for release notes: Form.Root's required submitButton replaces omitSubmitButton, and FilePanel.Button's required type. All form/mobile suites green on all three engines in Docker.

tests/src/end-to-end/platform/compositionSubmit.test.tsx
YousefED: Recommend keeping. It pins the platform contract the composition-guard *removal* rests on: the IME consumes the confirming Enter, so accepting a candidate must not submit and Enter afterwards must — through Chromium's real IME pipeline, not synthetic events. If someone reintroduces Enter key handlers or engine behavior shifts, this is the test that goes red naming the cause. Cheap (no screenshots).

tests/src/end-to-end/platform/implicitSubmit.test.tsx
YousefED: Recommend keeping, same reasoning one level down: it asserts the HTML implicit-submission rules per engine (single field submits; multiple fields need a submit control; the clipped button still counts; no double-submit) — the facts the whole Form.Root design is built on, and now the executable documentation for why submitButton is a required prop. If an engine deviated or the hidden-button pattern stopped qualifying, this names the cause instead of leaving a mystery e2e failure.

tests/src/end-to-end/mobile/mobileToolbar.test.tsx
YousefED: Proposed split criterion: what machinery is under test, not what UI appears in it. mobileToolbar.test.tsx = toolbar/popover machinery (show/hide with the keyboard, a popover's input surviving keyboard resizes, focus returning on toggle) — its link popover is just the vehicle, since it's the popover with an input; linkSubmit.test.tsx = the submission path (link created with no key event at all, focus staying in the right editor). Current placement already follows that line; I'd only rename the middle test ('link popover holds focus…' → 'a popover's input holds focus through keyboard resizes') to make the subject explicit. On necessity: the three mobileToolbar tests pin three distinct regressions (toolbar lifecycle, the Mantine hideDetached collapse, focus jump on toggle) — each fails differently, none is redundant with linkSubmit's two.

packages/mantine/src/form/TextInput.tsx
```diff
@@ -29,26 +28,44 @@ export const TextInput = forwardRef<
...
+  // `data-autofocus` below makes any such trap pick this same element
+  // instead of falling back to "first focusable", so the two mechanisms can
+  // never fight over where focus lands.
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const setRefs = useMergeRefs([inputRef, ref]);
+  useEffect(() => {
+    if (autoFocus) {
+      inputRef.current?.focus({ preventScroll: true });
+    }
+  }, [autoFocus]);
```
nperez0111: I think we need a hook here to prevent drift and duplication

nperez0111: I also hesitate on a focus side effect. I'm unsure if this is meant to be in a layout effect or on a setTimeout(0) or something, I've seen this done lots of different ways

packages/ariakit/src/input/TextInput.tsx
```diff
@@ -32,6 +31,19 @@ export const TextInput = forwardRef<
 
   assertEmpty(rest);
 
+  // Focus with `preventScroll`, rather than the native `autofocus`: these
```
matthewlipski: This feels like a smell to me. The Mantine/Ariakit/ShadCN components should be "dumb" adapters which render a primitive with some generic self-explanatory props. Having code for overwriting focus behaviour here is fixing a specific issue that the `TextInput` component should have no knowledge of.

packages/react/src/components/FilePanel/DefaultTabs/EmbedTab.tsx
```diff
@@ -37,25 +37,7 @@ export const EmbedTab = <
...
-            url: currentURL,
-          } as any,
-        });
-      }
-    },
-    [editor, props.blockId, currentURL],
-  );
-
-  const handleURLClick = useCallback(() => {
+  const embedURL = useCallback(() => {
```
matthewlipski: Rename to `handleSubmit` for consistency

packages/react/src/components/FormattingToolbar/DefaultButtons/FileRenameButton.tsx
```diff
@@ -133,7 +124,10 @@ export const FileRenameButton = () => {
         className={"bn-popover-content bn-form-popover"}
         variant={"form-popover"}
       >
-        <Components.Generic.Form.Root>
+        <Components.Generic.Form.Root
+          onSubmit={() => setPopoverOpen(false)}
+          submitButton={<ScreenReaderOnlySubmit />}
```
matthewlipski, line 125: If I understand correctly, the submit button is visually hidden here. Why not have an actual submit button though? Doesn't having it solve 90% of the problem? Since on mobile, pressing enter to commit changes in a text input is not as ubiquitous as it is on desktop, and generally you have buttons for that.

packages/react/src/editor/ComponentsContext.tsx
```diff
@@ -304,6 +313,31 @@ export type ComponentProps = {
     Form: {
       Root: {
         children?: ReactNode;
+        /**
```
matthewlipski, line 323: In general, I don't think we need this many comments. At some point it makes the code harder to read & comprehend rather than easier.

matthewlipski (commented):
Tested on Android and iOS - works very nicely! Though the PR description focuses on the enter handling/form submission fixes, the whole issue of the virtual keyboard handling on focus moving to the input is also fixed. What change in the code is actually responsible for this? Since I don't really see that mentioned in the description.

packages/mantine/src/form/TextInput.tsx
YousefED: Extracted: useAutoFocus in @blocknote/react now owns the ref + focus({ preventScroll: true }) effect and the rationale, and the three skins consume it — the drift you predicted had already happened (mantine's copy had grown a comment paragraph and a data-autofocus attribute the others lacked). On whether the pattern is sound, we audited the official implementations from source: React's own autoFocus is a bare .focus() at commit with no preventScroll — exactly the scroll-yank this exists to avoid; Mantine's focus-on-open is setTimeout + preventScroll; floating-ui's FloatingFocusManager is layout effect → microtask → rAF → focus({ preventScroll }). Ours sits mid-spectrum with unconditional preventScroll (stronger scroll-safety than floating-ui's initial-element case) and deliberately no extra deferral: there's nothing to wait for (floating-ui's microtask waits for its own tabIndex setters), and every added hop erodes the user-gesture window in which iOS Safari lets programmatic focus open the keyboard — the real-device suite validated the keyboard appears with this timing. data-autofocus is set only where the skin's UI library actually reads it (Mantine's trap, Ariakit's dialog initial-focus — verified in their source); shadcn's Base UI has no attribute convention, so that skin omits it. The comparison lives in the hook's jsdoc.

packages/ariakit/src/input/TextInput.tsx
YousefED: Agreed, and this landed already via the parallel thread: the focus block is gone from all three skins — @blocknote/react's useAutoFocus owns the behavior and the rationale, and a skin's entire involvement is one hook call plus mapping to its own library's convention where one exists (data-autofocus for Mantine's trap and Ariakit's dialog initial-focus; Base UI has none, so shadcn maps nothing). That last part is the reason the call sits in the skin rather than in the react-side callers: which attribute a UI library reads is exactly the kind of knowledge a dumb adapter is for.

packages/react/src/components/FilePanel/DefaultTabs/EmbedTab.tsx
YousefED: Renamed to handleSubmit.

packages/react/src/editor/ComponentsContext.tsx
YousefED: Trimmed — the Form contract docs went from 24 lines to 13: the contract stays (real <form>, preventDefault, what submitButton decides), the device-verified Android rationale moved out (it lives in the PR description and the e2e tests that pin it). Fair point generally; the public ComponentsContext docs are the one place we keep prose since custom-skin authors read them without the PR context.

packages/react/src/components/FormattingToolbar/DefaultButtons/FileRenameButton.tsx
YousefED: The hidden button is the default, not a ceiling — the required submitButton prop exists precisely so any caller can pass a visible type="submit" button instead, and the embed tab already does (its visible Embed button is the form's submit control). For the rename/caption popovers we kept this PR visual-change-free on purpose; and on mobile the visible affordance for these is the keyboard's own action key — which is the thing this PR makes actually work (Gboard only offers a submitting action inside a real <form>). If we want visible buttons in those popovers too, it's a per-caller one-liner now — but that's a design change worth its own decision.

packages/mantine/src/form/TextInput.tsx
YousefED: Now pushed (with the comments themselves given a trim pass too). One correction to my earlier note: on closer inspection Ariakit's initial-focus reads the attribute but focuses with a bare element.focus() plus a Safari scrollIntoView — the exact scroll-yank useAutoFocus exists to avoid, previously masked only by effect ordering. So instead of advertising the input to it, the ariakit skin now disables autoFocusOnShow outright (matching the other skins — no library moves focus to an input on open) and sets no attribute. Final policy in the hook's jsdoc: data-autofocus only where the library reads it AND focuses safely — which today is Mantine alone.

packages/react/src/components/FilePanel/DefaultTabs/EmbedTab.tsx
YousefED: All pushed: the rename, the doc trim (plus a second trim pass over the new comments in this batch), and the focus-ownership hardening that came out of the adapter-smell discussion.

packages/react/src/components/FormattingToolbar/DefaultButtons/FileRenameButton.tsx
YousefED: But yes @matthewlipski , might be a good idea, but for now I figured to keep design changes out of scope

packages/react/src/editor/PortalElementOverride.tsx
```diff
@@ -4,16 +4,17 @@ import {
   useCallback,
   useContext,
   useEffect,
-  useLayoutEffect,
+  useInsertionEffect,
```
nperez0111, line 7: til about this, didn't know it existed, also was there in React 18

nperez0111 (approved):

packages/react/src/editor/PortalElementOverride.tsx
YousefED: same. Would rather stick to more common APIs, but this specifically addressed an Ariakit bug iirc (covered in tests).

PR #3031, opened 2026-08-31
fix(core): Enter on Android and iOS, toolbar and side menu fixes on phones

### LINKED ISSUE #3001: Enter, newline, and down arrow broken on Android
### What’s broken?

Using the enter key on Android has some weird behavior: you never get a new line, sometimes you get a space, sometimes it breaks a table, sometimes it shifts the placeholder/ghost text. Similar with the down arrow key: the cursor does not get moved down to the next line.

I can reproduce this on desktop Chrome and Firefox when emulating a mobile device with devtools _only if_ it was emulating the mobile device at page load, and it stays broken after exiting mobile device emulation. Thus I believe this is triggered by some platform-detection by BlockNote rather than any true mobile text-editing bug.

https://github.com/user-attachments/assets/93203953-628a-44d8-b2af-04c7baa06c73

The enter key and down arrow should insert newlines or move down, respectively.

### Steps to reproduce

On Android:
1. Navigate to https://www.blocknotejs.org/demo
2. Focus editor and hit enter

On desktop Chrome/Firefox:
1. Open devtools, emulate a mobile device
2. Navigate to https://www.blocknotejs.org/demo
3. Focus editor and hit enter

### BlockNote version

Latest; reproduced on @blocknote/shadcn 0.52, 0.53, 0.54.

### Environment

Android 10; desktop Chrome/Firefox on Linux and MacOS in mobile emulation mode

### Additional context

_No response_

### Contribution

### Sponsor

YousefED:
Top of the stack, on #3030. Fixes #3001 (Enter on Android) and the "two columns after Enter" on iOS, through #3062.

## The bug

On Android, prosemirror-view deliberately ignores Enter keydowns (the IME reports composing keys as keyCode 229, so key identity can't be trusted), lets the browser split the paragraph and reads the result back from the DOM; on iOS it does the same with a 200 ms fallback. Since #2912 (0.53) BlockNote's node views ignored every mutation outside their content DOM, and the browser's new paragraph lands next to the content DOM, so the split was never read. On Android that showed as Enter inserting a space, doing nothing, or breaking tables (#3001). On iOS the fallback split the document while the stray paragraph stayed in the block, rendered next to the text by the flex block content.

## The fix

#3062 (merged into this layer, under review on main) replaces the blanket filter with one that ignores only what Dark Reader writes: attributes named `data-darkreader*` and inline `style` values containing `--darkreader`, measured with Brave and Dark Reader on the code block and toggle examples. Everything else reaches ProseMirror again, as before #2912; a node view without a content DOM keeps prosemirror-view's own default of ignoring all but the selection (the image block's resize drag otherwise reset the node selection and broke copying it, see #3062).

An earlier revision of this layer worked around the symptom by intercepting Enter on Android (`beforeinput` and `keypress`) and running the keymap chain directly. With the filter fixed, the only case the interception still covered was an Enter keypress on a selection across blocks, which prosemirror-view's own keypress handler turns into a no-op (also with Gboard on a Fairphone 5). That is accepted as a known gap for a rare pattern rather than kept as a workaround that reaches into prosemirror-view internals; the interception is gone.

## Tests

- `androidEnter.test.tsx`: a keyboard-delivered Enter and a CDP-driven IME newline commit, both pinning prosemirror-view's DOM read; red under #2912's filter. The cross-block case, `keyboardhandlers` "Enter when selection is not empty", is skipped on the android instance with the reason next to it.
- This layer also widens the android instance to the suites with distinct consumers of Enter (keyboardhandlers: the keymap chain; emojipicker: the suggestion menu's own key handling), held out of the test-infra layer because, before the filter fix, every test that presses Enter failed under the emulation. A follow-up commit fixes the instance itself: the shared setup had been forcing a scaled desktop-width iframe onto it (displacing positional input, long misread as "mouse idioms don't translate"), so it now tests true phone geometry with self-healing touch emulation, and the include list is grounded per entry on one principle: a suite runs there when it can go red for a mobile-conditional reason no other suite pins (form/ and copypaste/ dropped under that bar).
- Real keyboards, from development: Gboard sends keydown 229 plus `beforeinput: insertParagraph`, AOSP LatinIME sends 229 plus a real keydown; both take the DOM-read path. The device suite that made the discovery is parked on `mobile/emulator-layer` (#3034).

## Also on this layer (release review, 6 and 7 September)

- **Formatting toolbar hides when the user leaves the editor** (core `FormattingToolbar` extension). Its `show` state was recomputed only on selection and content changes and reset only by a pointerdown inside the editor or a drag; a blur left it, and ProseMirror keeps its selection across a blur. On a phone, tapping the page away from the editor closed the keyboard, the mobile controller unmounted, and the desktop controller mounted with the stale state and showed the desktop toolbar over a blurred editor (desktop never showed it because the desktop popover's outside-press dismiss resets the store, but that only runs while it is mounted). The extension now subscribes to `onFocusChange` with the editor's UI included: focus leaving resets the state, focus returning re-evaluates it unless a press or drag is in progress. Known edge: with focus inside the toolbar (a menu open), scrolling the selection out of view hides the toolbar and the browser drops that focus, so the toolbar is gone until the next selection change. Reproduced and verified on a Fairphone 5 over adb with the DOM read over CDP; `mobileToolbar.test.tsx` pins it (red before).
- **Side menu leaves the toolbar alone** (core `SideMenu` extension). iOS Safari delivers a tap as a hover first, one mouse move to the tap point, and drops the click when that hover changes the page. The side menu's document `mousemove` handler looked up the block under the pointer for any target inside the editor's UI too, found none under the mobile toolbar and hid a shown side menu, so the block type select and its items needed two taps whenever the side menu was showing. A move whose target is BlockNote UI outside the editor content now leaves the state as it is; on desktop that means hovering the formatting toolbar no longer moves or hides the side menu. Event and mutation logs from the simulator located it; `mobileToolbar.test.tsx` pins it (red before).
- **Viewport properties.** `useVirtualKeyboard` counts its mounted publishers and the last one out removes the `--bn-vv-*` properties from `<html>`; left behind, they pinned a `bn-scroll-container` to the keyboard-open height after a client-side navigation.
- **Tests**: the mobile toolbar's inputs are at least 16px (iOS zooms below that); the ariakit z-index pin asserts the overlap it relies on; the desktop browser instances' exclude list is built from vitest's defaults as `DESKTOP_EXCLUDE`.

Android, dragdrop, multicolumn and mobile suites green in Docker after the side menu change; android Enter and keyboard-handler suites green after the filter merge and the interception removal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android and iOS handling for Enter key input, including IME-entered line breaks.
  * Formatting toolbars now hide correctly when the editor loses focus or the keyboard is dismissed.
  * Mobile side menus remain stable when interacting with toolbars and menus.
  * Prevented Dark Reader styling updates from causing unnecessary editor refreshes or freezes.
  * Improved virtual-keyboard viewport cleanup when editors are removed.
  * Improved mobile scrolling and overscroll behavior in pinned toolbar containers.

* **Compatibility**
  * Added Android detection and expanded mobile browser coverage for more reliable behavior across devices.


nperez0111 (approved):

packages/react/src/components/FormattingToolbar/useVirtualKeyboard.ts
nperez0111, line 1: Don't love the complexity of what is happening here, but I take it that this fixes a bunch of bugs

nperez0111 (approved):

packages/react/src/components/FormattingToolbar/useVirtualKeyboard.ts
YousefED: Agree, but yes, added some workarounds in the latest iteration to fix bugs some bugs (mostly iOS).

Mostly to fix weird positioning when scrolling at the top (pull-to-refresh) / bottom ("rubber band")

matthewlipski: Unfortunately looks like the bottom scroll behaviour on iOS is still not quite right. Before, we had an issue where there was a gap at the very bottom of the document between the formatting toolbar and keyboard. Now, we can't scroll to the bottom of the page at all, and trying to do so results in a lot of jerkiness. Both cases are only applicable when not using a `.bn-scroll-container`.

Before:

https://github.com/user-attachments/assets/2c9be8e9-4545-4503-b6e9-659993aa8047

After:

https://github.com/user-attachments/assets/5412ebd0-5d93-4bb8-acff-6636960f5f54

PR #3034, opened 2026-09-01
test(device): local emulator layer — real Chrome/Gboard as normal CI

YousefED:
Fifth layer of the stack, on #3031. Adds the OS-emulator layer between browser emulation and BrowserStack hardware, per the testing-foundation discussion.

## What

The device suite's session becomes an interface with three backends; tests and page helpers are written once and run against whatever the machine can drive (`activeDevices()` probes availability):

- **`localAndroid.ts`** — a local Android emulator via Playwright's first-party (experimental) `_android` support: real Chrome as a Playwright page over CDP, input as genuine OS events via `adb shell` (Playwright's `device.input` needs a companion APK; `shell` doesn't). Page→screen coordinates come from a one-time calibration tap on an app page — never Chrome's initial page, whose 980px virtual viewport poisons the mapping. **This backend can press the on-screen keyboard itself**: `pressImeActionKey` drives Gboard's real action key with a verify ladder.
- **`browserstack.ts`** — the existing selenium client, unchanged behaviour, now one backend among several.
- **`localIos.ts`** — an iOS simulator via **Appium/XCUITest** (WebDriverAgent), driven by the same selenium client as BrowserStack: the actual iOS build and actual Safari, headless (XCUITest owns the HID stack, so the software keyboard works without the Simulator GUI), no tunnel. `mobile: tap` is the same channel BrowserStack iOS uses, so the gesture layer's chrome-offset ladders apply unchanged. Two dead ends are documented in code so nobody re-walks them: safaridriver's input is synthetic at the WebKit layer (never summons the keyboard; injecting real HID alongside trips Safari's "stop the current automated test session?" guardrail), and Appium's web-context clicks are synthetic too (`nativeWebTap` included).

`imeAction.device.test.ts` is the previously-manual release-checklist item as a regression test: **Gboard's real action key submits the link popover and focus stays in the editor** — the flow no cloud automation can exercise (BrowserStack has no channel to the on-screen keyboard).

## CI

`emulator-tests.yml` runs both targets on every PR and push to main — a KVM-backed API-35 emulator on ubuntu (**3m29s**) and an iPhone simulator via Appium on macos-15 (**5m47s**). Free minutes, no credentials, pinned actions, **both green**. Locally: boot any emulator (the setup boots the simulator itself) and `pnpm run test:device` — it runs whatever the machine can drive.

## Verified

Local matrix green: 11 tests across the API-35 emulator (Chrome 124 + Gboard) and the iPhone simulator (real iOS Safari), plus two consecutive iOS runs proving the boot/teardown cycle. CI green on both jobs. Bugs found along the way by this layer: `Array.findLast` crashes the editor on Chrome <97, and `isTouchDevice()` fails on Chrome 91's `pointer: coarse` — both filed for a support-matrix decision rather than silently patched.


YousefED:
Folded into the stack: the emulator/simulator suite now *is* the test-infra layer (#3029, with the device tests landing in #3030/#3031 alongside the fixes they verify). The final commits here also removed the BrowserStack backend entirely — the local targets reach everything it did and more (they can press the on-screen keyboard; no cloud channel can), as free per-PR CI with no credentials or tunnel. This PR's history is the reference for reviving the BrowserStack backend behind the same session interface if real-hardware coverage (OEM keyboards) ever earns automation.

PR #3039, opened 2026-09-01
fix(core): stop schema.extend() mutating the shared default specs

### LINKED ISSUE #2968: Playground editor state between examples leaks over
Reproduce:
- Go to https://playground.blocknotejs.org/basic/multi-column
- The go to basic example
- Suddenly, you can drop blocks side by side (you won't see the handle, but you can create multiple columns)

This might point to a more severe root issue where either editors are not cleaned up properly, or we leak some global state, so worth investigating

YousefED:
Fixes #2968

## Symptom

In the playground, open the multi-column example, then the basic example: you can now drop blocks side by side in the basic editor (no vertical drop indicator is shown, but columns get created).

## Root cause

Global state mutation, not editor cleanup. `CustomBlockNoteSchema.extend()` merged the added specs *into* the existing spec objects:

```ts
Object.assign(this.opts.blockSpecs, opts.blockSpecs);
```

`BlockNoteSchema.create()` with no arguments passes the module-level `defaultBlockSpecs` / `defaultInlineContentSpecs` / `defaultStyleSpecs` **by reference**, so `withMultiColumn(BlockNoteSchema.create())` wrote `column` and `columnList` into the shared defaults. Every editor created afterwards without an explicit schema (e.g. the basic example's `useCreateBlockNote()`) then inherited:

- the `column` / `columnList` node types, and
- the editor extensions those block specs register (`multiColumnDropHandler`, `columnResize`), which the `ExtensionManager` collects from `schema.blockSpecs`.

It did *not* inherit the `dropCursor: multiColumnDropCursor` editor option, which is why the drop works but no side indicator is shown.

This affects more than the playground: `BlockNoteSchema.create().extend({...})` is the pattern the docs recommend, and `withPageBreak` uses it too, so any app that creates one extended editor and later a default-schema editor got a polluted default schema.

## Fix

`extend()` merges into fresh objects and reassigns `this.opts` instead of writing into the inputs. This also stops it from mutating a caller-owned spec object passed to `create({ blockSpecs })`. The schema instance's builder semantics are unchanged (`extend` still mutates and returns `this`).

## Tests

New `packages/core/src/schema/schema.test.ts` with three cases — the shared defaults stay untouched, a caller-owned spec object stays untouched, and the end-to-end symptom (a default-schema editor created after an extended one has neither the block nor the extension it registers). All three fail on those assertions before the fix and pass after.

## Verification

- Core unit suite (766) and xl-multi-column unit suite (82) pass; `vp lint --type-aware` and `vp fmt` clean.
- Reproduced the real path in the source-aliased playground dev server (multi-column → "Basic Setup" via the sidebar, i.e. SPA navigation): **without** the fix the basic editor's schema has `["column","columnList"]` and tiptap extensions `["multiColumnDropHandler","columnResize","column","columnList"]`; **with** the fix all are empty.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed schema extensions so custom blocks, inline content, and styles no longer modify shared default configurations.
  - Prevented custom blocks and editor extensions from unexpectedly appearing in other editors using the default schema.

- **Tests**
  - Added coverage confirming schema extensions remain isolated and do not mutate provided specifications or later default-schema editors.

### DIFF
--- a/packages/core/src/schema/schema.test.ts
+++ b/packages/core/src/schema/schema.test.ts
@@ -1,0 +1,110 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { afterEach, describe, expect, it } from "vite-plus/test";
+
+import { BlockNoteSchema } from "../blocks/BlockNoteSchema.js";
+import {
+  defaultBlockSpecs,
+  defaultInlineContentSpecs,
+  defaultStyleSpecs,
+} from "../blocks/defaultBlocks.js";
+import { BlockNoteEditor } from "../editor/BlockNoteEditor.js";
+import { createExtension } from "../editor/BlockNoteExtension.js";
+import { createBlockSpec } from "./blocks/createSpec.js";
+import { createInlineContentSpec } from "./inlineContent/createSpec.js";
+import { createStyleSpec } from "./styles/createSpec.js";
+
+const editorsToCleanup: BlockNoteEditor<any, any, any>[] = [];
+
+afterEach(() => {
+  for (const editor of editorsToCleanup) {
+    editor.unmount();
+  }
+  editorsToCleanup.length = 0;
+});
+
+function createMountedEditor(
+  options: Parameters<typeof BlockNoteEditor.create>[0],
+) {
+  const editor = BlockNoteEditor.create(options);
+  editor.mount(document.createElement("div"));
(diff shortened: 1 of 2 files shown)


nperez0111 (approved):
I had found this separately on the nested blocks branch, but let's go with this first

PR #3046, opened 2026-09-03
feat: Portals cleanup

matthewlipski:
# Summary

This PR attempts to consolidate and clean up how portalling is handled in UI elements.

The trigger for this comes from the mobile formatting toolbar. It requires dropdowns to be portalled outside of it as it has `overflow: auto`. On desktop, `position: fixed` elements like the dropdowns are not clipped by an ancestor `overflow: auto` element, but this behaviour varies between platforms on mobile. Additionally, the mobile toolbar itself needs to be portalled to `document.body` as it should float on top of everything else on the page.

Originally, a proprietary `MobileToolbarPortalContext` would set a target element for the dropdowns to portal to. This was clunky as there is already the `BlockNoteView` `portalElements` prop which is supposed to handle portalling for editor UI elements. Alongside that there's `editor.portalElement`, which is set by `portalElements.default` and is used by `isWithinEditor` to check whether something is in the editor UI. However, this was also broken as it didn't take into account the other portals set by `portalElements`. The other portals were also broken as they don't receive proper styling/theming, as that relies on a container element with various attributes (things like a `.bn-root` class, `data-theme` attribute, etc). Overall, I think it's quite clear that there's a benefit to consolidating all of these things into one and fix some existing bugs along the way.

So this PR does so by introducing `PortalTarget` and `usePortalContext`, which replaces `MobileToolbarPortalContext` and is also how `portalElements` creates portals. It takes a `target` and exposes it in a `PortalContext` so any descendants always know where an element should be portalled to, by consuming said `PortalContext`. Additional `PortalTarget`s can also be added anywhere in the render tree to override the target element (useful especially for the mobile formatting toolbar where it and descendant dropdowns portal to `document.body`). The reason we need a `PortalTarget` component instead of just using a bare `PortalContext`, is because the portalled elements need to be inside a container element for styling - same issue that already exists with `portalElements`. Therefore, `PortalTarget` first checks if a `.bn-root` container element exists up the DOM tree, and creates it if necessary.

## Rationale

1. There is no unified way of creating portals internally.
2. `isWithinEditor` and `portalElements` have bugs which partially due to this.
3. `editor.portalElement` has a pretty convoluted implementation and is treated as a special case.

## Changes

- Added `PortalTarget` and `usePortalContext`.
- Refactored `react` package to always use `PortalTarget` & `usePortalContainer` to set/get portal elements.
- Replaced `editor.portalElement` with `editor._portalRoots`. This is now a map which contains all created portals, which `isWithinEditor` reads to guarantee that all editor-related elements are checked.

## Impact

N/A

## Testing

TODO

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A


packages/react/src/editor/EditorPortalProvider.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  return {
+    root: containerElement,
+    themingContainer: createPortal(
+      <div {...rootProps} ref={setContainerElement} />,
+      target,
+    ),
+  };
+}
```
nperez0111, line 74: I didn't look at this deeply, but my first thought is whether we can just always have a container? I would rather not rapidly mount/unmount elements based on effects of what their parent tree looks like.

packages/react/src/editor/PortalTarget.tsx
```diff
@@ -0,0 +1,106 @@
...
+  if (!target || needsContainer?.target !== target) {
+    return { root: null, themingContainer: null };
+  }
+
+  if (!needsContainer.value) {
+    return { root: target, themingContainer: null };
+  }
+
+  return {
+    root: containerElement,
```
YousefED: isn't this `null`, because it hasn't been set yet?

packages/react/src/editor/EditorPortalProvider.tsx
matthewlipski: We could, it would just clutter the DOM somewhat since we could get `.bn-root`s nested within other `.bn-root`s. Also for the record, the container element should only get mounted once

packages/shadcn/src/popover/popover.tsx
```diff
@@ -62,15 +62,15 @@ export const PopoverContent = forwardRef<
...
 
   const portalRoot = useContext(PortalRootContext);
-  // Default to the editor's portal element (which carries the color-scheme
-  // class) so popovers inherit light/dark mode instead of the document body's,
-  // and escape the mobile formatting toolbar's horizontal scroll clip.
-  const editor = useBlockNoteEditor();
+  // Default to the ambient portal target (a themed `.bn-root`) so popovers
+  // inherit light/dark mode instead of the document body's, and escape the
+  // mobile formatting toolbar's horizontal scroll clip.
+  const contextPortal = usePortalContext();
```
YousefED: We need better naming for this. what's the difference between `portalRoot` and `contextPortal`? `portalRoot` also comes from a context 🤷‍♂️

packages/react/src/editor/BlockNoteView.tsx
```diff
@@ -127,18 +136,20 @@ function BlockNoteViewComponent<
...
-    () => resolvePortalTarget(portalElements?.default) ?? null,
-    [portalElements?.default],
-  );
+  const editorDOMElement = useEditorDOMElement(editor);
+  const portalTarget =
+    useMemo(
+      () => resolvePortalTarget(portalElements?.default),
+      [portalElements?.default],
+    ) ??
+    editorDOMElement?.parentElement ??
```
YousefED, line 151: if this defaults to `editorDOMElement?.parentElement`, I don't think `portalTarget` is a descriptive name for this

YousefED: I'm also curious if there are race conditions here or whether we can initialize things simpler. 

`useEditorDOMElement` will be `null` initially, and then passed as `<PortalTarget>`. Than once things are mounted, it will be set to the `bn-container` (`BlockNoteViewContainer`)

packages/shadcn/src/popover/popover.tsx
matthewlipski: Updated naming, hopefully it's a bit clearer. Found it kinda difficult to have names that are not too long and also not too vague.

packages/react/src/editor/BlockNoteView.tsx
matthewlipski: To avoid confusion - `portalTarget` is put in a React a context so that any popovers further down the render tree can use it to figure out where they need to portal to. So `editorDOMElement?.parentElement` is not special just because it's the editor container element, `portalTarget` is just saying to any popovers that they should portal there if they need to. In any case, hopefully things are clearer with the updated naming.

PR #3054, opened 2026-09-04
feat: Remove `useRegisterPortalRoot` from ShadCN

matthewlipski:
# Summary

Generally, the components in the `ariakit`, `mantine`, and `shadcn` packages that are used for the `ComponentsContext` are supposed to be "dumb" adapters for primitives like buttons, menus, etc. In #3046, this contract was broken somewhat as ShadCN components required the use of `useEditorPortalElement`. This is because unlike the other 2 UI libs, which render dropdowns as siblings of their trigger elements, ShadCN portals them to `body` by default. This doesn't work for us as that puts them outside container elements which provide additional theming and styling (basically the same issue #3046 addresses). Therefore, the default portal location was overridden to the return of `useEditorPortalElement`.

To ensure consistency across UI libs then, this PR makes it so that the `portalRoot` prop, which tells an Ariakit/Mantine/ShadCN component if & where it should portal its dropdown to, is always passed the element returned by `useEditorPortalElement`.

Additionally, the `preventFocusOnOpen` prop has been added. This is so that for mobile, we can disable dropdowns from grabbing focus from the editor on open, as this would close the virtual keyboard. But on desktop, this should be enabled for better accessibility. This was previously set based on if `portalRoot` was defined, which isn't quite right. Yes, it works in our case because the only time we'd actually set `portalRoot` was for the mobile formatting toolbar. But there is no direct correlation between a dropdown needing to be portalled, and preventing focus from moving to a dropdown on open, so these 2 concerns have been separated.

## Rationale

See above.

## Changes

See above.

## Impact

N/A

## Testing

E2E snapshots will need updates, but I'm holding off on this until we decide whether this change is smth we want.

## Checklist

### DIFF
--- a/packages/mantine/src/menu/Menu.tsx
+++ b/packages/mantine/src/menu/Menu.tsx
@@ -16,7 +16,15 @@ const SubMenuContext = createContext<
 >(undefined);
 
 export const Menu = (props: ComponentProps["Generic"]["Menu"]["Root"]) => {
-  const { children, onOpenChange, position, portalRoot, sub, ...rest } = props;
+  const {
+    children,
+    onOpenChange,
+    position,
+    portalRoot,
+    preventFocusOnOpen,
+    sub,
+    ...rest
+  } = props;
 
   assertEmpty(rest);
 
@@ -38,9 +46,9 @@ export const Menu = (props: ComponentProps["Generic"]["Menu"]["Root"]) => {
     <MantineMenu
       withinPortal={!!portalRoot}
       portalProps={portalRoot ? { target: portalRoot } : undefined}
-      // Do not move focus to dropdown when portaled (mobile), as it blurs the
-      // editor's contentEditable and dismisses the on-screen keyboard.
-      trapFocus={portalRoot ? false : undefined}
+      // Do not move focus to the dropdown when requested (mobile), as it blurs
+      // the editor's contentEditable and dismisses the on-screen keyboard.
+      trapFocus={preventFocusOnOpen ? false : undefined}
       middlewares={{ flip: true, shift: true, inline: false, size: true }}
       onChange={onOpenChange}
       position={position}
--- a/packages/react/src/editor/ComponentsContext.tsx
+++ b/packages/react/src/editor/ComponentsContext.tsx
@@ -48,6 +48,13 @@ type ToolbarSelectType = {
   }[];
   isDisabled?: boolean;
   portalRoot?: HTMLElement | null;
+  /**
+   * When true, the surface must not move DOM focus onto itself when it opens.
+   * On mobile, stealing focus blurs the editor's `contentEditable` and
+   * dismisses the on-screen keyboard. Adapters map this to their own library's
+   * focus mechanism.
+   */
+  preventFocusOnOpen?: boolean;
 };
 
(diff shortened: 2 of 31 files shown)


packages/mantine/src/menu/Menu.tsx
```diff
@@ -38,9 +46,9 @@ export const Menu = (props: ComponentProps["Generic"]["Menu"]["Root"]) => {
     <MantineMenu
       withinPortal={!!portalRoot}
       portalProps={portalRoot ? { target: portalRoot } : undefined}
-      // Do not move focus to dropdown when portaled (mobile), as it blurs the
-      // editor's contentEditable and dismisses the on-screen keyboard.
-      trapFocus={portalRoot ? false : undefined}
+      // Do not move focus to the dropdown when requested (mobile), as it blurs
+      // the editor's contentEditable and dismisses the on-screen keyboard.
+      trapFocus={preventFocusOnOpen ? false : undefined}
```
YousefED, line 51: nice improvement, so much better to have explicit naming like this

packages/shadcn/src/menu/Menu.tsx
```diff
@@ -81,11 +84,9 @@ export const MenuDropdown = forwardRef<
...
   const ShadCNComponents = useShadCNComponentsContext()!;
 
-  const portalRoot = useContext(PortalRootContext);
-  // Default to the ambient portal target (a themed `.bn-root`) so the menu
-  // inherits light/dark mode instead of the document body's.
-  const editorPortalElement = useEditorPortalElement();
-  const container = portalRoot ?? editorPortalElement ?? undefined;
+  // The `portalRoot` supplied at the call site is a themed `.bn-root`, so the
+  // menu inherits light/dark mode instead of the document body's.
+  const container = useContext(PortalRootContext) ?? undefined;
```
YousefED: double check; but all scenarios keep working like this? because it feels like one option has been dropped here compared to `container = portalRoot ?? editorPortalElement ?? undefined;`

packages/shadcn/src/toolbar/Toolbar.tsx
YousefED, line 76: there's still a dependency on `useEditorPortalElement` here. can you check all occurences?

YousefED (commented):
If this works it's SO much cleaner. Love both improvements!!

packages/react/src/editor/ComponentsContext.tsx
```diff
@@ -48,6 +48,13 @@ type ToolbarSelectType = {
   }[];
   isDisabled?: boolean;
   portalRoot?: HTMLElement | null;
```
YousefED: It seems like the pattern is now that we always pass `portalRoot`. In that case, maybe let's make this non-optional so we can validate the codebase does this correctly?

packages/shadcn/src/toolbar/Toolbar.tsx
YousefED: tried, but was cumbersome for Toolbar and Badges. Let's keep these exceptions for now (added two comments in code)

packages/react/src/editor/ComponentsContext.tsx
YousefED: done, see commit

PR #3058, opened 2026-09-07
fix(core): link lookups at the edges of a link

YousefED:
Two link lookups in `StyleManager` disagreed about what happens at the edges of a link, because the link mark is non-inclusive.

**Edit and delete** (`editLink`, `deleteLink` → `getLinkMarkAtPos`): the position comes from the link toolbar and can sit at either edge of the link (caret at the end of a link is normal). The lookup now uses tiptap's `getMarkRange`, which looks at the node after the position and then the one before, so the link is found at both ends; the `+ 1` nudge the callers used, which only covered the left edge, is gone.

**Pre-filling the link form** (`getSelectedLinkUrl`): this answers "is the selected text a link?", so it deliberately reads only the node the selection starts in. Looking on both sides would pre-fill the previous link's URL when a selection starts right after a link. The bug it fixes: selecting only the last character of a link, or a one-character link, read as "no link" and opened the form empty.

Tests: `StyleManager.test.ts` (headless, the selection lookup at every edge; two cases fail on the old lookup) and `StyleManager.browser.test.ts` (mounted editor; editing a link with only its last character selected used to split the link).

Extracted from the mobile stack so it can be reviewed on its own; the stack currently carries a `+ 1` in `getSelectedLinkUrl` that this replaces.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved link selection handling when editing links, including selections at link boundaries and document endpoints.
  - Prevented link edits from unintentionally splitting link text.
  - Fixed link deletion so it removes link formatting while preserving the text.
  - Corrected link URL detection to avoid applying a preceding link’s URL when the cursor is immediately after it.
- **Tests**
  - Added coverage for link editing, deletion, URL detection, and boundary selections.


YousefED:
@nperez0111 this was claude generated, not directly related to what I was working on, so might be worth a look at any time. I didn't really review this atm

nperez0111 (approved):
Verified locally that this covers #3081 / #3073: editing at the end replaces the whole link, deleting at the end preserves the text and removes the link, and calls at doc.content.size do not throw. getMarkRange also handles one-character links, which #3081 still misses.

Added focused regression tests reusing the existing fixture. All 8 cases pass in Chromium, Firefox, and WebKit in Docker (24 checks); the 7 existing lookup unit tests and lint also pass. Positions outside the document remain outside this fix’s scope.

PR #3051, opened 2026-09-04
refactor(core): simplify the BlockInfo API into a single vocabulary for block plumbing

nperez0111:
## Summary

Reworks `BlockInfo` into a single, uniform vocabulary for reading and writing blocks in ProseMirror positions, removing most of the manual `+1`/`-1` position arithmetic sprinkled across the codebase.

### New BlockInfo model

`BlockInfo` is now a discriminated union over what a block contains:

- `block` — wrapper-level info (`block`, `children`)
- `content` / `hasContent` — adds `contentStart`, `contentEnd`, `contentKind` (read from the node spec via `NodeSpec.blockConfig`), `isContentEmpty`
- `children` — adds `childrenStart`/`childrenEnd` (`ChildrenInfo`)

### Producers reduced 6 → 4

- `getBlockInfoFromNode`, `getBlockInfoAt`, `getBlockInfoNearPos`, `getBlockInfoFromSelection`
- plus navigation helpers: `getParentBlockInfo`, `getPrevBlockInfo`, `getNextBlockInfo`, `getLastDescendantBlockInfo`
- new `blockEdgePos` / `blockEdgeSelection` / `tableContentCaretPos` helpers replace hand-rolled table caret arithmetic in `textCursorPosition.ts`

### New insertBlocks placements

`insertBlocks` now supports `"first-child"` and `"last-child"` in addition to `"before"`/`"after"`, via a shared `getInsertionPos` helper that descends into wrapper blocks as needed.

### NodeSpec.blockConfig

Block specs now attach their resolved `BlockConfig` onto the TipTap node spec (`extendNodeSchema`), so low-level code can read `contentKind` from the schema instead of re-deriving it by string-matching node names. Hand-written nodes (via `createBlockSpecFromTiptapNode`) are validated with `checkNodeMatchesConfig`.

### Adopted across the codebase

`KeyboardShortcutsExtension`, list keyboard shortcuts, `insertBlocks`/`mergeBlocks`/`moveBlocks`/`nestBlock`/`replaceBlocks`/`splitBlock`/`updateBlock`, selection & text-cursor position handling, paste/file-insertion transforms, and the AI package tests.

## Test plan

- `pnpm run lint` — clean
- `pnpm run test` — all green (core incl. new `getBlockInfoFromPos.test.ts` and `insertPlacement.test.ts`)

## Summary by CodeRabbit

* **New Features**
  * Added `"first-child"` and `"last-child"` placement options to `insertBlocks` for nested insertion.
[...]

Issue #2378, opened 2026-01-23
Toggle block content is moved to next toggle block


virgile-dev:
**Describe the bug**
Content from 1st toggle block is moved to the newly created toggle block

**To Reproduce**
1. Hit enter on a folded toggle block with content
2. see another toggle block appear with the content from the first toggle block

https://github.com/user-attachments/assets/721b5674-1360-437a-9b03-09a21f5ac95e

**Misc**

- Node version:
- Package manager:
- Browser:
- [x] I'm a [sponsor](https://www.blocknotejs.org/pricing) and would appreciate if you could look into this sooner than later 💖

woutervroege:
Thanks @virgile-dev.
Separating these issues going forward:

1. I am not able to reproduce this particular issue. Could you specify your environment (OS, Browser?). Would you able to reproduce this issue in our playground?

2. Think this is a duplicate of https://github.com/TypeCellOS/BlockNote/issues/2109

virgile-dev:
@woutervroege I just updated the issue with the current problem we're facing and which I was able to reproduce today on BlockNotes interactive demo : https://www.blocknotejs.org/demo#n3gbb

PR #3059, opened 2026-09-07
Container blocks: compartments, renderFrame, repair, validation, exporters

### LINKED ISSUE #2020: Enter Key Breaks Children in Custom Toggleable Blocks
**Describe the bug**
When pressing Enter inside a custom toggleable block, the children are broken unexpectedly.
- If the cursor is on the toggle title, all children should be preserved, but they are not.
- If the cursor is inside the children, pressing Enter should move the cursor to the next child (if it exists) or create a new default block.

**To Reproduce**
1. Open the example: https://www.blocknotejs.org/examples/custom-schema/toggleable-blocks
2. Place the cursor in the toggle title and press `Enter`.
3. Observe that the children are broken.

**Misc**

- Node version: 22.13.0
- Package manager: npm
- Browser: Firefox

https://github.com/user-attachments/assets/eb85ca78-7b74-497d-944e-27017ba6f675

nperez0111:
Stacked on #3051 (BlockInfo API refactor) — this branch is rebased onto `refactor/block-info-api` and adopts its vocabulary throughout (producers, `NodeSpec.blockConfig`, shared `getInsertionPos`); the parallel home-grown implementations are gone.

## What this adds on top of #3051

**Compartments** — `content: "inline"` + `children` coexist, giving blocks a real rich-text title with a body of child blocks (fixes #2020, #2378). Title/body editing behaves as one unit: Enter splits into the body, Backspace merges back, Shift-Tab stops at the body edge.

**`renderFrame`** — second hook beside `render` that draws the box around content + children (`{ dom, slot, update? }`). Returning `undefined` declines the frame (plain nesting) — the toggle pattern. Pure containers can draw their box in `renderFrame` alone with in-place `update`; React renders pure-container frames live and installs compartment frames as static snapshots.

**Derived repair** — dissolve-vs-pad replaces configured strategies: below-`min` anywhere-containers dissolve into survivors (counted on content, not padded empties), `containerOnly` blocks pad, emptied container children are dropped while emptied regular blocks are kept.

**Fail-fast validation** — bad `content`+`children` combos, regular blocks in `allow`, require-cycles, and missing `render`/`renderFrame` all throw at spec-definition time.

**Dropped as YAGNI** — `default`, `whenEmptied`, `boundary: sealed`, `rootDOM`, container `runsBefore` validation, `removeEmptyChildren` export.

**Examples/docs** — `09-container-block` rewritten as a Panel (live frame + flavor switcher), new `13-callout-block` headline demo (real inline title), container-blocks docs page updated.

## Test plan

- `vp run lint` (type-aware) — clean
- core 907, tests/unit 947, react, multicolumn, all five exporters, xl-ai — green
- e2e (browser suite) left for CI — Docker unavailable locally

## Summary by CodeRabbit

* **New Features**
  * Added support for custom container blocks that can hold nested child blocks, including titled containers and configurable child restrictions.
  * Added container-aware editing, insertion, nesting, moving, merging, keyboard navigation, and side-menu interactions.
  * Added React rendering and framing support for container blocks.
  * Added container-block examples for panels and callouts.

* **Bug Fixes**
  * Improved preservation of children and content when changing block types.
  * Improved HTML, Markdown, DOCX, ODT, PDF, email, and Typst export handling for containers.
  * Improved multi-column editing, resizing, dragging, and structural recovery.

* **Documentation**
  * Added comprehensive container-block guidance and updated custom block, insertion, and Typst export documentation.


examples/06-custom-schema/12-alert-blocks/vite.config.ts
```diff
@@ -0,0 +1,35 @@
...
+export default defineConfig(((conf: { command: string }) => ({
+  plugins: [react()],
+  optimizeDeps: {},
+  build: {
+    sourcemap: true,
+  },
+  resolve: {
+    alias:
+      conf.command === "build" ||
+      !fs.existsSync(path.resolve(__dirname, "../../packages/core/src"))
```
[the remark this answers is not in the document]
nperez0111: Fixed in 4f3ade8de: corrected the existence guard and core/react aliases in the shared generator and regenerated all 104 example Vite configs. Verified every resulting source path exists; generated changes are limited to the three path corrections per file.

examples/06-custom-schema/13-callout-block/vite.config.ts
```diff
@@ -0,0 +1,35 @@
...
+        : ({
+            // The repo-wide alias for the shared test-utils directory (private,
+            // so it only resolves inside the monorepo). Harmless for examples
+            // that don't use it.
+            "@shared": path.resolve(__dirname, "../../../shared/"),
+            // Comment out the lines below to load a built version of blocknote
+            // or, keep as is to load live from sources with live reload working
+            "@blocknote/core": path.resolve(
+              __dirname,
+              "../../packages/core/src/",
```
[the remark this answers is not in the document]
nperez0111: Fixed in 4f3ade8de: the generator and generated callout config now use ../../../packages for the existence guard and both source aliases. The existing guard previously disabled these aliases rather than enabling them with invalid paths.

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -874,6 +1039,69 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
+              // Everything after the cursor moves into the new block, so
+              // splitting the title mid-way puts its tail at the top of the
+              // body instead of handing the body to a new sibling.
+              const tail = blockInfo.content.node.cut(
+                state.selection.from - blockInfo.contentStart,
+              );
+              const newBlock = state.schema.nodes["blockContainer"].create(
+                undefined,
+                state.schema.nodes["paragraph"].create(undefined, tail.content),
+              );
```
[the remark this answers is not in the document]
nperez0111: This configuration is rejected by validateChildrenConfigs: blocks with inline or plain content only support children: { allow: "blocks" }. In particular, a titled block with allow: ["heading"] cannot be constructed. The existing regression tests cover this restriction, so the Enter handler does not need to select a restricted child type. Resolving as not applicable.

tests/src/unit/react/reactFrame.test.tsx
```diff
@@ -0,0 +1,465 @@
...
+          type: "frameAlert",
+          props: { framed: false },
+          content: "Title",
+          children: [{ type: "paragraph", content: "Body" }],
+        },
+      ]);
+      expect(html).not.toContain("alert-frame");
+      expect(html).toContain("Title");
+      expect(html).toContain("Body");
+      expect(activeFrames).toBe(0);
```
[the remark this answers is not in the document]
nperez0111: Addressed in 4f3ade8de by asserting activeFrames === 0 after each test has unmounted and destroyed its editor. This reports leaks in the test that caused them. Unconditionally resetting the counter would hide a lifecycle defect. All 8 React frame tests pass with the teardown assertion.

packages/core/src/schema/blocks/validateChildren.ts
```diff
@@ -0,0 +1,68 @@
+import { isContainerConfig } from "./children.js";
+import type { BlockConfig, ChildrenConfig } from "./types.js";
+
+type ValidatableConfig = Pick<BlockConfig, "type" | "content" | "placeable"> & {
+  children?: ChildrenConfig;
+};
+
+/** Reject declarations ProseMirror would accept with different semantics. */
```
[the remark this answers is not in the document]
nperez0111: This is an intentionally accepted edge case, as agreed during the PR review. A mutually required container-only cycle has no finite valid document; we are leaving that invalid custom-schema configuration to ProseMirror rather than restoring an additional DFS validator. Resolving with no code change; recursive configurations with a terminating alternative remain supported.

examples/06-custom-schema/09-container-block/src/styles.css
```diff
@@ -0,0 +1,106 @@
...
+  justify-content: center;
+  margin-top: 2px;
+}
+
+.panel-icon-button:hover {
+  background-color: var(--bn-colors-hovered-background);
+}
+
+.panel-icon-button:focus-visible {
+  outline: 2px solid currentColor;
```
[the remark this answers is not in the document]
nperez0111: Already addressed by the subsequent example simplification: the outline declaration and currentColor keyword are no longer present in this stylesheet. Current lint passes. Resolving as obsolete.

nperez0111:
Addressed the remaining review feedback in 4f3ade8de. In addition to the inline-thread replies, the outside-diff validation finding is fixed: children.allow entries must name configured blocks, with regressions for unknown names, internal node-group names, and inherited object properties. The empty-title Enter test now also verifies existing child IDs are preserved. Validation: 34 core tests and 8 React frame tests passed; lint and commit checks passed. The shared alias-generator fix regenerates 104 configs, each with only the same three path corrections.

docs/content/docs/features/export/typst.mdx
```diff
@@ -120,6 +120,22 @@ For a block with inline content, render it the way the default mappings do:
 `exporter.transformInlineContent(block.content).join("")` (inline results are
 markup strings, so plain concatenation composes them).
 
+### Container blocks
```
YousefED, line 123: I'm not sure this is a useful addition to the docs

docs/content/docs/reference/editor/manipulating-content.mdx
```diff
@@ -165,14 +165,16 @@ editor.insertBlocks(
...
+// Insert a paragraph as the last child of a container block
 editor.insertBlocks(
   [{ type: "paragraph", content: "Nested paragraph" }],
-  "existing-block-id",
+  "container-block-id",
   "last-child",
 );
 ```
 
+For [container blocks](/docs/features/custom-schemas/container-blocks), `"first-child"` inserts at the beginning of the container and `"last-child"` inserts at the end. Use `"before"` or `"after"` to insert next to the container instead.
```
YousefED, line 176: not specific to containers right? (I mean, first-child and last-child can be used for regular blocks as well?)

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,194 @@
...
+`children.allow:` Set to `"blocks"` to accept the editor's block types. You can also restrict a container to specific container types, as explained in [Restricting Children](#restricting-children).
+
+`propSchema:` Defines the container's props, just like for other custom blocks. Use these to customize its appearance or behavior.
+
+### Block Implementation
+
+`render:` Your React component defines how the block should look. With `content: "none"`, attach `contentRef` where the child blocks should appear. With `content: "inline"` or `"plain"`, attach it to the block's own editable text. You can add icons, buttons, or other elements around it:
+
+```tsx
+render: (props) => (
```
YousefED, line 49: should this be `renderFrame`?

docs/content/docs/features/custom-schemas/container-blocks.mdx
```diff
@@ -0,0 +1,194 @@
...
+});
+```
+
+If you create a panel without specifying its children, it starts with an empty paragraph. To let users insert panels themselves, add a [custom Slash Menu item](/docs/react/components/suggestion-menus#changing-slash-menu-items), as shown in the demo.
+
+## Combining Content and Child Blocks
+
+A block can have both its own text and child blocks. Use this for a question followed by hints, a checklist item with detailed instructions, a code sample followed by explanatory blocks, or a callout with a heading. In the demo below, we use the block's text as a callout title:
+
+<Example name="custom-schema/callout-block" />
```
YousefED, line 110: Let's streamline the Alert + Callout examples (we now have an inconsistent mix I think)

packages/xl-multi-column/src/blocks/Columns/index.ts
```diff
@@ -1,28 +1,68 @@
...
-    node: Column,
-    type: "column",
+    type: "column" as const,
+    propSchema: {
+      width: {
+        default: COLUMN_WIDTH_DEFAULT,
+      },
+    },
     content: "none",
+    children: { allow: "blocks" },
```
YousefED, line 17: previously, column could not contain ColumnList. Now, this is possible. Is there a way to prevent this?

packages/xl-docx-exporter/src/docx/docxExporter.test.ts
```diff
@@ -416,6 +417,163 @@ describe("exporter", () => {
   );
 });
 
+describe("custom container blocks", () => {
```
YousefED, line 420: Instead of adding these to all separate exporter tests, should we just extend the default testDocument?

packages/core/src/pm-nodes/BlockContainer.ts
```diff
@@ -88,4 +145,57 @@ export const BlockContainer = Node.create<{
...
+            (frameView.stopEvent?.(event) ?? false)
+          );
+        },
+        destroy: frameView.destroy?.bind(frameView),
+        selectNode: frameView.selectNode?.bind(frameView),
+        deselectNode: frameView.deselectNode?.bind(frameView),
+        ignoreMutation: frameView.ignoreMutation?.bind(frameView),
+      };
+      if (framed) {
+        ignoreNonContentMutations(nodeView);
```
YousefED, line 196: this caused a lot of issues recently (the dark-reader saga). See if it's still needed (and/or update with main first)

packages/core/src/pm-nodes/BlockContainer.ts
```diff
@@ -88,4 +145,57 @@ export const BlockContainer = Node.create<{
       contentDOM: block,
     };
   },
+
+  addNodeView() {
+    // Cast: this returns a plain ProseMirror node view, which tiptap's
+    // `NodeViewRenderer` type doesn't model.
+    return ((props: NodeViewRendererProps) => {
```
YousefED, line 152: I think we can skip the node view entirely for blocks that never draw a frame (e.g.: default blocks)

packages/core/src/pm-nodes/BlockContainer.ts
```diff
@@ -1,10 +1,67 @@
...
+  return {
+    dom,
+    contentDOM: frame?.slot ?? fallback,
+    destroy: frame?.destroy?.bind(frame),
+    update(node) {
+      if (frame?.update) {
+        return frame.update(nodeToBlock(node, props.view.state.doc)) !== false;
+      }
+      // Declined frames must also be reconsidered when their block changes.
+      return !renderFrame || node.eq(props.node);
```
YousefED, line 60: a frame with no update hook falls through to node.eq(props.node), which compares the whole subtree — so it rebuilds on every keystroke. Desired? Make update mandatory or "smarter"?

packages/core/src/pm-nodes/BlockContainer.ts
```diff
@@ -1,10 +1,67 @@
...
 import type { BlockNoteEditor } from "../editor/BlockNoteEditor.js";
+import { nodeToBlock } from "../api/nodeConversions/nodeToBlock.js";
+import { isDocumentFragment } from "../schema/blocks/internal.js";
 import { BlockNoteDOMAttributes } from "../schema/index.js";
+import { ignoreNonContentMutations } from "../schema/nodeViewMutations.js";
 import { mergeCSSClasses } from "../util/browser.js";
 import { suggestionMarks } from "./suggestionMarks.js";
 
+/** Adapts vanilla frames to the same lifecycle as framework node views. */
+function createFrameView(
```
YousefED, line 17: I found this file a bit difficult to navigate, I think it comes down to these things:

- createFrameView is returning a NodeView, but not actually used as NodeView (more like a "nested nodeview"?). Maybe different naming?
- We also return a NodeView for default nodes that don't define a frame, or, in claude's words:

> Return NodeView | undefined from the frame adapter. Today it always returns a NodeView, even with no frame, so the caller has to recover that from frameView.dom !== contentDOM — a sentinel encoded in object identity. Making undefined mean declined deletes the fallback parameter, both ?? contentDOM fallbacks, and the framed flag.

- If I understand correctly, some of this is because we want to retain the `bn-block-outer` shape. After a quick investigation, I think that structure is only used for animations (PreviousBlockType). We should probably get rid of this and do animations in a simpler way that don't require the dom nesting (unless the dom nesting is also needed for sth like bullets / children / etc). Anyway, that should be a separate project

packages/core/src/schema/blocks/containerAttributes.ts
```diff
@@ -0,0 +1,37 @@
...
+  blockType: string,
+  blockProps: Partial<Props<PSchema>>,
+  propSchema: PSchema,
+  id?: string,
+) {
+  if (!element) {
+    return;
+  }
+
+  for (const [prop, value] of Object.entries(blockProps)) {
```
YousefED, line 23: fyi,

The same rule, written five ways
All of them encode one policy: props become data-kebab attributes, except those at their default.

where	shape	removes?
containerAttributes.ts:23 (selected)	imperative loop	yes
blocks/internal.ts:237	imperative loop	no
blocks/internal.ts:63	tiptap renderHTML, returns {}	n/a
inlineContent/internal.ts:40	filter().map().forEach()	no
ReactBlockSpec.tsx:145 / ReactInlineContentSpec.tsx:97	filter().map() → fromEntries → JSX spread	no
Table/block.ts:280	imperative loop over the schema	yes
The closest pair is your selected code and blocks/internal.ts:237, which differ only in that one removes and one doesn't:


// containerAttributes.ts                        // blocks/internal.ts
for (const [prop, value] of ...) {               for (const [prop, value] of ...) {
  const attribute = camelToDataKebab(prop);        const spec = propSchema[prop];
  if (value === undefined ||                       if (value !== spec?.default) {
      value === propSchema[prop]?.default) {         blockContent.setAttribute(
    element.removeAttribute(attribute);                camelToDataKebab(prop), value);
  } else {                                         }
    element.setAttribute(attribute, String(value));
  }                                              }
}
The splits are real, but only two of them
Set-only vs set-and-remove is a genuine difference: three sites build a fresh element, so there's nothing to remove; two patch an element that already exists across updates. That's two behaviours, not six.
Table/block.ts iterates the schema, not the props. Everyone else iterates Object.entries(blockProps), which silently skips a prop absent from the object — the schema-driven loop clears a stale attribute even when the prop isn't present. That's arguably the more correct of the two, and it's the odd one out.
The rest — imperative vs filter/map, DOM vs JSX — is incidental. React genuinely needs a props object rather than DOM calls, but it could still get it from a shared propsToDataAttributes(blockProps, propSchema) returning Record<string, string>, then fromEntries at the call site.
Worth flagging, with a caveat
Three small things would collapse it: one function producing the entries, one applyDataAttributes(element, entries) for the set-and-remove case, and the React sites spreading the entries. That kills four of the five duplicates and puts the default-skipping rule — which is a serialization contract, since parseHTML has to mirror it exactly — in one place.

The caveat for this review: containerAttributes.ts is new in #3059, but four of the five sites predate it. So it's "the PR added a fifth copy of an existing pattern", not "the PR introduced duplication" — which makes it a reasonable "while you're here" note rather than a blocker, and the consolidation itself is better as its own change.

packages/core/src/schema/blocks/internal.ts
```diff
@@ -157,6 +157,39 @@ export function getBlockFromNodeView(
...
 }
 
+/**
+ * `Node.DOCUMENT_FRAGMENT_NODE`, inlined. Server-side rendering shims only
+ * `document` and `window` onto the global scope, so `Node` and
+ * `DocumentFragment` are undefined there and `instanceof` throws.
+ */
+const DOCUMENT_FRAGMENT_NODE = 11;
+
+export function isDocumentFragment(
```
YousefED, line 167: Is DocumentFragment support for frames needed? Fragments in render clearly earn their place — CheckListItem and the code block both rely on siblings without a wrapper. But for renderFrame I can't find a producer: nothing in the repo returns one, and React's path can't, since renderToDOMSpec returns firstElementChild. The only thing exercising it is the ContentFrame test fixture, and it costs us the display: contents wrapper in BlockContainer.ts. Since frames are new here, we could narrow renderFrame's dom to HTMLElement and drop that branch — unless there's a use case in mind?

packages/core/src/schema/blocks/createSpec.ts
```diff
@@ -81,280 +215,241 @@ export function getParseRules<
...
- * thing, e.g. `"(text)*"` and `"text*"`. Unwraps a parenthesized single
- * term, with or without a trailing quantifier, so equivalent spellings
- * compare equal. Anything with real structure (sequences, alternation) is
- * left as-is: unwrapping those would change the expression's meaning.
- */
-function normalizeContentExpression(expression: string): string {
-  const trimmed = expression.trim();
-  const match = trimmed.match(/^\(([A-Za-z_][A-Za-z0-9_]*)\)([*+?])?$/);
-  return match ? `${match[1]}${match[2] ?? ""}` : trimmed;
+export function containerRootDOM(output: {
```
YousefED, line 225: maybe drop (see other comment)

packages/core/src/api/exporters/html/util/serializeBlocksInternalHTML.ts
```diff
@@ -197,10 +206,30 @@ function serializeBlock<
...
-  bc.contentDOM?.appendChild(ret.dom);
+  // Frames wrap the content and its child group in static HTML too. The DOM
+  // render context lets interactive frames export without browser view state.
+  const renderFrame = impl.renderFrame<I, S>;
+  const frame = renderFrame?.call(
+    {
+      renderType: "dom",
+      props: undefined,
+      blockContentDOMAttributes:
+        editor._tiptapEditor.extensionManager.extensions.find(
```
YousefED, line 217: the _tiptapEditor lookup seems a little hacky. is that needed?

packages/core/src/api/exporters/html/util/serializeBlocksExternalHTML.ts
```diff
@@ -224,27 +226,37 @@ function serializeBlock<
...
-      { ...block, props } as any,
+      blockWithDefaults as any,
       editor as any,
     );
 
   const elementFragment = doc.createDocumentFragment();
 
-  if ((ret.dom as HTMLElement).classList.contains("bn-block-content")) {
+  // React renders can return a fragment around the root element.
+  const rootElement = containerRootDOM(ret);
```
YousefED, line 248: > would the codebase be simpler if "render" would never return a frame, but only renderframe does?

Not sure this is 100% correct, but I have the feeling the code will be easier to follow in that case. This is what Claude gave me (I think most of it still holds up if we migrate toggle blocks)

> Yes, and the collision is real — here's the evidence.

> The bug the merge causes
serialize returns { ...output, dom: frame.dom, childrenDOM: frame.slot }, and childrenDOM: frame.slot overwrites whatever the block's own render asked for. A block that both places its children (<details>) and draws a frame loses the first:


<div class="FRAME">
  <div class="SLOT">
    <div>
      <div class="bn-inline-content title">Title</div>
      <details class="DETAILS"></details>   ← empty
    </div>
    <p data-nesting-level="1">CHILD</p>     ← escaped to the frame slot
  </div>
</div>
The child should be inside <details>. It isn't, because two different concepts are being carried on one field.

Why your instinct is right
childrenDOM predates frames — it's main's existing mechanism for "this block decides where its children go", used by collapsible Heading and ToggleListItem. The PR reuses it to mean "the frame's slot", and render's return value now carries three things at once: the content DOM, an optional wrapper around it, and children placement. Once ret.dom may or may not be the content root, every caller has to discover which it got — that's exactly what forces closest(".bn-block-content"), and what makes the append + replaceWith dance necessary instead of a straight unwrap.

If render always returned just the block's content, the cascade unwinds:

blockContentRoot collapses back to rootElement — no DOM search, no class-based lookup
childrenDOM keeps one meaning, and the override bug above can't exist
serialize's else if (context && !external) branch disappears, so context?: { nestingLevel } stops doubling as a "we're exporting externally" mode flag
the two serializers become symmetric — internal already calls renderFrame itself
The cost is that the external serializer would have to call renderFrame too. That sounds like adding work, but it's really consolidating it: there are already two renderFrame call sites assembling their context differently, one of which scans editor._tiptapEditor.extensionManager.extensions to rebuild what the other has in closure. One shared applyFrame(block, editor, content) helper would serve both and fix that layering issue at the same time.

Caveat on scope
The override is only reachable for a block declaring both renderFrame and a childrenDOM from render — nothing in the repo does, and the two shipped childrenDOM users have no frame. So it's a latent bug, not a live one. But it's good evidence for the design point: the merge is unsound in a way that only shows up when the two features meet, which is precisely the kind of thing that surfaces after a custom-block author combines them.

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -560,52 +559,70 @@ export function prosemirrorSliceToSlicedBlocks<
     blockCutAtStart: string | undefined;
     blockCutAtEnd: string | undefined;
   } {
-    if (node.type.name !== "blockGroup") {
+    // Both `blockGroup` and container nodes (columnList, column, callout,
+    // ...) hold block children directly, so both can be processed here.
+    if (node.type.name !== "blockGroup" && !isContainerNode(node.type)) {
```
YousefED, line 564: nice fix!

packages/core/src/api/nodeConversions/nodeToBlock.ts
```diff
@@ -560,52 +559,70 @@ export function prosemirrorSliceToSlicedBlocks<
...
       throw new Error("unexpected");
     }
     const blocks: Block<BSchema, I, S>[] = [];
     let blockCutAtStart: string | undefined;
     let blockCutAtEnd: string | undefined;
 
     node.forEach((blockContainer, _offset, index) => {
```
YousefED, line 571: at this point we're not sure if it's a blockContainer, and it can be sth else (a container that's not a `blockContainer`), so better to use a different variable name

packages/react/src/schema/ReactBlockSpec.tsx
```diff
@@ -219,131 +259,277 @@ export function createReactBlockSpec<
         ? blockImplementationOrCreator(options as any)
         : blockImplementationOrCreator;
 
+    if (!blockImplementation.render) {
```
YousefED, line 262: the whole `createReactBlockSpec` function is quite difficult to read which makes it hard to review, but I think that's mostly pre-existing and some if it is actually cleaned up

packages/react/src/schema/ReactBlockSpec.tsx
```diff
@@ -219,131 +259,277 @@ export function createReactBlockSpec<
...
-                          element.className = mergeCSSClasses(
-                            "bn-inline-content",
-                            element.className,
-                          );
-                          element.dataset.nodeViewContent = "";
-                        }
-                      }}
-                    />
-                  </BlockContentWrapper>
+                  <NodeViewWrapper style={DISPLAY_CONTENTS}>
```
YousefED, line 465: can / should we make this more consistent with `BlockContentWrapper`?

i.e.: introduce a new component that we pass props, instead of `applyContainerAttributes` call above? (not sure if this is entirely possible)

packages/core/src/pm-nodes/BlockContainer.ts
```diff
@@ -1,10 +1,67 @@
-import { Node } from "@tiptap/core";
+import {
```
YousefED, line 1: would be cool if at some point we can migrate BlockContainer to the container API, I think that could remove a lot of (old) code

packages/core/src/api/blockManipulation/commands/splitBlock/__snapshots__/splitBlock.test.ts.snap
```diff
@@ -621,23 +621,6 @@ exports[`Test splitBlocks > Block has children 1`] = `
...
-    "id": "paragraph-with-children",
-    "props": {
-      "backgroundColor": "default",
-      "textAlignment": "left",
-      "textColor": "default",
-    },
-    "type": "paragraph",
-  },
   {
     "children": [
```
YousefED, line 625: This snapshot was rewritten, but I think for regular blocks with nested children, the original (main) is the intended behavior

Reproduce:

Parent block
--- nested

when pressing enter at end / middle of parent block, nested should also move to the newly inserted block (instead of enter jumping after it)

(not sure how it is for containers / frames)

tests/src/end-to-end/keyboardhandlers/__snapshots__/enterPreservesNestedBlocks.json
```diff
@@ -25,28 +25,6 @@
                   "text": "H"
                 }
               ]
-            }
```
YousefED, line 28: I think this is a smell (already mentioned in SplitBlock.test)

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -818,6 +700,63 @@ export const KeyboardShortcutsExtension = Extension.create<{
 
             return false;
           }),
+        // If the block is empty and the last child of a container or an
+        // owned-children body, moves the block out (double Enter exits the
+        // container). The block lands at the nearest enclosing position that
+        // accepts it. E.g. out of a column it skips the columnList, which
+        // holds only columns, and lands below it. Without this, Enter only
```
YousefED, line 707: In Notion and on `main`, enter creates new blocks. In this PR the following happens:
- cursor moves out of columnlist
- if the column is empty, it "dissolves"

At least for columns, the main+Notion behavior is better imo. Because people use columns for layouts somewhat, it can make sense to add some space / align with enters.

For other blocks I'm not sure (haven't tested / thought about those cases), so that would determine if we would need to make it configurable vs fully revert this

packages/core/src/extensions/tiptap-extensions/KeyboardShortcuts/KeyboardShortcutsExtension.ts
```diff
@@ -909,42 +915,27 @@ export const KeyboardShortcutsExtension = Extension.create<{
...
+          editor.getExtension(FilePanelExtension)?.store.state !== undefined)
+      ) {
+        // Let the browser navigate into and out of an open toolbar.
+        return false;
+      }
+      return shift ? unnestBlock(editor) : nestBlock(editor);
+    }
+
     return {
       Backspace: handleBackspace,
```
YousefED, line 933: Setup: a column layout, left column A1 + A2, right column B1.
Do: cursor at the start of B1 → press Backspace.

main: A1, A2, ``, B1 — an empty block appears between them.
#3059: A1, A2, B1.

I think the change is good, but the cursor isn't at the right position anymore, so that part is a regression

YousefED: (might be related to #3104)?

examples/06-custom-schema/09-container-block/src/Panel.tsx
```diff
@@ -0,0 +1,16 @@
+import { createReactBlockSpec } from "@blocknote/react";
+
+import "./styles.css";
+
+export const createPanel = createReactBlockSpec(
```
YousefED, line 5: when you change this to `children: { allow: "blocks", min: 2 }`, hitting tab on the second child throws an error

packages/core/src/schema/blocks/children.ts
```diff
@@ -0,0 +1,117 @@
...
+  children?: unknown;
+}): boolean {
+  return config.children !== undefined && config.content === "none";
+}
+
+// Whether `type` is a node that holds child blocks directly: a container
+// block's own node. A container is a child-holding node that is itself a
+// block; `blockGroup` also holds children but is not a block (it's regular
+// blocks' nesting machinery), so the `bnBlock` check excludes it.
+export function isContainerNode(type: NodeType): boolean {
```
YousefED, line 26: A "blockContainer" is not a containerNode, which tripped me a few times in the review. Not sure if we can come up with a better naming policy

packages/react/src/components/Popovers/BlockPopover.tsx
```diff
@@ -29,6 +29,34 @@ export const BlockPopover = (
           return undefined;
         }
 
+        // Containers anchor to their own root, not the child-block contentDOM.
+        if (isContainerNode(nodePosInfo.node.type)) {
```
YousefED, line 33: It's a smell that we have 3 branches here. BlockPopover shouldn't need to know about the structure of the dom / blocks.

I dove into it a bit, and I think we can simplify it with a shared helper that does sth like this:

```
const dom = editor.prosemirrorView.nodeDOM(nodePosInfo.posBeforeNode);
if (!(dom instanceof Element)) { return undefined; }
// A framework node view may wrap the block in elements of its own — React adds
// two `display: contents` wrappers — none of which carry the block's id.
const element = dom.getAttribute("data-id") === blockId
  ? dom
  : dom.querySelector(`[data-id="${blockId}"]`);
return element ? { element } : undefined;
```

In words: 
- `nodeDOM` should give the correct element for all cases
- although we do then get `block-outer` instead of `div.bn-block`, but I don't think layout-wise this matters (afaik they have the same size)
- only for a React Container block, we need to check if the element has data-id, because otherwise we get the `react-renderer` element which doesn't have a layout (`display:contents`) and can therefor not be used
-- This does require the removal of the duplicate `data-id` from the react-renderer (see separate PR)

YousefED: We should probably also check other callsites of nodeDOM and domAtPos and see if they're affected. I did a quick scan and there are some, but it seems a bit of a rabbithole so holding off further (maybe we need to change sth in the overall setup / or CSS). Quoting here:

---

Yes — and widening the net changed the conclusion. Here is the exhaustive list (repo-wide, *.ts/tsx/js/jsx/md/mdx, including tests, examples, playground and docs; nothing outside packages/*/src):


core/src/comments/extension.ts:333                     domAtPos
core/src/editor/managers/SelectionManager.ts:108       nodeDOM
core/src/extensions/DropCursor/utils.ts:77, :103       nodeDOM
core/src/extensions/LinkToolbar/LinkToolbar.ts:6       nodeDOM
core/src/extensions/SideMenu/dragging.ts:75,76,84,89   domAtPos
core/src/y/extensions/Suggestions.ts:29                nodeDOM
react/src/components/Popovers/BlockPopover.tsx:34,54   nodeDOM
react/src/components/Popovers/BlockPopover.tsx:60      domAtPos
react/src/components/TableHandles/…Controller.tsx:82,102  domAtPos
xl-ai/src/AIExtension.ts:457                           domAtPos
xl-multi-column/…/multiColumnDropCursor.ts:46          nodeDOM
17 lines, all covered. But I had scoped the audit to those two method names, and that was too narrow.

The bigger finding
I checked the sibling position→DOM APIs (coordsAtPos, posToDOMRect, posAtDOM, posAtCoords) and ProseMirror's own coordsAtPos is broken at a React container boundary, for the same reason:


coordsAtPos(59)      into the panel     -> 0,0,0,0
coordsAtPos(59, -1)  back at the para   -> 479,38,1102,38    ok
coordsAtPos(199, -1) back at the panel  -> 0,0,0,0
coordsAtPos(199)     into the para      -> 479,174,1102,174  ok
So posToDOMRect inherits it — a range starting at the container gives posToDOMRect(59, 244) → 0,0,1102,204, anchored at the viewport origin. That reaches SelectionManager.ts:114's non-NodeSelection fallback (a MultipleNodeSelection whose first block is the container, e.g. from dragging.ts:183). LinkToolbar, y/Suggestions and PositionPopover also call posToDOMRect, but their endpoints are inside text, so they're fine in practice.

Root cause, isolated
Not containers, and not nodeDOM — it's this PR's styles.css:117:


.bn-react-node-view-renderer.bn-container-node-view {
  display: contents;
}
That element is the React container's node view dom, and both PM and BlockNote ask it for its box. Overriding just that rule at runtime, with no code change:


display: contents  ->  nodeDOM rect 0,0,0,0        coordsAtPos(59) 0,0,0,0
display: block     ->  nodeDOM rect 479,38,623,136 coordsAtPos(59) 479,38,1102,38
The vanilla container is the control: columnList's node view dom is the author's boxed .bn-block-column-list, and every one of these APIs returns correct geometry for it, on this branch and on main.

This also means the BlockPopover approach (descend to the data-id element) patches the consumers we control but cannot fix coordsAtPos. The rule's comment is right that the wrapper "must not be a box of its own" — but display: contents buys that at the cost of all position-based geometry. The durable fix is for the React container's node view dom to be the author's root, as the vanilla path already manages; flipping the CSS to display: block fixes the geometry but reintroduces the extra box the comment is avoiding (it would break an author root that needs to be a flex/grid child, e.g. a React column).

Standing corrections to my previous message: findings 1–3 are symptoms of this one rule, not three independent nodeDOM bugs, and the affected surface is wider than the two method names. Finding 4 (dragging.ts, empty drag image, pre-existing on main for columnList) is separate and unrelated to display: contents.

packages/core/src/extensions/SideMenu/SideMenu.ts
```diff
@@ -46,21 +53,15 @@ function getBlockFromCoords(
...
-      }
-    }
-    return getDraggableBlockFromElement(element, view);
+    const block = getBlockFromElement(element, view);
+    return block
+      ? {
+          ...block,
+          // Controls in a container's chrome belong to that container, even
+          // when they share a row with one of its child blocks.
+          isControl: !!element.closest('[contenteditable="false"]'),
```
YousefED, line 62: Is this needed? and is it covered in tests?

packages/core/src/extensions/SideMenu/SideMenu.ts
```diff
@@ -46,21 +53,15 @@ function getBlockFromCoords(
...
       continue;
     }
-    if (adjustForColumns) {
-      const column = element.closest("[data-node-type=columnList]");
-      if (column) {
-        return getBlockFromCoords(
-          view,
-          {
-            // TODO can we do better than this?
-            left: coords.left + 50, // bit hacky, but if we're inside a column, offset x position to right to account for the width of sidemenu itself
```
YousefED, line 56: With the removal of the +50 hack, I'm unable anymore to click the + button of blocks in the 2nd or 3rd column of column blocks.

Note that a solution where I need to move the mouse exactly over the draghandle towards the + isn't sufficient, real users will move the mouse diagonally from a lower part of the block towards the +.

I'm not sure if there's a quick fix of this. Perhaps we can take inspiration from floating UI's `safePolygon`. Or, we can finish the sidemenu refactor (https://linear.app/blocknote/issue/BLO-1174/sidemenu-refactor) and as part of this move a lot more of the work the extension is doing to Floating UI (similar to how we did for the other components). Note though we don't want to register a popover / floating ui for every block, but I think there should be a way around that

packages/core/src/schema/blocks/types.ts
```diff
@@ -603,20 +699,79 @@ export type BlockImplementation<
...
+     * patched `dom` in-place and PM should keep the existing view.
      *
-     * Useful for blocks whose `render` builds custom DOM that needs to stay in
-     * sync with the node (e.g. a code block rendering a preview of its content).
+     * Only honored for container blocks (blocks with `children`), where
+     * recreating the node view would remount every child block: e.g. column
+     * resizing patches widths in place through this hook. Non-container
+     * blocks always recreate on attr changes (see
+     * https://github.com/TypeCellOS/BlockNote/pull/1904#discussion_r2313461464).
      */
```
YousefED: the hook is on the return
type for every block, and the paragraph above says it is honoured only
for containers. An author who writes one on a regular block gets no
signal - it is simply never called. Either the type should exclude it
or an error raised?

packages/core/src/exporter/Exporter.ts
```diff
@@ -70,6 +70,20 @@ export abstract class Exporter<
...
+    type: string;
+    children?: unknown[];
+  }): boolean {
+    const spec = this.schema.blockSpecs[block.type];
+    if (!spec && block.children?.length) {
+      throw new Error(
+        `Exporter has no block spec for block type "${block.type}", and blocks of that type in this document have children. Add its spec to the exporter schema so it can determine who renders the children.`,
+      );
+    }
+    return spec?.config.children !== undefined;
```
YousefED: should this use the existing isContainer*** helper?

Issue #2398, opened 2026-01-30
Make links visually distinguishable by default


rubzie:
When creating a link in BlockNote, the linked text looks identical to regular text. Users must manually apply styling to make links recognizable.

### Current Behavior
- Create a link with the link tool or `Cmd+K`
- Link has no visual distinction (no color change, no underline, no hover state)
- Must manually style each link

### Expected Behavior
_(tested on Blocknote site example and default implementation)_ 
Links should be visually distinguishable by default, like in Notion and Slite (both use color by default). 

### Proposed Solution
Add default link styling that can be overridden:
```typescript
const editor = useCreateBlockNote({
  linkStyling: {
    enabled: true, // default
    color: '#0066cc', // customizable
  }
});
```

Or include in the default theme:
```css
.bn-container .bn-inline-content a {
  color: #0066cc;
}
```

### Rationale
- **UX standard**: Notion, Slite, Google Docs all style links by default
- **Accessibility**: Links need visual distinction beyond hover state
- **User friction**: Manual styling for every link is tedious

### Current Workaround
Custom CSS override (requires users to know about CSS customization and makes default behavior bad UX for a reader)

YousefED:
Hi @rubzie!

This is likely caused by Tailwind unstyling all `<a>` tags.

You can reset this using CSS like:

```
.bn-editor a {
  color: revert;
  text-decoration: revert;
}
```

Thanks for pointing out the issue also occurs on our website; was triggered by our redesign last Thursday, you're quick!

[ ] We'll look into if this needs to be a default style in the editor Stylesheet (likely, yes. If not, add documentation)
[ ] investigate if this is also the case for other styles (for example, ones we set on `.demo .bn-editor` on the docs examples)
[ ] fix our website demo

-

rubzie:
Not quick, just (un)lucky ;)

PR #3060, opened 2026-09-07
fix(core): keep the browser's default link styling under CSS resets

YousefED:
## Summary

Links inside the editor lost their color and underline in apps that ship a CSS reset such as Tailwind's preflight (`a { color: inherit; text-decoration: inherit }`), including our own `/demo` page since the website redesign. Reported in #2398 (now discussion #2425).

- **Core stylesheet**: `:where(.bn-editor a) { color: revert; text-decoration: revert }` restores the user-agent link styling. `:where()` keeps the specificity at zero, so any `a` rule a consumer writes wins regardless of import order. Tailwind v4's preflight still loses because it lives in a cascade layer, and unlayered rules beat layered ones. Tailwind v3's preflight is unlayered and still wins; the docs show the one-line override. Consumers without a reset see no change.
- **Website**: drops the `.demo`-scoped workaround, which never covered the demo page.
- **Docs**: new "Links" section on Overriding CSS.
- **Typst exporter**: links are underlined like the editor's browser default (unit snapshots updated).

The shadcn skin keeps its own `.bn-shadcn .bn-editor a` rule, since that skin also serves Tailwind v3 apps where the zero-specificity rule loses.

## Verification

- Local docs app: the demo page's links compute to the browser default (blue, underlined) instead of the body text color. Example pages look the same as before.
- Cascade test against the live site's Tailwind v4 chunk: editor links revert in either stylesheet order, links outside the editor stay reset, a consumer rule inside `@layer base` loses, a bare consumer `a` rule wins, and `color-scheme: dark` yields the browser's dark link color.
- Docker e2e for static, typst PDF, links, theming, ariakit and shadcn: passing on chromium, firefox and webkit with no baseline changes.
- Lint, formatting and docs link validation are clean.

## Summary by CodeRabbit

* **Style**
  * Links in the BlockNote editor now display with browser-default visual styling, making them easier to distinguish from regular text.
  * Exported Typst documents now render links in BlockNote blue with underlining.

* **Documentation**
  * Added guidance on link styling, including compatibility considerations for Tailwind CSS and instructions for restoring link color and text decoration when needed.

### DIFF
--- a/docs/app/styles.css
+++ b/docs/app/styles.css
@@ -68,11 +68,6 @@ body {
   padding-block: 1rem;
 }
 
-.demo .bn-editor a {
-  color: revert;
-  text-decoration: revert;
-}
-
 .demo code.bn-inline-content {
   font-size: 1em;
   line-height: 1.5;
--- a/docs/content/docs/react/styling-theming/overriding-css.mdx
+++ b/docs/content/docs/react/styling-theming/overriding-css.mdx
@@ -42,3 +42,16 @@ BlockNote uses data attributes to target specific block types and properties:
 
 - `[data-content-type="blockType"]`: Targets blocks of type `blockType`.
 - `[data-propName="propValue"]`: Targets blocks with specific prop values. If the value is the same as the default value, the `data-propName` attribute will not be added.
+
+## Links
+
+Links in the editor use the browser's default link styling. BlockNote sets this with a rule of zero CSS specificity, so any `a` rule in your own stylesheet overrides it.
+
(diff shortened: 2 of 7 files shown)


nperez0111 (approved):

matthewlipski (approved):
LGTM, though I imagine adding our own styling that closely matches the browser's with higher specificity would also fix the issue. Tradeoffs between that approach and this one?

YousefED:
> LGTM, though I imagine adding our own styling that closely matches the browser's with higher specificity would also fix the issue. Tradeoffs between that approach and this one?

it's more prone to break users who have set application-wide link styling themselves

PR #3062, opened 2026-09-07
fix: ignore only Dark Reader's writes in node views

YousefED:
## Summary

#2912 made every block and inline content node view ignore all DOM mutations outside its content DOM and every attribute mutation, to stop the Dark Reader re-render loop (#2818). That also hid the browser's native paragraph split from ProseMirror. On Android and iOS, prosemirror-view leaves Enter to the browser and reads the split back from the DOM; the new paragraph lands next to the content DOM, so it was never read. On iOS the 200 ms fallback then split the document while the stray paragraph stayed in the block, rendered next to the text by the flex block content (the "two columns" after Enter on the demo). On Android, Enter, Backspace and Delete broke (#3001).

What Dark Reader actually writes, measured with Brave and Dark Reader on the code block and toggleable blocks examples: attribute changes only, inline `style` declarations starting with `--darkreader` and `data-darkreader-*` attributes. The filter now ignores exactly those and lets everything else reach ProseMirror.

## Testing

- `nodeViewMutations.test.ts` rewritten for the new rule.
- Android e2e instance, on the `mobile/android-enter` branch with its Enter interception switched off: with #2912's filter, 14 keyboard-handler cases fail (Enter, Backspace, Delete); with this filter only Enter on a non-empty selection fails, which #3031's interception covers.
- iOS simulator: with the blanket filter out of the way the native split is read and the DOM stays clean; this rule only ignores attribute writes, so the split (a child-list mutation) reaches ProseMirror the same way.
- Dark Reader by hand: `vp run dev`, open the code block example and the toggleable blocks example with Dark Reader on, type in the code block and change its language; the tab must stay responsive.

## Node views without a content DOM

Defining `ignoreMutation` replaces prosemirror-view's default, which ignores every mutation but the selection in a node view without a content DOM (an image block, say). The first revision answered "read it" there, so the image block's own DOM changes during a resize drag were read back, which reset the node selection and copied nothing: `copypaste` "Images should keep props" failed in CI, and in Docker without the restored default (green with it). #2912 had kept that default explicitly. With it restored, core node views behave as before #2912 in everything except Dark Reader's writes. React block and inline content specs are untouched by all of this: they never used the filter and run on tiptap's `NodeView` default.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved editor handling of Dark Reader-generated DOM changes to prevent unnecessary re-render loops.
  - Restored proper processing of native paragraph-split mutations on Android and iOS.
  - Ensured unrelated attribute, selection, and content mutations continue to reach the editor correctly.

- **Tests**
  - Expanded coverage for Dark Reader attributes, inline styles, child-list changes, selection mutations, and existing mutation handling.

### DIFF
--- a/packages/core/src/schema/blocks/createSpec.ts
+++ b/packages/core/src/schema/blocks/createSpec.ts
@@ -12,7 +12,7 @@ import {
   ExtensionFactoryInstance,
 } from "../../editor/BlockNoteExtension.js";
 import { nonFormattingMarks } from "../markGroups.js";
-import { ignoreNonContentMutations } from "../nodeViewMutations.js";
+import { ignoreDarkReaderMutations } from "../nodeViewMutations.js";
 import { PropSchema } from "../propTypes.js";
 import {
   getBlockFromNodeView,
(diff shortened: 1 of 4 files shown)


nperez0111:
I don't like to have to make this change, but we will go with it for now until https://github.com/darkreader/darkreader/pull/15970 or equivalent is merged

PR #3083, opened 2026-09-10
fix(core): keep the next block's children at their depth when merging blocks

### LINKED ISSUE #3018: Delete child breaks indentation for downstream child
### What’s broken?

Behaviour on BlockNote, when I hit backspace to delete a child in a bullet list. Same happens for other types of blocks (checklist, numbered list)

https://github.com/user-attachments/assets/078bf389-0089-44d2-b111-2934a19edf98

Behaviour on Notion

https://github.com/user-attachments/assets/57b443b4-b2a4-409e-a303-e67d7e3960f6

I'd expect a similar behaviour as Notion. 
Sometimes when I'm a meeting a editing a large bullet list, it can unindent dozens of elements

### Steps to reproduce

_No response_

### BlockNote version

_No response_

### Environment

_No response_

### Additional context

_No response_

### Contribution

### Sponsor

Ishkirat-Singh:
# Summary

Fixes #3018. With a list like `A > [B, C, D]` and the caret at the start of `C`, pressing Backspace turns `C` into a paragraph, then un-nests it (which makes `D` a child of `C`), then merges `C` into `B`. That last step put `D` at the root instead of keeping it under `A`, so every sibling after the caret lost a level of nesting.

## Rationale

`mergeBlocksCommand` lifted the next block's children to the next block's own level before stitching the text. The merge target is the deepest last descendant of the previous block, so whenever that target sits deeper than the next block, the children ended up one level shallower than they were. Keeping their depth is what users expect (and what Notion does): the children are appended to the merged block's ancestor at the depth they had, or nested under the merged block when there is no such ancestor.

## Changes

- `packages/core/src/api/blockManipulation/commands/mergeBlocks/mergeBlocks.ts`: take the next block's children out first (so the merge leaves no empty shell of the next block behind), stitch the text as before, then put the children back at their original depth. Both the Backspace and the Delete path go through this command.
- `mergeBlocks.test.ts`: two new cases, previous block nested once and nested twice, with explicit outline assertions.
- Snapshot `Second block has children` changes: `nested-paragraph-0` now stays nested under the merged `paragraph-1` instead of moving to the root. That is the same behaviour change applied to a same-depth merge, and it is intentional.

## Impact

Behaviour change only for merging a block that has children: the children keep their indentation instead of being un-indented by one level. Blocks without children merge exactly as before. I saw that #3050 and #3014 touch this file as part of the container refactor; the change is small, so I am happy to rebase onto whichever lands first.

## Testing

`mergeBlocks.test.ts` (14 pass), the full `packages/core` and `packages/react` suites, `vp lint` on the touched folder and `tsc --noEmit` for `@blocknote/core`, all run locally.

## Checklist

- [ ] The documentation has been updated to reflect the new feature (no docs change needed).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved block merging to preserve content and nested child blocks correctly.
  - Fixed merging behavior when blocks contain children or are nested within other blocks.
  - Ensured merged documents maintain the expected structure across single- and multiple-level nesting.

### DIFF
--- a/packages/core/src/api/blockManipulation/commands/mergeBlocks/__snapshots__/mergeBlocks.test.ts.snap
+++ b/packages/core/src/api/blockManipulation/commands/mergeBlocks/__snapshots__/mergeBlocks.test.ts.snap
@@ -1673,34 +1673,35 @@ exports[`Test mergeBlocks > Second block has children 1`] = `
     "type": "paragraph",
   },
   {
-    "children": [],
-    "content": [
-      {
-        "styles": {},
-        "text": "Paragraph 1Paragraph with children",
-        "type": "text",
-      },
-    ],
-    "id": "paragraph-1",
-    "props": {
-      "backgroundColor": "default",
-      "textAlignment": "left",
-      "textColor": "default",
-    },
-    "type": "paragraph",
-  },
-  {
     "children": [
       {
-        "children": [],
+        "children": [
+          {
+            "children": [],
+            "content": [
+              {
+                "styles": {},
+                "text": "Double Nested Paragraph 0",
+                "type": "text",
+              },
+            ],
+            "id": "double-nested-paragraph-0",
+            "props": {
+              "backgroundColor": "default",
+              "textAlignment": "left",
+              "textColor": "default",
+            },
+            "type": "paragraph",
+          },
+        ],
         "content": [
           {
             "styles": {},
-            "text": "Double Nested Paragraph 0",
+            "text": "Nested Paragraph 0",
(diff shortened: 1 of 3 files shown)


matthewlipski:
Hi, thanks for your contribution. Unfortunately, I think the changes you made are not quite at the right level as it's not the behaviour of `mergeBlocks` that's incorrect here, but rather what the keyboard handlers call. Additionally, I would prefer to have this be configurable. Therefore, I'm closing this PR in favor of #3124.

PR #3090, opened 2026-09-15
feat!: rebuild version history and customize snapshot actions

nperez0111:
# Summary

Rebuild version history around a single selected version and a consistent read-only preview lifecycle. Opening the sidebar selects the current version; users can name, filter, compare, restore, and delete versions, and applications can replace snapshot menu items while reusing the built-in actions.

## Rationale

Version history needs predictable selection, loading, and closing behavior across storage backends. Menu customization should preserve restore, preview, and async follow-up behavior without requiring applications to duplicate sidebar internals.

## Changes

- Add composable read-only restrictions and prevent link editing while the editor is locked.
- Unify versioning list, view, and loading state; handle overlapping requests, restores, closing, and scrolling to the first change. Update the in-memory, Yjs 13/14, and YHub adapters.
- Rebuild the sidebar with inline naming, named-version filtering, comparison controls, keyboard navigation, loading/error feedback, and updated translations and UI adapters.
- Expose composable snapshot menus, a default-items fragment, and action hooks colocated with their components. Support explicit menu hiding, presentation overrides, merged classes, and disabled items across Mantine, Ariakit, and Shadcn.
- Refresh attribution colors, deleted-block presentation, example histories, and the suggestion gallery's responsive and dark-mode styling. Fix nested bullet markers and guides using CSS inheritance, and remove issue notes for accepted single-user nesting behavior.
- Expand versioning, adapter, read-only, and sidebar regression coverage.

## Impact

This includes breaking changes to the versioning state/adapter contracts and sidebar customization API. Consumers using the previous store fields, capability flags, or sidebar configuration must migrate; the updated examples demonstrate the new contracts. Attribution colors and deleted-block rendering also change.

## Testing

- Focused core suites: 240 tests passed across 11 files, covering versioning, adapters, YHub, read-only restrictions, attribution colors, and diff scrolling.
- React sidebar and read-only link toolbar suites: 71 tests passed across 2 files.
- Commit hooks passed `vp check --fix`.
- Manually checked the two- and three-level nested bullet examples in Suggestions and Versioning modes in the running gallery, including markers, 24px indentation, and guide lines.
- Manually verified that the six supported single-user nesting examples no longer show issue badges or panels.

## Screenshots/Video

The updated versioning demos and suggestion gallery provide interactive before/after views for review. Manual visual checks used the running gallery.

## Checklist

- [x] The documentation has been updated to reflect the new feature.

## Additional Notes

The full browser suite was not rerun. Earlier deleted-block baseline updates are included, but the latest nesting, table-deletion, and dark-mode changes were checked manually; affected browser screenshots and exporter/static visual baselines have not been regenerated for those changes.

## Summary by CodeRabbit

- **New Features**
  - Redesigned version history sidebar with current-version display, named-version filtering, comparison controls, and keyboard navigation.
  - Added actions to name, rename, compare, restore, and delete versions.
  - Version previews now highlight and scroll to the first change, while keeping the editor read-only.
  - Added localized version-history labels across supported languages.
  - Added attribution support for block property changes and improved light/dark user colors.

- **Bug Fixes**
  - Link tools no longer open while the editor is read-only.


packages/core/src/editor/managers/StateManager.ts
```diff
@@ -205,8 +205,24 @@ export class StateManager {
...
+    // transaction instead, so a selector reading `isEditable` (the link
+    // toolbar's read-only gate, a host's own UI) sees the change through the
+    // same `transaction` event as any other state change — and nothing else
+    // fires.
+    this.editor._tiptapEditor.setEditable(editable, false);
+    this.notifyEditableChanged();
+  }
+
+  /** Recompute plugin editability and notify transaction subscribers. */
+  private notifyEditableChanged() {
```
YousefED: some sanity-check questions on how editable relates to readonly:

- the event (notify) doesn't trigger when ReadOnly changes right? should it?
- Can't / should we use our new ReadOnly system instead of calling `_tiptapEditor.setEditable` directly in `isEditable = false`?

packages/core/src/extensions/Versioning/scrollToFirstChange.ts
```diff
@@ -0,0 +1,126 @@
...
+    return false;
+  }
+
+  const target = findScrollTarget(root);
+  if (!target) {
+    return false;
+  }
+
+  // Inline marks sit inside block content; block-level marks wrap it.
+  const enclosingBlock = target.closest(".bn-block-content");
```
YousefED: Is toggling classnames on bn-block-content safe or could it conflict with Prosemirror (and it's MutationObserver)?

packages/core/src/y/versioning/yhub.ts
```diff
@@ -244,194 +254,157 @@ async function yhubFetch(
...
-    activityLimit = 50,
-    group,
-  } = options;
+  const { baseUrl, org, docId, headers = {} } = options;
 
   const activityUrl = `${baseUrl}/activity/v1/${org}/${docId}`;
   const changesetUrl = `${baseUrl}/changeset/v1/${org}/${docId}`;
   const rollbackUrl = `${baseUrl}/rollback/v1/${org}/${docId}`;
-  const ydocUrl = `${baseUrl}/ydoc/v1/${org}/${docId}`;
+  const documentUrl = `${baseUrl}/ydoc/v1/${org}/${docId}`;
```
YousefED: I haven't reviewed this file in depth (line-by-line), but couple of questions come up:

- This file is both a YHub API Client and an implementation of our VersioningEndpoints interface, right? Can we extract a YHub API REST client that's as "pure" as possible? (no blocknote logic around it, just fetch REST calls and parameter conversions)
- The version names are now stored in a Y.Array which I don't think should / will be the expected way to do this. Can you coordinate this with Kevin? https://github.com/orgs/TypeCellOS/projects/14?pane=issue&itemId=210266680&issue=TypeCellOS%7CProduct%7C90
- Is there any other "magic" we're doing in this file that we should validate with Kevin or you think should have an easier solution (on YHub's side)?

nperez0111: There isn't a versioning API in yhub yet, I'm not sure what Kevin is thinking on doing here, but until then, it needs to be stored somewhere, so on the Y.Array it will be

Not really any magic here, nothing that I'd upstream to Kevin.

packages/core/src/y/versioning/yhub.ts
```diff
@@ -1,637 +1,240 @@
...
-    entry.customAttributions?.some(
-      (a) => a.k === "type" && a.v === "version",
-    ) ?? false
-  );
+const ACTIVITY_PARAM_DEFAULTS: YHubQueryParams = {
+  order: "desc",
+  limit: 50,
+  groupMaxGap: 60 * 60 * 1000, // Start a version after an hour of inactivity.
+  groupMaxDuration: 12 * 60 * 60 * 1000, // Cap a version at twelve hours.
+  mergeUsers: true, // Supported by our YHub fork.
```
YousefED: still needed? upstream or drop?

packages/core/src/y/versioning/yhub.ts
```diff
@@ -1,637 +1,240 @@
...
-      if (!map) {
-        throw new Error(
-          "Cannot rename version: no live collaboration document is available.",
+    return {
+      refreshAfterRestoreMs: options.refreshAfterRestoreMs ?? 6000,
+      async list() {
+        const activity = await fetchActivity();
+
+        const rows = new Map<number, YHubSnapshot>();
+        for (const entry of activity) {
```
YousefED, line 110: I'm not really following here. can there be multiple activity items with the same id?

packages/core/src/y/versioning/yhub.ts
```diff
@@ -1,637 +1,240 @@
...
+              existing?.metadata,
+              entry.customAttributions,
+            ),
+          });
+        }
+
+        // Stored labels can have newer timestamps, but only activity defines current.
+        const newestActivityTo = Math.max(...rows.keys());
+
+        for (const [id, entry] of versions.readEntries()) {
```
YousefED, line 126: what's happening here?

packages/core/src/y/versioning/yhub.ts
```diff
@@ -1,637 +1,240 @@
...
+          getContentAt(to),
+          fetchNewestEntry(),
+          // Retained history includes deleted subtrees missing from the live doc.
+          client.getDocument({ gc: false }),
+        ]);
+
+        // Restore only this editor's fragment, preserving other editors and metadata.
+        const contentIds = collectFragmentIds(fragment, document);
+
+        await client.rollback({
```
YousefED, line 207: Isn't there an easier way to do a rollback? This whole `collectFragmentIds` seems, uhm, cumbersome to say the least?

YousefED (commented):
comments related to yhub

scripts/patch-y-prosemirror.sh
```diff
@@ -15,7 +15,7 @@ LOCAL_YPM="${1:-$(cd "$BLOCKNOTE_ROOT/../y-prosemirror" && pwd)}"
 
 # Version of @y/prosemirror to patch. Must match the version pinned in
 # pnpm-workspace.yaml (overrides + patchedDependencies) and package.json files.
-YPM_VERSION="2.0.0-6"
+YPM_VERSION="2.0.0-11"
```
YousefED, line 18: still needed / remove this entire script / keep for future?

paseo.json
```diff
@@ -0,0 +1,10 @@
+{
+  "worktree": {
```
YousefED: want to keep this file?

packages/core/src/extensions/Versioning/Versioning.ts
```diff
@@ -328,180 +39,206 @@ export const VersioningExtension = createExtension(
...
-         */
-        compareTo?: VersionSnapshotIdentifier;
-      },
-    ) => {
-      const snapshot = getSnapshot(id);
-
-      if (!snapshot) {
-        snapshotNotFoundError(id);
+    /** Sync versioning's read-only restriction when preview or restore state changes. */
+    function setStateSyncingReadOnly(
```
YousefED: we still call `store.setState` manually a bunch of times in this file. Shouldn't we wire all calls through this method?

packages/core/src/extensions/Versioning/Versioning.ts
```diff
@@ -328,180 +39,206 @@ export const VersioningExtension = createExtension(
...
-       * omitted, the live document is shown without a diff.
-       */
-      compareTo?: VersionSnapshotIdentifier;
-    }) => {
-      if (!serializeCurrentContent) {
-        throw new Error(
-          "previewCurrentVersion requires `serializeCurrentContent` to be " +
-            "provided to the VersioningExtension options.",
-        );
+    function refreshList(): Promise<LoadedVersioningList> {
```
YousefED: nitpick, but this is pretty intense (generation, listing, signal etc). Isn't there an easy util for this?

packages/core/src/extensions/Versioning/Versioning.ts
```diff
@@ -328,180 +39,206 @@ export const VersioningExtension = createExtension(
...
+      serializeCurrentContent,
+      getSnapshot: findSnapshot,
+      setView,
+      syncStatus,
+      editor,
+      scrollToFirstChangeEnabled,
+    });
 
-    const exitPreview = () => {
+    function syncStatus() {
```
YousefED: nice-to-have; this seems like `status` is a derived value of two stores. we can't set it up like that (derived / functional) instead of calling a method (`syncStatus`) to keep a property in-sync?

packages/core/src/extensions/Versioning/preview.ts
```diff
@@ -0,0 +1,254 @@
...
+      previewToken++;
+      setLoadingPreview(undefined);
+      setView({ mode: "live" });
+      // The controller is only asked to leave what it was asked to render. A
+      // preview still fetching never replaced the document, and the controller
+      // must not "restore" one it never touched (the Yjs controller rebuilds
+      // the whole document when asked).
+      const rendered = renderedView.mode !== "live";
+      renderedView = { mode: "live" };
+      if (rendered) {
```
YousefED: would this be a bug? (if the renderedView !== "live", but we call `exitPreview`?). Or could it happen when loading still loading for example (like above)? Zoomed out, mostly wondering if we can simplify the state management (quite some wiring between view states / loading / errors, etc)

packages/core/src/y/versioning/yhub.ts
nperez0111: We can talk with Kevin on this

YousefED (commented):
more small comments

scripts/patch-y-prosemirror.sh
nperez0111: keep for now, I will need to make some patches but wanted it clean for putting into docs

packages/core/src/y/versioning/yhub.ts
nperez0111: Problem is that with the way that version snapshots are stored in the document, we can't just rollback to a timestamp since we will rollback the versions too.

This is one more argument for having Kevin expose a versioning API.

packages/react/src/components/AttributionTooltip/formatChangeLabel.ts
```diff
@@ -41,6 +41,14 @@ export const defaultFormatChangeLabel: FormatChangeLabel = ({
   const toolbar = dictionary.formatting_toolbar as Record<string, unknown>;
```
YousefED, line 41: afaik it's not a good practice to re-use i18n labels from other sources. In different languages, a proper translation could differ based on the usage (e.g.: an action in the formatting toolbar could be different from how it should be shown in an attribution label).

Besides, when using i18n strings (except generic ones) in multiple locations, it makes it more difficult for translators to understand what their change (translation) would impact

It would also fix the Link workaround, simplify this method, and make it a simple dictionary lookup

packages/core/src/extensions/Versioning/scrollToFirstChange.ts
nperez0111: Used Element.animate which is something I did not know existed!

packages/core/src/y/versioning/yhub.ts
nperez0111: I've left this in for now, given that we also store comments on the document and I don't think a rollback should drop comment data. We will see if Kevin uptakes this, he seemed interested in offering a scoped rollback.

packages/core/src/extensions/Versioning/preview.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+
+  async function showPreview(
+    view: VersioningPreviewView,
+    target: PreviewTarget,
+    compareTo: VersionSnapshot | undefined,
+    getPrimaryContent: () => Promise<any>,
+  ) {
+    const request = ++latestPreview;
```
YousefED, line 104: do we need this manual request id plumbing? I can see two different options:
- simplify: use `store.state.view !== view` instead of `request !== latestPreview`
- add support for aborting requests. i.e.: make sure pending, outdated requests are nicely aborted instead of executed and ignored

packages/core/src/y/versioning/yhub.ts
```diff
@@ -1,637 +1,241 @@
...
-  customAttributions?: Array<{ k: string; v: string }>;
+const ACTIVITY_PARAM_DEFAULTS: YHubQueryParams = {
+  order: "desc",
+  limit: 50,
+  groupMaxGap: 60 * 60 * 1000, // Start a version after an hour of inactivity.
+  groupMaxDuration: 12 * 60 * 60 * 1000, // Cap a version at twelve hours.
+  // Fork-only params (not upstream YHub): `mergeUsers` coalesces co-authors
+  // within a window, `customAttributions` returns name metadata. Coordinate
+  // with Kevin before relying on these in production / upstreaming.
+  mergeUsers: true,
```
YousefED: TBD (as discussed)

packages/core/src/y/extensions/AttributionExtension.ts
```diff
@@ -101,23 +105,24 @@ export const getReferenceClientRects = (wrapper: Element): DOMRectList =>
  * none). The extension computes it; a React controller renders + positions it
  * (see `AttributionTooltipController`).
  */
-export type AttributionTooltipState = {
+export type AttributionChange =
```
YousefED, line 78: `getReferenceRect` and `getReferenceClientRects` don't seem used? Maybe more? (did we have a lint rule to detect these?)

packages/core/src/y/extensions/DiffVersioningExtension.ts
```diff
@@ -29,8 +29,9 @@ const diffAuthorId = (label: string) => DIFF_AUTHOR_ID_PREFIX + label;
 /** Fallback label used when a diff is rendered without a version name. */
 const DEFAULT_DIFF_LABEL = "This version";
```
YousefED: used? not translated

packages/core/src/editor/editor.css
```diff
@@ -195,3 +195,64 @@ For the ShowSelectionPlugin
   background-color: highlight;
   padding: 2px 0;
 }
+
+/* Shared loading indicator. Override the color and size on the host element. */
+.bn-loader,
```
YousefED, line 200: I don't think bn-loader is set anywhere?

packages/core/src/extensions/Versioning/types.ts
```diff
@@ -0,0 +1,234 @@
...
+  restoring: boolean;
+};
+
+/**
+ * Version storage, paired with a {@link PreviewController} for rendering.
+ * @typeParam Input - Live document handle supplied by `getCurrentDocument`.
+ * @typeParam Output - Serialized content fetched/restored and passed to the controller.
+ * @typeParam Attributions - Diff authorship data passed to the controller.
+ */
+export interface VersioningEndpoints<
```
YousefED, line 102: I find it a bit difficult to navigate the different concepts, particulary `getContent`, `getCurrentDocument`, `serializeCurrentContent`, Input / Output, but not sure we can make this simpler (or improve naming)

packages/core/src/i18n/locales/en.ts
```diff
@@ -414,6 +414,32 @@ export const en = {
     formatting_change_by: (formats: string, users: string) =>
       `Formatting change (${formats}) by: ${users}`,
   },
+  versioning: {
+    title: "History",
+    close: "Close",
+    save_version: "Save version",
```
YousefED: not used anymore?

packages/react/src/components/Versioning/VersionSnapshotContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+ */
+export type VersionSnapshotContextValue = {
+  /** The version this row shows. */
+  snapshot: VersionSnapshot;
+  /**
+   * The previous visible version in the filtered list, i.e. what a
+   * diff of this row is taken against by default. `undefined` for the oldest
+   * row.
+   */
+  previousSnapshot?: VersionSnapshot;
```
YousefED: I don't think `previousSnapshot` is used anywhere

packages/core/src/y/versioning/yhub.ts
YousefED: ok, good catch.

tests/src/end-to-end/y-prosemirror/__snapshots__/addRemoveBlocks.test.tsx.snap
```diff
@@ -26,23 +26,25 @@ exports[`suggestion mode: add an empty block 3`] = `
...
-          <paragraph backgroundColor="default" textAlignment="left" textColor="default"></paragraph>
-        </y-attributed-insert>
-      </blockContainer>
+      <y-attributed-attrs
+        changes="{&quot;id&quot;:{&quot;userIds&quot;:[],&quot;timestamp&quot;:null}}"
+      >
+        <blockContainer id="0">
+          <y-attributed-insert userIds="">
+            <y-attributed-attrs
+              changes="{&quot;backgroundColor&quot;:{&quot;userIds&quot;:[],&quot;timestamp&quot;:null},&quot;textColor&quot;:{&quot;userIds&quot;:[],&quot;timestamp&quot;:null},&quot;textAlignment&quot;:{&quot;userIds&quot;:[],&quot;timestamp&quot;:null}}"
```
YousefED, line 35: There are a bunch of prop-change attrs rendered now. This relates to the first bullet of https://github.com/TypeCellOS/Product/issues/87#issue-5142638233.

Either we need to fix it on the rendering (CSS / hover) side, or prevent yjs from outputting them (unlikely?). If we choose the former the snapshots would stay as-is I suppose, but at least the UX won't show these nested attributions

packages/react/src/components/AttributionTooltip/AttributionTooltipController.tsx
```diff
@@ -101,12 +101,7 @@ export const AttributionTooltipController = (props: {
     () =>
       state
         ? {
-            color: state.color,
-            className: state.className,
-            modificationType: state.modificationType,
-            contentType: state.contentType,
-            users: state.users,
-            format: state.format,
+            ...state,
```
YousefED, line 103: can we explain what's happening in the `reference` initialization above? (it seems significantly more complex than other usages of `<GenericPopover>`)

packages/react/src/components/Versioning/VersioningSidebar.tsx
```diff
@@ -1,232 +1,117 @@
...
-  // truth for what's rendered — including the "current version" entry that
-  // backends surface via `list()` — so the sidebar can't rely on the host
-  // having listed already.
-  useEffect(() => {
-    void list();
-  }, [list]);
+  // Read at mount only: the initial selection uses whatever comparison mode the
+  // panel opened with, and must not re-run when the user toggles it (the header
+  // re-previews for that).
+  const latest = useRef({ previewRow, run });
```
YousefED: is this needed?

examples/07-collaboration/11-versioning-yjs13/README.md
```diff
@@ -2,7 +2,7 @@
 
 This example shows how to use the `VersioningExtension` with collaborative editing using `yjs` (v13). Snapshots are stored in localStorage using Yjs state updates.
 
-**Try it out:** Edit the document, then click the "Version History" button to open the sidebar. From there you can save snapshots, preview older versions, rename them, and restore them.
+The sidebar opens on a document with a few versions already in its history, so you can preview them, rename them, and restore them right away. The editor is read-only while the sidebar is open: close it to edit the document, then reopen it with the "History" button and press "Save version" to add a version of your own.
```
YousefED: Save version doesn't exist anymore

examples/08-extensions/02-versioning/README.md
```diff
@@ -2,4 +2,4 @@
 
 This example shows how to use the `VersioningExtension` without any collaboration layer (no Yjs required). Snapshots are stored in memory using ProseMirror JSON.
 
-**Try it out:** Edit the document, then use the Version History sidebar to save snapshots, preview older versions, rename them, and restore them. You can hide the sidebar with the close button and reopen it with the "History" button.
+The sidebar opens on a document with a few versions already in its history, so you can preview them, compare them, rename them, and restore them right away. The editor is read-only while the sidebar is open: close it to edit the document, then reopen it with the "History" button and press "Save version" to add a version of your own.
```
YousefED: Save version doesn't exist anymore

packages/react/src/components/Versioning/VersioningPrimitives.tsx
```diff
@@ -0,0 +1,85 @@
+import { assertEmpty, mergeCSSClasses } from "@blocknote/core";
+import type { ComponentProps } from "../../editor/ComponentsContext.js";
+import {
+  forwardRef,
+  type ComponentType,
+  type ComponentPropsWithRef,
+} from "react";
+
+export const VersioningSidebarRoot = forwardRef<
```
YousefED: I think a lot more of the versioning markup (and CSS) should live in the skins
rather than here — the sidebar row is UI chrome, so it belongs in
mantine/ariakit/shadcn the way `Comment`/`Card` do for comments, built from each
library's own components, right? I understand this will be a bit of work (and the components setup is contested), but it's important to keep the codebase consistent

examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
```diff
@@ -237,16 +231,6 @@ export const scenarios: SuggestionScenario[] = [
   {
     kind: "single",
     id: "nest-bullet-existing",
-    feedback: [
-      {
-        severity: "low",
-        note: "Nested bullets all render as • instead of •/◦/▪ — the suggestion-mark wrappers (display: contents) break the depth-detecting CSS chains. Fix: compute each bullet's nesting level in JS and expose it as data-bullet-level, then pick the glyph with a wrapper-independent attribute selector (as numbered lists do with data-index).",
-      },
-      {
```
YousefED, line 245: this second one hasn't been fixed

examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
```diff
@@ -300,12 +284,6 @@ export const scenarios: SuggestionScenario[] = [
   {
     kind: "single",
     id: "delete-nested",
-    feedback: [
```
YousefED, line 303: hasn't been fixed (dont remove feedback)

examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
```diff
@@ -625,12 +591,6 @@ export const scenarios: SuggestionScenario[] = [
   {
     kind: "single",
     id: "nesting-change-parent-type",
-    feedback: [
```
YousefED, line 628: afaik hasn't been fixed (dont remove feedback)

examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
```diff
@@ -601,12 +573,6 @@ export const scenarios: SuggestionScenario[] = [
     kind: "single",
     id: "nesting-unindent",
     title: "Unindent a block",
-    feedback: [
```
YousefED, line 604: hasn't been fixed (dont remove feedback)

examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
```diff
@@ -577,12 +555,6 @@ export const scenarios: SuggestionScenario[] = [
   {
     kind: "single",
     id: "nesting-indent",
-    feedback: [
```
YousefED, line 580: hasn't been fixed (dont remove feedback)

packages/react/src/editor/styles.css
```diff
@@ -509,6 +631,10 @@ SideMenuController offsets its position to keep it centered on the line. */
   gap: 4px;
 }
 
+.bn-root[data-color-scheme="dark"] .bn-snapshot-comparing-to {
```
YousefED: we seem to mix `data-color-scheme="dark"` and `.dark`, maybe we can streamline? (unless they are different somehow?)

tests/src/unit/react/LinkToolbarReadOnly.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { TextSelection } from "@tiptap/pm/state";
+import { act, cleanup, fireEvent, render } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vite-plus/test";
+
+/**
+ * A read-only editor must not open the link toolbar: it only offers editing
+ * actions, and a version preview keeps the editor read-only for as long as the
+ * history panel is open.
+ */
+describe("LinkToolbarController in a read-only editor", () => {
```
YousefED: should live next to link toolbar?

tests/src/unit/react/versioning/VersioningSidebar.test.tsx
```diff
@@ -0,0 +1,1353 @@
+import { StrictMode, type ComponentType, type ReactNode } from "react";
```
YousefED: should live next to link versioningSidebar?

tests/src/unit/react/LinkToolbarReadOnly.test.tsx
```diff
@@ -0,0 +1,112 @@
+import { BlockNoteEditor } from "@blocknote/core";
+import { BlockNoteView } from "@blocknote/mantine";
+import "@blocknote/mantine/style.css";
+import { LinkToolbarController } from "@blocknote/react";
+import { TextSelection } from "@tiptap/pm/state";
+import { act, cleanup, fireEvent, render } from "@testing-library/react";
```
YousefED: this introduces a new pattern ("@testing-library/react"). (granted, the dependency was already there, but not used).

Let's make sure we want to introduce this. Similarly, should these be browser tests instead of jsdom? (I think I'd prefer that and start deprecating jsdom?)

examples/07-collaboration/14-suggestion-gallery/src/scenarios.ts
```diff
@@ -178,12 +178,6 @@ export const scenarios: SuggestionScenario[] = [
   {
     kind: "single",
     id: "add-nested-bullets",
-    feedback: [
```
YousefED, line 181: I don't think this has been fixed (despite the changes in Block.css)

examples/07-collaboration/11-versioning-yjs13/src/localStorageEndpoints.ts
```diff
@@ -1,11 +1,9 @@
 import * as Y from "yjs";
 import { toBase64, fromBase64 } from "lib0/buffer";
 
-import {
-  CURRENT_VERSION_ID,
-  sortSnapshotsNewestFirst,
-  type VersioningEndpoints,
-  type VersionSnapshot,
+import type {
```
YousefED, line 4: according to Claude `restoreSnapshot` in this file is inreachable. 

```
The v13 preview controller ships no `applyRestore`, so the extension leaves `restore` undefined and the menu item never renders — yet this endpoint and the README both still advertise restore. Either drop both, or give the adapter back a throwing `applyRestore` so the gap is visible instead of silent.
```

packages/core/src/extensions/Versioning/inMemoryVersioning.ts
```diff
@@ -122,100 +123,130 @@ export function createInMemoryPreviewController(
...
 
-      // Create a "Restored from …" snapshot of the current state before
-      // restoring, so the user can undo the restore.
+      // Capture the pre-restore state as its own version so the restore can be
+      // undone — the in-memory backend has no continuous history to fall back
+      // on the way a server-backed one does.
       const now = nextTimestamp();
       const backupId = String(nextId++);
-      const backup: VersionSnapshot = {
+      snapshots.push({
```
YousefED, line 218: `before restore` should probably be translated

examples/08-extensions/02-versioning/README.md
YousefED: feedback after testing this example:
- "show named versions only" doesn't make sense (there are only named versions
- tooltips show "inserted by <version name>" which is weird (e.g.: "inserterd by current version")
- "Compare with this version" on the currently selected item should probably either be removed or disabled. Screenshot:

Image

- When I name "current version", a new version pops up ("current version"). As the editor is not editable, I don't think there's any use for this? (current version will always be the same as the just renamed version). This also happens for the yhub demo

examples/07-collaboration/10-suggestion-multi-editor/src/App.tsx
```diff
@@ -25,7 +25,7 @@ provider2.awareness.setLocalStateField("user", {
   color: "#6eeb83",
 });
 
-const attrs = new Y.Attributions();
+const attrs = Y.createContentMap();
```
YousefED, line 29: I think this demo (suggestion-multi-editor) is now completely broken. Either fix or remove?

examples/07-collaboration/11-versioning-yjs13/README.md
YousefED: might be good to mark this example as deprecated / unsupported, because it's yjs 13. In general, I think it's worthwhile to think which example should have `docs: true`, whether naming is clear, or whether we should introduce a "versioning" folder or not

YousefED: Also, styling for this example is off?

packages/react/src/components/AttributionTooltip/formatChangeLabel.ts
nperez0111: I'd agree normally, but I did leave this sort of half-baked purposefully, because I am not sure that we know the exact wording for this so I did the cheapest thing by re-using the label. I'd rather get real feedback from users than to bloat the bundle and churn translations for this. It's an easy change to go back on

packages/core/src/extensions/Versioning/preview.ts
nperez0111: aborting requests is more plumbing, I tried that and got a huge diff on it.
promise comparison is fine, but does not handle ordering (i.e. lww)

packages/core/src/y/versioning/yhub.ts
nperez0111: Kevin added something under a different name, will address: https://github.com/yjs/yhub/commit/b622df09132ddd7d8299d31113814fde82c5af00

packages/core/src/extensions/Versioning/types.ts
nperez0111: It's internal. This is not a public API unless someone wanted to implement Y.js versioning differently. This was the simplest I could come up with, it's inherently a complicated and low-level problem

tests/src/end-to-end/y-prosemirror/__snapshots__/addRemoveBlocks.test.tsx.snap
nperez0111: Yea, I agree we should ask Kevin

PR #3094, opened 2026-09-17
fix: Scrolling to bottom of page on iOS

matthewlipski:
# Summary

This PR partially fixes [this](https://github.com/TypeCellOS/BlockNote/pull/3031#discussion_r4025280291) issue in #3031. You can now scroll to the bottom of the page again, but there is a lot of jitter once the bottom of the page is reached.

This is because the core fix to the iOS gap issue is the same. iOS has extra space below `body` for scrolling which the formatting toolbar cannot follow the virtual keyboard into when scrolling down. So whenever we scroll further than the bottom of `body`, we scroll back up to it. This "snap back" is what causes the jitter. The aforementioned issue was due to the scroll position at the bottom of `body` not taking the height of the virtual keyboard into consideration, which is fixed in this PR.

I tried an alternative approach which tried to prevent the "snap back" behaviour by preventing touch events from propagating into scroll events, so that we could prevent the scroll before it happens rather than snapping back to a previous position after it happens. However, this resulted in a ton of extra complexity as it had to account for things like dragging momentum, yet didn't meaningfully help with the jitter.

## Rationale

See [this issue](https://github.com/TypeCellOS/BlockNote/pull/3031#discussion_r4025280291)

## Changes

Updated `useVirtualKeyboard` hook.

## Impact

I'm not convinced this results in better UX than just having the gap on iOS. The jitteriness imo is more annoying from a UX perspective.

## Testing

Given we elected not to include iOS emulation in our automated tests, tests have not been added as they require an iOS environment to reproduce.

## Screenshots/Video

https://github.com/user-attachments/assets/63933fa8-6c02-4014-9e99-b6dd44ad8031

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mobile keyboard scrolling behavior in Safari.
  * Prevented document scrolling from being incorrectly clipped when the on-screen keyboard opens, during orientation changes, and as the keyboard closes.


packages/react/src/components/FormattingToolbar/useVirtualKeyboard.ts
```diff
@@ -115,23 +120,43 @@ export function useVirtualKeyboard(): boolean {
...
+        (window.innerHeight > height + 1 &&
+          Math.max(window.scrollY, vp?.pageTop ?? 0) + height <
+            html.scrollHeight - 1)
+      ) {
+        unclippedKeyboardHeight = height * scale;
+      }
+      const max = Math.max(
+        0,
+        html.scrollHeight - unclippedKeyboardHeight / scale,
+      );
```
nperez0111, line 148: Is there a way to simplify this logic, it's getting to be a bit of a mess, maybe it can be simpler to cache the maximum instead of of the unclippedKeyboarHeight instead. Presumably you want the absolute maximum and for it to go no further? Unsure, what we can do here

matthewlipski:
Closing as the jitter caused by these changes at the bottom of the page makes for a worse overall UX than just having the gap between the keyboard and formatting toolbar.

PR #3095, opened 2026-09-17
fix: Collaboration cursor clipping (BLO-1362)

### LINKED ISSUE #3079: Editor label cut off
### What’s broken?

We're using BlockNote together with Hocuspocus for real-time collaboration. This means we see the cursor of other people in the editor (together with their label).
On the edges of the editor, it does not work well though, as the labels are cut off instead of rendered in a different position. Also in tables, they are cut off.

The editor labels should always be rendered inside the editor. 
- So when the other persons cursor is on the **right edge of the editor** and there is not enough space, it should render the label to the left. 
- If the other persons cursor is at **the top of the editor** and there is not enough space to render it on top, it should render below.
- Inside a table it should not be cut off

### Steps to reproduce

1. Use BlockNote with a real-time collaboration backend (e.g. Hocuspocus)
2. Have multiple users in the same editor instance
3. Have a cursor near the edge of the editor (on top, to the right) or inside a table

### BlockNote version

_No response_

### Environment

_No response_

### Additional context

Image

Image

Image

### Contribution

### Sponsor

matthewlipski:
# Summary

This PR fixes an issue with the collaboration cursor label getting cut off by elements in the editor which clip overflow. To fix this, the labels are rendered under `editor.portalElement` (conflict with changes in #3052, will need updating), so that only `editor.portalElement` can clip them. Then, FloatingUI is used to flip the label orientation when one is clipped. This does require a FloatingUI dependency in `core`, but only `@floating-ui/dom`.

Closes #3079 

## Rationale

This is a bug.

## Changes

- Extracted shared collaboration cursor code to separate file.
- Updated styles.
- Portalled collaboration cursor label to `editor.portalElement`.
- Updated collaboration cursor to use FloatingUI for label positioning.

## Impact

Added dependency to `core` - increased bundle size.

## Testing

Added component tests.

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A

## Summary by CodeRabbit

* **New Features**
  * Collaboration cursor labels remain visible near editor edges, tables, and clipped areas.
  * Labels track their associated caret during scrolling, resizing, and horizontal table movement.
  * Long labels are truncated cleanly to prevent layout issues.
  * Improved cursor styling, placement, and activity states provide more consistent visibility across supported collaboration backends.

* **Bug Fixes**
  * Preserved custom cursor rendering while improving hover and inactive-cursor behavior.


packages/core/src/extensions/Collaboration/cursor.ts
nperez0111, line 1: Eh, I'm wary of this. I don't like having floating UI be in core. Can we do this without floating UI and instead just have the label be mounted to editor.portalElements?

Sure, it's nice to have the flip and all, but I'm not sure that it is worth it for this

packages/core/src/extensions/Collaboration/cursor.ts
```diff
@@ -0,0 +1,202 @@
...
+        const document = initialView.dom.ownerDocument;
+        const window = document.defaultView!;
+        // Capture scroll events from nested tables as well as outer scrollers.
+        root.addEventListener("scroll", updatePositions, true);
+        if (root !== document) {
+          document.addEventListener("scroll", updatePositions, true);
+        }
+        window.addEventListener("resize", updatePositions);
+        root.addEventListener("load", updatePositions, true);
+        const resizeObserver = new ResizeObserver(updatePositions);
```
nperez0111: Oh, am I misunderstanding something here. Why do we need to listen to events now? Does this have to do with the element not "following" for scrolls and stuff?

matthewlipski: This is to do with rendering the label at `editor.portalRoot` instead of inline. While that fixes any overflow issues within the editor like with tables described in the original issue, it basically detaches the label's positioning from the decoration which creates the cursor. But yea the listeners are not a great solution, I've now replaced them with CSS anchors.

packages/core/src/extensions/Collaboration/cursor.ts
```diff
@@ -0,0 +1,177 @@
...
+            for (const cursor of cursors.values()) {
+              hideCursor(cursor);
+              cursor.label?.remove();
+            }
+            cursors.clear();
+            view = undefined;
+          },
+        };
+      },
+    }),
```
nperez0111, line 175: If I'm getting this right, this plugin is needed because cursorBuilder will tell you only what to render but does not give you any sort of indication when to stop rendering. So you are using this plugin to reconcile the labels you've attached to the portal element with the current awareness states.So, if I understand this correctly, I think there might be another way about this without a ProseMirror plugin:

We could instead listen for awareness state changes directly and do clean up of awareness peers that are no longer relevant. Then on cursorBuilder we can just always create elements and if there exists a label that has the same clientID as the one we are creating we can just update that label element and re-use it (or even detach it if we want to bring in a new one without having worry about updating all the attributes like name and color). I think this will end up simplifying this to not have to use a ProseMirror plugin, and remove the need for this whole synchronization process.

PR #3097, opened 2026-09-18
docs: Update docs with mobile PR stack changes

matthewlipski:
# Summary

This PR updates the docs for all changes in the #2939 PR stack. Created a separate PR to ease review.

## Rationale

Ensures we didn't miss documenting any changes from the stack. The addition of the mobile formatting toolbar is the most obvious change, but there were a number of smaller supporting changes which flew under that radar.

## Changes

N/A

## Impact

N/A

## Testing

N/A

## Screenshots/Video

N/A

## Checklist

## Additional Notes

N/A


docs/content/docs/react/components/formatting-toolbar.mdx
```diff
@@ -1,11 +1,11 @@
 ---
 title: Formatting Toolbar
-description: The Formatting Toolbar appears whenever you highlight text in the editor.
+description: Format selected text or use the mobile toolbar above the on-screen keyboard.
```
YousefED: The "or" isn't valid I think

YousefED (commented):
looks good, 1 comment

YousefED (commented):
Some generic feedback on the docs @ https://blocknote-website-git-mobile-docs-update-typecell.vercel.app/docs/

https://blocknote-website-git-mobile-docs-update-typecell.vercel.app/docs/getting-started#mobile-compatibility
- The Android meta workaround is now listed on getting started, but the iOS scroll-container on the mobile formatting toolbar area. wdyt, might be better to make consistent?
- instead of terms like "it's good practice", I'd say "we recommend"

https://blocknote-website-git-mobile-docs-update-typecell.vercel.app/docs/react/components/formatting-toolbar#mobile-formatting-toolbar
- The warning goes into explaining technicalities. `It's pinned to the visual viewport with position: fixed, so multiple containers would overlap each other. Wrap all your scrollable page content in one.` I'm not sure this is useful. Less is more?

matthewlipski:
> * The Android meta workaround is now listed on getting started, but the iOS scroll-container on the mobile formatting toolbar area. wdyt, might be better to make consistent?

I would say the current structure is better tbh. Setting the meta tag is non-invasive, easy, and has no real down sides so it's applicable to everyone and hence makes sense to put in "Getting Started". The iOS scroll container fix is however more involved and should be a decision made by the developer, therefore won't be applicable to everyone in a way that a "Getting Started" section imo should be.

> * instead of terms like "it's good practice", I'd say "we recommend"

Sure, will fix

> * The warning goes into explaining technicalities. `It's pinned to the visual viewport with position: fixed, so multiple containers would overlap each other. Wrap all your scrollable page content in one.` I'm not sure this is useful. Less is more?

Think I agree, will remove the technical explanation

YousefED:
> I would say the current structure is better tbh. Setting the meta tag is non-invasive, easy, and has no real down sides so it's applicable to everyone and hence makes sense to put in "Getting Started". The iOS scroll container fix is however more involved and should be a decision made by the developer, therefore won't be applicable to everyone in a way that a "Getting Started" section imo should be.

Agree, but the way it was setup there wasn't any incentive for developers to click through to the mobile experience to get the best iOS user experience. Addressed this in the last commit

